PR #14, opened 2016-03-14
A patch to correct the text printing (aka display).

NikEfth:


KrisThielemans:
rejected due to alternative solution

PR #15, opened 2016-03-14
Root lm

NikEfth:
This code has been developed and tested to import ROOT data into STIR as listmode files.
It has been tested with a linear source (see attached)
screenshot from 2016-03-14 18-03-35

In addition I attach a compressed file with an example for lm_to_projdata and a header for a scanner with listmode ROOT data. 
[examples.tar.gz](https://github.com/UCL/STIR/files/172592/examples.tar.gz)


tniknejad:
I have tested the Root_lm codes to import my Gate data into STIR using the list mode format.
This code is for a full ring scanner but can be easily modified to be compatible with any other geometries. I used it for both full ring and partial ring scanners and it works quite well.

KrisThielemans:
thanks. I'll review this later.
By the way, why do we need ECAT?

NikEfth:
HI Kris, 
I have not finished, yet. 
I committed some changes, but don't merge, yet.

The ecat dependency is bad history, I thought that I needed that in order to compile the listmode things, since the interface was common. 
I am going to fix this before the final commit.

NikEfth:
Hi Kris, 

Feel free to review this. 
I implemented *most* of your requests. 
Until, your comments I am not going to work more on this, 
but please keep in mind that I need it for the Scatter and tweaks, improvements and 
correction can take place when I resume working on the scatter ;)

KrisThielemans:
nearly there. Travis check failed for some reason.

Also, please add a few lines to documentation/release_3.1.htm. nice to put it in the same PR.

NikEfth:
I don't think it failed. 
I think it never started. The log is blank.

KrisThielemans:
Ok. Well, do the commit for the release notes. It should trigger a new one

KrisThielemans:
great. one question. I don't think you need

> Number of rings                          := 4
> Number of detectors per ring             := 504

as those are in your GATE keywords. However, it might be good to leave them in anyway, as this could be use to encode "gaps" in a very simple way: if the num_rings isn't what GATE simulated, we add gaps between blocks, and similarly for num_dets_per_ring. It doesn't give us flexibility regarding where the gaps are, but we can always rotate stuff with the offset, so for now this would solve our problems, and would allow us to do the mMR with gaps.

seems to me that this can be done by modifying get_next_record (as in Bin_GATE.cxx on the other branch).

what do you think?

NikEfth:
Bummer!
Travis doesn't know about ROOT.

KrisThielemans:
It might be possible to install a docker image with ROOT. @dvolgyes, would that help?

dvolgyes:
Hi,

I am sorry for the slow response time.
ROOT is tricky, personally I had a lot of troubles with ROOT, usually when
I compiled GATE.

The cleanest way would be if ROOT would provide deb packages, and Travis
would use these.
ROOT/GATE/GEANT have a long history with enviromental variables, these all
can be set in the Travis config file.

More realistic, but really time consuming: ROOT could be compiled inside
Travis.
This takes a lot of time, and also tricky, but possible.

Finally, docker images can be executed inside Travis.
I never tried it, but it is certainly possible.

As far as I know CERN doesn't have official Docker images.
In this case, STIR could host a Dockerfile (maybe, actually: several docker
files).
ROOT image could be generated from this dockerfile,
and inside this dockerfile, STIR could be compiled.

This is a good start:
https://gist.github.com/betatim/17f9c3116aefc8fa6605

I will not have time to think on this until 5th September, but if the issue
is still present after the 5th, I will try something.

In the meantime: the Travis build could be configured to ignore ROOT.
(Later a ROOT and a ROOTless build also should be enabled.)

(Several docker files: if ROOT image will be generated, then it is not a
big deal
to add an official STIR docker image too. Obviously, the stable releases
should
be separated from the master branch, etc., but it is not too complicated.)

Best,
  David

NikEfth:
Hi David, 

Ubuntu and Debian have ROOT in their repositories. The name of the metapackage is root-system-bin (http://packages.ubuntu.com/search?keywords=root&searchon=names&suite=xenial&section=all) 
This should do the trick. 
Just keep in mind that we need v5.3x and certainly on the v6.x, I will add support to that later.

dvolgyes:
My apologies, I haven't reinstall my ROOT since around ubuntu 11.04, at this time I had to compile it. 
Anyway, in my understanding this is the expected behavior:
- DISABLE_CERN_ROOT_SUPPORT=1 or DISABLE_CERN_ROOT_SUPPORT=0 should enable/disable root support, 
- and Travis should compile both versions. 
Is that correct? 

I will upload an updated Travis file with these features. The CMake config will give warnings until the corresponding root configs are not merged. (It will complain that the cmake macro is not present, but it will compile everything.)

NikEfth:
Correct!

Thank you David.

NikEfth:
Looks like it cannot find a global installation of root-config. 
Should you export the ROOT_SYS/bin path in PATH? 

AVAILABLE ROOT LIBRARIES:

-- CERN ROOT not found. If you do have it, set the missing variables (missing:  CERN_ROOT_LIBRARIES CERN_ROOT_INCLUDE_DIRS)

KrisThielemans:
if you know where it is, better to set CERN_ROOT when calling CMake (see the travis.yml file)

NikEfth:
I applied a request to whitelist root-system-bin. 
I don't know how much time this gonna take.

PR #16, opened 2016-04-12
Fix swig projdata segments

ashgillman:
There was an issue with `projdata_to_3D()`, it was designed to skip over all the odd segments. I'm not sure if this was a deliberate choice, but these changes mean all segments are loaded.

I can look at changing if the intent was for segment 0 to be loaded first, followed by +1, -1, +2, -2, etc. (Like the Siemens format)

I also made one quick change to the SWIG readme.


ashgillman:
Tested in Python 3.4, But since the changes are only in the SWIG `.i` file I'm not sure how other versions of Python would be different. Similarly for MATLAB.

KrisThielemans:
I'd prefer to keep the  Siemens-like order, as this was designed to do. Possibly moer intuitive to people to have -1,+2 next to eachother. Also, we have to make sure that fill_proj_data_from_3D is the reverse.

It looks like we forgot a copy statement in the s!=0 case. could you test that, as opposed to your change? Would also be good to add a siimple test to test_buildblock.py (using ProjDataInMemory: create one, convert to arrray, fill with randoms, store in ProjData, get back, should be identical)

ashgillman:
Alright, I made the change and have tested manually: looks good.

I have looked at implementing the test, but I'm not sure how to make an empty ProjData object. Any suggestions? I can always just pop a small dummy interfile sinogram in the `src/swig/text/python` directory, but it might be nicer (if possible) to just define one programmatically. I'm just not sure how to get data into a ProjData object.

KrisThielemans:
you can create a ProjDataInMemory object like this

`    s=stir.Scanner.get_scanner_from_name("ECAT 962")
    projdatainfo=stir.ProjDataInfo.ProjDataInfoCTI(s,3,9,8,6)
    projdata=stir.ProjDataInMemory(projdata.get_exam_info(), projdatainfo);
`
setting data can go via `get_empty_segment_by_view()` and `set_segment()` or by the trick mentioned above.

ashgillman:
Great, thanks very much :D

Also, the test modules use py.test, which has since moved to a separate project, [pytest](http://pytest.org/latest/). The only issue I see is that it is difficult (relatively) to get a hold of the older version retrospectively.

What are your thoughts on allowing compatibility between both. The command line usage is the exact same (`py.test <test_file.py>`), the import statement changes to:

    try:
        import pytest
    except ImportError:
        # No pytest, try older py.test
        try:
            import py.test as pytest
        except ImportError:
            raise ImportError('Tests require pytest or py<1.4')

KrisThielemans:
good idea. there's a ton of these around, I just picked one.

ashgillman:
Test added. It makes a projection from sinograms in which all values are just equal to the segment number, converts it back and forth, and checks they are the same. It also sanity checks to make sure the projection was set up as described.

Something similar should be done for MATLAB at some stage I suppose, but I don't have it installed to play with.

KrisThielemans:
great. thanks!

PR #17, opened 2016-07-04
 register reconstruction functions

NikEfth:
In this branch I registered the *reconstruction classes in the recon_buildblock_registry so they can be invoked in realtime. 
In addition, I have added a small demo class General_Reconstruction which demonstrates how it is done and an executable reconstruct_test which implements the main() function.

ATB, 
Nikos Efthimiou


NikEfth:
I applied all the suggested corrections.

PR #18, opened 2016-07-06
 reconstruct without output

NikEfth:


KrisThielemans:
change warning message in Reconstruct.cxx. Currently
 "You have disabled the output. No files will be written in the\n" 
  "disk after or during reconsturction\n"); }
No newlines in messages anymore (certainly not at the end). "written to disk", and spelling mistake "reconsturction"

PR #19, opened 2016-07-08
 data input from a new abstract data class

NikEfth:
In this branch we made the proper changes to be able to set the input data in the reconstruction process in real-time ( e.g. some other code). 
To do so, the classes ProjData and CListmodeData are now derived from a common ExamInfo, the Reconstruction class has a new function set_input_dataset(shared_ptr<ExamInfo>) and in the 
PoissonLogLikelihoodWithLinearModelForMeanAndProjData and PoissonLogLikelihoodWithLinearModelForMeanAndListModeData there proper dynamic_cast takes place. 

Limitations: 
1) Don't handle DynamicProjData ( in the short future) 
2) ExamInfo should be a registered.


KrisThielemans:
I don't think it makes sense to derive from ExamInfo. The Info classes are always member of the Data classes. Also, it means that at present you will have 2 ExamInfos in the objects. A more logical choice would be to ceate a new abstract class DataSet and derive from that. Originally I thought ExamData but check also CCP terminology.

Use consistent functions. you have both
`void set_input_dataset(shared_ptr<ExamInfo> _this_dataset); 
`
and
`virtual void set_input_data(const shared_ptr<ExamInfo> &);
`

Are you sure you need
`this->proj_data_ptr.reset(dynamic_cast < ProjData * > (_this_dataset.get()) );
`
I think the shared_ptr class will do this for you.

NikEfth:
In CCP I found the object: study which would contain several series, some of which would be images, some raw data etc 
Between study and examdata, I would prefer examdata, but I am afraid of conflict between examinfo and examdata, they sound similar but in the current implementation they would sit in different levels. 
In my point of view study, is too generic and unless it is related to images as well does not make much sense.

KrisThielemans:
a Study would contain several Series. Series is a rather specific DICOM name though. (it's a Series as it can contain several files, e.g. one per slice and or time frame). 

We have ProjDataInfo and ProjData (which has a ProjDataInfo which describes it). From that perspective, having ExamInfo and ExamData makes sense (although I guess to be consistent it'd have to be ExamDataInfo). The ExamInfo describes the Exam, but it has no actual measured data.

It's a deficiency in STIR that we don't have Info objects for the images. it'd be useful in many places. 

Alternatively, we have no Info objects anywhere, just Data that might not contain the "large stuff", but that'd be a major change for ProjData. 

so, I guess I vote for ExamData

NikEfth:
So ExamInfo should move higher to ExamData. 
In addition we could use the {data format:= sinogram || coincidencelist } to register the two data types.

NikEfth:
@KrisThielemans 
\o/

PR #23, opened 2016-08-09
Array from projdata

NikEfth:
Added functionality to fill 1D and 3D arrays from projdata. 

I can see that the comparison brings some older work from the _reconstruct_without_output that I had accidentally done/pushed in my master. 
I can see that is safe.


KrisThielemans:
close due to moving functionality to ProjData

PR #24, opened 2016-08-16
Projdata functions to arrays

NikEfth:
This pull request includes modifications in ProjData and MultipleProjData to allow them to be sorted in arrays. 

In addition a test class is included.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,30 @@
+# Language selection
+language: cpp
+
+# Both clang and gcc can be tested. The more is the better.
+compiler:
+  - clang
+  - gcc
+
+# Latest supported Ubuntu
+dist: trusty
+
+# No need for sudo
+sudo: false
+
+
+# Compilation dependencies
+addons:
+  apt:
+    packages:
+    - libboost-all-dev
+    - swig
+
+# Actual compilation script
+# cmake parameters could be refined
+script:
+  - mkdir build
+  - cd build
+  - cmake ..
+  - make -j 8
+  - make test
--- a/examples/C++/General_Reconstruction.cxx
+++ b/examples/C++/General_Reconstruction.cxx
@@ -1,0 +1,57 @@
+#include "stir/recon_buildblock/General_Reconstruction.h"
+#include "stir/CPUTimer.h"
+#include "stir/HighResWallClockTimer.h"
+
+#include <iostream>
+START_NAMESPACE_STIR
+
+General_Reconstruction::
+General_Reconstruction()
+{
+    this->set_defaults();
+}
+
+void
+General_Reconstruction::set_defaults()
+{
+
+}
+
+void
+General_Reconstruction::initialise_keymap()
+{
+    this->parser.add_start_key("General reconstruction");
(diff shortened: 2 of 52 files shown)


KrisThielemans:
- does MultipleProjData.cxx need to include ProjDataInMemory.h?
-  size_t get_projData_size() needs to be  std::size_t get_proj_data_size() 
- get_num_sinograms() and size_all() need to return std::size_t. You will need explicit casts in their implementation. I'd put their implementation in .cxx
- stir.i has `int num_sinos = proj_data.get_num_sinograms()` would need std::size_t or conversion
- test_export_array.cxx has wrong Copyright header (Hammersmith Imanet)
- rather unfortunate white-space differences, making it difficult to see relevant changes.

NikEfth:
@KrisThielemans 
> - does MultipleProjData.cxx need to include ProjDataInMemory.h?
I am sorry is a leftover from some tests. 
> - size_t get_projData_size() needs to be std::size_t get_proj_data_size()
> - get_num_sinograms() and size_all() need to return std::size_t. You will need explicit casts in their implementation. I'd put their implementation in .cxx
> - stir.i has int num_sinos = proj_data.get_num_sinograms() would need std::size_t or conversion
I set this in order. But I am a bit confused, why we use size_t were, MultipleProjData::get_num_proj_data() returns an unsigned, and all Array constructors use integers. Should we overload the constructors to define sizes with size_t? 
> - test_export_array.cxx has wrong Copyright header (Hammersmith Imanet)
I corrected it mentioning only UCL 2016.
> - rather unfortunate white-space differences, making it difficult to see relevant changes.
🎱 I have not edited (consciousnessly) the reconstruction class. I Checked them, they are all white spaces.

KrisThielemans:
-	Still need to change get_projData_size to get_proj_data_size() (to follow STIR naming convention)
-	Indeed confusing that sometimes we return int, unsigned and std::size_t. Largely historical. Array often uses int as it has signed indices, but gradually size stuff moved to std::size_t to be consistent with the C++ library. size_all() definitely needs std::size_t. get_num_whatever() could be unsigned (as it’s maybe not a “size"). Let’s try to be as consistent with the rest of STIR as we can.
-	If no changes in PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx, can you revert it? It’d confuse people (and me).
-	In the test function, you still write the array to file. As far as I can see, this is not really a test on the fill_from, copy_to functionality, but just some writing. It creates a dependency on IO stuff which we don’t need. In fact, this test would probably fail on Windows anyway (need to open the stream binary). I’d cut it out.
- Are we sure that the Reconstruction commit should be in this pull-request? Maybe you should base your PR from something else? Could maybe create a new one with a single clean commit (if it takes less than 10 mins).
- By the way, I notice that the last commit message is relatively nonsensical except for the 2 of us :-;

NikEfth:
I tried to correct that leak in the Reconstruction classes, but I wonder what will happen when we merge with the scatter branches... :/

NikEfth:
Please review PR #25

PR #28, opened 2016-08-18
Travis integration: recon_test_pack enabled, preparation for OpenMP

dvolgyes:
These two commits enable automatic tests after the compilation.
OpenMP tests at the moment fail, so OpenMP compilation is not enabled (and would be tricky with clang because reliable openmp requires at least clang 3.8 which is not yet present in Travis).

Travis shows a test error, which is fixed in PR#27. (Merge first PR#27, then this request.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,3 +44,5 @@ script:
   - ./run_test_simulate_and_recon_with_motion.sh
   - ./run_scatter_tests.sh
   - ./run_tests.sh --nointbp
+  - cd $TRAVIS_BUILD_DIR/recon_test_pack/SPECT
+  - ./run_SPECT_tests.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,4 +43,4 @@ script:
   - ./run_test_simulate_and_recon.sh
   - ./run_test_simulate_and_recon_with_motion.sh
   - ./run_scatter_tests.sh
-  - ./run_tests.sh
+  - ./run_tests.sh --nointbp
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,15 @@ compiler:
   - clang
   - gcc
 
-# Latest supported Ubuntu
+# Enviroment variables:
+global:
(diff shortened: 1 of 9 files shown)


dvolgyes:
Theoretically, the pull request can be re-started after the PR#27 is merged. 
If PR does not pass the check after the merge of PR#27, then I can extend/refine this PR.

KrisThielemans:
- ./run_tests.sh will need --nointbp as otherwise it will fail (known bug)
- ./run_ecat_tests.sh can only be run if we have the ECAT library, which is getting near to impossible, so disable that one as well.
- some unfortunate white space differences in the .sh scripts. can you set your editor to not do this until we have sorted that one out?

dvolgyes:
I will correct them tomorrow.

dvolgyes:
I fixed the run_ecat_tests.sh to exit with proper error codes if necessary. Someone might use it at his/her local version. But the test is not called from Travis CI. 

FBP3DRP fails if it runs with OpenMP using more than 1 cores. (Double free bug.) Due to this bug, and the challenges with clang+openmp combination: OpenMP is disabled in the automated tests.

KrisThielemans:
thanks. could add the SPECT test...

question: I guess by adding "set -e", the scripts that use if to test on return status are now aborted before? Not ideal, but would need more work.

dvolgyes:
Yes, it is less than ideal, but could be improved.
The idea is that any segfault or other problems are also caught this way,
not just regular exits.
(Also any shell error, for instance, an error during  a "cp" command, etc.)
The exit status could be explicitly checked, just it is a bit more
complicated.

This one checks the exit codes and prints 0, 2 and 3 but terminates at the
single "false" line.

#!/bin/bash
set -e
EXAMPLE=0
echo $EXAMPLE
false && EXAMPLE=1 || EXAMPLE=2
echo $EXAMPLE
true && EXAMPLE=3 || EXAMPLE=4
echo $EXAMPLE
false
echo "Never reached"

It is a design question what and how should be tested.
As an improvement, my idea is this:
- testing, if the scrip runs in Travis ($TRAVIS=true)
- enable "set -e" only in these cases.
What do you think?

SPECT: Sure, I missed it.

KrisThielemans:
good idea. I'll accept if the build works.

PR #30, opened 2016-08-21
Fixing header guard bug.

dvolgyes:
That's a quite trivial bug, but it is a good occasion to discuss something related:
shouldn't we aim more modern c++ constructs?

Like  #pragma once 
which would eliminate this kind of errors, 
(See https://en.wikipedia.org/wiki/Pragma_once )
or the c++11 features like the auto specifier,  constexpr,  or range-based loops?

### DIFF
--- a/src/include/stir/Shape/Ellipsoid.h
+++ b/src/include/stir/Shape/Ellipsoid.h
@@ -25,7 +25,7 @@
   \author Kris Thielemans
 */
 #ifndef __stir_Shape_Ellipsoid_h__
-#define __stir_Shape_Elliposoid_h__
+#define __stir_Shape_Ellipsoid_h__
 
 
 #include "stir/RegisteredParsingObject.h"


KrisThielemans:
C++11 features would simplify the code. I've been reluctant due to compiler support in the past but would be happy to go this way for STIR 4.0.

dvolgyes:
I am not sure if the master branch is the 4.0 development branch or it is the 3.1, etc.
Anyway, my suggestion is this: when you decide, write a letter, and summarize what features are accepted or rejected.  (c++11/14/17, etc.)

The best would be to have a coding_standards.txt in the actual development branch which summarizes briefly the requirements. 
E.g. C++98 and boost 1.36+
or C++11 except lambda expressions , boost 1.50+ 

C++11 has many features from former boost releases (under different namespaces, so there is no collision), so at the transition, it should be decided which features should be used from the standard and which from boost.

Anyway, I think this header bug is done. If it works, it could be closed from my side.

PR #32, opened 2016-08-22
Pre Scatter correction 1

NikEfth:
I try to bring the scatter correction in small pieces of clean code. 
Here are some prerequisites.

### DIFF
--- a/documentation/release_3.1.htm
+++ b/documentation/release_3.1.htm
@@ -184,7 +184,7 @@ using the default output file format.</li>
   </ul>
 </li>
 <li> Support for Energy Resolution in Scanners, has been added. 
-In order ensure backwards compatibility, at the momement, it is completely optional.</li>
+In order to ensure backwards compatibility, at the moment, it is completely optional.</li>
 <li> Information on energy windows has been added in the ExamInfo. </li>
 </ul>
 
--- a/documentation/release_3.1.htm
+++ b/documentation/release_3.1.htm
@@ -183,7 +183,9 @@ using the default output file format.</li>
    <li>added <code>get_proj_data_info_sptr()</code> (also to <code>Sinogram</code> et al)</li>
   </ul>
 </li>
-
+<li> Support for Energy Resolution in Scanners, has been added. 
+In order ensure backwards compatibility, at the momement, it is completely optional.</li>
+<li> Information on energy windows has been added in the ExamInfo. </li>
 </ul>
 
 <h3>Other code changes</h3>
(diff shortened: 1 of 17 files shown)


NikEfth:
I have to highlight that, the test did not fail on my code. 
It timed out.

KrisThielemans:
travis build fails.

- Scanner.h: add to doxygen comment that the energy resolution of the system as at the reference energy (both in variable and set function). add somewhere that by default the energy resolution is set to a negative value (indicating "unknown")
- `Scanner::set_energy_resolution` and `set_reference_energy`, just use `(const float)`  as argument (no reference). Makes little actual difference but no need for the extra indirection.
-  CreateTailMaskFromACFs.h: some spelling mistakes in doxygen

 > In the latter case, where the output is to be used from some other code the
 >  output file can be omitted, in order to avoid unnecessary outputs.

and usage message for create_tail_mask_from_ACFs  executable should not be in this .h but still the executable .cxx
- `CreateTailMaskFromACFs::set_input_projdata` etc should be called `_sptr` I guess like rest of STIR?
- stir_math.h: don't we have an include file with similar functions already?

KrisThielemans:
in log message of run_test_simulate_and_recon.sh (in the Travis build): 

> WARNING: KeyParser warning: unrecognized keyword: energy resolution
> WARNING: KeyParser warning: unrecognized keyword: reference energy (in kev)

that's weird...

Probably because you haven't added the reading/writing yet. By the way, I wouldn't write the energy keywords when one of them is invalid.

NikEfth:
Before my commit, the class class pow_times_add (which is now in stir_math.h) was defined in stir_math.cxx. 

Actually, I could transfer process_data, as well. It would make sense.

KrisThielemans:
sure, I thought we had a .h with other FunctionObjects but can't seem to find it.  Maybe do a search for unary_function or so.

Moving process_data would make sense, except that its interface is a bit specific. Let's leave it for later (unless you actually use it elsewhere)

KrisThielemans:
great. I suggest to change

> pdfs.get_exam_info_ptr()->get_high_energy_thres() > -1 &&
>  pdfs.get_exam_info_ptr()->get_low_energy_thres() > -1
to
> pdfs.get_exam_info_ptr()->get_high_energy_thres() > 0 &&
>  pdfs.get_exam_info_ptr()->get_low_energy_thres() >=0
etc.

KrisThielemans:
is this one ready?
add a few lines to documentation/release_3.1.htm. I guess this belongs in the developers section, saying that everything is backwards compatible and that actual scanner energy resolutions still need to be filled in.

PR #33, opened 2016-08-23
Cosmetic fixes and some leaks.

dvolgyes:
Some cosmetic changes, e.g. switching from char array to strings,
and some memory leak fixes.
The leaks are not serious (the code will exit anyway) but
I try to reduce the number of warnings, so I could focus on the more serious ones.
Deprecated header import, deprecated source:  I think we don't need them anymore.

Let's see if the build passes the tests.


KrisThielemans:
thanks. But I do think we need getopt.h. It's there for Windows.

dvolgyes:
Well, that's true, getopt is not present on windows.
For now I reverted the getopt part but on midterm (in another PR) I would like to replace it with boost::program_options.

KrisThielemans:
thanks. I'll accept as squash-commit to avoid people getting confused by the getopt removal/inclusion.

let's discuss boost::program_options on the devel list first.

PR #34, opened 2016-08-28
Travis CI: enabling ROOT support

dvolgyes:
ROOT 5.34 will be installed always (both for Linux and OSX),
but builds will be generated with and without enabling ROOT in cmake.


NikEfth:
Looks good, 
I will merge ROOT_lm with your branch and push, to check

PR #35, opened 2016-08-31
PostFiltering.h created.

NikEfth:
Many times we need to use this from somewhere else.

### DIFF
--- a/examples/samples/postfilter_Metz.par
+++ b/examples/samples/postfilter_Metz.par
@@ -1,4 +1,4 @@
-PostFilteringParameters :=
+PostFiltering parameters :=
   Postfilter type :=  Separable Cartesian Metz
 Separable Cartesian Metz Filter Parameters :=
 x-dir filter FWHM (in mm):= 5
--- a/examples/samples/postfilter_median3x3x3.par
+++ b/examples/samples/postfilter_median3x3x3.par
@@ -1,4 +1,4 @@
-PostFilteringParameters :=
+PostFiltering parameters :=
   Postfilter type :=Median   
   Median Filter Parameters :=
   mask radius x := 1   
--- a/recon_test_pack/postfilter_truncate_circular_FOV.par
+++ b/recon_test_pack/postfilter_truncate_circular_FOV.par
@@ -1,4 +1,4 @@
-PostFilteringParameters :=
+PostFiltering parameters :=
 postfilter type := Truncate To Cylindrical FOV
 Truncate To Cylindrical FOV Parameters:=
 strictly_less_than_radius:=0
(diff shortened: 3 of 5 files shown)


NikEfth:
I think that this is the last preScatter PR.

KrisThielemans:
A few suggestions:

- Add some doxygen for the class
- The file (and class) shouldn't be in the doxygen group `utilities`
- Let's rename to `filter_sptr`
- add start key "Postfilter parameters" (i.e. with space). Need both for backwards compatibility.
- need to say that constructor's inline to avoid linking trouble (or define it in the class def)
- could template in terms of DiscretisedDensity<3,float> 

More advanced stuff:
- Not so nice to have public members.
- better to derive from ParsingObject?

Even more advanced:

it seems a lot of effort to have to define a class which just wraps a single object for parsing. Other examples of this are for forward_project etc. Could nearly automate this:

> template <class T> MakeParsingObject: public ParsingObject
> {
>  inline MakeParsingObject:(std::string start, std::string type=name)
> {
>   this->add_start_key(name + " Parameters");
>  this->add_parsing_key(type " type", &object_sptr);
>   this->add_stop_key("END " + name + " Parameters");
> }

> void set(shared_ptr<T>);
> shared_ptr<T> get() const;
> // maybe add convenience function to forward function call
> // hard to do this well though (multiple args, return-type,...)
> // maybe just rely on get()
> template <class ArgsT> inline void operator()(ArgsT args) const 
> { (*this->object_sptr)(args); }
> //or what about this
> inline const shared_ptr<T>& operator->() const 
> { return this->object_sptr; }
>   protected:
>     shared_ptr<T>  object_sptr;
>  // etc
> };

Then we can do
>
> MakeParsingObject<DataProcessor<DiscretisedDensity<3,float> > > postfilter("Postfilter");
> postfilter.parse(filename);
> postfilter->apply(blabla);

ok, wouldn't be able to add a 2nd starting_key at present I guess (or can we do `postfilter.add_starting_key()` just in postfilter.cxx?)

NikEfth:
I am going to schedule the "more advanced" case for next week. 👍

KrisThielemans:
travis failed (I didn't check why).

I forgot to say that in the doxygen header you need to say stir::PostFilter (otherwise it doesn't link to the class def).

I have a preference for template in terms of DataT as opposed to being specific to DiscretisedDensity. DataProcessor is also templated that way... Will immediately allow using this class for DynamicDiscretisedDensity etc for instance

NikEfth:
The error was with the start_key. 
The one used was Postfilterparameters and not Postfilter parameters, I added both. 

I agree with DataT.

PR #41, opened 2016-09-05
Root extension

NikEfth:
Two new classes InputStreamFromROOTFile(ForCylindricalPET) & (ForECATPET) should address the issue with the different scanner types in GATE, at least these two. 
I added a new hroot example, too. 

I will focus now to create a test class.


NikEfth:
When Travis succeed you may accept the PR, if you have no further remarks.
Of course, still doesn't support ROOT.  

Though I don't use anymore the get_total_number_of_events() I feel that it could stay for future reference. 
I didn't extent the tests in the test_root_GATE.sh but  cleared the code. Allow me to do this at a later point, if really needed.

KrisThielemans:
looks great. I'll wait for Travis.
Can you add a few lines in the release notes as well?

thanks!

NikEfth:
I will add them tonight 

One more thing, how do I add to Travis the new test?

KrisThielemans:
check the travis.yml file in the root. should be obvious I think.

dvolgyes:
Hi,

Travis is quite easy.
Basically, you have to edit the .travis.yml file.
The main point is this:
you need a program which exits with non-zero exit code.
It can be anything, like a Makefile (make test, etc.)
(Or several such lines.)
If any of them return with nonzero exit codes, then the test fails.
You could use any magic, like grep/sed/other command line magic,
but too many tools could compromise portability, e.g. OSX build.

Bash/sh scripts usually removes the exit codes, but if you add
"set -e"
then it will check executed commands, and the script will return with exit
code,
if the executed command returned with non-zero value.
(Check one of the current test scripts. I think I left a comment there
about Travis.)

So the most straightforward it to extend the current STIR test-script set.
If it is a new test file, then call it from after the current tests.

Be careful with the directories, check the 'cd' commands in the current
setup.
(You have to use Travis variables, because there is no guarantee that the
compilation will always be performed in the same dir.)

NikEfth:
Done. 
Thank you.

KrisThielemans:
travis tests fail

 ./run_root_GATE.sh
This script should work with STIR version >3.0. If you have
a later version, you might have to update your test pack.
Please check the web site.
GATE support has not been installed in this system. Aborting.

I guess you'll have to disable the running of the test if DISABLE_CERN_ROOT is on or something

KrisThielemans:
Hmmm. I don’t really like that. Better to disable it from travis.

dvolgyes:
The question is: do you want to test it in Travis?

If yes:
 - add ROOT/GATE to the build process (e.g. as deb/homebrew dependency)
 - add  and/or enable cmake compilation flag(s)
 - call the test script at the end of .travis.yml

If no: (locally still could be tested)
 - add the test script
 - add the compilation flag

I quite dislike silent fails. If something is missing, then I prefer
non-zero exit codes.
I have seen many projects, even in radiotherapy, where codes silently fail,
not even with a warning message,
and they use some strange built-in defaults. (Which is very bad for a MC
simulation in RT.)

Travis doesn't have to test everything, but if it does test something, then
we should not introduce a hack for it,
because it will backfire sooner or later.

Of course, this is just my opinion.
(I try to follow the conversations, but I have limited time to interact
nowadays,
and I know it is easier to give advices than actually doing the job.)

NikEfth:
Hi @dvolgyes , 

Regarding the first step, to add ROOT sources for apt, I have tried to whitelist root-system-bin but Travis returns messages that cannot find the source. 

Which version of Linux to you use?

KrisThielemans:
hi.
let's leave the running of these tests on Travis for later. just disable the test in travis.yml when DISABLE_ROOT_CONFIG is set.

NikEfth:
I hope that this will work.

NikEfth:
One small correction, there was a leftover var.

NikEfth:
It is ready :) 
A squashed merge would be fine.

KrisThielemans:
so... I understand now. DISABLE_ROOT_CERN=0 fails on Linux as the package is not allowed. I guess the only sensible thing there is to remove this run from the travis.yml?

it seems that the mac version failed because of a spurious time-out. I've restarted the job to check.

NikEfth:
ok, I removed the line in travis config file.

KrisThielemans:
This commit gets rid of the test completely. Instead i thought we'd test on "$TRAVIS_OS_NAME" == "osx"  our something.

However i see that I the previous build the osx gcc test timed out after 10 mins again, while the osx clang test succeeds. Outta it normal that it takes so long?

Sorry...

NikEfth:
No this is not normal, 
Although I have noticed that in ~30% of the times that I tried to run the 
script in my laptop the lines which read data from ROOT files hanged, doing nothing. 

I thought that it was an issue in my installation. Guess not. 
The solution is to terminate the procedure and re-run. :/

NikEfth:
Should note that this behavior has not appeared during lm_to_projdata.

KrisThielemans:
not during lm_to_projdata? But when then?

NikEfth:
So what are we going to do, because I need some to go on?

NikEfth:
I am sorry, I had not refreshed the page. 
Well somewhere in these lines: 

true_root_num=$(root -l ${INPUT_ROOT_FILE} << EOF | grep "Number of trues stored in this time period" | grep -o -E '[0-9]+'
Coincidences->Draw(">>eventlist","eventID1 == eventID2 && comptonPhantom1 == 0 && comptonPhantom2 == 0","goff");
Int_t N = eventlist->GetN();
cout<<endl<<"Number of trues stored in this time period:"<< N<<endl;
EOF)

dvolgyes:
One remark about the timeouts:
You can increase the timeout to larger numbers, e.g. 20 minutes or more:
https://docs.travis-ci.com/user/common-build-problems/

Look for "travis_wait".

It is just a band-aid, but if there is no other problem,then it could help.
(However, run-time regressions often indicate something deeper.)

NikEfth:
Hi David, 

Could you please have a look at these: 
[](https://github.com/travis-ci/apt-package-whitelist/issues/3396)
[](https://github.com/travis-ci/apt-package-whitelist/issues/3329)

dvolgyes:
I am not quite sure about the whitelisting process but I think the issue
lies in the strange deb package generation.
The source for all of the root packages is this:
http://packages.ubuntu.com/trusty/science/root-system
(See the right column)
But from this source package, dozens of packages are generated.
This might be the issue. (Or Travis just did not recognize the source
package, because of the different names.)

Bigger issue could be that a single root package might not be enough,
and you might need to whitelist many of them.
If you check the .dsc file, it containts the most relevant information such
as which packages are built from this source.

Or it might be if you ask for whitelisting for "root-system" and you give
the source to this package,
then the dependencies will be accepted automatically. I have no idea.

KrisThielemans:
Let’s forget about the white listing for the moment. I'm just worried about the time out.


Are you saying it times out in the script itself? Some scripting thing or root itself then. 

Could first try to simplify the script (use an input file for the root command). as it times out on your system, you don't have to wait for travis. you can indeed just run these few lines then without the rest of the script...

NikEfth:
This time-out had nothing to do with ROOT. 
It timed-out during compilation.

NikEfth:
@KrisThielemans 
Checks passed!

KrisThielemans:
wonderful. thanks!

Issue #40, opened 2016-09-05
Remove set additive and multiplicative functions from iterative reconstruction


NikEfth:
virtual void set_additive_proj_data_sptr(const shared_ptr<ExamData>&);
virtual void set_normalisation_sptr(const shared_ptr<BinNormalisation>&);
 
IterativeReconstruction has no clue about additive data nor BinNormalisation… they are implemented in terms of
 
this->objective_function_sptr->set_additive_proj_data_sptr(arg);
 
but GeneralisedObjectiveFunction doesn’t have a clue about these either..
 
So, these additions have made GeneralisedObjectiveFunction and IterativeReconstruction really rather specific to our problems. 

Question is if we need them… 
Kris

PR #42, opened 2016-09-05
 fix issue #40 cleanup reconstruction class

NikEfth:
Issue discussed at #40


KrisThielemans:
thanks. let me know when Travis finishes.

By the way, it would have made sense to add them to AnalyticReconstruction but as you didn't have an implementation we leave it for now.

NikEfth:
Right, 

I will add this in my todo list.

It seems that Travis finished. 

Nikos

PR #43, opened 2016-09-09
Get covariance value on the fly

NikEfth:
This PR has a new function which allows us to perform the power method and get the first eigenvalue&vector without knowing explicitly the CoV matrix.

### DIFF
--- a/src/include/stir/numerics/MatrixFunction.inl
+++ b/src/include/stir/numerics/MatrixFunction.inl
@@ -110,7 +110,7 @@ namespace detail
   template <class elemT, class vecT>
   inline
   void
-  matrix_matrixT_multiply_help(vecT& retval, const Array<2,elemT>& m, const vecT& vec, vecT& cov)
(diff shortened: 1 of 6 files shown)


NikEfth:
Probably I should note that for aesthetic reasons I added a mean() function in the Array. 
Completely optional otherwise.

src/include/stir/numerics/MatrixFunction.inl
```diff
@@ -106,6 +106,44 @@ namespace detail
...
+      float norm_factor = 1.f / m_max_row;
+      int k = m_min_row;
+
+      for(int i=m_min_col; i<=m_max_col; ++i)
+      {
+          for(int j= m_min_col; j<= i; ++j)
+          {
+              k = m_min_row;
+              cov = m[k][i] * m[k][j];
+              for (++k; k<= m_max_row; ++k)
```
NikEfth: @KrisThielemans 
I would definitely consider caching this loop. 
We could cache half the CoV values and never loop over it again.

PR #53, opened 2016-10-09
Listmode reconstruction

NikEfth:
This is ongoing work. 
But I would like to start some discussion, 
please feel free to review and comment.


NikEfth:
On my side, Travis didn't fail. 
Could you restart the tests?

NikEfth:
I will defend my case, as only one osx timed out, for no obvious reasons

src/buildblock/ProjDataFromStream.cxx
```diff
@@ -194,6 +194,96 @@ ProjDataFromStream::get_viewgram(const int view_num, const int segment_num,
...
+{
+    return get_bin_value(this_bin.segment_num(),
+                         this_bin.axial_pos_num(),
+                         this_bin.view_num(),
+                         this_bin.tangential_pos_num());
+}
+
+
+float
+ProjDataFromStream::get_bin_value(const int segment_num,
```
KrisThielemans: do we need the version with 4 ints? Scope for confusion in argument order. (Of course it's there in Bin as well, but at least it's in 1 place)

src/buildblock/ProjDataFromStream.cxx
```diff
@@ -194,6 +194,96 @@ ProjDataFromStream::get_viewgram(const int view_num, const int segment_num,
...
+                                  const int view_num,
+                                  const int tang_pos_num) const
+{
+    if (sino_stream == 0)
+    {
+        error("ProjDataFromStream::get_viewgram: stream ptr is 0\n");
+    }
+    if (! *sino_stream)
+    {
+        error("ProjDataFromStream::get_viewgram: error in stream state before reading\n");
```
KrisThielemans: error message says "get_viewgram".

Also, delete \n from end of error messages (also a few lines above)

src/buildblock/ProjDataFromStream.cxx
```diff
@@ -371,6 +461,93 @@ ProjDataFromStream::set_viewgram(const Viewgram<float>& v)
 }
 
 
+std::vector<std::streamoff>
+ProjDataFromStream::get_offsets_bin(const int segment_num,
```
KrisThielemans: I think it might be best to rewrite this in terms of get_offsets. That way the ugly code is in only one place? Or we could rewrite get_offsets in terms of this one of course.

src/buildblock/ProjDataInMemory.cxx
```diff
@@ -160,10 +160,11 @@ write_to_file(const string& output_filename) const
...
 ProjDataInMemory::get_bin_value(Bin& bin)
 {
-   Viewgram<float> viewgram = get_viewgram(bin.view_num(),bin.segment_num()); 
+   Viewgram<float> viewgram = get_viewgram(bin.view_num(),bin.segment_num());
     
-   return viewgram[bin.axial_pos_num()][bin.tangential_pos_num()]; 
-
+   return viewgram[bin.axial_pos_num()][bin.tangential_pos_num()];
+//    return get_bin_value(bin.segment_num(), bin.axial_pos_num(),
+//                         bin.view_num(), bin.tangential_pos_num());
```
KrisThielemans: not sure why you commented it out now. I don't mind having get_bin_value in ProjDataFromStream so you can use it there.

src/include/stir/Bin.h
```diff
@@ -79,6 +79,12 @@ class Bin
   
   //! accumulate voxel's contribution during forward projection 
   inline Bin&  operator+=(const float dx);
+  //! multiply bin values during normalisation -- apply()
```
KrisThielemans: ok to add these, but remove the comments on apply/undo. It only makes sense for people who think about BinNormalisationFromProjData...
maybe should then add a subtract as well for consistency.

src/include/stir/ProjData.h
```diff
@@ -151,6 +152,10 @@ class ProjData : public ExamData
   //! Set sinogram
   virtual Succeeded 
     set_sinogram(const Sinogram<float>&) = 0;
+  //! Get Bin value
+   virtual float get_bin_value(const int view_pos, const int segment_pos, const int axial_pos, const int tang_pos) const = 0;
+  //! Get Bin value
+  virtual float get_bin_value(const Bin& this_bin) const = 0;
```
KrisThielemans: Unless we really need this I would comment this out for the moment. It breaks compatibility for all those non-ProjDataFromStream objects.
we could flag it up as something to implement in the future.

src/include/stir/ProjDataGEAdvance.h
```diff
@@ -73,7 +73,18 @@ class ProjDataGEAdvance : public ProjData
     Sinogram<float> get_sinogram(const int ax_pos_num, const int sergment_num,const bool make_num_tangential_poss_odd=false) const; 
     Succeeded set_sinogram(const Sinogram<float>& s);
  
-    
+    float get_bin_value(const int segment_num,
```
KrisThielemans: cannot do this. Will generate errors for any decent compiler. That's why i wouldn't force implementation. (there's no way we can test if an ProjDataGEAdvance mods are correct as we don't have such files anymore)

src/include/stir/listmode/CListModeDataECAT8_32bit.h
```diff
@@ -18,7 +18,7 @@
   \file
   \ingroup listmode
   \brief Declaration of class stir::ecat::CListModeDataECAT8_32bit
-    
+
```
KrisThielemans: lots of white-space diffs in this file. remove

src/recon_buildblock/ProjMatrixByBinSPECTUB.cxx
```diff
@@ -205,32 +205,32 @@ set_up(
...
+    vol.thcm = image_info_ptr->get_voxel_size().z()/10.f;  // Image: slice thickness (cm)
 
 	//..... geometrical and other derived parameters of the volume structure...............	
 	vol.Npix    = vol.Ncol * vol.Nrow;
 	vol.Nvox    = vol.Npix * vol.Nsli;
 
-	vol.Ncold2  = (float)vol.Ncol / (float)2.;  // half of the matrix Nvox (integer or semi-integer)
-	vol.Nrowd2  = (float)vol.Nrow / (float)2.;  // half of the matrix NbOS (integer or semi-integer)
-	vol.Nslid2  = (float)vol.Nsli / (float)2.;  // half of the number of slices (integer or semi-integer)
+    vol.Ncold2  = (float)vol.Ncol / (float)2.f;  // half of the matrix Nvox (integer or semi-integer)
```
KrisThielemans: either (float)2 or 2.F. why both? So I think we don't need these changes?

src/include/stir/recon_buildblock/ProjMatrixByBin.h
```diff
@@ -155,7 +156,7 @@ class ProjMatrixByBin :
 
   
 protected:
-  shared_ptr<DataSymmetriesForBins> symmetries_ptr;
+  shared_ptr<DataSymmetriesForBins> symmetries_sptr;
```
KrisThielemans: unfortunate but necessary change. Add it to release notes for developers (as breaks any derived class from ProjMatrixByBin)

src/buildblock/ProjData.cxx
```diff
@@ -289,6 +289,25 @@ ProjData::get_related_viewgrams(const ViewSegmentNumbers& view_segmnet_num,
   return RelatedViewgrams<float>(viewgrams, symmetries_used);
 }
 
+std::vector<float>
+ProjData::get_related_bin_values(const std::vector<Bin>& r_bins) const
```
KrisThielemans: do we need it? If not, comment out. I'd rather have a RelatedBins interface later.

src/buildblock/ProjDataInMemory.cxx
NikEfth: This was tmp, just till I make sure other things are working properly.

src/include/stir/ProjData.h
NikEfth: Now they don't make much sense, as I removed the bin-wise operations. But I will keep them for future reference.

src/include/stir/ProjDataGEAdvance.h
NikEfth: What would you like me to do here?

src/include/stir/Bin.h
NikEfth: Now they don't make much sense, as I removed the bin-wise operations. But I will keep them for future reference.

dvolgyes:
Some remarks about the subset organization:
- indeed, the event-based subset selection has two major advantages:
  - usually speeds up the data reading
  - usually makes the subsets balanced

But:
- if you read the file, and you randomly select what you use and what you
don't, then you can make the code even slower then in MLEM.
  (I have seen this pattern in a CERN collaboration.)
- sometimes structured subsets are more useful, e.g. when you use GPU and
you can use texture memory and other tricks.

Another aspect is the actual subset pattern. For simplicity, most
commercial scanners use fixed NxM scheme, e.g. 3 iterations with 11
subsets, etc.
This is suboptimal choice. For initial state: small subsets are more
useful. E.g. start with 12 subsets.
But after a while the number of events will be a major obstacle. So after
e.g. 5 iterations, you would go for 8 subsets instead of 12. And so on.

So it might be useful to accept a function for number of subsets, e.g.
subsets = 4; // 4 is a very simple function
subsets = int( f 10 / iter  ); // this is a bit more complicated

Finally, when you calculate the update, it is a multiplicative factor, so
for the volume update you have something like this:
x+ = update* x

The trick is that you can use a power of the update factors. For p=1, it is
always stable, and converges to a fix point, for p>2, it is unstable.
For 1<p<2 it is conditionally stable, but many people use larger factors,
e.g. 1.5.
x+ = update^p* x

Changing this factor as function of iteration is also useful. But it is
more stable when you have less subsets.
I often used 4 (structured, unbalanced) subsets, and 1.2-1.7 as p.
(Decreasing p as iteration number grows.)

Anyway, it is just a remark. I don't want to change the course right now,
but they are easy tricks to try,
so it wouldn't take long time to try them. (But I would add new convergence
test for this one.)

NikEfth:
I applied all comments but two. 
I left the get_offset as a weekend exercise,  and for the subsets we might need to have a small chat, about the details. 
I added and one Travis test for the lm reconstruction, with a small lm file. But the normalisation factors were 103MB and I could not upload them. If you want we might upload them to a third-party server.

KrisThielemans:
We could create a fake norm file for the test, like we do elsewhere (e.g. in examples/PET_simulation/simulate_data.sh , or a more realistic one like the randoms in the STIR exercises).

NikEfth:
From the lm file about 13.000 events are finally used. The only thing you can see in the reconstructed images is a round thing. It is just something to be compared with the proj reconstruction for validation. 
Actually, is normalisation that important?

KrisThielemans:
Agreed, norm is not important as the image is nonsense, but as a code check it is a good thing to have.

NikEfth:
Done, 
This branch has a corrections in the run_root_GATE test script, as well.

NikEfth:
P.S. 
I don't know if a squash merge would be a good idea, here. 
Because I branched my TOF stuff from this point.

recon_test_pack/run_test_listmode_recon.sh
```diff
@@ -73,6 +73,35 @@ fi
 # first delete any files remaining from a previous run
 rm -f my_*v my_*s my_*S
 
+echo "=== calculate backround data"
```
KrisThielemans: "background"

NikEfth:
All together, with the proj_data_info in the CListmode, or I should leave that there and just comment out the rest?

KrisThielemans:
no, I meant only the ROOT stuff. it's up to you really but commenting out looks to me like the easiest option...

However, I thought about another problem. All CListModeData classes now have to set `CListModeData::proj_data_info_sptr`. There's 2 problems associated with that:
- We are now breaking backwards compatibility. 
- there's actually no way that this is enforced in the code. I see now where your assert() comes from (at least it would check that), but it means it will only be flagged up at run-time.

I guess the 2nd problem already existed with scanner_sptr and indeed exam_info_sptr (Ottavia's GE stuff was initially broken because of this). Any ideas? 

The least thing we can do is to add it to the release notes to tell developers).

One drastic solution would be to replace the member with a function CListModeData::get_proj_data_info_sptr() which is pure virtual. That way, derived classes have to implement it and if they don't, the developer will know at compile time.

Also, the scanner_sptr maybe is now obsolete really. changing that problably needs a fair amount of edits, so let's just put in a comment accordingly.

sigh, always something

NikEfth:
:) There is the conflict I was expecting

KrisThielemans:
@NikEfth where are we with this PR? out-of-date? bits merged elsewhere?

NikEfth:
Hmmm, I thought that this had merged months, ago. Obviously not... I will check today. Probably out dated by now.

KrisThielemans:
@NikEfth could you check this first week of September or so?

KrisThielemans:
hi. this might not be easy as there are a lot of things here which don't seem to have anything to do with listmode. A lot of this seems out-dated. 

I'd remove changes to ROOT related input and tests from here for example (unless you're sure of course that they're needed).

Do we need changes to `CListModeData:get_proj_data_info_sptr()`?

Also, I don't understand why you want separate projector support in PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin. The name implies it only works with a ProjMatrixByBin... Obviously in principle you might want to use 2 different projectors, but in general it's a bad idea. Not sure if it's worth it (and we'd have to rename the class).

NikEfth:
The main differences I added since then was the separate projectors in order to Pre and Post smoothing. I haven't finished transferring the code, yet. I have the rest on another branch. Should I stop this ?

KrisThielemans:
ah. ok, that was a good reason to have 2 projectors. I'm not sure how to handle that then. I guess it means renaming the class (and potentially creating a derived class that has the current name which is backwards compatible, although maybe nobody is using this yet as it only works since recently).

But I'd in any case avoid changes to files that we don't want to change :-;

Issue #54, opened 2016-10-12
BUG: get_total_number_of_events() from listmode files, is not declared for CListModeDataECAT.


NikEfth:
The get_total_number_of_events() is not declared which leads to compilation error when ECAT support is activated.

PR #55, opened 2016-10-12
Fix of the aforementioned bug.

NikEfth:
Fix for bug #54


KrisThielemans:
this will be a problem for every derived class. It's another example of adding a member to the abstract class which creates trouble.

I guess best resolution is to move this implementation with an error into CListModeData as the default, state clearly in the doxygen that the default just throws an error and not require the derived class to implement it.

NikEfth:
Indeed, this is far better.

KrisThielemans:
great. add a line to the release notes please. then I accept as squashed commit?

NikEfth:
Done. 
Sure fine with me

PR #58, opened 2016-10-14
Generate Doxygen documentation with CMake

jafische:
I have added a small section into the CMakeLists.txt file such that the documentation can be built using "make doc" if enabled in CMake. That had the advantage for me that I could automatically strip all CMake include paths from the filenames in the documentation, i.e. also those in the STIR local directory. It might be handy for other people as well. I have added variables to set the momentary version of STIR as a CMake variable. The template file Doxygen.in in mainly the old Doxygen file that was included before, only that certain things are parametrized now in terms of CMake variables.
The code was inspired by this Blog entry: https://majewsky.wordpress.com/2010/08/14/tip-of-the-day-cmake-and-doxygen/.


KrisThielemans:
great. seems to work for me.

Not sure why you set
`INPUT                  = @PROJECT_SOURCE_DIR@ \
			 @PROJECT_BINARY_DIR@`
I'd expect only the SOURCE.

Should we add STIR_LOCAL to this?

I guess this is going to fail if people don't have dot. Is this easy to check? (If it isn't found, set HAVE_DOT=NO).

KrisThielemans:
oh, and could you add a line to documentation/release_3.1.htm

jafische:
Thank you for your comments. I have added the check for dot (FindDoxygen has a variable for it). By default, the CMake option for building the option is on or off depending on whether dot has been found or not. I don't know anymore why I put @PROJECT_BINARY_DIR@. I removed it and added @STIR_LOCAL@, which seems to work for me.

PR #59, opened 2016-11-04
Sensitivity Distributable Computation

jafische:
Hi Kris,
it might be easier when this is a merge request. I have checked the speed-up for my example case in the MPI case:
MPI(-np)	Sensitivity Calculation CPU Time (s)
2	2.99
4	1.09
8	0.44
The sensitivity image is exactly the same except for a a pixel wide strip at the axial edges. I do not know where this comes from.
Maybe this is helpful sometimes.
Cheers
Jannis


NikEfth:
This is a great addition, 
but I think it should be moved higher, in Poisson..ModelForMean

KrisThielemans:
The travis build failed. could you please the log files (should be doable for you by clicking on the links on the github site)

jafische:
Hi Kris,
it's a bit strange, I don't understand why it fails. Looking at the logs of the 4 cases where it fails, the problem is in all of them related to not finding the Boost library headers:

CMake Error at /usr/share/cmake-2.8/Modules/FindBoost.cmake:1202 (message):
  Unable to find the requested Boost libraries.
  Unable to find the Boost header files.  Please set BOOST_ROOT to the root
  directory containing Boost or BOOST_INCLUDEDIR to the directory containing
  Boost's headers.

Since nothing in the CMakeFiles was changed and this is on the build level independent of the actual code, I would like to just try and run it again. Who knows what happend. Or do you have an idea or have encountered something similar before?

Cheers

Jannis

jafische:
Digging further into it: it seems as Travis did not install libboost-dev-all in these builds:

---
Installing APT Packages (BETA)
Disallowing packages: libboost-all-dev, swig, libgomp1, root-system-bin
If you require these packages, please review the package approval process at: https://github.com/travis-ci/apt-package-whitelist#package-approval-process
---

which is even stranger as libboost-all-dev is in the list of approved packages for Ubuntu Precise Penguin. Can we somehow force a rerun? My theory would be that it was a hick-up of the Travis system at that time...

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -1063,6 +1133,28 @@ void RPC_process_related_viewgrams_accumulate_loglikelihood(
...
+                                                            RelatedViewgrams<float>* measured_viewgrams_ptr,
+                                                            int& count, int& count2, double* log_likelihood_ptr,
+                                                            const RelatedViewgrams<float>* additive_binwise_correction_ptr,
+                                                            const RelatedViewgrams<float>* mult_viewgrams_ptr)
+{
+
+  assert(output_image_ptr != NULL);
+  assert(measured_viewgrams_ptr != NULL);
+
+  RelatedViewgrams<float> viewgrams = measured_viewgrams_ptr->get_empty_copy();
```
KrisThielemans: this is ok but somewhat wasteful in the (most useful) case where mul_viewgrams_ptr is not 0. I suggest something like

```
if( mult_viewgrams_ptr )
    back_projector_sptr->back_project(*output_image_ptr, *mult_viewgrams_ptr);
else
{
  RelatedViewgrams<float> viewgrams = measured_viewgrams_ptr->get_empty_copy();
  viewgrams.fill(1.F);
  back_projector_sptr->back_project(*output_image_ptr, viewgrams);
}
```

KrisThielemans:
rebuild succeeded. false alarm... sorry

jafische:
Done. :)

KrisThielemans:
sorry for huge delay. I forgot about this one...

Could you add something to `documentation/release_3.1.htm`? Then it should be ready.

Do you prefer a merge or squashed-merge?

jafische:
Hi Kris,
many thanks. I uncomment in [line 61 of src/recon_buildblock/distributable.cxx](https://github.com/UCL/STIR/blob/master/src/recon_buildblock/distributable.cxx#L61). Was there a reason for it to be commented? Just to be sure...
Cheers, Jannis

KrisThielemans:
The line was uncommented as we didn't need `PoissonLogLikelihoodWithLinearModelForMeanAndProjData.h`, so no reason it include it then! I guess you do need it (and I actually thought it would need it for MPI even in the previous version).

I guess we should move that line within the `#ifdef STIR_MPI`.

KrisThielemans:
sorry. closed by accident...

jafische:
Oops, sorry about the HTML. Is fixed now. I also moved the include to the #ifdef STIR_MPI block.

jafische:
Just a thought: maybe TravisCI does not compile with MPI?

dvolgyes:
Travis should be compatible with MPI (on a single machine, but you cannot
test a cluster setup). However, the MPI libs/compiler flags should be
properly installed/set. Probably a new test environment is needed for MPI /
openmp / etc. (Just like for gcc/clang have separate environments).

I am on holiday without a proper dev environment but if it is needed, I
could help late next week.

KrisThielemans:
it'd be great to test MPI on Travis. I guess that'd be a different PR (unless Jannis wants to fix this...).
@jafische can you confirm that this version properly works with MPI? You'd need to run `make test` and the recon_test_pack (with appropriate settings, see the README).

jafische:
I did a check with the recon_test_pack under the following conditions
1) CMake: default options except STIR_MPI=ON, STIR_MPI_TIMINGS=ON, DISABLE_CERN_ROOT_SUPPRT=ON (otherwise I'd have to compile with C++11 support because of my ROOT version 6).
2) sh run_tests.sh --mpicmd "mpirun -np 8" --nointbp
Unfortunately, the sensitivity calculation is NOT good. The result seems to depend on the number of subsets used. It's almost but not entirely good with n_subsets=1. I'm trying to find out, what's the problem.

jafische:
Hi Kris, so what I could find out so far: for some reason, the iteration when using the distributable computation only runs over views 0 to 24 and segments 0 to 3 instead of 0 to 95 and -3 to 3 in the setup of the test. However, this is something managed by the distributable code, not by my additions. Do you have a hint, where the problem might lie? If not, I can dig deeper anyhow. Thanks! Jannis

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -738,6 +738,29 @@ add_subset_sensitivity(TargetT& sensitivity, const int subset_num) const
   const int min_segment_num = -this->max_segment_num_to_process;
   const int max_segment_num = this->max_segment_num_to_process;
 
+#ifdef STIR_MPI
+     TargetT* dummy_image_pointer = construct_target_ptr();
```
KrisThielemans: this will be a memory leak. Either user a `shared_ptr`, or can you just used the sensitivity image itself (I guess its values will never be used)

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -738,6 +738,29 @@ add_subset_sensitivity(TargetT& sensitivity, const int subset_num) const
   const int min_segment_num = -this->max_segment_num_to_process;
   const int max_segment_num = this->max_segment_num_to_process;
 
+#ifdef STIR_MPI
```
KrisThielemans: ideally we use the new `distributable` function always, i.e. get rid of the old code (just put it in `#if 1 ... #endif`). This will also mean support for OPENMP (and easier debugging)

src/recon_buildblock/distributable.cxx
```diff
@@ -304,6 +304,8 @@ void distributable_computation(
     task_id=task_do_distributable_loglikelihood_computation;
   else if (RPC_process_related_viewgrams == &RPC_process_related_viewgrams_gradient)
     task_id=task_do_distributable_gradient_computation;
+  else if (RPC_process_related_viewgrams == &RPC_process_related_viewgrams_sensitivity_computation)
+  task_id=task_do_distributable_sensitivity_computation;
```
KrisThielemans: clean-up indentation (use spaces) and remove next comment

KrisThielemans:
hi. I've had a look but do not see where the problem could be. Note that `distributable_computation` will only print "basic" view-segment numbers (because it's using the symmetries), which might have confused you.

Is this a problem that appeared after the merge with master, or did you not run the recon_test_pack before?

jafische:
Thanks for looking.
The recon_test_pack was not run before, I did comparisons manually, but without subsets. Problem is there before and after the merge of the current master into the devSensitivityMPI branch. The difference between the reference sensitivity image and the newly generated one is a multiplicative factor of 12, i.e. voxel-wise RPTsens_seg3_PM/my_RPTsens_seg3_PM = 12 +- 0.1 for all voxels. 

Concerning symmetries/ViewSegmentNumbers: I had the program print the numbers processed by adding a line in the for loop at https://github.com/UCL/STIR/blob/master/src/recon_buildblock/distributable.cxx#L427, i.e. vs_nums_to_process does not include the same range as the manual nested loop starting in https://github.com/UCL/STIR/blob/master/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L742.

KrisThielemans:
I think I've found the problem. `distributable_computation` will start from a zero output image, while `add_subset_sensitivity` should add to the existing data. This was a bit of a bad choice possibly, but was done to be able to accomodate the old "don't use subset sensitivities" option.

I suppose the best option would be to have a new function "compute_subset_sensitivity" which would just call your new distributable version, let `add_subset_sensitivity` call that, and only call `add_subset_sensitivity` when necessary. However, this is going to be somewhat tricky as lots of other classes use this as well.

So, I think for the moment we should do something like this.
```C++
  shared_ptr<TargetT> sensitivity_this_subset_sptr(sensitivity.clone());
  distributable_sensitivity_computation(..., *sensitivity_this_subset_sptr, ...);
  // now add result to existing data
 {
  typename TargetT::full_iterator result_iter = sensitivity.begin_all();
  typename const TargetT::full_iterator end_iter = sensitivity.end_all();
  typename TargetT::full_iterator this_subset_iter = sensitivity_this_subset_sptr->begin_all();
  while (result_iter != end_iter)
     *result_iter++ += this_subset_iter++;
  }
```

(sorry that the addition is so complicated, but we don't know what the `TargetT` will be. We should introduce some functions for this at some point).

jafische:
Hi Kris,
I fixed it with your suggestions and help. Thanks a lot! There was another problem with the zero_seg0_end_planes option: filling the viewgram with 1.F in the RPC function would override the zeroed end planes. So the sensitivity projection data is now filled in the add_subset_sensitivity(...) function. The recon_test_pack reports all tests to be OK.
Cheers, Jannis

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -738,6 +739,46 @@ add_subset_sensitivity(TargetT& sensitivity, const int subset_num) const
   const int min_segment_num = -this->max_segment_num_to_process;
   const int max_segment_num = this->max_segment_num_to_process;
 
+#if 1
+     shared_ptr<TargetT> sensitivity_this_subset_sptr(sensitivity.clone());
+
+     // have to create a ProjData object filled with 1 here because otherwise zero_seg0_endplanes will not be effective
+     shared_ptr<ProjData> sens_proj_data_sptr(new ProjDataInMemory(this->proj_data_sptr->get_exam_info_sptr(), this->proj_data_sptr->get_proj_data_info_sptr()));
+     for (int segment_num = min_segment_num; segment_num <= max_segment_num; ++segment_num)
```
KrisThielemans: all of this can just be
```
sens_proj_data_sptr->fill(1.0F);
```

jafische:
Done. I've never done a squash merge, but from what I've just read, I don't see why it wouldn't work. As you have probably seen, I replaced the iterator-stuff with one line using std::transform(...) from <algorithm>.

KrisThielemans:
great. I'll wait till Travis is fine.

Squash-merge is fine from the UCL/STIR side. The question is for you. If you have merged this with other stuff in your local repo, then a squash merge can create conflicts (when you merge UCL/STIR master back into yours). But I'll ignore that and go ahead (unless you stop me)

PR #64, opened 2016-12-21
OSX compilerflag

dvolgyes:
Travis is still running (wait with the merge until it is finished), but it looks like this is the way how OS specific compiler flags can be specified.


KrisThielemans:
superseded by PR #65

PR #67, opened 2017-01-21
Fix unsigned long 32bit problem

jafische:
Dear Kris,
sorry for this problem. Here is a fix, the unsigned long long should be at least 64bit on all architectures as defined by the C++ standard. This then allows for the bit shift and a bit field of size 48.
I was wondering if this could somehow be checked by the Travis system. My naive try to add -m32 to the compiler flags does not work because then a lot of libraries are missing...
Cheers,
Jannis


KrisThielemans:
used boost::uint64_t instead. see 80dce30fd6744ee52489caf1d766626e6d74f3cb

PR #73, opened 2017-04-10
Fix number of args in lm_generate_headcurve

bathomas:


KrisThielemans:
thanks. shall we rename this to something more sensible? what about list_lm_countrates? (sorry, should have suggested that first)

bathomas:
Changed to lm_generate_countrates as I thought it was more in keeping with the old name. Now outputs comma-separated rather than tab-separated file.

Issue #61, opened 2016-12-03
LmToProjData crash with arc-corrected template


KrisThielemans:
most (but not all) scanners have list-mode data that corresponds to non-arccorrected data. Currently the code crashes (with segmentation fault) when given a wrong template proj_data. In debug mode, this is checked with an assert resulting in

> /src/include/stir/listmode/CListEventCylindricalScannerWithDiscreteDetectors.inl:80: 
>virtual void stir::CListEventCylindricalScannerWithDiscreteDetectors::get_bin(stir::Bin&, const stir::ProjDataInfo&) const: 
>Assertion `dynamic_cast<ProjDataInfoCylindricalNoArcCorr const*>(&proj_data_info) != 0' failed.

We need to write an error message. However, this is not easy, as we probably don't want to do this check in stir::CListEventCylindricalScannerWithDiscreteDetectors::get_bin as it'd slow it down.

eliseemond:
we need to add a method to CListEvent that allows us to check if a particular template can be used or not. LmToProjData then should call that one once.

KrisThielemans:
fixed in 5b1e726925331ea26ed712ffeb5699ba894ef619

PR #78, opened 2017-06-06
Fix for LmToProjData (issue #61)-check if compatible template

eliseemond:
see #61 

Jannis (@jafische), can you check if the fix in src/include/stir/listmode/CListRecordSAFIR.h is correct?
The new method "is_valid_template", used within lm_to_projdata to check whether the input template is compatible, returns always true in the case of CListEventSAFIR as the crash should never occur.
Are you happy with this modification?

Cheers,
Elise

### DIFF
--- a/src/include/stir/listmode/CListEventCylindricalScannerWithDiscreteDetectors.h
+++ b/src/include/stir/listmode/CListEventCylindricalScannerWithDiscreteDetectors.h
@@ -70,6 +70,13 @@ public:
   */
   inline virtual void get_bin(Bin& bin, const ProjDataInfo& proj_data_info) const;
 
+  //! This method checks if the template is valid for LmToProjData
+  /*! Used before the actual processing of the data (see issue #61), before calling get_bin()
+   *  Most scanners have listmode data that correspond to non arc-corrected data and
+   *  this check avoids a crash when an unsupported template is used as input.
+   */
+  inline virtual bool is_valid_template(const ProjDataInfo&) const;
+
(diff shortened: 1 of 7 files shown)


jafische:
Hi Elise,
let me rephrase to make sure I understand it correctly: the function should return true, if the class delivers non-arc-corrected data to STIR? This is indeed true for CListRecordSAFIR as it returns the LORAs2Points with the two points being the physical crystal coordinates as specified in the crystal map.
So yes, I'm happy :).
Cheers,
Jannis

KrisThielemans:
hi, no, the function should return true if the class can support the template (as Elise I think nicely wrote in the doxygen). In most cases, the CList code can only handle non-arccorrected data (and therefore crashed when given an arccorrected template), but we're not sure if SAFIR would be able to support arccorrected data directly.

jafische:
Hi Elisa, hi Kris,

I would say the SAFIR code can support both because it's up to the template projection data ("virtual scanner") object to decide, in which bin an event will be placed. The SAFIR class used the default implementation of  stir::CListEvent::get_bin(Bin&, const ProjDataInfo &) and only implements LORAs2Points<float> stir::CListEvent::get_LOR(). This does not yet depend on whether the data is arc-corrected or not. So I guess, it depends on how stir::CListEvent::get_bin(...) behaves. @KrisThielemans: What do you think?

Cheers

Jannis

KrisThielemans:
sounds ok to me, so I'll accept this.

PR #84, opened 2017-07-11
fixes #82

### LINKED ISSUE #82: set_get_position fails for ROOT listmode files
Going back to a previous location in a ROOT file currently fails if you first read to the end of the file (bug found with @eliseemond ).

This currently probably only affects you if you use `LmToProjData` if you use `num_segments_in_memory` to use less memory. however, it also breaks the TOF unlister for ROOT (which is currently under development).

eliseemond:
With the prospect of handling TOF listmode unlisting properly, this fix adds the possibility to go back to the beginning of the ROOT listmode when the end of the file was already reached (e.g. in a TOF loop in LmToProjData where the file must be parsed as many times as the number of TOF bins).


KrisThielemans:
@NikEfth please check

PR #94, opened 2017-08-25
Fix gate support #56

NikEfth:


src/recon_buildblock/IterativeReconstruction.cxx
```diff
@@ -397,16 +397,16 @@ reconstruct()
 {
   this->start_timers();
 
-  this->target_data_sptr.reset(this->get_initial_data_ptr());
-  if (this->set_up(this->target_data_sptr) == Succeeded::no)
+  shared_ptr<TargetT > target_data_sptr(this->get_initial_data_ptr());
```
KrisThielemans: why do we do this? It's normally using `Reconstruct::target_data_sptr`. Changing this to use a local variable is scary.

src/recon_test/recontest.cxx
```diff
@@ -71,24 +70,21 @@ int main(int argc, const char *argv[])
...
     HighResWallClockTimer t;
     t.reset();
     t.start();
 
-
     if (reconstruction_method_sptr->reconstruct() == Succeeded::yes)
     {
         t.stop();
         std::cout << "Total Wall clock time: " << t.value() << " seconds" << std::endl;
+        return Succeeded::yes;
```
KrisThielemans: a `main()` function has to return `EXIT_SUCCESS`

KrisThielemans:
By the way, I cannot understand why the original code occasionally worked... the test should always have failed for a root file, so we should never have been able to read them. @NikEfth, do you understand this?

src/listmode_buildblock/CListModeDataROOT.cxx
```diff
@@ -156,7 +156,7 @@ open_lm_file()
         t.read(mem,4);
         ss << mem;
 
-        if ( !sig.compare(ss.str()) )
+        if (sig.compare(ss.str()) )
```
KrisThielemans: actually, I think we need
```C++
sig.compare(ss.str())!=0
```
Presumably the previous code didn't work due to strange behaviour of ! on the return value of `std::string::compare`

src/recon_buildblock/IterativeReconstruction.cxx
NikEfth: I tried to track this down and goes back to 13/07/2016. It was part of two commits 2d4e3067f32542a5e8ec297260ece44b832c7b76 and efed646d94b4ab030366f6c7170146c1126b5f12 to be able to reconstruct without output.

NikEfth: I cannot remember how this didn't come to UCL/STIR as back then we had accepted this, as part of the scatter correction

src/recon_test/recontest.cxx
NikEfth: I will correct this when we see what will happen with the rest code

src/recon_buildblock/IterativeReconstruction.cxx
NikEfth: I could revert them both if you're concerned.

src/listmode_buildblock/CListModeDataROOT.cxx
NikEfth: This is a very strange error. I never had a problem.

KrisThielemans:
Dropping this PR. Will fix #56 elsewhere. Other commits need to be in a separate PR.

PR #101, opened 2017-08-30
various small bug and warning fixes.

dvolgyes:
I found a few things which should be fixed.
- Mostly incorrect format string (int vs. unsigned),
- but there is a nasty missing 'break' and a similar missing comma,
which caused an incorrect array initialization. 
- and a few missing constructor initialization parts.

### DIFF
--- a/src/listmode_utilities/lm_fansums.cxx
+++ b/src/listmode_utilities/lm_fansums.cxx
@@ -102,6 +102,7 @@ set_defaults()
   store_prompts = true;
   delayed_increment = -1;
   interactive=false;
+  current_frame=0;
 }
 
 void 
@@ -292,7 +293,7 @@ write_fan_sums(const Array<2,float>& data_fan_sums,
                const unsigned current_frame_num) const
 {
   char txt[50];
-  sprintf(txt, "_f%d.dat", current_frame_num);
+  sprintf(txt, "_f%u.dat", current_frame_num);
   std::string filename = output_filename_prefix;
   filename += txt;
   ofstream out(filename.c_str());
--- a/src/listmode_utilities/lm_fansums.cxx
+++ b/src/listmode_utilities/lm_fansums.cxx
@@ -85,7 +85,6 @@ private:
   std::string frame_definition_filename;
   bool store_prompts;
   int delayed_increment;
-  int current_frame;
 
   bool interactive;
 
@@ -102,7 +101,6 @@ set_defaults()
   store_prompts = true;
   delayed_increment = -1;
   interactive=false;
-  current_frame=0;
 }
 
 void 
--- a/src/local/recon_buildblock/ProjMatrixByBinWithPositronRange.cxx
+++ b/src/local/recon_buildblock/ProjMatrixByBinWithPositronRange.cxx
@@ -159,7 +159,7 @@ calculate_proj_matrix_elems_for_one_bin(
   const Bin bin = lor.get_bin();
 
   assert(lor.size() == 0);
-     
+
   ProjMatrixElemsForOneBin zoomed_lor(bin);
   ProjMatrixElemsForOneBin zoomed_lor(bin);
 
@@ -170,10 +170,9 @@ calculate_proj_matrix_elems_for_one_bin(
        ++iter)
     {
       ProjMatrixElemsForOneBinValue
-      
+    }
 }
 
-         
 
 END_NAMESPACE_STIR
 
--- a/src/local/recon_buildblock/ProjMatrixByBinWithPositronRange.cxx
+++ b/src/local/recon_buildblock/ProjMatrixByBinWithPositronRange.cxx
@@ -28,6 +28,7 @@
 #include "stir/IndexRange.h"
 #include <algorithm>
 #include <math.h>
+#include <boost/static_assert.hpp>
 
 START_NAMESPACE_STIR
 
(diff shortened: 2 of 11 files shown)


src/listmode_utilities/lm_fansums.cxx
```diff
@@ -102,6 +102,7 @@ set_defaults()
   store_prompts = true;
   delayed_increment = -1;
   interactive=false;
+  current_frame=0;
```
KrisThielemans: this actually seems a variable that's never used. let's remove it instead.

src/local/recon_buildblock/ProjMatrixByBinWithPositronRange.cxx
```diff
@@ -159,7 +159,7 @@ calculate_proj_matrix_elems_for_one_bin(
   const Bin bin = lor.get_bin();
 
   assert(lor.size() == 0);
-     
+
```
KrisThielemans: this file actually is incomplete. the crucial code isn't there (I started it in 2004 apparently). let's not modify this file therefore.

src/listmode_utilities/lm_fansums.cxx
dvolgyes: With this one I agree, I will make it disappear soon.

src/local/recon_buildblock/ProjMatrixByBinWithPositronRange.cxx
dvolgyes: I would argue against this one. I really would like to get rid of the code completely, or make it compilable, even if it is incomplete. (We could insert an assert(false) to make sure nobody uses it before major revision.)
The reason is that I would like to gradually rid of the syntactical errors, and introduce a static code analyzer to Travis, e.g. cppcheck. This code analyzers parse all the C++ files, and find nasty bugs, like the "break" in this PR, and I would like to avoid maintaining a list of excluded files.

So I don't mind to add a assert(false), but I would like to get this code compilable. :)

KrisThielemans: ok, but an assert wouldn't be good enough for me...

I don't want to get rid of it, as it's contributed by GE. Recovering it will be harder.

dvolgyes: By the way, the same static analyzes found a few strange asserts, like:
```assert(-1 >> 1 == -1);```
(source: src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx:176, among others.)

- First of all, do we really need negative shifts? (C/C++ standards say it is undefined behavior.)
- If yes, shouldn't be this in the CMake conditions somewhere?
  It depends on the compiler, and it should complain at configuration time.
- or worst case: boost static assert, e.g.
  ```BOOST_STATIC_ASSERT( -1 >> 1 == -1);```

   (It actually works, e.g.  BOOST_STATIC_ASSERT( -1 >> 1 == -2); cause compilation error.)

Well, I understand that some asserts are useful, but I would argue for more static asserts and less runtime asserts in general. (E.g. there are asserts about type sizes, etc., These are the prime examples for boost static asserts.)

In this particular case, it feels bad, and it is really a portability issue. (Anytime, any compiler can change this, and it is a smaller kind of miracle that most compilers do the same thing at this very moment.)

dvolgyes: GE: I understand this. In git, it is not too hard to recover, but yes, it is a bit more challenging. 
And I respect that you don't like assert(false). What do you recommend?
- leave it that way, and make an exception list?
- or something else, e.g. boost static assert? (which makes it syntactically correct but incompilable)
- or a longer remark around the (static) assert stating that it is incomplete, and before use one should finish the code?
- or something else?

KrisThielemans: that bit of code should be changed of course, but I don't have the time (and it's a bit scary).

I'm not 100% sure that BOOST_STATIC_ASSERT will do the same thing as a run-time assert, but if it does, then I'm happy to do that. 

a CMake test would be better but even more effort sadly. not sure how much mileage you have for this kind of stuff

dvolgyes: (Remark: the whitespace change in this file is irrelevant, I can make it disappear. What I argue for is an extra }., but it is a few lines below in the same file, see the next edit.)

KrisThielemans: the file isn't in the CMakeLists.txt so I'm not 100% sure if it matters really if it's incompilable. a boost static assert seems fine to me.

it's rather low down my priority list I'm afraid.

dvolgyes: I made an update, hopefully it is a good compromise.
CMakelists.txt: it matters for me because I would like to set up static analyzer which doesn't use cmakelists.txt. It is a medium term plan, I would like to do it around/after MPI is included in Travis. But I need to fix syntax errors for this.

dvolgyes: I missed one of your comments, and I think you also thought I want to change something else than I did. So the current code reflects my current idea about the file. 
Basically, I would give an extra } sign, and an extra static assert with comment, but I would keep everything else intact, including the runtime assert line.

PR #102, opened 2017-09-01
little bit of code modernization

dvolgyes:
Most part of the PR is code modernization.

First of all, vsnprintf is supported since VS 2005, and C99/C++11 compilers must implement it, 
so it is time to drop the ```#ifdef``` workarounds.

Second, some asserts could and should be replaced with compile time checks.

Third: "\n" should be replaced with std::endl. I didn't do it everywhere, there are 3.5 thousand such lines,
but it least I fixed them in the function I already edited for other reasons. 
(But if somebody has patience, there are gems like this: ```"...filename>\n"<<endl;```)

Fourth: M_PI, and similar constants shouldn't be defined locally but rather take it from a library. Visual Studio has/had issues with cmath, so let's use BOOST.
(Another big part should be: checking the arithmetic functions, especially atan, atan2 and acos, because their use is quite strange sometimes, but this remains a task for the future.)

Finally, there is a tiny bug about variadic function parameters: the va_start macro should have a va_end pair, but it is missing from one of the files.

### DIFF
--- a/src/buildblock/error.cxx
+++ b/src/buildblock/error.cxx
@@ -53,6 +53,8 @@ void error(const char *const s, ...)
   char tmp[size];
   const int returned_size= vsnprintf(tmp,size, s, ap);
   std::stringstream ss;
+  va_end(ap);
+
   if (returned_size<0)
 	  ss << "\nERROR: but error formatting error message" << std::endl;
   else
--- a/src/buildblock/warning.cxx
+++ b/src/buildblock/warning.cxx
@@ -53,6 +53,7 @@ void warning(const char *const s, ...)
   char tmp[size];
   const int returned_size= vsnprintf(tmp,size, s, ap);
   std::stringstream ss;
+  va_end(ap);
   if (returned_size < 0)
 	  ss << "\nWARNING: error formatting warning message" << std::endl;
   else
--- a/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx
+++ b/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx
@@ -40,6 +40,8 @@
 #include <sstream>
 #endif
 
+#include <boost/static_assert.hpp>
+
 #ifndef STIR_NO_NAMESPACES
 using std::endl;
 using std::ends;
@@ -141,8 +143,6 @@ ProjDataInfoCylindricalNoArcCorr::parameter_info()  const
 }
 
 /*
-   TODO make compile time assert
-
    Warning:
    this code makes use of an implementation dependent feature:
    bit shifting negative ints to the right.
@@ -151,6 +151,8 @@ ProjDataInfoCylindricalNoArcCorr::parameter_info()  const
    This is ok on SUNs (gcc, but probably SUNs cc as well), Parsytec (gcc),
    Pentium (gcc, VC++) and probably every other system which uses
    the 2-complement convention.
+
+   Update: compile time assert is implemented.
 */
 
 /*!
@@ -173,8 +175,8 @@ void
 ProjDataInfoCylindricalNoArcCorr::
 initialise_uncompressed_view_tangpos_to_det1det2() const
 {
-  assert(-1 >> 1 == -1);
-  assert(-2 >> 1 == -1);
+  BOOST_STATIC_ASSERT(-1 >> 1 == -1);
+  BOOST_STATIC_ASSERT(-2 >> 1 == -1);
 
   const int num_detectors =
     get_scanner_ptr()->get_num_detectors_per_ring();
@@ -220,8 +222,8 @@ void
 ProjDataInfoCylindricalNoArcCorr::
 initialise_det1det2_to_uncompressed_view_tangpos() const
 {
-  assert(-1 >> 1 == -1);
-  assert(-2 >> 1 == -1);
+  BOOST_STATIC_ASSERT(-1 >> 1 == -1);
+  BOOST_STATIC_ASSERT(-2 >> 1 == -1);
 
   const int num_detectors =
     get_scanner_ptr()->get_num_detectors_per_ring();
--- a/src/include/stir/common.h
+++ b/src/include/stir/common.h
@@ -103,7 +103,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <cmath>
-
+#include <boost/math/constants/constants.hpp>
 
 //*************** namespace macros
 #ifndef STIR_NO_NAMESPACES
@@ -295,7 +295,7 @@ START_NAMESPACE_STIR
 
 //! The constant pi to high precision.
 /*! \ingroup buildblock */
-const double _PI = 3.14159265358979323846264338327950288419716939937510;
+const double _PI = boost::math::constants::pi<double>();
 
 //! returns the square of a number, templated.
 /*! \ingroup buildblock */
--- a/src/recon_buildblock/SPECTUB_Tools.cxx
+++ b/src/recon_buildblock/SPECTUB_Tools.cxx
@@ -27,6 +27,7 @@
 #include <boost/math/special_functions/fpclassify.hpp>
 #include "stir/error.h"
 #include <boost/format.hpp>
+#include <boost/math/constants/constants.hpp>
 
 using namespace std;
 //using std::string;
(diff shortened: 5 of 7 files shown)


src/buildblock/error.cxx
```diff
@@ -36,13 +36,6 @@
 
 #include "stir/TextWriter.h"
 
-/* Warning: vsnprintf is only ISO C99. So your compiler might not have it.
-   Visual Studio can be accomodated with the following work-around
-*/
-#ifdef BOOST_MSVC
```
KrisThielemans: you want like it but can we just use
```
#if defined(BOOST_MSVC) && (BOOST_MSVC<1400)
```
I have no clue if VS 2003 would still work, but I don't want to introduce a largely unnecessary versoin-dependency until we jump to C++-11.

src/buildblock/error.cxx
```diff
@@ -53,13 +46,15 @@ void error(const char *const s, ...)
   char tmp[size];
   const int returned_size= vsnprintf(tmp,size, s, ap);
   std::stringstream ss;
+  va_end(ap);
+
   if (returned_size<0)
-	  ss << "\nERROR: but error formatting error message" << std::endl;
+	  ss << std::endl << "ERROR: but error formatting error message" << std::endl;
```
KrisThielemans: I disagree with this. see overall comment

src/buildblock/warning.cxx
```diff
@@ -36,13 +36,6 @@
 
 #include "stir/TextWriter.h"
 
-/* Warning: vsnprintf is only ISO C99. So your compiler might not have it.
-   Visual Studio can be accomodated with the following work-around
-*/
-#ifdef BOOST_MSVC
```
KrisThielemans: same of course

dvolgyes:
The best solution for me if we get rid of error/warning sooner or later. As I see all the disagreement is about these two, so I just remove this commit from the PR.

dvolgyes:
@KrisThielemans, do you see any unresolved concern with this PR?
I have nothing more to contribute to this one, so I propose merging it.

KrisThielemans:
@dvolgyes . I've had a report of @eliseemond (using Eclipse user on Ubuntu. sorry, no version info right now) with the following error message
```
/usr/include/boost/math/constants/constants.hpp:300:3: note: use -std=gnu++11 or -fext-numeric-literals to enable more built-in suffixes
/usr/include/boost/math/constants/constants.hpp: In static member function 'static constexpr T boost::math::constants::detail::constant_root_two_pi<T>::get(const mpl_::int_<5>&)':
/usr/include/boost/math/constants/constants.hpp:301:3: error: unable to find numeric literal operator 'operator""Q'
   BOOST_DEFINE_MATH_CONSTANT(root_two_pi, 2.506628274631000502415765284811045253e+00, "2.50662827463100050241576528481104525300698674060993831662992357634229365460784197494659583837805726611600997267e+00")
```
This is with gcc (no version info yet) with `-std=c++11`. Any ideas?

dvolgyes:
It's really strange.
I will try to figure out something but
I would say that boost is broken somehow, maybe two versions are installed
at the same time or something similar.
What would be useful:
architecture (32bit/64), OS version, boost version. In worst case we can
calculate it from Pi (which was used in previous lines, therefore should
work.)
But it is really strange,  have never seen this before.

KrisThielemans:
not much to find on the internet but others do see this as well, e.g. https://gist.github.com/mikaelarguedas/7731b4100e7ed78b6aab73f052ffac79

If the root_two_pi is the only problem (but I don't think so), I'd suggest to just replace `boost::math::constants::root_two_pi<float>()` with `sqrt(2*boost::math::constants::pi<float>())`. @eliseemond, can you try this at some point? That would be only in `src/recon_buildblock/SPECTUB_Weight3d.cxx` I think (but I though you had more error messages).

eliseemond:
Hi there,
Replacing boost::math::constants::root_two_pi<float>() with sqrt(2*boost::math::constants::pi<float>()) did not change anything. However, compiling with -std=gnu++11 instead of -std=c++11 solved my compilation issues. Would it be because of some gnu extensions used somewhere? https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Extensions.html

Cheers,

Elise

NikEfth:
@KrisThielemans there is the todo to change the erf function with boost::erf. Should I do this?

dvolgyes:
Maybe it is gnu extension, but it is annoying.
One of the reasons for boost is to avoid compliler/platform specific issues.
And in general, it is quite OK.
Could you detail your setup?
compiler version, os version and architecture (32/64bit, x86/arm/etc) boost
version?

David

eliseemond:
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609
boost 1.58

Eclipse IDE for C/C++ Developers
Version: Oxygen Release (4.7.0)
Build id: 20170620-1800

Architecture: x86_64

KrisThielemans:
This would be a boost bug then. See https://svn.boost.org/trac10/ticket/9240. Claimed to be fixed 4 years ago, not sure in which version.

It probably doesn't show up on Travis because CMake enables `gnu+11` if asking for C++11 (which I find crazy).

I guess Elise could upgrade her boost version to see if it disappears, but I'm not so happy to introduce a STIR dependency on this.

@eliseemond, could you clarify which `boost::math::const`  this creates an error? (probably all of them)

KrisThielemans:
@NikEfth, how old is `boost::erf`? low priority?

NikEfth:
I see that is around from 1.46, could be earlier. I need the inverse erf to test for normality of the timing.

KrisThielemans:
turns out `erf` was introduced in 1.35.0 http://www.boost.org/doc/libs/1_35_0/libs/math/doc/html/index.html so if you need it, I'm fine with introducing it (as our current minimum boost is 1.36.0). This needs a separate PR of course (branched off master). The scatter stuff will need it, so that's going to be a pain given the different branches. So please keep `stir/numerics/erf.h` and have inline implementations of `stir::erf` and `stir::erfc`.

eliseemond:
@KrisThielemans @dvolgyes 

You can have a look at the entire build log for the boost errors with -std=c++11.
I have updated my boost libraries to the latest version (1.65.1) and tried to compile with -std=c++11 again. Still same compilation error. Same thing happened when I updated gcc. It might be something with cmake? (I have cmake 3.5.1)

[build_log.txt](https://github.com/UCL/STIR/files/1308203/build_log.txt)

KrisThielemans:
nothing to do with CMake. that just provides some options to your compilers. Even if they're a bit funny, it still shouldn't break. To me this is definitely a boost bug (somewhat amazing).

Could you check what compiling the following does
```
#include <boost/math/constants/constants.hpp>

float f{}
{ 
  const float K0 = 1.0f/boost::math::constants::root_two_pi<float>(); 
  return K0*2;
}
```
something like
```
g++ -c -std=c++11 test.cxx
```
If it works, try to find out what compile options are used for g++ for your STIR (not sure how that do that in Eclipse, as `CMAKE_CXX_FLAGS` might be incomplete). if you have makefiles, you can do
```
make VERBOSE=On
```

eliseemond:
It did compile yes. 
Actually the verbose was already enabled in the previous log (hard to see with so many errors though). I have added the cmake verbose. 

/usr/bin/c++    -I/home/eemond/root/include -I/home/eemond/UCL-STIR_ubuntu/build/src/include -I/mnt/c/Users/azome/UCL-STIR_test/src/include  -std=c++11 -O3 -DNDEBUG -ffast-math   -Wall -Wno-deprecated -std=gnu++11 -o CMakeFiles/buildblock.dir/Array.cxx.o -c /mnt/c/Users/azome/UCL-STIR_test/src/buildblock/Array.cxx

As you can see, both "-std=c++11" and "-std=gnu++11" are there. Weird

[testlog2.txt](https://github.com/UCL/STIR/files/1308325/testlog2.txt)

KrisThielemans:
both options seems to be what was on the link as well. You'll have it because you added `c++11` by hand (`CMAKE_CXX_FLAGS`), but current STIR `CMakeLists.txt` will add `gnu++11` if ROOT 6 is detected.

So, try to have all of those options for the test programme, and if it fails, cut away the cruft.

but not during the weekend...

eliseemond:
`g++ -c -std=c++11 -std=gnu++11 test.cxx`
is indeed not compiling.

I did not realise that CMakeLists.txt changed and that I did not have to manually add the c++11 flag.

PR #109, opened 2017-09-19
Fix #87

### LINKED ISSUE #87: SAFIRCListmodeInputFileFormat::can_read should not write errors
At present, when a non-SAFIR listmode file is used, `SAFIRCListmodeInputFileFormat::can_read` will write warnings/error messages when it tries to parse the non-SAFIR listmode file.

suggested solution: modify `actual_do_parsing` to
```
bool actual_do_parsing(std::istream& input, bool write_warning = true)
```
and pass it to the `parse` call. (Obviously for all signatures). Then call it appropriately in `can_read`.

@jafische could you have a look?

jafische:
Hi Kris,
I have fixed the can_read issue. I have at the same time removed the ability to call can_read() with an ifstream as argument --> always returns false now as in the ECAT file format libraries. This makes it a bit cleaner and does not impair functionality as we probably always call the parser with a filename of the parameter file.
Your original suggestion did not work, because a) ParsingObject does not have the parse(...,write_warning) syntax, only KeyParser does, and b) Even KeyParser::parse(..., false) will print the warning in question (https://github.com/UCL/STIR/blob/master/src/buildblock/KeyParser.cxx#L270).
The code now checks for the CListModeDataSAFIR Parameter at the beginning of the file, whitespace is ignored. If this is successful, the file is parsed and the file signature in the binary file checked as done before.
Cheers, Jannis

### DIFF
--- a/src/include/stir/IO/SAFIRCListmodeInputFileFormat.h
+++ b/src/include/stir/IO/SAFIRCListmodeInputFileFormat.h
@@ -31,6 +31,7 @@
 #ifndef __stir_IO_SAFIRCListmodeInputFileFormat_H__
 #define __stir_IO_SAFIRCListmodeInputFileFormat_H__
 
+#include <algorithm>
 #include <cstring>
 #include <string>
 #include <iostream>
@@ -85,18 +86,18 @@ public:
 	//! Checks in binary data file for correct signature (can be either "SAFIR CListModeData" or "MUPET CListModeData").
 	virtual bool can_read( const FileSignature& signature, const std::string& filename) const
 	{
-		char* buffer = new char[20];
-		
-		std::ifstream par_file(filename.c_str(), std::ios::binary);
-		par_file.read(buffer, 20);
-		if( strncmp(buffer, "CListModeDataSAFIR", 18) ) { 
-			delete[] buffer;
+		// Looking for the right key in the parameter file
+		std::ifstream par_file(filename.c_str());
+		std::string key;
+		std::getline(par_file, key, ':');
+		key.erase(std::remove_if(key.begin(), key.end(), is_whitespace), key.end());
+		if( !boost::iequals(key, std::string("CListModeDataSAFIRParameters")) ) { 
 			return false;
 		}
 
 		bool can_parse = actual_do_parsing(filename);
 		std::ifstream data_file(listmode_filename.c_str(), std::ios::binary);
-		buffer = new char[32];
+		char* buffer = new char[32];
 		data_file.read(buffer, 32);
 		bool cr = (!strncmp(buffer, "MUPET CListModeData\0", 20) ||  !strncmp(buffer, "SAFIR CListModeData\0", 20)) && can_parse;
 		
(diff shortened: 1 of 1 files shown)


src/include/stir/IO/SAFIRCListmodeInputFileFormat.h
```diff
@@ -77,43 +80,36 @@ class SAFIRCListmodeInputFileFormat : public InputFileFormat<CListModeData>, pub
...
+		std::ifstream par_file(filename.c_str());
+		std::string key;
+		std::getline(par_file, key, ':');
+		key.erase(std::remove_if(key.begin(), key.end(), is_whitespace), key.end());
+		if( !boost::iequals(key, std::string("CListModeDataSAFIRParameters")) ) { 
+			return false;
+		}
+
 		bool can_parse = actual_do_parsing(filename);
 		std::ifstream data_file(listmode_filename.c_str(), std::ios::binary);
```
KrisThielemans: seems you should get out if parsing failed. otherwise who knows what you'll be reading.

KrisThielemans:
Windows builds on Appveyor are failing with
```
  C:\projects\stir\src\include\stir/IO/SAFIRCListmodeInputFileFormat.h(179): error C2672: 'std::isspace': no matching overloaded function found 
  C:\projects\stir\src\include\stir/IO/SAFIRCListmodeInputFileFormat.h(179): error C2780: 'bool std::isspace(_Elem,const std::locale &)': expects 2 arguments - 1 provided 
```
this is a bit weird given the Microsoft doc https://msdn.microsoft.com/en-us/library/y13z34da(v=vs.120).aspx but I'd rather not dig into this. 

Probably best is to use `standardise_interfile_keyword` (in `stir/interfile_keyword_functions.h`). (Or, as I originally suggested, to let it do the parsing silently when doing the check, but that'd require more work).

src/include/stir/IO/SAFIRCListmodeInputFileFormat.h
```diff
@@ -90,16 +90,17 @@ class SAFIRCListmodeInputFileFormat : public InputFileFormat<CListModeData>, pub
...
 			return false;
 		}
 
 		bool can_parse = actual_do_parsing(filename);
+		if( !can_parse ) return false;
 		std::ifstream data_file(listmode_filename.c_str(), std::ios::binary);
 		char* buffer = new char[32];
 		data_file.read(buffer, 32);
-		bool cr = (!strncmp(buffer, "MUPET CListModeData\0", 20) ||  !strncmp(buffer, "SAFIR CListModeData\0", 20)) && can_parse;
+		bool cr = (!strncmp(buffer, "MUPET CListModeData\0", 20) ||  !strncmp(buffer, "SAFIR CListModeData\0", 20));
```
KrisThielemans: seems to me that this could use writing 
```
warning(boost::format("SAFIR tried to read file '%1%' but this seems to be in the wrong format") % listmode_filename);
```
otherwise it will be a silent warning and people will get very puzzled.

if you think this is worth adding, please test it on a file first (as I've had that the warning itself crashed because I mistyped the format string).

PR #120, opened 2017-11-02
Add missing include

tokkot:
With boost 1.63.0, this file won't compile without including boost/format.hpp.


KrisThielemans:
thanks. multiple PRs for the same issue. I happened to accept PR #124 first.

Issue #122, opened 2017-11-07
Compilation fails with boost 1.63 on boost::format


NikEfth:
Probably `boost::format` has changed location and it has to be included.

KrisThielemans:
closed with #124

PR #123, opened 2017-11-07
Fix on issue#122

NikEfth:
FIxes issue #122


KrisThielemans:
thanks. as before, there's too much being dragged-in here. I guess the only change we need is in the `test*cxx`?

NikEfth:
sigh .... No in generalised objective function. I will clean my repo at the weekend to loose all the the history. Maybe re-clone. 
Actually I am not sure what should I do. How can I keep in sync with the UCL/master without merging to the upstream which will end up in the history.

NikEfth:
In addition, please note that I cannot see this history on any other page than this. Not even when I create the PR.

KrisThielemans:
hi, a full git tutorial is out-of-scope here. However, the trick for this kind of small stuff is to branch-off from UCL/master for the PR. You put the fix on the branch. Then, you create the PR, and either merge that branch straight on your development stuff, or you wait till I accept. The wait can be a pain of course. However, if the PR is really small, it should be not too long :-)

sounds easy... of course, for other stuff we're talking much longer development cycles, and then it becomes much trickier. Still, trying to keep a "one feature per branch" is I think the best. You can then merge lots of feature-branches togetehr to actually do something useful.

So, please close this PR and generate a new one.

PR #127, opened 2017-12-11
improve Python test infractructure

KrisThielemans:


KrisThielemans:
Currently getting
```
../install/python/stir.py:28: in <module>
>       _stir = swig_import_helper()
../install/python/stir.py:24: in swig_import_helper
>               _mod = imp.load_module('_stir', fp, pathname, description)
E               ImportError: dynamic module does not define init function (init_stir)
```
This might have to do with [wrong naming of the shared library](https://stackoverflow.com/questions/35947972/swig-setup-py-importerror-dynamic-module-does-not-define-init-function-init) or so, but it happens only on Travis.

KrisThielemans:
now a syntax in a cleaner .travis.yml.
```
/home/travis/.travis/job_stages: eval: line 72: syntax error near unexpected token `fi'
/home/travis/.travis/job_stages: eval: line 72: `  fi'
```
@casperdcl, I tried to follow your scheme but apparently failed. suggestions?

KrisThielemans:
The above error was caused because STIR was built with `python3` but the travis file used `python` (version 2)

PR #130, opened 2017-12-20
ViewOffset Support Enabled

pw351:
Added View offset in ProjDataInfoCylindrical and ProjDataInfoCylindricalNoArcCorr.

DataSymmetries in phi turned off for scanners with view offset.


pw351:
This works in simple reconstructions to have a view offset, but it seems a lot of changes are still needed to be made in LORCoordinates class to have interchangeability in different types of LOR coordinates available.

src/test/test_proj_data_info.cxx
```diff
@@ -168,19 +169,19 @@ test_generic_proj_data_info(ProjDataInfo& proj_data_info)
 			  if (diff_tangential_pos_num>max_diff_tangential_pos_num)
 			    max_diff_tangential_pos_num=diff_tangential_pos_num;
 			}
-		      if (!check(org_bin.get_bin_value() == new_bin.get_bin_value(), "round-trip get_LOR then get_bin: value") ||
```
KrisThielemans: why are all these commented out?

src/test/test_proj_data_info.cxx
```diff
@@ -215,11 +216,11 @@ test_generic_proj_data_info(ProjDataInfo& proj_data_info)
 			  if (diff_tangential_pos_num>max_diff_tangential_pos_num)
 			    max_diff_tangential_pos_num=diff_tangential_pos_num;
 			}
-		      if (!check(org_bin.get_bin_value() == new_bin.get_bin_value(), "round-trip get_LOR then get_bin (LORAs2Points): value") ||
```
KrisThielemans: these are commented out as well?

KrisThielemans:
I'm not sure why there would need to be changes in LORCoordinates. Those are in a continous coordinate system, where phi=0 is well defined. I would have guessed this needs zero changes. (We could modify the classes to allow a non-zero phi-offset, but I wouldn't know why that would be useful).

src/test/test_proj_data_info.cxx
pw351: These were commented out because when view offset was introduced the bin view numbers and so on changes, so org_bin is no longer same as new_bin. I am not sure how to address this, but in next request, I have disabled this test it only for non-zero view offset scanners. Probably need to review this test.

src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx
```diff
@@ -21,7 +21,7 @@
   \file
   \ingroup projdata
 
-  \brief Implementation of non-inline functions of class 
+  \brief Implementation of non-inline functions of class
```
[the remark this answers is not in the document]
pw351: This file is now consistent with new code (master using BOOST_STATIC_ASSERT etc)

src/recon_buildblock/DataSymmetriesForBins_PET_CartesianGrid.cxx
```diff
@@ -40,16 +40,16 @@ START_NAMESPACE_STIR
 
 //! find correspondence between axial_pos_num and image coordinates
 /*! z = num_planes_per_axial_pos * axial_pos_num + axial_pos_to_z_offset
-   compute the offset by matching up the centre of the scanner
```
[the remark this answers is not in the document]
pw351: How should I revert these white spaces.
For some reason, it shows I have modified and added these spaces, whereas I did not, not sure how to go around this.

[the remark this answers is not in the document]
KrisThielemans: The problem is likely your editor that wants to be smart. Sadly, the original file had somewhat inconsistent white-space settings, and spaces at the end of the line etc. Your editor is cleaning that up. That's normally great, but in this case it generates a lot of changes such that we don't see anymore what is the real change.

You can do a `git diff -w` to check what the actual difference is. https://github.com/UCL/STIR/pull/130/files?w=1 claims that there are really only very few. That'd make me happy :-)

I'm afraid the trick would then be to use a stupid editor (or a smart one where you can disable white-space changes), get the original file, edit it to put the changes again. The painful thing is of course to be sure that you do the correct thing...

sorry.

src/test/test_proj_data_info.cxx
```diff
@@ -111,6 +111,9 @@ test_generic_proj_data_info(ProjDataInfo& proj_data_info)
 #ifdef STIR_OPENMP
 #pragma omp parallel for schedule(dynamic)
 #endif
+  float view_offset=proj_data_info.get_scanner_ptr()->get_default_intrinsic_tilt();
```
KrisThielemans: are you really sure that you need to disable this test? It makes me feel very uncomfortable. I think with your current changes in the other files, it should work... If it doesn't, we probably need to fix something.

By the way, you'd have to keep the `pragma omp` just above the `for` loop of course.

pw351: I am also a bit puzzled about this. In the test for a given bin.view_num, get_LOR calculates phi which should be phi+view_offset. Then the second part of the test uses get_bin to return the value of bin.view_offset from that phi which should give the actual value. But somewhere it seems this breaks as test_proj_data_info gives the following error:
 
 Error. round-trip get_LOR then get_bin (LORAs2Points): segment
	Problem at    segment = 15, axial pos 0, view = 189, tangential_pos_num = 30
	round-trip to segment = -15, axial pos 0, view = -3, tangential_pos_num = -30
Max Deviation:  segment = 30, axial pos 0, view = 192, tangential_pos_num = 62

I am unsure why this segment swap occurs with the current code.

pw351:
Nikos, yes I think that could be a way to go about. I did merge it to
master first according to best of my knowledge.
But anyways, I will revert back.

KrisThielemans:
hi
I really don't know what happened here. It seems that PR is now reverting a lot of the updates that we made since it was created. Maybe a merge went wrong. An alternative explanation is that github shows us the changes w.r.t. current master (as opposed to the changes from when you branched off). Maybe @casperdcl can tell us if that's the case. If it's the latter, another merge could clear this up. You'd need to have an "upstream" remote (set to UCL/STIR), fetch it, and then merge upstream/master to your branch. (I guess a rebase would be clearner but as you're merging all this branch elsehwere, that would create havoc, I'm sure).

sorry for the trouble.

casperdcl:
yes PRs always compares branches, not commits, so it's always against the latest `master`. Usually if there are conflicts the PR won't say "This branch has no conflicts with the base branch" as it currently (oddly) does.

KrisThielemans:
right. let's merge UCL/STIR/master on here then and have a look before we decide to throw it all away. @pw351, can you do that?

casperdcl:
To be clearer, I'm pretty sure this is "a merge gone wrong", i.e. `master` was merged in this branch, discarding `master`'s changes. That would make git & github believe this branch supercedes and has no conflicts with master.

Looking at the commit history, it'll be a pain to fix.

KrisThielemans:
I guess that'd be fastest. I'd prefer though that you commit your changes, and then @NikEfth redoes his. It clarifies ownership, but also makes my life easier as it allows me to see what fixes he made.

(You could potentially do this by cherry-picking commits, but somewhat easy to make mistakes that way, unles your name is Casper!)

@NikEfth, if you redo this, would it be possible to do it with an editor that doesn't change white-space. (I know, it's still on our to do list...)

thanks

NikEfth:
@KrisThielemans  Sure, I will try another editor.

PR #136, opened 2018-01-12
Changed all comparisons with null pointers to is_null_ptr()

rijobro:
Comparisons with == 0 etc. caused errors on OSX


KrisThielemans:
Appveyor failure seems just a time-out. Code is fine, so we'll merge.

PR #140, opened 2018-01-29
class FilePath

NikEfth:
This class will provide a cross-platform interface for very simple path manipulation.
Initially, the only new function will be append(Path) which will create a new path. 
Few extras have been added like checks if the path is a directory and if it is writable. 
Progress on going.


KrisThielemans:
we'd have to be careful with this. We can easily spend a lot of time on it, and then get something the same as boost::filesystem. That'd be a waste of time of course. I would suggest to keep this really really small. I'd rather us to concentrate on other stuff.

Of course, it'd have to be cross-platform which won't be easy. Re-using whatever is in `utilities.h` would therefore be crucial. Note that the tests would have to be as well (on Windows "/bla" is not an absolute pathname).

I'm not sure `Path` conveys the meaning of this class. `FilePath`? (maybe doesn't cover directories).

Please follow STIR coding and naming conventions (which are in the developer's guide).

src/buildblock/FilePath.cxx
```diff
@@ -0,0 +1,386 @@
...
+    if( stat( my_string.c_str(), &info ) != 0 )
+        error(boost::format("FilePath: Cannot access %1%.")%my_string);
+    else if( info.st_mode & S_IFDIR )
+        return true;
+
+    return false;
+}
+
+
+bool FilePath::is_regfile() const
```
KrisThielemans: I'd rather not have abbreviations if you don't mind.

src/buildblock/FilePath.cxx
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    //    if( stat( my_string.c_str(), &info ) != 0 )
+    //        return false;
+    //    else
+    //        return true;
+
+}
+
+FilePath FilePath::append(FilePath p)
```
KrisThielemans: argument should be `const FilePath&`. Does the function change the current member? If so, it shouldn't return a `FilePath`. If not, it has to be a `const` member.

src/buildblock/FilePath.cxx
```diff
@@ -0,0 +1,386 @@
...
+{
+    return append(p.get_path());
+}
+
+FilePath FilePath::append(std::string p)
```
KrisThielemans: same as above.

This is a very complicated function. I have no idea what it does... At some point, it will need some doxygen of course.

src/buildblock/FilePath.cxx
```diff
@@ -0,0 +1,386 @@
...
+        if(n != *c) buff+=n; else
+            if(n == *c && buff != "") { v.push_back(buff); buff = ""; }
+    }
+    if(buff != "") v.push_back(buff);
+
+    return v;
+}
+
+std::string::size_type
+FilePath::find_pos_of_filename(std::string _s) const
```
KrisThielemans: `const std::string&`

src/buildblock/FilePath.cxx
```diff
@@ -0,0 +1,386 @@
...
+        return false;
+    else
+        return ptr != filename_with_directory;
+#else // defined(__OS_UNIX__)
+    // absolute names start with '/'
+    return filename_with_directory[0] == '/';
+#endif
+}
+
+void FilePath::prepend_directory_name(std::string p)
```
KrisThielemans: const ref argument

src/buildblock/FilePath.cxx
```diff
@@ -0,0 +1,386 @@
...
+        my_string.erase(0);
+    new_name = p + separator + my_string;
+#else // defined(__OS_UNIX__)
+    // append / if necessary
+    new_name = merge(p, my_string);
+#endif
+    my_string = new_name;
+}
+
+std::string FilePath::merge(std::string first, std::string sec)
```
KrisThielemans: const ref args

src/buildblock/FilePath.cxx
```diff
@@ -0,0 +1,386 @@
+#include "stir/FilePath.h"
```
KrisThielemans: needs copyright and license info.
seems this contains code from utilities.cxx so will need copyright from there as well

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,176 @@
...
+    //! \warning Possibly dangerous... when real paths are used.
+    inline void operator=(const std::string& other) {
+        this->my_string = other;
+        this->initSeparator();
+    }
+
+private:
+    //! Checks on wether the path exits, is writable and accessible.
+    void checks() const;
+    //! Split the path on char c. If c not defined then default on the current separator.
```
KrisThielemans: I think this is a `static` function

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,176 @@
...
+    const std::vector<std::string> split(const std::string& s, const char* c ="");
+
+    //! Merge two strings using the correct OS separator
+    std::string merge(std::string first, std::string sec);
```
KrisThielemans: I think this is a `static` function

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,176 @@
...
+    bool is_regfile() const;
+    //! Returns true if the path is writable. 
+	//! \warning As far as I understand is only a *NIX feature. Write permissions do not exist in this level at Windows. 
+    bool is_writable() const;
+    //! Returns true if the path is absolute
+    static bool is_absolute(const std::string& _filename_with_directory);
+    //! Returns true if the path already exists
+    static bool exist(std::string s);
+
+    static std::string get_current_dirname();
```
KrisThielemans: I think better `get_current_working_directory`

src/buildblock/FilePath.cxx
```diff
@@ -0,0 +1,384 @@
...
+#if defined(__OS_WIN__)
+		nError = _mkdir(new_path.c_str());
+#else
+		nError = mkdir(new_path.c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
+#endif
+
+        if (-1 == nError)
+        {
+            printf("Error creating directory!n");
+            exit(1);
```
KrisThielemans: cannot call `exit`. Use `error`

src/buildblock/FilePath.cxx
```diff
@@ -0,0 +1,384 @@
...
+    }
+    else
+    {
+        error(boost::format("FilePath: Maybe %1% does not exist?")%my_string);
+    }
+}
+
+std::string FilePath::get_current_working_directory()
+{
+    char buffer[100];
```
KrisThielemans: 100 is dangerously small. At least use the const in `utilities.h` for max filename size. Alternatively,
`get_current_dir_name` seems to allocate the buffer itself (will then have to deallocate).
not sure on Windows.

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,187 @@
...
+    //! Returns true if my_string points to a regular file
+    //! Returns true if the path is writable. 
+	//! \warning As far as I understand is only a *NIX feature. Write permissions do not exist in this level at Windows. 
+    bool is_regular_file() const;
+    bool is_writable() const;
+    //! Returns true if the path is absolute
+    //! This funtion has been copied from the Functions for filename manipulations
+    static bool is_absolute(const std::string& _filename_with_directory);
+    //! Returns true if the path already exists
+    static bool exist(std::string s);
```
KrisThielemans: one more const ref. should be called `exists` I guess

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,187 @@
...
+    FilePath append(const FilePath& p);
+    //! \brief Create a new folder,by string, and return its path
+    //! \details This functions creates appends path p to my_string.
+    //! It supports multiple new folder levels.It will try to avoid
+    //! errors as permissions and non existing root path.
+    FilePath append(const std::string& p);
+    //! Append the separator
+    static void append_separator(std::string& s);
+    //! If path is no absolute then prepend the p string
+    //! This funtion has been copied from the Functions for filename manipulations
```
KrisThielemans: remove the `!`. no point in having that appear in the doxygen-generated files. Misspelling `function`

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,187 @@
...
+    //! This function will return the the bit that is after the last dot.
+    std::string get_extension() const ;
+
+    inline bool operator==(const FilePath& other);
+
+    inline bool operator==(const std::string& other);
+
+    inline void operator=(const FilePath& other);
+
+    //! \warning Possibly dangerous... when real paths are used.
```
KrisThielemans: cryptic comment. why would it be dangerous? real path?

NikEfth:
@KrisThielemans done! 
But I have a question. For Windows all separators are possible? You test for all them ... even in combination

src/buildblock/FilePath.cxx
KrisThielemans: no copyright info in the .cxx file though? (I think you put it in the .h, but there it should just be your own (as it doesn't contain any code from `utilities.h` I guess)

src/buildblock/FilePath.cxx
KrisThielemans: this still needs addressing. On Windows there's [GetCurrentDirectory](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364934(v=vs.85).aspx). don't bother using it though, just put it in a comment.

some comments on all this at https://stackoverflow.com/questions/2868680/what-is-a-cross-platform-way-to-get-the-current-directory. Note the one amount `getcwd` returning `NULL`.

src/include/stir/FilePath.h
KrisThielemans: I think you've fixed all these.

src/include/stir/FilePath.inl
```diff
@@ -0,0 +1,70 @@
...
+*/
+
+/*!
+  \file
+  \ingroup buildblock
+
+  \brief Implementations of inline functions for class stir::FilePath
+
+  \author Nikos Efthimiou
+  \author Kris Thielemans
```
KrisThielemans: remove me here and PARAPET.

src/include/stir/FilePath.inl
```diff
@@ -0,0 +1,70 @@
...
+    else
+        return false;
+}
+
+inline void FilePath::operator=(const FilePath& other) {
+    this->my_string = other.my_string;
+    this->separator = other.separator;
+}
+
+//! \warning Possibly dangerous... when real paths are used.
```
KrisThielemans: I really don't understand what's dangerous here... sorry.

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,192 @@
...
+  \ingroup buildblock
+
+  \brief Declaration of class stir::FilePath
+  This is a class implementing basic filesytem functionality. Parts of this class were
+  copied from Functions for filename manipulations.
+
+  Maximum effort was put in order to be cross platform.
+
+  \author Nikos Efthimiou
+  \author Kris Thielemans
```
KrisThielemans: remove me and PARAPET here.

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,192 @@
...
+
+    This file is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License for more details.
+
+    See STIR/LICENSE.txt for details
+*/
+
+/*
```
KrisThielemans: so these lines should go to the .cxx

KrisThielemans:
Note that on Windows that \ or / situation is confusing. Visual Studio supports both. other things don't. oh well. as long as the tests work, I'm fine.

KrisThielemans:
don't forget to add a line in the `Release3.1.htm`(developers section)

KrisThielemans:
Appveyor Visual Studio 2013 fails the test with
```
      Start 13: test_filename_functions
13/37 Test #13: test_filename_functions .........................................***Exception: Exit code 0xc0000409
 ```
(Visual Studio 2015 is ok).

Change `.appveyor.yml` to use `ctest -VV -C <config>` to see what's going on  (that's a useful thing anyway)

File: src/buildblock/FilePath.cxx (hunk 7 above)
Comment (NikEfth) [reply src/buildblock/FilePath.cxx]: ok

File: src/buildblock/FilePath.cxx (hunk 12 above)
Comment (NikEfth) [reply src/buildblock/FilePath.cxx]: Yep, that why I use that pointer to guard the string

**Comment by NikEfth** [comment]
That was not more insightful. I am downloading vs13 ...

File: src/include/stir/FilePath.inl (hunk 17 above)
Comment (NikEfth) [reply src/include/stir/FilePath.inl]: I the case that we have a real path, set during construction, that passes tests like exists and then with the  assignment we set a fake string. If we don't repeat checks (e.g. the one I added about the separator) we can run into trouble.

File: src/include/stir/FilePath.inl (hunk 16 above)
Comment (NikEfth) [reply src/include/stir/FilePath.inl]: ok

**Comment by KrisThielemans** [comment]
you're brave! error was
```
Testing various filename utilities (using Windows filesystem conventions)
13: Error. 
13: 
13: ERROR: FilePath: Maybe C:\projects\stir\build\src\test\my_test_folder_a\ does not exist?
```
If you cannot get it to installI suppose you could insert a `dir C:\projects\stir\build\src\test\` after the ctest in the Appveyor file.

File: src/include/stir/FilePath.inl (hunk 17 above)
Comment (KrisThielemans) [reply src/include/stir/FilePath.inl]: ok. I think you need to document what the checks do (I get the impression that setting the flag at constructor time means you will presume checks need to be done always?), that `append` will create the dir (kind of non-obvious from the name if you ask me). and then add something like the above in the doc as well. (By the way, why not run the check then here?)

File: src/include/stir/FilePath.h (hunk 13 above)
Comment (NikEfth) [reply src/include/stir/FilePath.h]: ok

**Comment by NikEfth** [comment]
brave or stupid?

File: src/include/stir/FilePath.inl (hunk 17 above)
Comment (NikEfth) [reply src/include/stir/FilePath.inl]: Well, the initial notion was to run the check on the constructor. Then I saw the potential problem and I added the checks pretty much everywhere needed. Maybe a bit redundant now that I think about it. 
Append is the command in both boost::filesystem and C17

File: src/include/stir/FilePath.h [hunk 20]
```diff
@@ -0,0 +1,170 @@
...
+
+*/
+
+#include "stir/common.h"
+#include <stdio.h>
+#include <iostream>
+#include <string>
+#include <boost/format.hpp>
+// N.E: Not sure I want this here. Included for the max_filename_length.
+#include "stir/utilities.h"
```
KrisThielemans: I think you only want this in the .cxx. Probably also the case for the includes below.

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,170 @@
...
+public:
+    FilePath();
+
+    //!
+    //! \brief FilePath
+    //! \param __str Name of path
+    //! \param _check Perform checks to the __str
+    //!
+    //! If the initial path exists in the disk _check should be left true.
+    //! It will check that the path is writtable
```
KrisThielemans: "writable"

Document what will happen if the check fails. (presumably call `error()`)

If you can be bothered, you could write `\p _check` etc wherever you use a parameter (not after the `\param` keyword)

src/test/test_filename_functions.cxx
```diff
@@ -5,6 +5,7 @@
 
   \brief Test program for filename functions defined in utility.h
 
+  \author Nikos Efthimiou
```
KrisThielemans: add Hull to copyright

src/include/stir/FilePath.h
NikEfth: ok

src/include/stir/FilePath.h
NikEfth: ok

src/test/test_filename_functions.cxx
NikEfth: :)

NikEfth:
the failure is openmp related

src/include/stir/FilePath.inl
KrisThielemans: still to do

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,159 @@
...
+  copied from Functions for filename manipulations.
+
+  Maximum effort was put in order to be cross platform.
+
+  \author Nikos Efthimiou
+
+*/
+
+#include "stir/common.h"
+#include <stdio.h>
```
KrisThielemans: do we need stdio.h?  definitely only in .cxx

src/include/stir/FilePath.h
```diff
@@ -0,0 +1,159 @@
...
+#include <iostream>
+#include <string>
+#include <boost/format.hpp>
```
KrisThielemans: boost/format and iostream move to .cxx

NikEfth:
Squash is fine 
done with the last comments :)

PR #170, opened 2018-03-15
Update PLSPrior (and GeneralisedPrior a bit too)

rijobro:
- Added public set/get functions to PLSPrior. 
- Changed set_up function to match base class.
- Move file opening from post_processing to set_up

### DIFF
--- a/src/include/stir/recon_buildblock/GeneralisedPrior.h
+++ b/src/include/stir/recon_buildblock/GeneralisedPrior.h
@@ -93,6 +93,7 @@ protected:
   /*! Has to be called by initialise_keymap in the leaf-class */
   virtual void initialise_keymap();
 
+  bool _already_set_up;
 };
 
 END_NAMESPACE_STIR
--- a/src/include/stir/recon_buildblock/GeneralisedPrior.h
+++ b/src/include/stir/recon_buildblock/GeneralisedPrior.h
@@ -93,6 +93,9 @@ protected:
   /*! Has to be called by initialise_keymap in the leaf-class */
   virtual void initialise_keymap();
 
+  //! Check that the prior is ready to be used
+  virtual void check(DataT const& current_estimate) const;
+
   bool _already_set_up;
 };
 
--- a/src/recon_buildblock/PLSPrior.cxx
+++ b/src/recon_buildblock/PLSPrior.cxx
@@ -209,7 +209,10 @@ template <typename elemT>
 void
 PLSPrior<elemT>::
 set_anatomical_image_sptr (const shared_ptr<DiscretisedDensity<3,elemT> >& arg)
-{ this->anatomical_sptr = arg; }
+{
+    this->anatomical_sptr = arg;
+    this->anatomical_filename = ""; // Clear filename in case it was set.
+}
 
 template <typename elemT>
 void
@@ -261,7 +264,10 @@ template <typename elemT>
 void
 PLSPrior<elemT>::
 set_kappa_sptr(const shared_ptr<DiscretisedDensity<3,elemT> >& k)
-{ this->kappa_ptr = k; }
+{
+    this->kappa_ptr = k;
+    kappa_filename = ""; // Clear filename in case it was set.
+}
 
 //! Set kappa filename
 template <typename elemT> void PLSPrior<elemT>::set_kappa_filename(const std::string& filename)
@@ -269,7 +275,6 @@ template <typename elemT> void PLSPrior<elemT>::set_kappa_filename(const std::st
   kappa_filename  = filename;
   this->kappa_ptr = read_from_file<DiscretisedDensity<3,elemT> >(kappa_filename);
   info(boost::format("Reading kappa data '%1%'") % kappa_filename  );
-  kappa_filename = ""; // Clear at end
 }
 
 //! Set anatomical filename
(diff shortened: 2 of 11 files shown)


src/recon_buildblock/PLSPrior.cxx
```diff
@@ -62,8 +62,16 @@ PLSPrior<elemT>::initialise_keymap()
 
 template <typename elemT>
 Succeeded
-PLSPrior<elemT>::set_up ()
+PLSPrior<elemT>::set_up (shared_ptr<DiscretisedDensity<3,elemT> > const& target_sptr)
 {
+    base_type::set_up(target_sptr);
+
+    if (kappa_filename.size() != 0)
```
KrisThielemans: I have thought about this more, and I don't think reading the data here is a good idea. It means that calling `set_up` twice will read the data twice, and that errors when reading will only be flagged when doing the set_up. Furthermore, it introduces surprising behaviour (in my opinion) related to what to do with the filename vs the member.

So, please move the reading to the `set_kappa_filename` function, and call it from `post_processing`

src/recon_buildblock/PLSPrior.cxx
```diff
@@ -105,17 +112,10 @@ PLSPrior<elemT>::post_processing()
 {
   if (base_type::post_processing()==true)
     return true;
-  if (kappa_filename.size() != 0)
-    this->kappa_ptr = read_from_file<DiscretisedDensity<3,elemT> >(kappa_filename);
```
KrisThielemans: so, keep the if, and replace the next line with `set_kappa_filename` (which will then read)

src/recon_buildblock/PLSPrior.cxx
```diff
@@ -498,7 +498,6 @@ PLSPrior<elemT>::
 compute_gradient(DiscretisedDensity<3,elemT>& prior_gradient,
                  const DiscretisedDensity<3,elemT> &current_image_estimate)
 {
-
   if (!this->anatomical_sptr->has_same_characteristics(current_image_estimate))
```
KrisThielemans: move this to a `check` function (needs to be one in `GeneralisedPrior`)

src/include/stir/recon_buildblock/GeneralisedPrior.h
```diff
@@ -93,6 +93,7 @@ class GeneralisedPrior:
   /*! Has to be called by initialise_keymap in the leaf-class */
   virtual void initialise_keymap();
 
+  bool _already_set_up;
```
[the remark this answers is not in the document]
rijobro: It is set to false in GeneralisedPrior::set_defaults (which is called by the derived classes).

rijobro:
Changes have been made to the other priors, let me know what you think.

src/recon_buildblock/PLSPrior.cxx
```diff
@@ -105,23 +105,29 @@ PLSPrior<elemT>::post_processing()
 {
   if (base_type::post_processing()==true)
     return true;
-  if (kappa_filename.size() != 0)
-    this->kappa_ptr = read_from_file<DiscretisedDensity<3,elemT> >(kappa_filename);
 
-  if (anatomical_filename.size() != 0){
-    this->anatomical_sptr = read_from_file<DiscretisedDensity<3,elemT> >(anatomical_filename);
```
KrisThielemans: shouldn't we read the anatomical image here as well?

src/recon_buildblock/PLSPrior.cxx
```diff
@@ -258,7 +264,21 @@ PLSPrior<elemT>::
...
+  this->kappa_ptr = read_from_file<DiscretisedDensity<3,elemT> >(kappa_filename);
+  info(boost::format("Reading kappa data '%1%'") % kappa_filename  );
+}
 
+//! Set anatomical filename
+template <typename elemT> void PLSPrior<elemT>::set_anatomical_filename(const std::string& filename)
+{
+  anatomical_filename  = filename;
+  this->anatomical_sptr = read_from_file<DiscretisedDensity<3,elemT> >(anatomical_filename);
+  info(boost::format("Reading anatomical data '%1%'") % anatomical_filename  );
```
KrisThielemans: do we have a set_anatomical_sptr()? Should there set the filename to zero., same for kappa

rijobro: There are three ways to set the anatomical/kappa images. 
1. parser
2. set_filename
3. set_sptr
The parser will call the set_filename function in post_processing. The set_filename will clear the filename at the end. set_sptr will never alter the filename. So in all three cases, the image will be read and the filename will be cleared.

src/recon_buildblock/PLSPrior.cxx
rijobro: added in.

src/recon_buildblock/PLSPrior.cxx
KrisThielemans: the removal of the filename doesn't look ok to me. A user might be surprised that if they set a filename, it suddenlly isn't valid anymore. This will in particular occur with the `parameter_info` from `ParsingObject`. It seems that `set_sptr` has to clear the filename (as it doesn't know what it is), but `set_filename` keeps it. as the parser will use `set_filename`, everything should still be fine.

rijobro: done.

PR #178, opened 2018-04-24
Major update related to deriving images from ExamData

### LINKED ISSUE #176: derive DiscretisedDensity and DynamicDiscretisedDensity from ExamData
we need ExamInfo for images as well.

KrisThielemans:
This derives DiscretisedDensity and DynamicDicretisedDensity from
ExamData, such that they have access to an ExamInfo object. This
enables the reconstruction and other code to set time_frame info
and anything else in ExamInfo for reconstructed images (and similar
for projection).

Interfile IO has been adapted to write ExamInfo. stir_ecat7 has also
had some updates but those haven't been tested yet.

list_image_info has now several options like list_projdata_info, and
can now display ExamInfo.

The code should be backwards compatible. I have added constructors,
as opposed to changing existing ones. The existing constructors
will not set the ExamInfo obviously.

There is no explicit test code for this update yet, although I have
checked that everything works via the recon_test_pack.
(tests in VoxelsOnCartesianGrid should be added after moving ExamData to
buildblock).

Closes #176


NikEfth:
This looks great and checks out in my tests. 
I tested it using the image/data viewer. 
But in the process I tried to make a `ExamData* read_interfile_data()` (maybe not very useful for you but for me could reduce dramatically the IO)  but that was not possible because there is an asymmetry. 
`ProjData::ExamData` have the `Viewgram<T>::Array<T>` as members while `DiscretisedDensity<T>::Array<T>` inherits `ExamData` :/.

PR #181, opened 2018-05-01
View offset

pwadhwa351:


pwadhwa351:
I think I am done with the commits. :)

NikEfth:
This is wrong. 
It includes parts of my changes. @pwadhwa351 Do the same thing but start from scratch. from a clean master. And do not give the same name twice.

pwadhwa351:
@NikEfth Can you please specify which changes you are referring to which you made?

NikEfth:
I should not be able to see this : 
```
commit b3a955831d9c6b5512e52c4910866e0ab09bfdee
Merge: 124e0889 cca8e608
Author: pw351 <pwadhwa351@gmail.com>
Date:   Fri Apr 27 15:33:02 2018 +0100

    Merge pull request #1 from NikEfth/Viewoffset
    
    Fixed all tests in test_proj_data_info with tilt (view_offset)
```

f

KrisThielemans:
@nikefth, I don't see that commit. In this PR, I see only 4 commits from @pwadhwa351. It restores the PR to what I remember. Possibly you have some links to the old PR? (not sure why thought. this is PR 181, the old one was #130)

NikEfth:
I can see that from 
```git log```

**Comment by KrisThielemans** [comment]
not sure what you did then. you'd have to add a remote, fetch this new branch, and do a git log on that. I guess you might have merged onto an existing branch.
as far as I understand, the commits that I see on github are what's in this branch. (Would be really very surprising otherwise)

**Comment by NikEfth** [comment]
Alright, I see now.  so @pwadhwa351 and @pw351 are different accounts with the same brach ... ok

**Comment by NikEfth** [comment]
OK, so let me explain the situation. I made two commits and a PR to Palak. Commit: ed71b4eb9fc986773ed0b02232f470843524a687 and 0af025699f5ea4d43556575bc86e41fed4676aee. 
In both `make test` succeeds but `test_simulate_and_recon.sh` fail for different reasons.
In the first commit the `calculate_attenuation_coefficient `fail because it needs `--PMRT` to run. Iterative reconstructions run smoothly and of course analytic fail.
In the second commit I added the PMRT flag but now I get a timeout on the `Running precorrections`.

I am sorry I will not be able to finish it this week.

**Comment by KrisThielemans** [comment]
I had a look at ed71b4eb9fc986773ed0b02232f470843524a687 and added some comments there (not ideal but best I could do). There's a few small things I doubt that can be correct, but the main change here is that the code now takes the offset into account in the translation between views and detectors. I'm not sure if this is the best way though, see also [my comment in the code](https://github.com/UCL/STIR/commit/ed71b4eb9fc986773ed0b02232f470843524a687#r28834541).

I believe that changing the relation between views and detectors is dangerous though. It would have to be taken into account in any code for normalisation etc (detector efficiencies for the first and middle detector would now no longer correspond to the first view). I think that spreads things too far.

Instead, I believe it is better to use the following convention:
- "continuous" coordinates such as `phi` and `psi` (the one (around the ring") are always in the standard coordinate system (with 0 being up or whatever it is).
- "discrete" coordinates in bins always start from 0
- we have a `phi_offset` (aka `view_offset`) and `psi_offset` (aka `det_num_offset`) such that 
    - `phi = phi_offset + view * view_increment`
    - `psi = psi_offset + det_num * det_num_increment`

As long as these 2 offsets are consistent, we should be fine. Sadly, because of the interleaving, the relation between these 2 is maybe not trivial, but at least initially I'd vote for `psi_offset = 2*phi_offset`. That would be backwards compatible for the offset=0 case so probably our only possible choice.

In summary, I think we need a new function `get_psi_offset` (implemented in terms of the view offset), and used that in ProjDataInfoCylindericalNoArcCorr whereever we convert between `psi` and `detx`. I would hope that we don't need any other changes.

File: src/include/stir/ProjDataInfo.h [hunk 1]
```diff
@@ -348,9 +350,10 @@ class ProjDataInfo
   
 protected:
   virtual bool blindly_equals(const root_type * const) const = 0;
+//PW Shifted scanner_ptr from private to protected as it is being used in derived class for intrinsic tilt.
+  shared_ptr<Scanner> scanner_ptr;
```
KrisThielemans: this isn't necessary. Let's use `get_scanner_ptr()` in the derived classes (like in most places already)

KrisThielemans:
I've tried this and can confirm that for the arccorr case, there are round-trip problems, giving for instance
```
Error. round-trip get_LOR then get_bin (LORAs2Points): segment
        Problem at    segment = -15, axial pos 0, view = 177, tangential_pos_num = -31
        round-trip to segment = 15, axial pos 0, view = -15, tangential_pos_num = 31
```
showing that the view_num is negative, which shouldn't be the case. so we indeed need some of @NikEfth's fixes.

KrisThielemans:
I've now added a few "modulo" operations (`to_0_2pi`) and then the `get_psi_offset` works. All test now work, except the analytic reconstructions as they use the incremental interpolating backprojector which doesn't take this into account. that'll be a pain.

thanks @NikEfth for the discussion on this!

KrisThielemans:
shifted to 5.0 as output will be incompatible for some scanners (with non-zero view-offset)

KrisThielemans:
I've "fixed" the tests by running FBP on data with view-offset = 0. I expect this to work now.

Things to do:
- [ ] check if the scatter code does the proper thing
- [ ] make CMake option to disable view-offset, or change release notes

PR #108, opened 2017-09-18
Listmode reconstruction

NikEfth:
This  branch is clean and passes all test at my laptop. In particular the new test a listmode reconstruction in comparison to a sinogram.

PR #186, opened 2018-05-12
HKEM sino

danieldeidda:
Implementation of the sinogram based kernelised-loglikelihood


KrisThielemans:
thanks for this!

Can you please revert `PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx` to the original? This shouldn't have anything to do with this PR. The relevant fix is presumably in #108 which should hopefully be merged soon.

I saw travis failures. Let me know when this is ready for me to review. (it might take me a while...)

danieldeidda:
**Can you please revert PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx to the original? This shouldn't have anything to do with this PR. The relevant fix is presumably in #108 which should hopefully be merged soon.**

Sorry, this happened also for the PLS and I only now realised I had uploaded this file on my forked master by mistake, and every time I create a new branch and commit this also come as another commit. Is there a way to remove this? for the HKEM_sino I have copy-pasted from master, but I would like this not to happen in the future.

KrisThielemans:
thanks for reverting PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx. However, there's a lot of white-space differences left in that file. Best to do this with a "git checkout master blabla/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx"

KrisThielemans:
hi, not sure what's happening here. you seem to have committed a more recent version from somewhere. probably not the right `master` therefore. I assumed that it would point to UCL/STIR/master but that was obviously not guaranteed.

we can play various games with git now, but maybe easiest to just download from
https://raw.githubusercontent.com/UCL/STIR/master/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx
and commit that. 

sorry.

KrisThielemans:
amazing. still white-space diffs. did you open the file at all before committing? sorry about the mess.

anyway, I'll try to fix this later.

could you have a look at the Codacy report?
```
src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
The scope of the variable 'dim_y' can be reduced.
1188 {
1189    double kmean=0;
1190    double kStand_dev=0;
1191    double dim_x=0, dim_y=0, dim_z=0;
1192    int nv=0;
1193    const int min_z = (*anatomical1_sptr).get_min_index();
```

KrisThielemans:
I haven't checked the code, but generally in C++ you want to define the variable where you use it. so stuff like `for (int i=0...`.
not terribly important of course, but at some point we'll want to clean up most of these Codacy flags. could as well do it now...

examples/samples/OSMAPOSL_HKEM.par
```diff
@@ -0,0 +1,65 @@
+OSMAPOSLParameters :=
+
```
KrisThielemans: add a comment what this file is

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
```diff
@@ -0,0 +1,471 @@
...
+  \f$\lambda_j\f$, can then be described using the kernel matrix
+
+
+
+  \f[
+   \lambda_j=  \sum_{l=1}^L \alpha_l k_{jl}
+  \f]
+
+  where \f$k_{jl}\f$ is the \f$jl^{th}\f$ kernel element of the matrix, \f$\boldsymbol{K}\f$.
+  The resulting algorithm is the following:
```
KrisThielemans: thanks for the extensive documentation. However, this line and stuff below is inaccurate. You are talking about MLEM or OSEM versions of HKEM. However, you are implementing the objective function here. it could be used with OSSPS presumably (or other algorithms).

Either adjust the formulas, or say that for OSEM, the algorithm would be...

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
```diff
@@ -0,0 +1,471 @@
...
+  bool distributed_tests_enabled;
+  bool message_timings_enabled;
+  double message_timings_threshold;
+  bool rpc_timings_enabled;
+  //#endif
+  //@}
+
+
+
+  std::string sens_filenames;
```
KrisThielemans: shouldn't be public variables I guess

danieldeidda:
Hi, thanks for the feedback, I will try to address everything focusing on each point. For example:

derive from PoissonLogLikelihoodWithLinearModelForMeanAndProjData. I think you'd just need to add a few variables/member functions, and I believe in compute_sub_gradient_without_penalty_plus_sensitivity do some stuff and then call the same function of the base class. (same for the value). I think 90% of the code could go. In fact, if you'd do this later on as a template, you could get the listmode version almost for free.

I had thought to do this but when I was planning the implementation I was not to sure of how many methods I would have needed to override. So I would follow this suggestion.

one question: Presumably all the functions that I did not modify can be removed. However, it seems that I cannot use the distributable functions like "distributable_compute_gradient()" etc. Is there a specific way of doing this?

KrisThielemans:
thanks.

Regarding calling `distributable_compute_gradient`, this is indeed currently not an "exposed" function. We could do that if necessary, but can't you simply call `base_type::compute_sub_gradient_without_penalty_plus_sensitivity` (where `base_type=PoissonLogLikelihoodWithLinearModelForMeanAndProjData`)?

danieldeidda:
**but can't you simply call base_type::compute_sub_gradient_without_penalty_plus_sensitivity (where base_type=PoissonLogLikelihoodWithLinearModelForMeanAndProjData)?**

Can I do that? because I am actually overriding this by calling all the kernel functions.

KrisThielemans:
sure, but the base_type version just calls `distributable_computation_gradient` anyway. So in your `HKEM::compute_sub_gradient_without_penalty_plus_sensitivity` where you currently call `distributable_computation_gradient``, call instead `base_type::compute_sub_gradient_without_penalty_plus_sensitivity`. might have to play with arguments of course.

danieldeidda:
OK, this works. I will make one big commit with all the changes in the ncoming days.

examples/samples/OSMAPOSL_HKEM.par
```diff
@@ -0,0 +1,65 @@
...
+; Daniel: here we have the possibility to choose the parameters which define the kernel matrix and the name of the anatomical image.
+
+hybrid:=${hybrid}
+sigma_m:= ${kpar}
+sigma_p:=${kPpar}
+sigma_dm:=${Dmpar}
+sigma_dp:=${Dpar}
+neighbours_num:= ${N}
+anatomical1_image_filename:=${mr}
```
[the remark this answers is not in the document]
danieldeidda: no underscore even for the sigmas? I put one because I am working on a version wich allows the use of more anatomical images to use for the kernel.

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -0,0 +1,1919 @@
...
+
+//    this->current_kimage_filename =  ;
+}
+
+
+    const std::string current_sensitivity_filename =
+      boost::str(boost::format(this->get_subsensitivity_filenames ()) % subset_num);
+
+    //if(get_subiter_num ()<=this->get_num_subsets()){
+         shared_ptr<TargetT> sens_sptr(read_from_file<TargetT>(current_sensitivity_filename));
```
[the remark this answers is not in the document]
danieldeidda: Yes, the sensitivity is calculated before reconstruction so both if you read or recalculate the sensitivity this works. However, I seem to remember that I did this because I was having problem to access the sensitivity from the code.
I am open to suggestion though...

examples/samples/OSMAPOSL_HKEM.par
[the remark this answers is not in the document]
KrisThielemans: sure. underscore in sigmas etc is fine.

`anatomical1` is fine, although it'd be better to then use something like
```
number of anatomical images := xxx
anatomical image filename[1] := ...
```
That requires a little bit of extra code at present (10 lines or so) but I could point you to an example if you want to do this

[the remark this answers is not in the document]
danieldeidda: Yes, an example would be great. I do not know I will be adding this feature but it is good to think about it. At the moment then I will just remove the 1.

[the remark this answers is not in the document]
danieldeidda: *I do not know when...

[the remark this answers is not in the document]
KrisThielemans: following might work
```C++
// class def
 int num_anatomical_images;
 std::vector<std::string> anatomical_image_filenames;
private:
 void read_anatomical_image_info();

// constructor or wherever you initialise the parser
  add_key("number of anatomical images", 
KeyArgument::INT, (KeywordProcessor)&HKEM::read_anatomical_image_info,&num_anatomical_images);
  add_key("anatomical image filename", & anatomical_image_filenames);

// extra member
void HKEM::read_anatomical_image_info()
{
  set_variable();
  anatomical_image_filenames.resize(num_anatomical_images);
}
```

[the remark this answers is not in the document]
danieldeidda: I guess then I can also define a vector for the actual images?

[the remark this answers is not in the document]
KrisThielemans: yes of course.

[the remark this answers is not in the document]
KrisThielemans: you could in your current version always restrict the length to 1. However, then it will be future proof

ashgillman:
Hi,

Sorry I have only just seen these comments from Friday. I have now subscribed to the PR so should get future updates.

Regarding the code duplication, Daniel and I discussed this on Friday. I think there is enough similarity that inheriting from `PoissonLogLikelihoodWithLinearModelForMeanAndProjData` might be the best option. I think the latter option could be a little bit tedious and just as prone to breaking when `PoissonLogLikelihoodWithLinearModelForMeanAndProjData` changes. (i.e., `KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData` will probably often have to change if `PoissonLogLikelihoodWithLinearModelForMeanAndProjData` changes whether it inherits or uses as a member, however either way at least specific changes aren't duplicated).

I have not yet looked at the tests run in CI, but if there is some test for `PoissonLogLikelihoodWithLinearModelForMeanAndProjData`, we should add another for `KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData`.

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
```diff
@@ -0,0 +1,266 @@
...
+      (e.g. for instance when using asymmetric attenuation).
+  */
+
+ private:
+
+/*! Create a matrix containing the norm of the difference between two feature vectors, \f$ \|  \boldsymbol{z}^{(n)}_j-\boldsymbol{z}^{(n)}_l \| \f$. */
+/*! This is done for the PET image which keeps changing*/
+  void  calculate_norm_matrix(TargetT &normp,
+                              const int &dimf_row,
+                              int &dimf_col,
```
KrisThielemans: `const int dimf_col`? also below

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
```diff
@@ -0,0 +1,266 @@
...
+  const std::string get_anatomical_filename() const;
+  const int get_num_neighbours() const;
+  const int get_num_non_zero_feat() const;
+  const double get_sigma_m() const;
+  double get_sigma_p();
+  double get_sigma_dp();
+  double get_sigma_dm();
+  const bool get_only_2D() const;
+  int get_subiter_num();
+  double get_kSD();
```
ashgillman: make private? Maybe just a field rather than have a getter/setter?

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
```diff
@@ -0,0 +1,266 @@
...
+    This can be used as alternative to the parsing mechanism.
+   \warning After using any of these, you have to call set_up().
+   \warning Be careful with changing shared pointers. If you modify the objects in
+   one place, all objects that use the shared pointer will be affected.
+
+  */
+  //@{
+
+  void set_subiter_num(int subiter_num);
+  void set_kSD(double kSD);
```
ashgillman: private

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
```diff
@@ -0,0 +1,266 @@
...
+  virtual double
+    actual_compute_objective_function_without_penalty(const TargetT& current_estimate,
+                                                      const int subset_num);
+
+protected:
+  //! Filename with input projection data
+  std::string input_filename,kernelised_output_filename_prefix;
+  std::string current_kimage_filename;
+  std::string sens_filenames;
+ int subiter_num;
```
ashgillman: Also private, no getter/setter. Name -> `iteration_count` so doesn't necessarily imply alignment with iterations/subsets

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
ashgillman: name -> `anatomical_sd`

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
```diff
@@ -0,0 +1,266 @@
...
+  //! sets any default values
+  /*! Has to be called by set_defaults in the leaf-class */
+  virtual void set_defaults();
+  //! sets keys for parsing
+  /*! Has to be called by initialise_keymap in the leaf-class */
+  virtual void initialise_keymap();
+  //! checks values after parsing
+  /*! Has to be called by post_processing in the leaf-class */
+  virtual bool post_processing();
+
```
ashgillman: add a `set_up()`

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
ashgillman: to private, rename, field

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -0,0 +1,992 @@
...
+  this->sigma_m=1;
+  this->sigma_p=1;
+  this->sigma_dp=1;
+  this->sigma_dm=1;
+  this->anatomical_image_filename="";
+  this->only_2D = 0;
+  this->kernelised_output_filename_prefix="";
+
+  this->subiter_num=0;
+  this->kSt_dev=0;
```
ashgillman: to `set_up` ( this->subiter_num=0;
+  this->kSt_dev=0;)

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -0,0 +1,992 @@
...
+{
+
+
+    set_subiter_num (this->subiter_num+=1);
+
+    unique_ptr<TargetT> kImage_uptr(current_estimate.get_empty_copy());
+    TargetT& kImage=*kImage_uptr;
+
+    if(this->num_non_zero_feat==1){
+        fast_compute_kernelised_image (kImage, *current_estimate.clone(), current_estimate);
```
ashgillman: No clone, unique_ptr

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -0,0 +1,992 @@
...
+                                    }
+    else{
+      compute_kernelised_image(kImage, *current_estimate.clone(),
```
ashgillman: No clone, unique ptr

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -0,0 +1,992 @@
...
+                                                              current_estimate);
+         }
+         else{
+                 compute_kernelised_image(ksens, *sens_sptr,
+                                                             current_estimate);
+
+         }
+
+
+        this->set_subset_sensitivity_sptr (shared_ptr< TargetT >(ksens.clone()),subset_num);
```
ashgillman: should this be cloned? unique ptr?
@KrisThielemans Daniel and I have discussed for this example but may need some guidance

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -0,0 +1,992 @@
...
+
+ PoissonLogLikelihoodWithLinearModelForMeanAndProjData<TargetT>::compute_sub_gradient_without_penalty_plus_sensitivity (
+             gradient,
+             kImage,
+             subset_num);
+
+  if(this->num_non_zero_feat==1){
+      fast_compute_kernelised_image(gradient,*gradient.clone(),
```
ashgillman: unique ptr

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -0,0 +1,992 @@
...
+                                             current_estimate);
+
+  }
+  else{
+      compute_kernelised_image(gradient,*gradient.clone(),
```
ashgillman: unique ptr

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -0,0 +1,992 @@
...
+actual_compute_objective_function_without_penalty(const TargetT& current_estimate,
+                                                  const int subset_num)
+{
+  double accum=0.;
+
+  shared_ptr<TargetT> kImage_sptr(current_estimate.get_empty_copy());
+  TargetT& kImage=*kImage_sptr;
+
+  if(this->num_non_zero_feat==1){
+      fast_compute_kernelised_image (kImage, *current_estimate.clone(), current_estimate);
```
ashgillman: No clone, unique ptr

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -0,0 +1,992 @@
...
+                                  }
+  else{
+    compute_kernelised_image(kImage, *current_estimate.clone(),
```
ashgillman: No clone, unique ptr

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
```diff
@@ -0,0 +1,266 @@
...
+  void estimate_stand_dev_for_anatomical_image(double &SD);
+
+/*! Compute for each voxel, jl, of the PET image the linear combination between the coefficient \f$ \alpha_{jl} \f$ and the kernel matrix \f$ k_{jl} \f$\f$ */
+/*! The information is stored in the image, kImage */
+  void compute_kernelised_image(TargetT &kImage, TargetT &Image,
+                                                            const TargetT &current_estimate);
+
+ /*! Similar to compute_kernelised_image() but this is the special case when the feature vectors contains only one non-zero element. */
+ /*! The computation becomes faster because we do not need to create norm matrixes*/
+void fast_compute_kernelised_image(TargetT &kImage, TargetT &Image,
```
ashgillman: second arg const

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
```diff
@@ -0,0 +1,266 @@
...
+  void  calculate_norm_const_matrix(TargetT &normm,
+                              const int &dimf_row,
+                              int &dimf_col);
+
+/*! Estimate the SD of the anatomical image to be used as normalisation for the feature vector */
+  void estimate_stand_dev_for_anatomical_image(double &SD);
+
+/*! Compute for each voxel, jl, of the PET image the linear combination between the coefficient \f$ \alpha_{jl} \f$ and the kernel matrix \f$ k_{jl} \f$\f$ */
+/*! The information is stored in the image, kImage */
+  void compute_kernelised_image(TargetT &kImage, TargetT &Image,
```
ashgillman: second arg const

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
```diff
@@ -0,0 +1,266 @@
...
+                                                            const TargetT &current_estimate);
+
+ /*! Similar to compute_kernelised_image() but this is the special case when the feature vectors contains only one non-zero element. */
+ /*! The computation becomes faster because we do not need to create norm matrixes*/
+void fast_compute_kernelised_image(TargetT &kImage, TargetT &Image,
+                                                          const TargetT &current_estimate);
+
+};
```
ashgillman: Make a function `compute_kernelised_image()` which calls either `compact_compute_kernelised_image()` or `full_compute_kernelised_image()`

Replace all calls

ashgillman:
@KrisThielemans Another question - for testing, it looks like recon_test_pack is used. Should we just add a test using this Obj Fun there?

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
KrisThielemans: I guess ksens is already a new image? (haven't checked). if so, no need to clone again of course.

KrisThielemans:
> for testing, it looks like recon_test_pack is used. Should we just add a test using this Obj Fun there?

yes, unless you can come up with a ctest. (one interesting ctest is to check if the gradient and objectiv  function actually match. There's some stuff for this, but I think @rijobro generalised this a bit already).

an issue with the recon_test_pack is for the "compare with known output", we currently need to have an image on github. That's not nice so ideally we don't do that... Hence my feeling is to use the ROI-based `test_simulate_and_recon.sh`. I'd have to check if this is straightforward or not, but will leave this to you...

src/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
danieldeidda: Yes it is a new image. If I remove clone() i get an error like:  "no matching function..."

danieldeidda: Yes, this is what I actually want to do, as the kernel has to be applied to the sensitivity images somewhere. The other option could be applying the kernel is where the division between the gradient and the sensitivity is made. However, I think this is going to be difficult because I believe this is done in the OSMAPOSLreconstruction class which is probably not ideal.

danieldeidda: Anyway the following works and it is a little faster:

 compute_kernelised_image(ksens, *sens_sptr,
                                          current_estimate);
        *ksens_sptr=ksens;

        this->set_subset_sensitivity_sptr (ksens_sptr,subset_num);

src/include/stir/recon_buildblock/KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData.h
danieldeidda: @KrisThielemans  I added set_up as follow:
```
template<typename TargetT>
Succeeded
KernelisedPoissonLogLikelihoodWithLinearModelForMeanAndProjData<TargetT>::
set_up(shared_ptr<TargetT>  const& target_sptr)
{
    base_type::set_up(target_sptr);

    PoissonLogLikelihoodWithLinearModelForMeanAndProjData<TargetT>::set_up (target_sptr);
    this->subiteration_counter=0;
    this->anatomical_sd=0;
    return Succeeded::yes;
}
```
but when I run the images are all zeros

KrisThielemans:
I have some doubts about the sensitivity stuff. I'm going to write some stuff but really it needs a discussion among all of you. A few things to consider:

The question is if this class is computing the emission image, or the kernel coefficients (alpha).  For some cases (like fixed kernel, or similary an "ROI-value" estimator) it would make sense to effectiviely add the kernel to the system model, and therefore have an objective function in terms of the alphas. However, I guess for the hybrid method it has to be in terms of the emission image (as the kernel is changing).

So, the question is: what are the parameters of the objective function, and I think for the hybrid method it is the emission image (as you implemented it).

Now arguably the sensitivity image is part of the objective function, but it is a rather weird thing as it is only there (and used in OSEM MLEM) because we're using a linear model (hence the name of the base class). Nevertheless, it seems then that the "sensitivity" has to be the one corresponding to the relevant term in the objective function (which is the "normal one").

As you write, the difficulty now comes about when implementing the specific HKEM algorithm that uses OSMAPOSL, and therefore needs the "alpha"-sensitivity.

In a sense HKEM alternates between finding the kernel, constructing a new objective function in terms of the alphas, and then doing a (sub)-iteration. The difficulty comes because you're sticking the kernel construction inside the HKEM objective function, while actually it should be an algorithm. That seems to lead to very different class design where you have a HKEM algorithm that takes 3 components: an objective function (in emission space), a way to construct the kernel, and an optimiser for an objective function in alpha-space.

This seems to lead us rather far though...

danieldeidda:
kernel_diag2

The class itself is calculating the alpha image, now if we look at the algorithm formula (let's pretend the formula is for sinograms) in the figure what the class does is to apply the calculated kernel to:
1) current_estimate=alpha_f;
2) gradient;
3) sensitivity= sum_f p_ij;
the application of the kernel to the 1) can also be seen as the PET image for the previous subiteration, because the K^(n+1)*alpha^(n+1) is going to be used in the (n+2) iteration. This is the reason why I include the counter, just to know when I have to write the PET image. "and then doing a (sub)-iteration" Actually there isn't an extra subiteration, the objective function parameter is defenitely alpha but the calculation of the PET image comes for free when you apply the kernel in the denominator on the right of the formula, all it does is to save it as the previous iteration. The way the kernel was though is probably MLEM specific and probably needs to be different for different algorithms.

Including the kernel in the system model would be good but it will need a way to be updated iteration after iteration.

I am not sure about the last paragraph, but I guess you are suggesting to create a new algorithm, say KOSMAPOSL, where all the kernel function are created and applied to the relevant images and to use the normal Poissonloglikelihood... as objective function.   

From Ash

We have discussed, and we agree that potentially we should implement a KOSMAPOSLReconstructor. But the motivation is a little different. The issue may be in wrapping into SIRF, where it is expected that one can call `get_current_estimate()`, but this would return alpha.


So KOSMAPOSLReconstructor and inherit from OSMAPOSLReconstructor, but override `get_current_estimate()`.


We are a little confused still as to the issue exactly with sensitivity. I think Daniel'd architecture of having KernelisedPoisson... modify the sensitivity image seems to make sense in this context, where that factor is changing iteration to iteration.

KrisThielemans:
>  in the figure what the class does is to apply the calculated kernel to:
let's just be clear. The formula is an MLEM formula. The current class is an objective function class (that so happens to compute a part of the formula).

If `OSMAPOSL` with your current `HKEMObjFunc` really computes alpha, this will indeed be quite confusing. It is consistent with my guess that the HKEM algorithm in the figure shouldn't really be implemented via OSMAPOSL of a new objective function.

This seems to need some brainstorming. If we do this properly, the set-up would be useful for other cases as well. The (non-hybrid) kernel method for instance is a version of regularisation using "sieves" (Snyder et al).

non-hybrid kernel could be done as 
- have normal objective function (emission image)
- create new "alpha" objective function that modifies the projectors by adding a "preforward" and "postbackward" step. (those go from alpha to emission and vice versa)
- run OSMAPOSL (or whatever) on that objective function
- finalise by going from alpha to emission

hybrid-kernel would then just run one sub-iteration of OSMAPOSL, update the kernel, recompute new sensitivity.

This is a major redesign though. It'd need some thought on doing this without recomputing stuff. It would also need @rijobro 's modifications (not commited yet) to the post/pre stuff that makes that far more efficient.

ashgillman:
That is an interesting proposal. I agree it potentially makes for a much more flexible framework.

But it does seem to require a large amount of refactoring. Would you consider for this PR having a KOSMAPOSL, which derives OSMAPOSL with the one override (we think, Daniel is currently attempting an implementation), with the intent for this to be abstracted and/or refactored in the future. Currently, I think we are getting close to a "minimal diff" PR. Implementing such changes could significantly increase the number of changes.

KrisThielemans:
not 100% sure how KOSMAPOSL would look like. are you saying you have a KOSMAPOSL and the current HKEM objective function? that'be rather weird.

can you give a brief pseudo-code implementation?

ashgillman:
I guess we propose that in KOSMAPOSL, an objective function can optimise an image surrogate rather than the image directly. The HKEM formulation is synonymous with MLEM, but optimises the coef image instead of the activity estimate directly.

Hi Kris,

Daniel and I had another chat this morning and I believe we are on the same page.

The proposed KOSMAPOSL inherits OSMAPOSL, and overrides `update_estimate()`. We believe then we should be able to use the standard PoissonLog... objective function. `KOSMAPOSL.update_estimate()` must apply the kernel in 3 ways:
  - Supply the kernel applied to the alpha as current estimate to Obj Fun (actually this is free, as this is just the emission estimate).
  - Apply the kernel to the calculated gradient (`multiplicative_update_image_ptr`) to get a kernelised gradient.
  - Apply the kernel to the sensitivity, to get a kernelised sensitivity (but without actually modifying any state, this would be a temporary variable).

This would also mean all parameters would have to be moved to the reconstructor rather than the obj fun.

To us, this idea seems good, and should be compatible with SIRF, but it might take a little bit of coding so we thought best to get some feedback before comitting. Any thoughts much appreciated :)

Regarding your proposal, would this require us to wait for @rijobro's modifications to be merged?

KrisThielemans:
I'm not sure if you can write the new `update_estimate` in terms of the old one. if you have to copy all that, it seems like a bad idea. would need to write some formulas...

ashgillman:
Okay, so we had a meeting with Harry this morning and I think the following
rather long message summarises the conversation.

There are two dimensions to decide upon for the implementation:
1. where HKEM fits and
2. how high in the class hierarchy.

# Where HKEM fits.
## 1. An Objective Function
This is similar to the current implementation.

algo

In this picture, the objective function is responsible for the circular HKEM
update loop. Although it doesn't actually apply the updates in the bottom, but
when the Reconstructor asks for the `gradient` and the `sensitivity`, the
Objective Function supplies kernelised versions.

There seem to be two objections to the current implementation:
1. It modifies the sensitivity, which is unusual for a objective function
    - However, I don't actually think this is an issue. The objective function
      seems to be "in charge" of handling the sensitivity image, so I see no issue
      with it modifying this.
2. It means that the Reconstructor's `current_estimate` is actually a
   coefficient image, rather than the emission image. This won't wrap well in
   SIRF.
   - A solution to this is to also modify OSMAPOSLReconstructor (and
     IterativeReconstructor?) to apply the kernel

## 2. A Reconstructor
In this implementation, we would implement the kernel handling into the
reconstructor, and use a normal objecive function. The details are outlined in
detail in a subsequent section, but notionally:

algo

We pass the emission image:
image
to the Obj Fun, and get back the gradient,
image
and sensitivity
image
We then calculate (in `update_estimate()`) kernelised versions of each:
image
image
We then update by calculating coef image, and unkernelise
image
to emission image.

## 3. A projector
We could also potentially implement the method with modified projectors,
forward:
image
and backward:
image
This would also, as per option (1), require modifications to the Reconstructor
to apply the kernel.

I think this is what Kris has suggested.

# How high in the class hierarchy
Currently, the implementation would sit at the bottom of the class hierarchy
(regardless of where implemented).

However, Harry suggests that actually, HKEM/KEM is a generalisation on OSEM,
where OSEM is KEM with an identity kernel. So perhaps, it makes sense to
implement the changes not in a specific inheriting kernelised class, but
actually higher up. This way, its functionality would be compatible with all
other STIR functionality (e.g., MCIR, Patlak recon, etc.).

# My Preference

Personally, I like sticking everything into the reconstructor, since it seems
that no matter what, this will have to be modified in order to apply the kernel
to the coefficient image to get the emission image.

I think Harry's suggestion is reasonable, perhaps instead of the KOSMAPOSL I
propose below the changes could be directly implemented in OSMAPOSL.

# Harry's preference

I think Harry likes keeping the current implementation (Obj Fun), but
implementing this higher in the hierarchy.

We have had a quick look at how this might work implementation-wise though and
it gets a bit tricky, as many Objective functions override and calculate their
own sensitivity and gradient calculations.

# KOSMAPOSLReconstructor implementation details
The reason I like modifying `update_estimate()` is that is seems to be a single
part of the code that touches everything needed for HKEM (i.e., the 3 areas
where the kernel is applied).

I think we can get away without duplicating all of `update_estimate()` by
abstracting these three points.

## 1. Gradient
replace
```
  this->objective_function().
    compute_sub_gradient_without_penalty_plus_sensitivity(
        *multiplicative_update_image_ptr, current_image_estimate, subset_num);
```
https://github.com/UCL/STIR/blob/master/src/iterative/OSMAPOSL/OSMAPOSLReconstruction.cxx#L415-L418
with
```
this->compute_sub_gradient_without_penalty_plus_sensitivity(
    *multiplicative_update_image_ptr, current_image_estimate, subset_num);
```
which would simply just call the relevant function from the objective function.
```
OSMAPOSLReconstruction::compute_sub_gradient_without_penalty_plus_sensitivity(
    T *multiplicative_update_image_ptr, T current_image_estimate, T subset_num) {
      this->objective_function().
        compute_sub_gradient_without_penalty_plus_sensitivity(
            *multiplicative_update_image_ptr, current_image_estimate, subset_num);
}
```
and allow us to override:
```
KOSMAPOSLReconstruction::compute_sub_gradient_without_penalty_plus_sensitivity(
    T *multiplicative_update_image_ptr, T current_image_estimate, T subset_num) {
      this->objective_function().
        compute_sub_gradient_without_penalty_plus_sensitivity(
            *multiplicative_update_image_ptr, current_image_estimate, subset_num);
     this->compute_kernelised_image(*multiplicative_update_image_ptr, *multiplicative_update_image_ptr, current_image_estimate)
}
```


## 2. Sensitivity
replace
```
    const TargetT& sensitivity = this->objective_function().get_subset_sensitivity(subset_num);
```
https://github.com/UCL/STIR/blob/master/src/iterative/OSMAPOSL/OSMAPOSLReconstruction.cxx#L422-L423
with
```
    const TargetT& sensitivity = this->compute_sensitivity(subset_num);
```
So now, `OSMAPOSL::compute_sensitivity()` runs the original line of code, and `KOSMAPOSL::compute_sensitivity()` will do that same but also apply the kernel again.


## 3. Update
Two options:
  - wrap around `OSMAPOSLReconsturctor::update_estimate()` and just update after
    the calculations
    OR
  - As with the last two, abstract
    https://github.com/UCL/STIR/blob/master/src/iterative/OSMAPOSL/OSMAPOSLReconstruction.cxx#L542-L551
    into `OSMAPOSLReconstructor::apply_multiplicative_update()`.

## Modified OSMAPOSLReconstructor
This would be very similar to the KOSMAPOSLReconstructor, just having some check
as to weather the kernelised flags have been set. In this way, methods
inheriting OSMAPOSLReconstructor get kernelised implementation free.

danieldeidda:
Following @ashgillman  very clear message, I think the easiest is probably option 2 and 3, but the 2 is probably easiest in terms of coding (although I might be wrong). As Ash said we probably could just include all the kernel related functions in the OSMAPOSLreconstruction.cxx and modify the update estimate function so that, when the kernel flag is activated, it will apply the kernel to the current_estimate, gradient and sensitivity, in this way we can keep KEM and HKEM together as one can be seen as a special case of the other. Another benefit is probably the fact that we would not need to implement a different class for LM reconstruction because we could use all the objective functions in STIR in theory.

KrisThielemans:
hi all, sorry for the delay.

Implementing things such that they are usable for other cases is clearly desirable. Note that `OSMAPOSL` doesn't depend on the objective function (it needs to be a Poisson with linear model of course), so modifying `OSMAPOSL` or deriving `KOSMAPOSL` from it will both allow it to work on listmode data. However, `OSMAPOSL` is templated in terms of `TargetT`. This is used for the Patlak stuff for instance. While you can of course implement a kernelised version of Patlak, I don't think this would be trivial to write generically (as the anatomical image would be 3D, while the parametric image is 4D). Indeed, you don't know what `TargetT` will be. No guarantees that it is an image in any shape or form.

So, sticking the kernelised stuff inside `OSMAPOSL` is not feasible (even though initially it looks like it's the least amount of work).

The current implementation of modifying the objective function fails for known reasons (it could be generalised to listmode etc by templating it, but that wouldn't solve the other problems).

So, we're left with either modifying the projector, or creating `KOSMAPOSL` as @ashgillman describes above. I prefer the former really because of its generality, but then again it cannot be fully templated either. Indeed, in current STIR, `GeneralisedObjectiveFunction` doesn't give you access to the projectors anyway.

So I guess @ashgillman's suggestion is best, especially as we have an extensive description. Modifications to `OSMAPOSL` are small and might make it a tiny bit more readable anyway. 2 comments: 
- rename `compute_sensitivity` to `get_subset_sensitivity` (no reason to change the name, and you could avoid recomputation in the non-hybrid case). 
- introduce `apply_multiplicative_update` (cleaner code)

Nevertheless, it might be worth just sketching the "projector-way". There's going to be 2 ways: one templated in the underlying algorithm, another with using it as a member. the latter might be easier than you think, see https://github.com/UCL/STIR/blob/4e0a2e89f2dfda87106cb13d14081b70d50dc0d4/src/recon_test/recontest.cxx#L70

ashgillman:
Hi Kris,

Thanks for the info.

Re: Projector
Sorry, I am not sure exactly how the code you link there would be relevant. If we implemented a custom projector for HKEM, would we not just pass its parameters to the Obj Fun? And then, how to instruct to apply the final kernel application for coef->emission image? Should the algorithm check which projector was used (perhaps a bit ugly in code)? Or should the user specify the final step (error prone, although little more so than applying unmatched projectors..)?

One suggestion I might have for this method is to apply it in a similar way to PSF (i.e., via filters):

Consider:
```
OSMAPOSLParameters :=
; example file for OSEM (21 subsets)

objective function type:= PoissonLogLikelihoodWithLinearModelForMeanAndProjData
PoissonLogLikelihoodWithLinearModelForMeanAndProjData Parameters:=

input file := ${INPUT}

; additive projection data to handle randoms etc
additive sinogram := ${ADDITIVE_SINOGRAM} 

zero end planes of segment 0:= 1

; very awkward way to specify PSF in both forward and back projector
projector pair type := Separate Projectors
 projector pair using separate projectors parameters :=
 forward projector type := Pre Smoothing
  Pre Smoothing Forward Projector Parameters:=
  original forward projector type := Matrix
    Forward Projector Using Matrix Parameters :=
    Matrix type := Ray Tracing
	number of rays in tangential direction to trace for each bin :=10
	do symmetry 90degrees min phi := 0
    End Forward Projector Using Matrix Parameters :=
  ; image-based blurring
  filter type :=  Separable Cartesian Metz
  Separable Cartesian Metz Filter Parameters :=
    x-dir filter FWHM (in mm):= 4.3
    y-dir filter FWHM (in mm):= 4.3
    z-dir filter FWHM (in mm):= 4.3
    x-dir filter Metz power:= 0.0
    y-dir filter Metz power:= 0.0
    z-dir filter Metz power:=0.0
    ; truncate it a bit to avoid negative lobes
    x-dir maximum kernel size := 5
    y-dir maximum kernel size := 5
    z-dir maximum kernel size := 1

  END Separable Cartesian Metz Filter Parameters :=
    
  End Pre Smoothing Forward Projector Parameters:=
 back projector type:= Post Smoothing
  Post Smoothing Back Projector Parameters:=
  original back projector type := Matrix
    Back Projector Using Matrix Parameters :=
    Matrix type := Ray Tracing
	number of rays in tangential direction to trace for each bin :=10
	do symmetry 90degrees min phi := 0
    End Back Projector Using Matrix Parameters :=
  ; image-based blurring
  filter type :=  Separable Cartesian Metz
  Separable Cartesian Metz Filter Parameters :=
    x-dir filter FWHM (in mm):= 4.3
    y-dir filter FWHM (in mm):= 4.3
    z-dir filter FWHM (in mm):= 4.3
    x-dir filter Metz power:= 0.0
    y-dir filter Metz power:= 0.0
    z-dir filter Metz power:=0.0
    ; truncate it a bit to avoid negative lobes
    x-dir maximum kernel size := 5
    y-dir maximum kernel size := 5
    z-dir maximum kernel size := 1

  END Separable Cartesian Metz Filter Parameters :=
  End Post Smoothing Back Projector Parameters:=
end projector pair using separate projectors parameters := 

; norm and acf
Bin Normalisation type := From ProjData
  Bin Normalisation From ProjData :=
   normalisation projdata filename:= ${MULTFACTOR}
  End Bin Normalisation From ProjData:=

; if the next parameters are enabled, 
; the sensitivity will be computed and saved
; use %d where you want the subset-number (a la printf)
; we do this here for illustration, but also for re-use later on (to save some time)
; CAREFUL: use correct number of subsets in name to avoid confusion
subset sensitivity filenames:= PSF_sense_%d.hv
recompute sensitivity := 1
use subset sensitivities := 1


zoom:= 1
xy output image size (in pixels) := 344

end PoissonLogLikelihoodWithLinearModelForMeanAndProjData Parameters:=

enforce initial positivity condition:=0
output filename prefix := ${OUTPUT_PSF}
number of subsets:= 21
start at subiteration number:= 1
start at subset:= 0
number of subiterations:= 210
save estimates at subiteration intervals:= 21

END :=
```

Since the kernel application is effectively a semi-local filter, we could implement it this way, and have a KernelFilter which is applied pre-forward, post-backward and post-recon. The complication here is that we would need actually a "stateful filter" that remembers the kernel and is able to update it. And all three filters would have to either be the same entity, or share their state. Actually this would be true for any projector-based implementation I guess.

In summary, I think to implement as a projector or filter, we'd need to solve two problems:
1. How to ensure the final step is applied (user specified, i.e., post filter, or special case check in algorithm)
2. How to share state between the projectors (so they all use the latest kernel estimate)

The positive of this approach is that it seems quite "mathematically satisfying", essentially hiding all the kernel bits into semi-local stateful filters.

ashgillman:
Oh and you have to have some stage also where you update the kernels. This would have to be separate, likely within the algorithm.

KrisThielemans:
this is along different lines than I was thinking, whcih was more like this
```
Hybrid Kernel Reconstruction parameters:=
  ; original recon method for kernel weights (alpha)
  reconstruction type := bla
  bla Parameters :=
   input:=
  etc etc
 End bla Parameters :=

  ; some parameters for kernel

 num sub-iterations:=
 output := ...
End:=
```
So now you have `HKRecon::set_up() that gets the projectors from `bla::get_objective_function()`, and add the pre/post stuff to it. `HKRecon::update_estimate` which calls `bla::update_estimate` (which updates `alpha`) and then constructs the image. For the hybrid method, it then updates the kernel (and therefore presumably the pre/post methods, although if these happen to use shared ptrs, it'd come for free (but quite dangerously so).

maybe not much else to do...

it'd have to be understood that any filtering/penalty-stuff specified in the `bla` recon would apply on the `alpha` image. Any filters specified in `HKRecon` would apply on the image.

you could then have different ways to construct the kernels (from MR, hybrid, fixed for the "sieves" method), but the rest of the algorithm remains the same.

HKEM would simply set `bla=OSMAPOSL`.

ashgillman:
Faster than before or faster than OSMAPOSL?

danieldeidda:
faster than before :D
OSEM is much faster

ashgillman:
Great, so I'll do another code review one evening this week, hopefully tomorrow :)

ashgillman:
Just wondering about the `only_2D` stuff. Does OSMAPOSL do 2D recons? In KOSMAPOSL, are we perhaps able to determine this from the data rather than from the parameter file?

danieldeidda:
Yes  OSMAPOSL is able to reconstruct 2D data. I am not sure whether there is a parameter which is associated with 2D recon.

KrisThielemans:
`only_2D` is used in the quadratic penalty to construct a penalty only in the transverse plane. (which is necessary for a single slice recon of course). It's not used anywhere else. If you don't use it, get rid of it.

danieldeidda:
I do use it, in a similar way as I used it for PLS. If activated the calculation of the kernel is adapted to a single slice.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1119 @@
...
+    else{
+      full_compute_kernelised_image (kImage, Image,
+                               current_estimate);
+}
+}
+
+template <typename TargetT>
+void 
+KOSMAPOSLReconstruction<TargetT>::
+update_estimate(TargetT &current_alpha_coefficent_image)
```
ashgillman: I think the point of the refactoring of OSMAPOSLReconstruction is so that we wouldn't have to modify update_estimate. Did you run into trouble with the other approach?

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -0,0 +1,259 @@
...
+  void set_write_update_image(const int);
+
+
+  //! prompts the user to enter parameter values manually
+  virtual void ask_parameters();
+
+ protected:
+  //! Filename with input projection data
+  std::string input_filename,kernelised_output_filename_prefix;
+  std::string current_kimage_filename;
```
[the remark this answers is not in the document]
danieldeidda: I had thought that too when we discussed but then I tried and the results are different. This is because we thought that KOSMAPOSL would just use the lambda image, but in this way the kernel will depend on lambda instead of alpha, it is like writing lambda=f(lambda) instead of  lambda=f(alpha). It is necessary to apply the extra step if we want the PET image, that is why we can avoid printing the alpha from the parfile and just print the PET image using the "current_kimage_filename".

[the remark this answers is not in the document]
ashgillman: Hmmm, I think I see the issue. The proposal was essentially to "mask out" alpha and pass lambda back and forth. But we need alpha to calculate the kernel, and also to apply the update to. Perhaps we can hide alpha inside KOSMAPOSL as some state. But then the `update_estimate` works a little differently to how OSMAPOSL uses it.

src/iterative/KOSMAPOSL/KOSMAPOSL.cxx
```diff
@@ -0,0 +1,77 @@
...
+*/
+/*!
+
+  \file
+  \ingroup main_programs
+  \brief main() for KOSMAPOSLReconstruction
+
+  \author Daniel Deidda
+  \author Ashley Gillman
+  \author Matthew Jacobson
```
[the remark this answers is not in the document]
danieldeidda: I don't know let's see what @KrisThielemans says

ashgillman:
I'd propose the following further changes. Let me know if you have thoughts @danieldeidda. This is all with the ultimate goal of not having to reimplement `update_estimate()`

1. It seems a little awkward in `KOSMAPOSLReconstruction` passing `current_alpha_coefficient_image` around everywhere, especially given that the kernel depends on it. Actually it means we'd have to pass the alpha image e.g., into the `get_subset_sensitivity()`, which would be weird in `OSMAPOSLReconstruction`. Perhaps it makes sense to modify `IterativeReconstruction::reconstruct` (`IterativeReconstruction.cxx L432`) to also have a `start_of_iteration_processing(*target_data_sptr)`. This could either precalculate the kernel, or save a reference to alpha for later on-the-fly kernel calculation.
2. Also, `IterativeReconstruction::end_of_iteration_processing()` could have a protected `maybe_save_current_estimate()` function refactored out for this part of the code. Then, this could be overwritten in `KOSMAPOSLReconstruction` to apply the kernel.
3. Perhaps something like `private KOSMAPOSLReconstruction::get_kernel_value(x, y, z)` could be added to simplify kernel code. But more importantly, kernel calculations would use a member reference to alpha rather than having it explicitly passed. This way we don't have to pass the kernel explicitly and can apply the kernel when we override `KOSMAPOSLReconstruction::compute_sub_gradient_without_penalty_plus_sensitivity()` and `KOSMAPOSLReconstruction::get_subset_sensitivity()`

I think with those three changes we may be able to remove `update_estimate()`

danieldeidda:
` It seems a little awkward in KOSMAPOSLReconstruction passing current_alpha_coefficient_image around everywhere, especially given that the kernel depends on it. Actually it means we'd have to pass the alpha image e.g., into the get_subset_sensitivity(), which would be weird in OSMAPOSLReconstruction. Perhaps it makes sense to modify IterativeReconstruction::reconstruct (IterativeReconstruction.cxx L432) to also have a start_of_iteration_processing(*target_data_sptr). This could either precalculate the kernel, or save a reference to alpha for later on-the-fly kernel calculation.`

Actually we can see KOSMAPOSL as the equivalent of OSMAPOSL in the feature space so I would think is fine to pass alpha. Regarding the sensitivity, we don't have to pass alpha to get_sens... actually the ksens is created only for the operation but the sensitivity itself will not change.

`Perhaps something like private KOSMAPOSLReconstruction::get_kernel_value(x, y, z) could be added to simplify kernel code. But more importantly, kernel calculations would use a member reference to alpha rather than having it explicitly passed. This way we don't have to pass the kernel explicitly and can apply the kernel when we override KOSMAPOSLReconstruction::compute_sub_gradient_without_penalty_plus_sensitivity() and KOSMAPOSLReconstruction::get_subset_sensitivity()`

I am not sure about the get_kernel_value(), the whole point of having the compact_compute_kernelised_image() was to avoid calculating the kernel matrix which will put in memory a matrix that is at least 27 times bigger than the image.

I would say if you can try this and it runs with no significant changes in terms of speed or memory, and of course the images between the two methods are exactly the same, the we can go for it.

ashgillman:
> Actually we can see KOSMAPOSL as the equivalent of OSMAPOSL in the feature space so I would think is fine to pass alpha. Regarding the sensitivity, we don't have to pass alpha to get_sens... actually the ksens is created only for the operation but the sensitivity itself will not change.

Yes, I agree to pass alpha image to `update_estimate()`. I'd like to override:

```
template <typename TargetT>
const TargetT&
OSMAPOSLReconstruction<TargetT>::get_subset_sensitivity(const int subset_num)
{
  return this->objective_function().get_subset_sensitivity(subset_num);
};
```

With:

```
template <typename TargetT>
const TargetT&
KOSMAPOSLReconstruction<TargetT>::get_subset_sensitivity(const int subset_num)
{
  // Calc normal sensitivity
  const TargetT& sensitivity = base_type::get_subset_sensitivity(subset_num);
  // Kernelise sensitivity
  unique_ptr< TargetT > ksens_ptr(sensitivity.get_empty_copy());
  compute_kernelised_image(*ksens_ptr, sensitivity,
                           current_alpha_coefficent_image);
  return ksens_ptr;
};
```
For example.

But I don't really want to have to modify `get_subset_sensitivity` to have an extra parameter `current_alpha_coefficent_image`. This could be avoided if `current_alpha_coefficent_image` was a private member.

Does this make sense?

> I am not sure about the get_kernel_value(), the whole point of having the compact_compute_kernelised_image() was to avoid calculating the kernel matrix which will put in memory a matrix that is at least 27 times bigger than the image.

I had hoped that `get_kernel_value()` might just cover this part of the code:
```
const int min_dx = max(distance[0][0].get_min_index(), min_x-x);
const int max_dx = min(distance[0][0].get_max_index(), max_x-x);


l=(z-min_z)*(max_x-min_x +1)*(max_y-min_y +1) + (y-min_y)*(max_x-min_x +1) + (x-min_x);


for (int dz=min_dz;dz<=max_dz;++dz)
  for (int dy=min_dy;dy<=max_dy;++dy)
    for (int dx=min_dx;dx<=max_dx;++dx)
      {
        m=(dz-min_dz)*(max_dx-min_dx +1)*(max_dy-min_dy +1) + (dy-min_dy)*(max_dx-min_dx +1) + (dx-min_dx);

        if (get_hybrid()){

            if(current_estimate[z][y][x]==0){
                continue;

            }
            else{

            kPET=exp(-(*this->kpnorm_sptr)[0][l][m]/square(current_estimate[z][y][x]*get_sigma_p())/2)*
                exp(-square(distance[dz][dy][dx]/grid_spacing.x ())/(2*square(get_sigma_dp())));
}

}
else{
    kPET=1;

}

kanatomical=exp(-(*this->kmnorm_sptr)[0][l][m]/square(anatomical_sd*sigma_m)/2)*
            exp(-square(distance[dz][dy][dx]/grid_spacing.x ())/(2*square(sigma_dm)));

kImage[z][y][x] += kanatomical*kPET*Image[z+dz][y+dy][x+dx];

pnkernel += kanatomical*kPET;
}
if(current_estimate[z][y][x]==0){
    continue;}

```
Which I think is shared. It wouldn't actually change any functionality (although it might make it easier in the future to have a `precalculate_kernel` option). I must admit i haven't completely thought through that part, but it wouldn't actually be necessary to remove the `update_image` override.

ashgillman:
It would be wise if you could check this diff @danieldeidda https://github.com/UCL/STIR/pull/186/commits/9e5bfbb793a7879aa622872e361e1a4777c9b098

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1162 @@
...
+            sqrt(square(x*grid_spacing.x())+
+                 square(y*grid_spacing.y())+
+                 square(z*grid_spacing.z()));
+        }
+
+  // get anatomical standard deviation over all voxels
+
+  // calculate kernelised image
+
+  const int min_z = (*anatomical_sptr).get_min_index();
```
ashgillman: Should `{min,max}_{x,y,z}` be calculated using `current_alpha_estimate` as in the full method?

ashgillman: @danieldeidda

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1162 @@
...
+                        + (dy-min_dy)*(max_dx-min_dx +1)
+                        + (dx-min_dx);
+
+                      if (get_hybrid()){
+                        if(current_alpha_estimate[z][y][x]==0){
+                          continue;
+                        }
+
+                        kPET
+                          = exp(-(*this->kpnorm_sptr)[0][l][m]
```
ashgillman: Why is the first index always 0? This matrix seems never to be used elsewhere, why have this index?

ashgillman: @danieldeidda

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: it is actually the same since the two images have the same dimensions. actually, it might be better to write as private member min and max because it is reused in different places

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: this is a 2D matrix, I remember having problems when I used array<float,2>, so i made it a targetT

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -676,284 +1064,226 @@ full_compute_kernelised_image(TargetT& kernelised_image_out,
...
 
+template <typename TargetT>
+double
+KOSMAPOSLReconstruction<TargetT>::
+calc_intensity_kernel_compact(int x, int y, int z,
+                              int dx, int dy, int dz,
+                              const TargetT& prior_image, double sigma,
+                              double precalc_denom) {
+  if (precalc_denom == 0) {
+    precalc_denom = prior_image[z][y][x];
```
danieldeidda: Do you do this only for the PET? because for the MR it should be the SD

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1487 @@
...
+
+// worker functions
+
+
+namespace { // priave namespace for internal functions
+
+unsigned int ravel_index(int x, int y, int z,
+                         int min_x, int min_y, int min_z,
+                         int max_x, int max_y, int max_z) {
+  unsigned int ravelled_index
```
ashgillman: Perhaps we shouldn't allocate a variable here and just return the value? Also, declare the function inline?

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1487 @@
...
+  return exp(-distance_sq / (2 * sigma*sigma));
+}
+double gaussian_kernel(double distance, double sigma) {
+  // std::cout << "gaussian_kernel(" << distance_sq << ", " << sigma << ")" << std::endl;
+  return exp(-distance*distance / (2 * sigma*sigma));
+}
+
+
+Array<3, float>
+precalculate_patch_euclidean_distances(int num_neighbours, bool only_2D,
```
ashgillman: Should probably return by reference? Isn't call from an inner loop though so shouldn't be causing your profiling issues

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1487 @@
...
+
+  const DiscretisedDensityOnCartesianGrid<3,float>* current_anatomical_cast =
+    dynamic_cast< const DiscretisedDensityOnCartesianGrid<3,float> *>
+      (this->get_anatomical_prior_sptr ().get());
+  // TODO - which spacing to use? Need both?
+  const CartesianCoordinate3D<float>& grid_spacing =
+    current_anatomical_cast->get_grid_spacing();
+
+  // std::cout << "a" << std::endl;
+  Array<3, float> distance
```
ashgillman: constant?

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1487 @@
...
+    = precalculate_patch_euclidean_distances(num_neighbours, only_2D, grid_spacing);
+  // std::cout << "b" << std::endl;
+
+  if (!use_compact_implementation && this->get_hybrid()) {
+    // Going to need the full PET regional normalised differences
+    int dimf_row = this->num_voxels;
```
ashgillman: const

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1487 @@
...
+    int dimf_col = this->num_non_zero_feat-1;
```
ashgillman: const

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -0,0 +1,312 @@
...
+  int subiteration_counter;
+  double anatomical_sd;
+  mutable Array<3,float> distance;
+  /*! Create a matrix containing the norm of the difference between two feature vectors, \f$ \|  \boldsymbol{z}^{(n)}_j-\boldsymbol{z}^{(n)}_l \| \f$. */
+  /*! This is done for the PET image which keeps changing*/
+    void  calculate_norm_matrix(TargetT &normp,
+                                const int &dimf_row,
+                                const int &dimf_col,
+                                const TargetT& pet,
+                                Array<3,float> distance);
```
ashgillman: pass by ref?

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -0,0 +1,312 @@
...
+  double calc_pet_kernel(int x, int y, int z,
+                         int min_x, int min_y, int min_z,
+                         int max_x, int max_y, int max_z,
+                         int dx, int dy, int dz,
+                         int min_dx, int min_dy, int min_dz,
+                         int max_dx, int max_dy, int max_dz,
+                         const TargetT& current_alpha_estimate,
+                         bool use_compact_implementation,
+                         const CartesianCoordinate3D<float>& grid_spacing,
+                         Array<3, float> distance);
```
ashgillman: pass by ref?

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -0,0 +1,312 @@
...
+
+  double calc_anatomical_kernel(int x, int y, int z,
+                                int min_x, int min_y, int min_z,
+                                int max_x, int max_y, int max_z,
+                                int dx, int dy, int dz,
+                                int min_dx, int min_dy, int min_dz,
+                                int max_dx, int max_dy, int max_dz,
+                                bool use_compact_implementation,
+                                const CartesianCoordinate3D<float>& grid_spacing,
+                                Array<3, float> distance);
```
ashgillman: pass by ref?

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1487 @@
...
+}
+
+/***************************************************************/
+// Here start the definition of few functions that calculate the SD of the anatomical image, a norm matrix and
+// finally the Kernelised image
+
+template<typename TargetT>
+void KOSMAPOSLReconstruction<TargetT>::
+calculate_norm_matrix(TargetT &normp, const int& dimf_row, const int& dimf_col,
+                      const TargetT& pet, Array<3,float> distance)
```
ashgillman: pass by ref?

KrisThielemans: definitely, and `const&` presumably

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
KrisThielemans: definitely, and `const&` presumably

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
KrisThielemans: definitely, and `const&` presumably

KrisThielemans:
regarding the sizes of the image. I think easiest for now to do a test that `has_same_characteristics` is true, if not abort, and mark it as a TODO (I would call `zoom_image`)

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
danieldeidda: OK I change all the call to distance and it is already significantly faster. Almost there!

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: what should be cost? num_non_zero_feat?

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: why constant? this should change for every voxel

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1493 @@
...
+            return false;
+        }
+    estimate_stand_dev_for_anatomical_image(this->anatomical_sd);
+
+    info(boost::format("SD from anatomical image calculated = '%1%'")
+         % this->anatomical_sd);
+
+
+    if(num_non_zero_feat>1){
+    shared_ptr<TargetT> normp_sptr(this->anatomical_sptr->get_empty_copy ());
```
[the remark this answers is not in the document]
danieldeidda: I am not sure to understand this, sorry

danieldeidda:
This size problem is quite weird. it seems that actually if you use xy output image size (in pixels) := -1 the reconstructed image is actually different from the simulated PET wich should be 256. Ha @KrisThielemans  suggests I added a check with has_same_characteristics() and error if false.

For now the best is probably to define the xy output image size (in pixels) equal to the MR size (256 for the 2D image I sent you)

danieldeidda:
With all the changes made we passed from 5.6x slower to less the 2x. I rerun the profiler and I can see that the function taking the most of the time is _ ieee 754 exp_fma. I presume this is the function for the calculation of exp(). This is called 637857684 times whereas in the old code is called 334 678 356 time.

I believe this is due to the new functions:
calc_anatomical_kernel()
calc_intensity_kernel_compact()
calc_pet_kernel()

KrisThielemans:
> xy output image size (in pixels) := -1 the reconstructed image is actually different from the simulated PET

not sure what you're saying but STIR recon doesn't know what you did for simulation of course. Default size is to cover the FOV, default voxel size is set by the arc-corrected bin size, which is a pain.

danieldeidda:
> xy output image size (in pixels) := -1 the reconstructed image is actually different from the simulated PET 

>  not sure what you're saying but STIR recon doesn't know what you did  for simulation of course. Default size is to cover the FOV, default  voxel size is set by the arc-corrected bin size, which is a pain.

I see. I was being silly than. I assumed that when you create a simulated data using forward_project this will keep somehow information from the true image I use.

danieldeidda:
> With all the changes made we passed from 5.6x slower to less the 2x. I rerun the profiler and I can see that the function taking the most of the time is _ ieee 754 exp_fma. I presume this is the function for the calculation of exp(). This is called 637857684 times whereas in the old code is called 334 678 356 time.
> 
> I believe this is due to the new functions:
> calc_anatomical_kernel()
> calc_intensity_kernel_compact()
> calc_pet_kernel()

I think  the problem is that this functions are called  hundreds of milion times in a loop. It is probably better to avoid these calls and perform all calculation directly

danieldeidda:
At the moment speed is still 30% lower than before but much better than 200%. I thinks some improvement can still be done. Suggestions are very welcome!

ashgillman:
I made a few comments, but I couldn't see anything too obvious performance-wise. Did you run it through the profiler again? Anything interesting?

ashgillman:
> I think the problem is that this functions are called hundreds of milion times in a loop. It is probably better to avoid these calls and perform all calculation directly in this way the run time is the same as previously.

I put these into a separate function to improve readability and reduce replicated code. If the functions are declared inline (or even otherwise), the compiler should try and compile the code without the function overhead (i.e., as if the code from the function was in place of the function call, like you say).

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1244 @@
...
+void KOSMAPOSLReconstruction<TargetT>::
+calculate_norm_const_matrix(TargetT &normm,
+                            const int& dimf_row, const int &dimf_col)
+{
+
+    calculate_norm_matrix(normm,dimf_row,dimf_col,*this->anatomical_prior_sptr);
+
+}
+template<typename TargetT>
+void KOSMAPOSLReconstruction<TargetT>::estimate_stand_dev_for_anatomical_image(double& SD)
```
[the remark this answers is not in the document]
danieldeidda: this creates problems

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -0,0 +1,1244 @@
...
+                                  }
+                               }
+                       }
+
+       SD= sqrt(kStand_dev / (nv-1));
+}
+
+
+
+//template<typename TargetT>
```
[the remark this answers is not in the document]
danieldeidda: sure

danieldeidda:
> I made a few comments, but I couldn't see anything too obvious performance-wise. Did you run it through the profiler again? Anything interesting?

the profiler says that there are a number of calls to exp() which is double to the previous version. But honestly it does not seem from the code. Any idea at all?

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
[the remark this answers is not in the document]
KrisThielemans: you seem to use it to return a double. probably better to do that as
```
double estimate_stddev() const
```

KrisThielemans:
I'd trust the profiler...

ashgillman:
Great work!

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -0,0 +1,291 @@
...
+
+  hybrid:=1
+  sigma m:= 1                                ;is the parameter $\f \sigma_{m} \f$;
+  sigma p:=1                                 ;is the parameter $\f \sigma_{p} \f$;
+  sigma dm:=1                                ;is the parameter $\f \sigma_{dm} \f$;
+  sigma dp:=1                                ;is the parameter $\f \sigma_{dp} \f$;
+  number of neighbours:= 3                   ;is the cubic root of the number of voxels in the neighbourhood;
+  anatomical image filenames:={filename1,filename2}        ;are the filenames of the anatomical images, currently can only do one image;
+  number of non-zero feature elements:=1     ;is the number of non zero elements in the feature vector;
+  only_2D:=0                                 ;=1 if you want to reconstruct 2D images;
```
KrisThielemans: as above

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -0,0 +1,291 @@
...
+  /*! Create a matrix similarly to calculate_norm_matrix() but this is done for the anatomical image, */
+  /*! which does not  change over iteration.*/
+    void calculate_norm_const_matrix(TargetT &normm,
+                                const int dimf_row,
+                                const int dimf_col);
+
+  /*! Estimate the SD of the anatomical image to be used as normalisation for the feature vector */
+    double estimate_stand_dev_for_anatomical_image();
+
+  /*! Compute for each voxel, jl, of the emission image the linear combination between the coefficient \f$ \alpha_{jl} \f$ and the kernel matrix \f$ k_{jl} \f$\f$ */
```
KrisThielemans: get rid of these lines?
in any case, the remaining doxygen lines will likely mess up the doxygen doc

src/include/stir/OSMAPOSL/OSMAPOSLReconstruction.h
```diff
@@ -159,6 +162,9 @@ class OSMAPOSLReconstruction:
 
  protected:
 
+  //! operations prior to the iterations
```
KrisThielemans: let's delete this now

src/include/stir/OSMAPOSL/OSMAPOSLReconstruction.h
```diff
@@ -194,15 +200,28 @@ class OSMAPOSLReconstruction:
...
+  virtual const TargetT& get_subset_sensitivity(const int subset_num);
+
+  virtual void apply_multiplicative_update(
+    TargetT& current_image_estimate, const TargetT& multiplicative_update_image);
  
 private:
+
   friend void do_sensitivity(const char * const par_filename);
 
+  //! the principal operations for updating the image iterates at each iteration
```
KrisThielemans: let's delete this now

src/iterative/KOSMAPOSL/KOSMAPOSL.cxx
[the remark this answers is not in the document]
KrisThielemans: sure, but she's already gone? If so, resolve this please

examples/samples/KOSMAPOSL_hkem.par
```diff
@@ -0,0 +1,106 @@
+KOSMAPOSLParameters :=
+; Example file for using [Hybrid] Kernelized Expectation Maximisation (HKEM or KEM).
+; See documentation of KOSMAPOSLreconstruction
+; for more info.
+
+;the following disable the alpha coefficient output:
```
KrisThielemans:
```suggestion
;the following disables the alpha coefficient output:
```

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -0,0 +1,291 @@
+//
+//
+/*
+    Copyright (C) 2000 PARAPET partners
```
KrisThielemans: I think it would be fair to remove copyright statements upto 2012 (and the PARAPET one). There is hardly (no?) code left from OSMAPOSL.

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -0,0 +1,291 @@
...
+    See STIR/LICENSE.txt for details
+*/
+/*!
+  \file
+  \ingroup OSMAPOSL
+  \brief Declaration of class stir::KOSMAPOSLReconstruction
+
+  \author Daniel Deidda
+  \author Ashley Gillman
+  \author Matthew Jacobson
```
KrisThielemans: as above. probably remove Matthew and PARAPET here

PR #189, opened 2018-05-13
Refactoring towards one class per translation unit for listmode_buildblock

NikEfth:
Motivation: 
1. Building shared libraries
2. Issue #6 

I saw that compiling shared libraries without ROOT support was possible. 
So I started looking for the problem with ROOT and I ended changing some things (for the better I hope).
Steps: 
1. The CListTime and CListEvent have their own headers, instead of being in CListRecord.  
2. Likewise CListTimeROOT and CListEventROOT. 
3. Cleaned unnecessary includes and clean up

 Pls review and if you prefer it this <del>was</del> way I will carry on with this.


NikEfth:
Travis failed with: `Error: Failed to download resource "gcc"`

src/include/stir/listmode/CListEvent.h
```diff
@@ -0,0 +1,123 @@
+//
+//
+/*!
+  \file
+  \ingroup listmode
+  \brief Declarations of classes stir::CListRecord, stir::CListTime and stir::CListEvent which
```
KrisThielemans: needs adapting

src/include/stir/listmode/CListRecordROOT.h
```diff
@@ -28,134 +28,45 @@
...
-#include "stir/listmode/CListEventCylindricalScannerWithDiscreteDetectors.h"
+#include "stir/listmode/CListEventROOT.h"
+#include "stir/listmode/CListTimeROOT.h"
 #include "stir/Succeeded.h"
-#include "stir/round.h"
-#include "boost/static_assert.hpp"
-#include "stir/DetectionPositionPair.h"
-#include <boost/cstdint.hpp>
+#include "stir/Scanner.h"
+#include <boost/shared_ptr.hpp>
```
KrisThielemans: cannot use boost/shared_ptr as we're moving to std::shared_ptr. in fact, we're probably not anyway as we don't have boost:: below. So remove this include

src/include/stir/listmode/CListRecordROOT.h
```diff
@@ -28,134 +28,45 @@
...
+    bool is_time() const;
     //! Returns always true
-    bool inline is_event() const;
+    bool is_event() const;
     //! Returns always true
-    bool inline is_full_event() const;
+     inline bool is_full_event() const;
 
-    virtual CListEventROOT&  event()
+    virtual CListEvent&  event()
```
KrisThielemans: better to keep this as it was (and all similar functions below) I believe. This should fall under the "covariant return types" rule for virtual functions a far as I understand. (if it compiled anyway...)

src/include/stir/listmode/CListRecordROOT.h
```diff
@@ -165,36 +76,17 @@ class CListRecordROOT : public CListRecord // currently no gating yet
...
-        raw[1] = event2;
-
-        return Succeeded::yes;
-    }
-
+                                      const int& event1, const int& event2);
 private:
-    CListEventROOT  event_data;
-    CListTimeROOT   time_data;
+    CListEventROOT*  event_data;
```
KrisThielemans: this looks terrible. a raw pointer? why?

src/include/stir/listmode/CListEvent.h
NikEfth: I think I have adapted most doxygen strings.

src/include/stir/listmode/CListRecordROOT.h
NikEfth: reverted. But I have to be honest I cannot understand how this works.

src/include/stir/listmode/CListRecordROOT.h
NikEfth: Sure :S is was only temporal

NikEfth:
Two major points left: 
1. Should we add Travis checks for shared libraries? 
2. Should I go ahead do the same for the local folder?

NikEfth:
One travis failed to download root. All the others were alright.

src/include/stir/listmode/CListRecordROOT.h
```diff
@@ -28,135 +28,33 @@
...
-    //! \brief timeA
-    //! \details The detection time of the first of the two photons, in seconds
-    double timeA;
-
-    //!
-    //! \brief timeB
-    //! \details The detection time of the second of the two photons
-    double timeB;
-};
+using namespace std;
```
KrisThielemans: cannot do this in an include file. use `std::` explicitly

NikEfth:
I believe this is ok. I did a further review over all modifications and I did not find any weird stuff.

KrisThielemans:
@nikefth, this PR is mostly obsolete due to #327. @danieldeidda moved most of `CList*` to `List*` to be able to create SPECT  versions. He followed your lead and put `ListEvent` etc in separate files, therefore essentially moving things around. Are you aware of anything that should be kept from this PR?

NikEfth:
no I don't think that there is more to it.

KrisThielemans:
ok. I'll close this then. thanks for the work. It pointed the way forward!

Issue #96, opened 2017-08-27
ROOT/GATE support needs further work


KrisThielemans:
- [x] Header format is only partially documented (not for ECAT for instance)
     - done via `examples/samples/*hroot`
- [x] InputStreamFromROOT constructors don't see all fields (should also do stuff from post_processing(), but that's a virtual function so cannot call this from a constructor)
    - done by moving checks to `set_up()`
- [x] parsing of user-defined scanner has no checks nor defaults
     - done for instance in #569. Created #576 for the rest
- [ ] Issue #95

KrisThielemans:
see also #79

KrisThielemans:
closing this issue as only remaining items have their own issues

Issue #106, opened 2017-09-15
ROOT file in different directory


KrisThielemans:
The current code doesn't pass the directory of the hroot file to the code that actually reads it. Therefore, it will just try to read the .root file in the current directory, not the directory of the .hroot file.

KrisThielemans:
It seems I can only fix this if the `name of data file` keyword is moved out of `InputStreamFromROOTFile` into `CListModeDataRoot`. I think we need to
- move code in `InputStreamFromROOTFile::post_processing` to a new function `InputStreamFromROOTFile::open(const string& filename)`.
- move `name of data file` 
- add a call to `InputStreamFromROOTFile::open` in `CListModeDataRoot::open_lm_file`
- add relative-path code (from Interfile) to `CListModeDataRoot`.

However, this will break the parsing of existing hroot files. We could leave the current option is for backwards-compatibiliity if really required (not sure how many people this would affect, ROOT support wasn't yet in STIR 3.0).

@NikEfth, any suggestions?

Issue #95, opened 2017-08-27
ROOT/GATE support ignores scanner geometry if a known system is used


KrisThielemans:
If the `originating_system` is not `user_defined_scanner`, `CListModeDataROOT` ignores scanner keywords such as `inner_ring_diameter` etc. It should at least check consistency.

PR #190, opened 2018-05-13
Tests introduced in CListModeDataROOT for the Scanner geometry

### LINKED ISSUE #79: keyword parsing in .hroot files
`CListModeDataROOT`  initialises the scanner's `num_rings` and `num_detectors_per_ring` from the module et al keywords (read by `InputStreamFromROOTFileForCylindricalPET` et al.), but it initialises the max_ring_difference  and num_views parameters for its `ProjDataInfoCylindricalNoArcCorr` object from independent `num_rings` and `num_detectors_per_ring` members (parsed from the corresponding keywords in the .hroot).

This seems very dangerous to me. I think we should remove the explicit keywords from the .hroot file.

@NikEfth, I think this is a left-over from the (currently disabled) code to do axial compression and view-mashing directly when reading the ROOT-file, but that should use different keywords in any case (if we ever enable it, I don't really see why we would, except for testing of the list-mode recon).

### LINKED ISSUE #80: InputStreamFromROOTFileForCylindricalPET uses wrong conventions for blocks and buckets?
Terminology in STIR (from Scanner.h) :
```
      \li \c crystal the smallest detection unit
      \li \c block several crystals are grouped in a block, this can be in
          3 dimensions (see layer). This information could be useful for finding the
          geometry of the scanner, but we would plenty more
          size info presumably.
      \li \c layer Some scanners have multiple layers of detectors to give
          Depth Of Interaction information
      \li \c bucket several \c blocks send detected events to one \c bucket.
          This has consequences for the dead-time modelling. For instance,
          one bucket could have much more singles than another, and hence
          presumably higher singles-dead-time.
      \li \c singles_unit (non-standard terminology)
          Most scanners report the singles detected during the acquisition.
          Some scanners (such as GE scanners) report singles for every crystal,
          while others (such as CTI scanners) give only singles for a 
          collection of blocks. A \c singles_unit is then a set of crystals
          for which we can get singles rates.
```
In [GATE CylindricalPET terminology](http://wiki.opengatecollaboration.org/index.php/Users_Guide_V7.2:Defining_a_system#CylindricalPET), this seems to lead naturally to
- Module = bucket
- submodule = block

with number of modules per RSector=1. (I think GATE would be able to have one extra level compared to STIR).

This seems consistent with `InputStreamFromROOTFileForCylindricalPET::get_num_axial_crystals_per_singles_unit`. However, this is different from other functions implemented in InputStreamFromROOTFileForCylindricalPET.inl, e.g.
```
int get_num_dets_per_ring() const
{
    return static_cast<int>( this->crystal_repeater_z * this->module_repeater_z *
                             this->submodule_repeater_z);
}
int get_num_axial_blocks_per_bucket_v() const
{
    return this->module_repeater_z;
}
int get_num_axial_crystals_per_block_v() const
{
    return static_cast<int>(this->crystal_repeater_z *
       this->module_repeater_z);
}
```
This seems to lead to some strange arrangement. I think the correct way is
```
int get_num_axial_blocks_per_bucket_v() const
{
    return this->submodule_repeater_z;
}
int get_num_axial_crystals_per_block_v() const
{
    return static_cast<int>(this->crystal_repeater_z );
}
```
@NikEfth what do you think?

Of course, I wouldn't know what to do with the RSector stuff.

NikEfth:
* closes #79 
<del>* In addition, fixes an unreported bug in `InputStreamFromROOTFileForCylindricalPET.inl`, where the `num_rings` got returned instead of `num_det_per_ring`. <\del>
* closes #80


NikEfth:
I actually have to check if #80 is fully covered.

NikEfth:
Random error `Error: Failed to download resource "gcc"`

NikEfth:
These check boxes will close #96 
- [ ] Header format is only partially documented (not for ECAT for instance)
- [ ] InputStreamFromROOT constructors don't see all fields (should also do stuff from post_processing(), but that's a virtual function so cannot call this from a constructor)
- [ ] parsing of user-defined scanner has no checks nor defaults
- [x] Issue #95

NikEfth:
> parsing of user-defined scanner has no checks nor defaults
I believe that this has been at least partially addressed. Setting an arbitrary default I think is a bad option. 
I will leave this unchecked for now

src/include/stir/IO/InputStreamFromROOTFileForCylindricalPET.h
```diff
@@ -1,10 +1,3 @@
-/*!
```
KrisThielemans: need to keep this. this is the doc for the file, which is spearate for the one for the class

src/include/stir/IO/InputStreamFromROOTFileForCylindricalPET.h
```diff
@@ -30,6 +23,52 @@
 
 START_NAMESPACE_STIR
 
+/*!
+  \file
```
KrisThielemans: remove `\file` here. this is the doc for the class

src/include/stir/IO/InputStreamFromROOTFileForECATPET.h
```diff
@@ -1,10 +1,3 @@
-/*!
-  \file
```
KrisThielemans: same as above

src/include/stir/IO/InputStreamFromROOTFileForECATPET.h
```diff
@@ -29,6 +22,48 @@
 
 START_NAMESPACE_STIR
 
+/*!
+  \file
```
KrisThielemans: same as above

src/listmode_buildblock/CListModeDataROOT.cxx
```diff
@@ -70,27 +73,39 @@ CListModeDataROOT(const std::string& hroot_filename)
...
         {
-            error(boost::format("Unknown value for originating_system keyword: '%s. Abort.") % originating_system );
+            warning(boost::format("CListModeDataROOT: Unknown value for originating_system keyword: '%s.\n WIll try to "
+                                  "figure out the scanner's geometry from the parameters") % originating_system );
+            give_it_a_try = true;
         }
-        warning("I've set the scanner from STIR settings and ignored values in the hroot header.");
-        // TODO at least check consistency
+        else
+            info("CListModeDataROOT: I've set the scanner from STIR settings and ignored values in the hroot header.");
```
KrisThielemans: looks like this should be `warning`, not `info`

src/include/stir/listmode/CListModeDataROOT.h
```diff
@@ -51,6 +51,20 @@ START_NAMESPACE_STIR
...
+  Scanner::User_defined_scanner and set the members: num_rings, num_detectors_per_ring,
+  max_num_non_arccorrected_bins, inner_ring_diameter, average_depth_of_interaction,
+  ring_spacing and bin_size, manually.
+
+  In any case this scanner must checkout the geometry created by the repeaters section
+  (which will initialise the InputStreamFromROOTFile.
+  And of course, match the Scanner used by other parts of STIR (e.g. ProjDataInfo e.t.c).
+
+  Finally, in the case that Scanner::Unknown_scanner the algorithm will use the flag
+  give_it_a_try and will try to create the scanner from the members. Subject to concistency
```
KrisThielemans: I wouldn't mention `give_it_a_try` in the class doc, as it's an internal variable, and the reader might think it's a flag she can set.

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -76,13 +76,32 @@ InputStreamFromROOTFile::post_processing()
...
     stream_ptr->SetBranchAddress("time2", &time2);
-
+    stream_ptr->SetBranchAddress("axialPos",&axialPos);
     stream_ptr->SetBranchAddress("eventID1",&eventID1);
     stream_ptr->SetBranchAddress("eventID2",&eventID2);
     stream_ptr->SetBranchAddress("energy1", &energy1);
     stream_ptr->SetBranchAddress("energy2", &energy2);
     stream_ptr->SetBranchAddress("comptonPhantom1", &comptonphantom1);
     stream_ptr->SetBranchAddress("comptonPhantom2", &comptonphantom2);
+    stream_ptr->SetBranchAddress("globalPosX1",&globalPosX1);
```
KrisThielemans: why are we reading all these? can only take extra time. should it be optional?

KrisThielemans:
@eliseemond, maybe you should try this on one of your files to see if it still works. I would guess that output has to be identical.

KrisThielemans:
by the way, appveyor failure is likely timeout.

src/IO/InputStreamFromROOTFile.cxx
KrisThielemans: I can't remember what I said :-(

I think it's a useful feature but presumably would have to be optional. I don't know how ROOT works, but I can only imagine that it's going to slow down the reading. Also, it means that we'd handle only ROOT files where all these fields are present. (I have no clue if you can switch them off in GATE but I would think you could to compactify the ROOT file that way).

I'd be happy enough to comment it out for the moment, or introduce an extra bool to switch on/off the reading.

I

KrisThielemans: I had a look at my comment

> InputStreamFromROOT constructors don't see all fields

I think what I meant that the constructor can be used to set various class variables, but not all (only `singles_readout_depth`?). As there are no `set_` functions, it means that the class can only be used via parsing. Not a major problem, but it's just "asymmetric".

KrisThielemans:
I spotted a few more problems.

- `singles_readout_depth` is not in your `verbatim` example in the doxygen header
- `InputStreamFromROOTFile::set_defaults` is empty, and the default constructor doesn't set most variables. This means that the parsed variables are undefined if they didn't occur in the .hroot file (and we have no way of telling the parser that they are required keywords). It seems that it'd be best to set them in `set_defaults` and call this from the default constructor.  The same applies for `InputStreamFromROOTFileForCylindricalPET` etc.
- `InputStreamFromROOTFileForCylindricalPET::set_defaults` could set some values to -1, and `post_processing` (or `set_up`?) could check for it.

Most of this is some work. Feel free to postpone (and just create a new issue for it)

KrisThielemans:
this will need a merge with master now, after accepting #106 as it generates conflicts.

src/IO/InputStreamFromROOTFile.cxx
NikEfth: :ok: I created a set function and added the option to activate all fields

NikEfth:
I think I applied all your suggestions.

src/listmode_buildblock/CListModeDataROOT.cxx
KrisThielemans: still to do

src/listmode_buildblock/CListModeDataROOT.cxx
```diff
@@ -75,27 +78,39 @@ CListModeDataROOT(const std::string& hroot_filename)
...
     }
-    else
+    // If the user provide a Scanner name then, the local variables will be ignored and the Scanner
+    // will be the selected.
+    else if (this->originating_system == "User_defined_scanner" ||
+             give_it_a_try)
     {
-        info("Trying to figure out the scanner geometry from the information "
+        warning("CListModeDataROOT: Trying to figure out the scanner geometry from the information "
```
KrisThielemans: possibly this should be `info`?

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -89,14 +98,36 @@ InputStreamFromROOTFile::set_up(const std::string & header_path)
 
     stream_ptr->SetBranchAddress("time1", &time1);
     stream_ptr->SetBranchAddress("time2", &time2);
-
+    stream_ptr->SetBranchAddress("axialPos",&axialPos);
```
KrisThielemans: why is this here? Is it used at all? I can't find any reference to it. should it be in the list below of optional values?

src/include/stir/IO/InputStreamFromROOTFileForCylindricalPET.inl
```diff
@@ -106,4 +106,64 @@ get_num_trans_crystals_per_singles_unit() const
     return 0;
 }
 
+void
+InputStreamFromROOTFileForCylindricalPET::set_crystal_repeater_x(int& val)
```
KrisThielemans: all of these need to be `int val`

src/include/stir/IO/InputStreamFromROOTFileForECATPET.h
```diff
@@ -82,6 +123,12 @@ class InputStreamFromROOTFileForECATPET : public
     //! Get the number of crystals per block
     inline virtual int get_num_trans_crystals_per_singles_unit() const;
 
+    inline void set_crystal_repeater_x(int&);
```
KrisThielemans: all of these need to be `int`

KrisThielemans:
I'm merging master back into here now. Somewhat non-trivial due to changes in `CListModeData` there. Only file affected seems `CListModeDataROOT.cxx`. @NikEfth, you'll have to check with something like
```
git diff 915b0dd0a3159b2f1820d91145660d938d20478d -- ./src/*/*ROOT*cxx
git diff 915b0dd0a3159b2f1820d91145660d938d20478d -- ./src/include/stir/*/*ROOT*
```

src/listmode_buildblock/CListModeDataROOT.cxx
KrisThielemans: let's leave this

KrisThielemans:
@NikEfth, I've fixed the remaining (tiny) issues, including `set_crystal_repeater_x(int& val)`. As far as I'm concerned, this is good to go. Please check my changes though.

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -89,14 +99,36 @@ InputStreamFromROOTFile::set_up(const std::string & header_path)
 
     stream_ptr->SetBranchAddress("time1", &time1);
     stream_ptr->SetBranchAddress("time2", &time2);
-
-    stream_ptr->SetBranchAddress("eventID1",&event1);
-    stream_ptr->SetBranchAddress("eventID2",&event2);
+    stream_ptr->SetBranchAddress("axialPos",&axialPos);
```
KrisThielemans: I think this should be in the optional ones.

KrisThielemans:
found a few other `set_` functions and fixed those. One outstanding question on `axialPos`.

Note that Travis on one of the OSX versions seems to fail with a time-out in compiling ROOT. we'll ignore this.

src/IO/InputStreamFromROOTFile.cxx
NikEfth: moved few lines below

NikEfth:
Hi Kris, thank you for this. I had forgot that there where outstanding requests here. Everything looks fine.
Cheers,

PR #191, opened 2018-05-14
fixes ROOT file in different directory #106

### LINKED ISSUE #106: ROOT file in different directory

NikEfth:
Introduced virtual InputStreamFromROOTFile::set_up(const std::string&). The function takes over functionality previously in post_processing(). But in addition the full path of the header file location can be used. 

FilePath is used for the merging of the strings. As this is the first time the class is used small modifications had to be made to make the functionality relevant. 

1. new function get_as_string() const; which returns the current my_string, wether it is path or file. 
2. If my_string has just a filename, function get_path_only) returns an empty string.

### DIFF
--- a/src/IO/InputStreamFromROOTFile.cxx
+++ b/src/IO/InputStreamFromROOTFile.cxx
@@ -71,12 +71,12 @@ InputStreamFromROOTFile::post_processing()
 Succeeded
 InputStreamFromROOTFile::set_up(const std::string & header_path)
 {
-    // Read the 4 bytes to check whether this is a ROOT file, indeed.
+
     FilePath f(filename,false);
     f.prepend_directory_name(header_path);
 
-    std::string fullfilename = f.get_as_string();
-
+    const std::string fullfilename = f.get_as_string();
+    // Read the 4 bytes to check whether this is a ROOT file.
     FileSignature signature(fullfilename.c_str());
     if ( strncmp(signature.get_signature(), "root", 4) != 0)
     {
--- a/src/IO/InputStreamFromROOTFile.cxx
+++ b/src/IO/InputStreamFromROOTFile.cxx
@@ -19,6 +19,7 @@
 #include "stir/IO/InputStreamFromROOTFile.h"
 #include "stir/IO/FileSignature.h"
 #include "stir/error.h"
+#include "stir/FilePath.h"
 
 START_NAMESPACE_STIR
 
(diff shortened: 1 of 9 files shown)


NikEfth:
This check boxes will close #96 
- [ ] Header format is only partially documented (not for ECAT for instance)
- [ ] InputStreamFromROOT constructors don't see all fields (should also do stuff from post_processing(), but that's a virtual function so cannot call this from a constructor)
- [ ] parsing of user-defined scanner has no checks nor defaults
- [ ] Issue #95

src/IO/InputStreamFromROOTFileForCylindricalPET.cxx
```diff
@@ -155,8 +155,15 @@ InputStreamFromROOTFileForCylindricalPET::initialise_keymap()
 bool InputStreamFromROOTFileForCylindricalPET::
 post_processing()
 {
-    if (base_type::post_processing())
-        return true;
+    return false;
```
KrisThielemans: safer to keep these lines as they were. you never know if `base_type::post_processing` will do something

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -63,16 +64,28 @@ InputStreamFromROOTFile::initialise_keymap()
...
+
+Succeeded
+InputStreamFromROOTFile::set_up(const std::string & header_path)
 {
     // Read the 4 bytes to check whether this is a ROOT file, indeed.
-    FileSignature signature(filename.c_str());
+    FilePath f(filename,false);
+    f.prepend_directory_name(header_path);
+
+    std::string fullfilename = f.get_as_string();
```
KrisThielemans: could be `const std::string` I guess

src/listmode_buildblock/CListModeDataROOT.cxx
```diff
@@ -60,6 +61,9 @@ CListModeDataROOT(const std::string& hroot_filename)
     if(!this->parser.parse(hroot_filename.c_str()))
         error("CListModeDataROOT: error parsing '%s'", hroot_filename.c_str());
 
+    FilePath f(hroot_filename);
+    root_file_sptr->set_up( f.get_path_only());
```
KrisThielemans: should check return value

src/IO/InputStreamFromROOTFile.cxx
NikEfth: :ok:

src/IO/InputStreamFromROOTFileForCylindricalPET.cxx
NikEfth: :ok:

src/listmode_buildblock/CListModeDataROOT.cxx
NikEfth: :ok:

PR #192, opened 2018-05-14
write_to_file() moved from ProjDataInMemory to ProjData

### LINKED ISSUE #46: ProjDataInMemory::write_to_file()
Currently write_to_file is in ProjDataInMemory a better place would be the ProjData.

NikEfth:
closes #46 ProjDataInMemory::write_to_file()

### DIFF
--- a/src/include/stir/ProjData.h
+++ b/src/include/stir/ProjData.h
@@ -301,7 +301,9 @@ public:
   inline int get_num_sinograms() const;
   //! Get the total size of the data
   inline std::size_t size_all() const;
-  
+  //! writes info to a file in Interfile format
+  Succeeded write_to_file(const std::string& filename) const;
+
 protected:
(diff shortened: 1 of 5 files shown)


NikEfth:
One more instance of `Error: Failed to download resource "gcc"`

src/include/stir/ProjData.h
```diff
@@ -301,7 +301,9 @@ class ProjData : public ExamData
   inline int get_num_sinograms() const;
   //! Get the total size of the data
   inline std::size_t size_all() const;
-  
+  //! writes info to a file in Interfile format
```
KrisThielemans: I think `info` is a bit confusing as it sounds like it's only `ProjDataInfo`. Say `data` or so?

NikEfth: :ok:

PR #198, opened 2018-05-18
This code might handle comparisons of floats over large scales, in a better way. 

### LINKED ISSUE #183: RunTests::check_if_equal unsafe for floats/double
current code uses
```c++
((fabs(b)>tolerance && fabs(a/b-1) > tolerance)
      || (fabs(b)<=tolerance && fabs(a-b) > tolerance))
```
However, this fails if `tolerance` is large, e.g. for large magnitude numbers, as `fabs(a/b/-1)` should always be small.

NikEfth:
Absolute and relative comparison, avoiding divisions, the tolerance is applied on the largest on the two numbers. closes #183

### DIFF
--- a/src/include/stir/RunTests.h
+++ b/src/include/stir/RunTests.h
@@ -3,6 +3,7 @@
     Copyright (C) 2000-2005, Hammersmith Imanet Ltd
     Copyright (C) 2013, Kris Thielemans
     Copyright (C) 2013, University College London
+    Copyright (C) 2018, University of Hull
     This file is part of STIR.
 
     This file is free software; you can redistribute it and/or modify
@@ -23,6 +24,7 @@
   \ingroup buildblock 
   \brief defines the stir::RunTests class
 
+  \author Nikos Efthimiou
   \author Kris Thielemans
   \author PARAPET project
 */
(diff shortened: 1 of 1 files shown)


src/include/stir/RunTests.h
```diff
@@ -353,8 +353,15 @@ bool RunTests::check(const bool result, const std::string& str)
 bool
 RunTests::check_if_equal(const double a, const double b, const std::string& str)
 {
-  if ((fabs(b)>tolerance && fabs(a/b-1) > tolerance)
-      || (fabs(b)<=tolerance && fabs(a-b) > tolerance))
+    double diff = fabs(b-a);
```
KrisThielemans: `const double`

src/include/stir/RunTests.h
```diff
@@ -353,8 +353,15 @@ bool RunTests::check(const bool result, const std::string& str)
...
+    if (diff <= tolerance)
+            return true;
+
+    double _a = fabs(a);
+    double _b = fabs(b);
+    float largest = (_b > _a) ? _b : _a;
```
KrisThielemans: `double`, but why not use `std::max(fabs(a),fabs(b))`. far simpler

KrisThielemans:
I'm a bit torn on this. the first test on `fabs(a-b)` uses a tolerance which is "absolute", while the second one is a relative one. In many cases, it won't make sense to have these to be the same.

I feel we should probably split this into 2 functions, one with a relative tolerance, and one with an absolute one. Or probably better: have 2 tolerances, a relative and an absolute one. Should we then satisfy both?

All this needs some checking of current tests on what they expect.

NikEfth:
I did some tests over many different ranges to come up with this formula. I was unsure where such test function would fit the best, if you would like to have it. Having two does not make sense to me, because the comparison function should be agnostic to the comparison.

NikEfth:
`Error: Failed to download resource "gcc"`

PR #214, opened 2018-07-11
Ge signa updates

pwadhwa351:


KrisThielemans:
out of date so we will abandon this

PR #224, opened 2018-08-15
Read multicomponent nifti images with ITK

rijobro:
This is useful for reading deformation/displacement field images. 

To implement this, the ITK reader was extended to include VoxelsOnCartesianGrid<CartesianCoordinate3D<float> > as well as DiscretisedDensity<3,float>.

Testing is impossible without test data. Since STIR cannot not write multicomponent niftis we cannot create the data either. Could possibly include an example image (submodule?)


src/test/IO/test_IO_ITKMulticomponent.cxx
```diff
@@ -0,0 +1,128 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License for more details.
+
+    See STIR/LICENSE.txt for details
+*/
+
+#include "stir/IO/test/test_IO.h"
+
+#ifndef STIR_NO_NAMESPACES
```
KrisThielemans: I'd rather not have these, and just use `std::` all the time below.

src/test/IO/test_IO_ITKMulticomponent.cxx
```diff
@@ -0,0 +1,128 @@
+//
+//
+/*!
+
+  \file
+  \ingroup test
+
+  \brief A simple program to test the stir::OutputFileFormat function.
```
KrisThielemans: I guess needs adapting

src/test/IO/test_IO_ITKMulticomponent.cxx
```diff
@@ -0,0 +1,128 @@
...
+  \author Kris Thielemans
+  \author Richard Brown
+
+
+
+  To run the test, you should use a command line argument with the name of a file.
+  This should contain a test par file.
+  See stir::OutputFileFormatTests class documentation for file contents.
+
+  \warning Overwrites files STIRtmp.* in the current directory
```
KrisThielemans: doesn't do this yet. remove?

src/test/IO/test_IO_ITKMulticomponent.cxx
```diff
@@ -0,0 +1,128 @@
...
+
+  \todo The current implementation requires that the output file format as also
+  readable by stir::read_from_file. At least we should provide a
+  run-time switch to not run that part of the tests.
+*/
+/*
+    Copyright (C) 2002- 2011, Hammersmith Imanet Ltd
```
KrisThielemans: not much present from those days i think. delete?

src/test/IO/test_IO_ITKMulticomponent.cxx
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+END_NAMESPACE_STD
+
+USING_NAMESPACE_STIR
+
+int main(int argc, char **argv)
+{
+    cerr << "\nTesting multi-component ITK reading is currently impossible since "
```
KrisThielemans: I guess the test doesn't really know it's ITK that's being used, so I'd remove all mention of ITK here.

src/test/IO/test_IO_ITKMulticomponent.cxx
```diff
@@ -0,0 +1,128 @@
...
+
+/*!
+  \ingroup test
+  \brief A simple class to test the OutputFileFormat function.
+
+  The class reads input from a stream, whose contents should be as
+  follows:
+
+  \verbatim
+  Test OutputFileFormat Parameters:=
```
KrisThielemans: appropriate?

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -81,46 +109,124 @@ ITKImageInputFileFormat::can_read(const FileSignature& /*signature*/,
...
+             origin,
+             voxel_size);
+
+    // copy data
+    STIRImageMulti::full_iterator stir_iter = image_ptr->begin_all();
+    typedef itk::ImageRegionConstIterator< ITKImageMulti > IteratorType;
+    IteratorType it (itk_image, itk_image->GetLargestPossibleRegion() );
+    for ( it.GoToBegin(); !it.IsAtEnd(); ++it, ++stir_iter) {
+        itk::Point<double,3U> itk_coord;
+        itk_coord[0] = -it.Get()[0]; // Seems we need an extra flip in the x-axis. don't know why...
```
KrisThielemans: we need to figure this one out before merging...

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -185,7 +291,91 @@ read_file_itk(std::string filename)
...
+          return NULL;
+      }
+
+      // Not a DICOM file, so we just read a single image
+      typedef itk::ImageFileReader<ITKImageMulti> ReaderType;
+      ReaderType::Pointer reader = ReaderType::New();
+      reader->SetFileName(filename);
+      reader->Update();
+
+      if (reader->GetImageIO()->GetPixelType() != itk::ImageIOBase::VECTOR) {
```
KrisThielemans: this doesn't do anything useful as far as I can see. It isn't a test on the `reader` really, it's a test on the `ITKImageMulti` type. But you have that under control. Remove this test.

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -185,7 +291,91 @@ read_file_itk(std::string filename)
...
+  info(boost::format("Opening a multi image: %1%") % filename);
+#endif
+
+  typedef itk::GDCMImageIO       ImageIOType;
+  ImageIOType::Pointer dicomIO = ImageIOType::New();
+
+  try
+    {
+      // We can't deal with DICOMs
+      if (dicomIO->CanReadFile(filename.c_str())) {
```
KrisThielemans: DICOM doesn't define deformation types etc.
we could stick to a single template definition but doing a check on the type, e.g. via `GetPixelType` presumably)

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -81,46 +109,124 @@ ITKImageInputFileFormat::can_read(const FileSignature& /*signature*/,
...
+                min_indices,
+                true);
+    }
+
+    return image_ptr;
+}
 
 //To read any file format via ITK
-template<typename TImageType> VoxelsOnCartesianGrid<float>*
+template<>
```
KrisThielemans: I'd rather stick to a single templated function (would need 2 template args, the ITK and STIR image types). see comments for the multi-version below on how that could work.

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -81,46 +109,124 @@ ITKImageInputFileFormat::can_read(const FileSignature& /*signature*/,
...
   typedef itk::GDCMImageIO       ImageIOType;
   ImageIOType::Pointer dicomIO = ImageIOType::New();
   try
     {
       if (!dicomIO->CanReadFile(filename.c_str()))
         {
           // Not a DICOM file, so we just read a single image
-          typedef itk::ImageFileReader<TImageType> ReaderType;
-          typename ReaderType::Pointer reader = ReaderType::New();
+          typedef itk::ImageFileReader<ITKImageSingle> ReaderType;
```
KrisThielemans: would need to revert this

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -81,46 +109,124 @@ ITKImageInputFileFormat::can_read(const FileSignature& /*signature*/,
...
+}
 
 //To read any file format via ITK
-template<typename TImageType> VoxelsOnCartesianGrid<float>*
+template<>
+STIRImageSingle*
 read_file_itk(std::string filename)
 {
+#ifndef NDEBUG
+  info(boost::format("Opening a single image: %1%") % filename);
```
KrisThielemans: I guess you used this as debugging. the message is rather cryptic and won't work anymore if it's back to 1 template. Delete, or generalise to something like `Using ITK to read ...` (but then probably more useful to have a different message for DICOM and the normal stuff).

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -185,7 +291,91 @@ read_file_itk(std::string filename)
...
+                 BasicCoordinate<3,int> &min_indices,
+                 BasicCoordinate<3,int> &max_indices,
+                 CartesianCoordinate3D<float> &origin,
+                 const typename ITKImageType::Pointer itk_image_orig)
+{
+    typedef itk::OrientImageFilter<ITKImageType,ITKImageType> OrienterType;
+    typename OrienterType::Pointer orienter = OrienterType::New();
+    orienter->UseImageDirectionOn();
+    orienter->SetInput(itk_image_orig);
+    orienter->SetDesiredCoordinateOrientation(itk::SpatialOrientation::ITK_COORDINATE_ORIENTATION_RAS);
```
KrisThielemans: add a comment, or even check, that this is only ok for `HFS`.

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -210,66 +400,32 @@ struct removePtr<itk::SmartPointer<Type> >
...
-    make_coordinate(0, -y_size/2, -x_size/2);
-  const BasicCoordinate<3,int> max_indices = 
-    min_indices + make_coordinate(z_size, y_size, x_size) - 1;
-
   // find STIR origin
   // Note: need to use - for z-coordinate because of different axis conventions
-  CartesianCoordinate3D<float> origin(-static_cast<float>(itk_image->GetOrigin()[2]), 
-				      static_cast<float>(itk_image->GetOrigin()[1]), 
-				      static_cast<float>(itk_image->GetOrigin()[0]));
+  CartesianCoordinate3D<float> stir_coord(-static_cast<float>(itk_coord[2]),
```
KrisThielemans: need a comment here on `HFS` as well

src/IO/ITKImageInputFileFormat.cxx
rijobro: This is actually very important. If you don't set the `intent_code` to `VECTOR` when you write the nifti image, then ITK will read the image differently.

I made this mistake before, and although the ITK reader didn't complain, the images weren't read in the way that I expected.

src/IO/ITKImageInputFileFormat.cxx
rijobro: Why does this need to be reverted? I thought `TImageType` and `ITKImageSingle` are the same so these are equivalent?

src/IO/ITKImageInputFileFormat.cxx
rijobro: Maybe I've misunderstood, but I used specialisations since methods of reading `single` and `multi` were very different.

Also, if I templated this method (`read_file_itk`) wrt `ITKImageType`, then I would also have to put the ITK type into the template of `read_from_file`. But I can't do that, since `read_from_file` is in the registeries. Am I wrong?

rijobro:
Changes made for all comments except the two replies I made.

src/IO/ITKImageInputFileFormat.cxx
KrisThielemans: yes, but I believe (but didn't check) that this test will always be true as you're using `itk::ImageFileReader<itk::VectorImage<float, 3> >`. Presumably, the test on `intent_code` for nifti sits in the ITK instantiation for vector images (or it should).

src/IO/ITKImageInputFileFormat.cxx
KrisThielemans: There is no technical problem to template `ITKImageInputFileFormat` in both STIR and ITK image type. The registry is STIR image type specific, but the derived classes can have any type whatsoever, e.g.
```c++
RegisterInputFileFormat<ITKImageInputFileFormat<
         VoxelsOnCartesianGrid<CartesianCoordinate3D<float> >, 
         somefunkyITKtype, someothertypes > >
```
question is if the instantiation are really different. I'm not sure. In any case `read_file_itk` should be a `static` function

rijobro: They are quite different, so I don't think they should be combined.

Ok for making the function prototype `static` (apparently not allowed for specialisations). Just for clarification, were you asking me to do this because you thought `read_file_itk` was a member function? (Because it's not.)

KrisThielemans: `read_file_itk` made static means it's a function local to the .cxx (not exposed to the linker), which is very different from a static member function. Sadly, `static` is an overloaded C++ keyword.

Fine to leave the 2 instantiations for `read_file_itk`

KrisThielemans:
I think only comment to do is the `HFS` check one.

The conversion functions between ITK and STIR could be moved elsewhere. It would be useful when we'd want to call an ITK filter. I'm not so sure in which library this would belong though, so I guess just leave for later.

rijobro:
I put in the comment for HFS, do you want something more? How would I do a check? This information isn't stored in nifti images.

KrisThielemans:
comment is fine for now. we will need to change this with @ashgillman 's changes anyway.

Nifti doesn't store patient orientation, but ITK could for image formats that do store it. Presumably, ITK provides meta-date about this. Possibly it can only be only be done for DICOM data in ITK, which would be sad.

ready to squah-merge?

KrisThielemans:
actually, travis is all red. codacy also gives 3 good suggestions.

KrisThielemans:
and by the way, I'd prefer `const std::string&` arguments. Could avoid a (almost irrelevant) copy :-;

rijobro:
Seems that the Travis set up doesn't have the `verbose` flag for `ctest`, so I can't see what went wrong. I suppose I'll have to wait for the AppVeyor output.

rijobro:
Also, how do I see Codacy suggestions?

KrisThielemans:
Codacy: can't you see them when clicking on the Details next to the Codacy Review below? Here's the list (badly formatted)

```
Class 'IOTests_ITKMulticomponent' has a constructor with 1 argument that is not explicit.
46    IOTests_ITKMulticomponent(const std::string &multi)
47    { _multi = multi; }
48
Related code pattern
Class 'classname' has a constructor with 1 argument that is not explicit.

Class 'classname' has a constructor with 1 argument that is not explicit. Such constructors should in general be explicit for type safety reasons. Using the explicit keyword in the constructor means some mistakes when using the class can be avoided.

Variable '_multi' is assigned in constructor body. Consider performing initialization in initialization list.
47 { _multi = multi; }
```

KrisThielemans:
Appveyor failure is
```
ERROR: no file format found that can read file 'C:/projects/stir/examples/samples/multicomponent_nifti.nii'
```
we don't install ITK here. I guess same error on Travis. The test has to be added only if ITK is available.

KrisThielemans:
still 2 codacy issues
```
 src/IO/ITKImageInputFileFormat.cxx
Uninitialized variable: axis
199 for (unsigned axis; axis<3; ++axis) {
```
not sure if it's guaranteed that `axis` is initialised to zero, but weird anyway. By the way, best to use `0U` and `3U` to avoid warnings.
```
src/test/IO/test_IO_ITKMulticomponent.cxx
Class 'IOTests_ITKMulticomponent' has a constructor with 1 argument that is not explicit.
46 IOTests_ITKMulticomponent(const std::string &multi) :
```
add the `explicit` keyword in front of the constructor. Otherwise we have an implicit convertor, which would be rather weird.

KrisThielemans:
I gues this is ready to merge? squash ok?

rijobro:
Thanks for the comments @ashgillman (the flip axis stuff is gone now).

@KrisThielemans , if you're happy with this I think you can merge it. I tested it on data that had been shifted in the x-, y-, and z-directions, as well as an arbitrary rotation.

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -210,66 +402,33 @@ struct removePtr<itk::SmartPointer<Type> >
...
-  const int y_size =  itk_image->GetLargestPossibleRegion().GetSize()[1];
-  const int x_size =  itk_image->GetLargestPossibleRegion().GetSize()[0];
-  const BasicCoordinate<3,int> min_indices = 
-    make_coordinate(0, -y_size/2, -x_size/2);
-  const BasicCoordinate<3,int> max_indices = 
-    min_indices + make_coordinate(z_size, y_size, x_size) - 1;
-
   // find STIR origin
   // Note: need to use - for z-coordinate because of different axis conventions
-  CartesianCoordinate3D<float> origin(-static_cast<float>(itk_image->GetOrigin()[2]),
```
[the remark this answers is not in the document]
KrisThielemans: yes. that'll be for you to fix...

src/IO/ITKImageInputFileFormat.cxx
KrisThielemans: I've made them static

PR #225, opened 2018-08-21
Construct parametric image from separated components

rijobro:
This is useful if, like me, you are given a parametric as its individual components (e.g., two nifti images) and need to construct a single parametric image so that you can use it with STIR.

### DIFF
--- a/src/modelling_utilities/make_parametric_image_from_components.cxx
+++ b/src/modelling_utilities/make_parametric_image_from_components.cxx
@@ -1,0 +1,102 @@
+//
+//
+/*
+  Copyright (C) 2018, University College London
+  This file is part of STIR.
+
+  This file is free software; you can redistribute it and/or modify
+  it under the terms of the GNU Lesser General Public License as published by
+  the Free Software Foundation; either version 2.1 of the License, or
+  (at your option) any later version.
+
+  This file is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU Lesser General Public License for more details.
+
+  See STIR/LICENSE.txt for details
+*/
+
+/*!
+  \file
+  \ingroup utilities
+  \brief Create parametric image from individual components
+  \author Richard Brown
+
+  \par Usage:
+  \code 
+  make_parametric_image_from_components output_parametric_image slope intercept
+  \endcode
+
+  In the case of Patlak, slope is V_0 and intercept is K_i.
+
+*/
+
+#include <iostream>
+#include "stir/DiscretisedDensity.h"
+#include "stir/modelling/ParametricDiscretisedDensity.h"
+#include "stir/IO/read_from_file.h"
+#include "stir/is_null_ptr.h"
+#include "stir/IO/OutputFileFormat.h"
+#include "stir/Succeeded.h"
+
+#include "stir/ProjDataInfo.h"
+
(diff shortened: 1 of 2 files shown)


rijobro:
Done. However, only implemented for 2 kinetic parameters for the moment.

src/modelling_utilities/make_parametric_image_from_components.cxx
```diff
@@ -47,45 +47,52 @@ int main(int argc, char *argv[])
...
-        const Succeeded success = OutputFileFormat<ParametricVoxelsOnCartesianGrid>::default_sptr()->write_to_file(argv[1], param_im);
-        if (success == Succeeded::no)
-            throw std::runtime_error("Failed writing.");
+        std::vector<VoxelsOnCartesianGrid<float> > params;
+
+        // Loop over all parameters
+        for (int i=2; i<=argc; ++i) {
+
+            // Read
+            shared_ptr<DiscretisedDensity<3,float> > im(read_from_file<DiscretisedDensity<3,float> >(argv[2]));
```
KrisThielemans: always `argv[2]`?

src/modelling_utilities/make_parametric_image_from_components.cxx
```diff
@@ -47,45 +47,52 @@ int main(int argc, char *argv[])
...
+            // Check
+            if (is_null_ptr(im)) throw std::runtime_error("Failed to read file: " + std::string(argv[2]) + ".");
```
KrisThielemans: argv[2]

src/modelling_utilities/make_parametric_image_from_components.cxx
```diff
@@ -47,45 +47,52 @@ int main(int argc, char *argv[])
...
+
+            // Convert to VoxelsOnCartesianGrid
+            if (is_null_ptr(dynamic_cast<VoxelsOnCartesianGrid<float>*>(im.get())))
+                throw std::runtime_error("Failed to convert parameter to VoxelsOnCartesianGrid.");
+
+            VoxelsOnCartesianGrid<float> param = *dynamic_cast<VoxelsOnCartesianGrid<float>*>(im.get());
+
+            params.push_back(param);
```
KrisThielemans: are you sure you need a copy? why not `*im`?

src/modelling_utilities/make_parametric_image_from_components.cxx
```diff
@@ -47,45 +47,52 @@ int main(int argc, char *argv[])
...
+
+            // Check characteristics match (compare new with first)
+            std::string explanation;
+            if (!param.has_same_characteristics(params.at(0),explanation))
+                throw std::runtime_error("Kinetic images do not have same characteristics (" + std::string(explanation) + ").");
+        }
+
+        // At the moment, only implemented for 2 parameters
+        if (params.size() == 2) {
```
KrisThielemans: throw error otherwise (could have done it above, but doesn't matter).
can get the value 2 from the `ParametricVoxelsOnCartesianGrid ` type possibly, but I don't know. In any case, put a `\todo` in the doxygen and a run-time check in the code as otherwise some day we'll be very surprised :-;

src/modelling_utilities/make_parametric_image_from_components.cxx
rijobro: Good spot!

src/modelling_utilities/make_parametric_image_from_components.cxx
rijobro: ok, done.

PR #227, opened 2018-08-22
No need to give output template to get dynamic images from parametric

rijobro:
There's no need (as far as I can see) to have to supply a dynamic image as input to get_dynamic_images_from_parametric_images. We have all the information we need from the Patlak and the Parametric image.

### DIFF
--- a/documentation/release_3.1.htm
+++ b/documentation/release_3.1.htm
@@ -174,6 +174,9 @@ was in HFS position. This is currently not checked (and for many file formats, n
   </li>
   <li>Changed the return value of <tt>DynamicProjData::read_from_file</tt> from a raw ptr to a <tt>unique_ptr</tt>.
   </li>
+  </li>
+  <li>Output template for <tt>get_dynamic_images_from_parametric_images</tt> is now only optional. Writing of the output is longer restricted to ecat7.
+  </li>
 </ul>
 <h4>Python (and MATLAB) interface</h4>
 <ul>
--- a/documentation/release_3.1.htm
+++ b/documentation/release_3.1.htm
@@ -175,7 +175,7 @@ was in HFS position. This is currently not checked (and for many file formats, n
   <li>Changed the return value of <tt>DynamicProjData::read_from_file</tt> from a raw ptr to a <tt>unique_ptr</tt>.
   </li>
   </li>
-  <li>Output template for <tt>get_dynamic_images_from_parametric_images</tt> is now only optional. Writing of the output is longer restricted to ecat7.
+  <li>Output template for <tt>get_dynamic_images_from_parametric_images</tt> is now optional. Writing of the output is no longer restricted to ecat7.
   </li>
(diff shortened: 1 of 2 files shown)


KrisThielemans:
We need it to get time frames?

rijobro:
We can get that from the Patlak (which is created from a paramter file that contains the tdefs). Thus, we can set the time frames with:
```
const TimeFrameDefinitions tdefs = patlak_plot.get_time_frame_definitions();
```

src/modelling_utilities/get_dynamic_images_from_parametric_images.cxx
```diff
@@ -75,29 +77,27 @@ int main(int argc, char *argv[])
...
+                                          scanner_sptr,
+                                          voxels_sptr);
+
+      patlak_plot.get_dynamic_image_from_parametric_image(dyn_image,*par_image_sptr);
 
   // Writing image 
   std::cerr << "Writing dynamic-image in '"<< argv[1] << "'\n";
-  Succeeded writing_succeeded=dyn_image.write_to_ecat7(argv[1]);
+
+  Succeeded writing_succeeded = OutputFileFormat<DynamicDiscretisedDensity>::default_sptr()->write_to_file(argv[1], dyn_image);
```
KrisThielemans: can we have an optional output-format-parameter-file here?

rijobro:
If output file is given, use it (keeps backwards compatibility).
User can also set output file type.

documentation/release_3.1.htm
```diff
@@ -174,6 +174,9 @@ <h3>Changed functionality</h3>
   </li>
   <li>Changed the return value of <tt>DynamicProjData::read_from_file</tt> from a raw ptr to a <tt>unique_ptr</tt>.
   </li>
+  </li>
+  <li>Output template for <tt>get_dynamic_images_from_parametric_images</tt> is now only optional. Writing of the output is longer restricted to ecat7.
```
KrisThielemans: "is now optional. Writing of the output is no longer..."

src/modelling_utilities/get_dynamic_images_from_parametric_images.cxx
```diff
@@ -22,16 +23,26 @@
   \ingroup utilities
   \brief Multiplies Parametric Images with the Model Matrix creating Dynamic Images
   \author Charalampos Tsoumpas  
+  \author Richard Brown
 
   \par Usage:
   \code 
-  get_dynamic_images_from_parametric_images output_parametric_image input_dynamic_image [par_file] 
+  get_dynamic_images_from_parametric_images output_parametric_image input_dynamic_image [par_file] [output_format_par_file]
```
KrisThielemans: can't be correct. where's the optional template? 2 options would need to be as `[option1 [ option2]]`. Needs to be backwards compatible as well. I haven't checked how it was...

PR #235, opened 2018-09-10
Change of Origin

ashgillman:
Progress to implement a change of origin.

# Features

- [X] `Scanner` now has a field specifying where the origin is defined with respect to
    - for mMR, this is redefined to middle of gantry
    - defaults for other scanners to middle of first ring
- [X] `ProjMatrixByBinUsingRayTracing` generates consistent projections regardless of origin by leveraging DiscretisedDensity to convert between gantry and bed/index space.
- [ ] Allow for shifts in origin
- [ ] Bed position
- [ ] Hard code mMR offset
  - And verify between two scanners.
- [ ] ITK IO to take into account patient position.

### DIFF
--- a/src/include/stir/LORCoordinates.inl
+++ b/src/include/stir/LORCoordinates.inl
@@ -7,11 +7,14 @@
   \brief Implementations for LORCoordinates.h
   \warning This is all preliminary and likely to change.
   \author Kris Thielemans
+  \author Ashley GIllman
 
 
 */
 /*
     Copyright (C) 2004- 2013, Hammersmith Imanet Ltd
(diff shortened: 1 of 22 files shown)


src/include/stir/LORCoordinates.inl
```diff
@@ -308,6 +311,54 @@ LORAs2Points(const LORInAxialAndNoArcCorrSinogramCoordinates<coordT>& coords)
 }
 
 
+template <class coordT>
+float
+LORAs2Points<coordT>::
```
KrisThielemans: currently not needed

ashgillman:
As discussed last week, this PR will be replaced by #241 #242 #243 #244

PR #237, opened 2018-09-12
Postfilter dynamic images as well as single

rijobro:
Gives the user the possibility to post filter a dynamic image as well as a single image. Simply give use the boolean flag to signal that your file is a dynamic image. If the flag is not used, it will assume the image is single, such that backwards compatibility is not broken.

### DIFF
--- a/src/utilities/postfilter.cxx
+++ b/src/utilities/postfilter.cxx
@@ -35,7 +35,7 @@
   This program enables calling any ImageProcessor object on input data, 
   and writing it to file. It can take the following command line:
   \verbatim
-   postfilter [[-verbose] <output filename > <input header file name> <filter .par filename> [--dynamic]
+   postfilter [[-verbose] <output filename> <input header file name> <filter .par filename> [--dynamic]
   \endverbatim
   This is done to make it easy to process a lot of files with the same 
   ImageProcessor. However, if the number of command line arguments is not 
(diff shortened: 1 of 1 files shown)


src/utilities/postfilter.cxx
```diff
@@ -28,12 +29,13 @@
...
   \author Matthew Jacobson
   \author PARAPET project
+  \author Richard Brown
   
 
   This program enables calling any ImageProcessor object on input data, 
   and writing it to file. It can take the following command line:
   \verbatim
-   postfilter [[-verbose] <output filename > <input header file name> <filter .par filename>
+   postfilter [[-verbose] <output filename > <input header file name> <filter .par filename> <dynamic image (0/1, default: 0)>
```
KrisThielemans: best to use same options as `stir_math`, i.e. `[--parametric || --dynamic]`.

src/utilities/postfilter.cxx
```diff
@@ -73,7 +76,7 @@ using std::endl;
 
 START_NAMESPACE_STIR
 
-DiscretisedDensity<3,float>* ask_image(const char *const input_query)
+DiscretisedDensity<3,float> * ask_image_single(const char *const input_query)
```
KrisThielemans: I guess I'd template it to avoid a copy

KrisThielemans:
travis failure on OSX is
```
==>Installingroot
==>Downloading https://homebrew.bintray.com/bottles/root-6.14.02.sierra.bottle.
curl: (56) SSLRead() return error -9806
Error: Failed to download resource "root"
```
probably temporary and certainly nothing to do with this PR

rijobro:
How would you propose giving two optional flags? One for output format type and the other for dynamic images?

PR #240, opened 2018-09-14
Motion corrected parametric objective function

rijobro:
Requires on a lot of the other recent pull requests (i.e., reading of multicomponent images and projectors that allow motion.


src/recon_buildblock/CMakeLists.txt
```diff
@@ -130,7 +130,7 @@ endif()
 
 # TODO what to do with IO?
 # modelling_buildblock currently needed for ParametricDensity and Patlak (TODO get rid of this somehow?)
-target_link_libraries(recon_buildblock modelling_buildblock display numerics_buildblock listmode_buildblock data_buildblock buildblock spatial_transformation_buildblock )
+target_link_libraries(recon_buildblock modelling_buildblock display numerics_buildblock listmode_buildblock data_buildblock buildblock spatial_transformation_buildblock local_motion_buildblock)
```
rijobro, line 133: We recently did a PR to some of the `Experimental` code, and we had a better way of adding libraries. Think we used `PARENT_SCOPE` from the Experimental `CMakeLists.txt`.

KrisThielemans: I cannot quite recall what this was about, but it seems not so important at this point. we'd need to differentiate if `STIR_EXPERIMENTAL=ON` or not, unless we move motion things.

rijobro: You're right, not important, but thought the other way was better: https://github.com/UCL/STIR/pull/580/files#diff-5e304515000fc33e14a8d811e63d04d1R27-R30.

KrisThielemans: seems different. this was creating a list of all registries/libraries. We cannot let `recon_buildblock` depend on `STIR_LIBRARIES` as it includes itself. We can also not let it depend on everything except itself, as then we create huge circles. I can't see how we can avoid doing this manually.

Probably better though would be to do
```cmake
if (BUILD_EXPERIMENTAL)
target_link_libraries(PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.cxx.o local_motion_buildblock)
endif()
```
if such a thing is possible.

src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx
```diff
@@ -159,9 +181,137 @@ post_processing()
...
+    }
+    // If separate components
+    else {
+        disp_field_headers.resize(3);
+        for (int i=0; i<3; ++i)
+            if (disp_field_headers[i].parse(_disp_field_filenames[i+1].c_str()) == false)
+                error("PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData::post_processing Error parsing %s", _disp_field_filenames[i+1].c_str());
+    }
+
+    // Check there are as many motion fields as sinograms
```
AnderBiguri, line 249: Is this required? 

I suppose that at some point we might not want to have 1 DVF == 1 sino, but may want to reuse some DVFs?

KrisThielemans: yes, but re-using becomes a bit tricky how to enter it which frame to re-use where. That would all sit in the input part, which arguably sits here. Until we would do that, I would leave this test in.

KrisThielemans:
we'll have to merge `master` into this. Also the PR has to go to `master`, not `release_4` as currently set. @AnderBiguri do you want me to do this?

AnderBiguri:
@KrisThielemans I can do it if you want. I will need to get my own branch to PR at some point, if I will work on it, as I likely can't push to rijobro.

KrisThielemans:
I have merged, but there will be some problems with the normalisation. The code currently tries to allow a single `normalisation_sptr` object (for instance used in `get_normalisation()` and also for the calibration check, but actually the `set_up` ignores that and uses the `_normalisations_sptr`. They can't be both right.

AnderBiguri:
I have been looking at the code and that comment for 15 minutes until I realized that there is `normalisationS` and `normalisation`. Perhaps that needs a change of name, for sanity?

It seems that one is for when you have a single normalization for all sinograms, and the other one for individual per dynamic-frame normalization. 

It makes sense in practical terms to allow both, as one might want to have a single normalization for all algorithm, albeit raises some questions about what to do with motion  and CTACs and if we should add something for that..... But thats for later. 

The code in `set_up()` (you mean `set_up_before_sensitivity()`  right?) seems to handle this well, i.e.:
https://github.com/UCL/STIR/blob/840a5cf32851984f4d7d80321ace1849fee6a702/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx#L536-L540

but indeed `get_normalisation()` will only return the single one. I can see it used in `get_exam_info_uptr_for_target()`, but I assume its used somewhere else (albeit I cant find it in Github), which would be an issue. 
What do you suggest happens with all this?

AnderBiguri:
Seems to me that changing what `get_normalisation()` returns may fix it, but I guess the problem is that `exam_info_uptr` either is not able to handle this array of normalisations, or will need to just store 1 and risk being wrong.

KrisThielemans:
I'm not sure why we need the `get_normalisation()` function and where it came from. Could make it private and see what happens...

> I guess the problem is that `exam_info_uptr` either is not able to handle this array of normalisations, or will need to just store 1 and risk being wrong.

`ExamInfo` doesn't know about normalisation, so that's no problem. "risk being wrong" is not something I'd want to do

AnderBiguri:
Sorry, I meant `get_normalisation_sptr()`, which is used once, as opposed to `get_normalization()`, which is not. And as you say, is likely useless (right now it does exactly the same as `get_normalisation_sptr()`).

The only place (inside the `PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData`  class) that I see it used is in here (338), which I now realise its not directly for a examinfo. 

https://github.com/UCL/STIR/blob/840a5cf32851984f4d7d80321ace1849fee6a702/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx#L332-L349

However, as it stands, I am not sure what this will return in the case of having dynamic normalization files , which seems to be the only accepted way of getting normalization, I think:

https://github.com/UCL/STIR/blob/840a5cf32851984f4d7d80321ace1849fee6a702/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx#L187-L199


Obviously I am just trying to get my head around this and how to fix it and I have too many ends of the puzzle piece unknown.

AnderBiguri:
I have this question: 

https://github.com/UCL/STIR/blob/840a5cf32851984f4d7d80321ace1849fee6a702/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx#L656

Why are gradients initialized with 1s? Are they not overwritten? Tried to go down the rabbit hole to answer that, but it goes very deep hehe.

KrisThielemans:
I see no reason why they're filled with 1. `git blame` says that this is a question that's at least 14 years old... I'm somewhat surprised that this is a rabbit hole though. `PatlakPlot::get_dynamic_image_from_parametric_image` should be a relatively easy function to track down. It's not a gradient, so very unlikely any content in `dyn_image_estimate` would be relevant.

Possibly you mean [higher up](https://github.com/UCL/STIR/blob/840a5cf32851984f4d7d80321ace1849fee6a702/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx#L619-L622), but there it's a call to `get_dynamic_image_from_parametric_image` so should again be overwritten.

Harder is https://github.com/UCL/STIR/blob/840a5cf32851984f4d7d80321ace1849fee6a702/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx#L629-L637
This will [call](https://github.com/UCL/STIR/blob/aa815ea9207e5bd1d65503a379426e5d3a94b983/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L621), which will call `distributable_computation`. I see
https://github.com/UCL/STIR/blob/aa815ea9207e5bd1d65503a379426e5d3a94b983/src/recon_buildblock/distributable.cxx#L385-L386
which implies that the image is first set to 0. That then implies that the above `fill(1)` is irrelevant. It certainly doesn't make any sense to fill it with `1` anyway!

It would be better to document what the `gradient` function does (although I guess it makes some sense that if it doesn't say that it accumulates, it will overwrite). We could add a test for it as well, which would make things a bit more explicit (and more rabbit-hole-proof). But I suppose that doesn't belong in this PR.

AnderBiguri:
Ah yes, that is exactly what I meant, what happens in the gradient ones. Good to know, I stopped my search in the distributable codes, one step short. ouch. Will remove those fill then.

KrisThielemans:
of course, we're now in the danger of mixing the "clean-up the direct Patlak" effort with "add motion to direct Patlak" effort. Possibly somewhat dangerous. Might be better to do the clean-up in a different PR, then merge it back here. All depends on how much clean-up you plan to do...

AnderBiguri:
No, I meant just remove the `fill()` that are unnecessary in this PR in this code. Otherwise as you say, I'd to a separate PR. I'll try to keep the PR on topic.

rijobro:
@AnderBiguri Looks like you're doing a good job wrapping your head around all this!

It might be useful to get an example working (based on the parameter file we discussed by email). That way you can move things around (e.g., remove `fill()`) and you would notice if the algorithm started falling over.

AnderBiguri:
I am clearly on the need of more C++ knowledge. So, please bear with me.

This line errors at compilation
https://github.com/UCL/STIR/blob/29fee74ed2501e075f6a50fa7856988d8b9fa5e9/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx#L571

```
/home/anderbiguri/SIRF/build/sources/STIR/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx:571:3: error: passing ‘std::__shared_ptr_access<const stir::ExamInfo, (__gnu_cxx::_Lock_policy)2, false, false>::element_type {aka const stir::ExamInfo}’ as ‘this’ argument discards qualifiers [-fpermissive]
```

If I do not understand this wrong, this is because `target_sptr->get_exam_info_sptr()` returns `shared_ptr<const ExamInfo>`, thus we can not modify the info in the `ExamInfo`, because its `const`, so `set_time_frame_definitions` is not allowed. 

But how can I do it though? both `get_exam_info` and `get_exam_info_sptr()` return a `const ExamInfo` or a `shared_ptr<const ExamInfo>`, so I can not modify it. 


That said, I think even if there was a way around it, this would still be a problem, because the function is defined with an input of  `shared_ptr<const TargetT > const& target_sptr`. I am not sure about this, but this means that `target_sptr` can not do any `set()` stuff, right (if they are programmed properly of course)? we can not modify any member of `target_sptr`, is that correct? 


I guess  a second question is then:  how do I add those `TimeFrameDefinitions` to the `ExamInfo`? I suspect making the input of the method a non-const is not a doable thing, as its some sort of derived (wrong word?) method, but this goes a bit into the area of STIR that is still cloudy in my mind. Pointers appreciated.

rijobro:
@KrisThielemans and I spent some time trying to secure some of the returns of shared pointers (https://github.com/UCL/STIR/pull/487, https://github.com/UCL/STIR/pull/480, https://github.com/UCL/STIR/pull/470). Looks like I didn't update this PR after we did that.

I guess you can just use the `get` method, create a copy, make any changes, and then replace the exam info?

AnderBiguri:
I tried the following:

```
  ExamInfo my_exam=target_sptr->get_exam_info();
  my_exam.set_time_frame_definitions(tdefs);
  const ExamInfo const_exam=my_exam; // possibly unnecessary
  target_sptr->set_exam_info(const_exam);
```
And I get a similar error:

```
/home/anderbiguri/SIRF/build/sources/STIR/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx:574:3: error: passing ‘std::__shared_ptr_access<const stir::ParametricDiscretisedDensity<stir::VoxelsOnCartesianGrid<stir::KineticParameters<2, float> > >, (__gnu_cxx::_Lock_policy)2, false, false>::element_type {aka const stir::ParametricDiscretisedDensity<stir::VoxelsOnCartesianGrid<stir::KineticParameters<2, float> > >}’ as ‘this’ argument discards qualifiers [-fpermissive]
```

Which is why I asked about the second part. This seems to be because `target_sptr` is const also? Or I am interpreting this worng (or my code is wrong)

rijobro:
Ah sorry missed that in your original message. Can you remove the `const` from the `target_sptr` argument into the method? I guess also in the `.h` file.

AnderBiguri:
I don't seem to be able to get it working. Compilation errors are confusing (to me) but I think that what they are saying is that this derived class does not match the base class, which is what I was suspecting. 

```
/home/anderbiguri/SIRF/build/sources/STIR/src/include/stir/RegisteredParsingObject.inl:40:23: error: invalid new-expression of abstract class type ‘stir::PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData<stir::ParametricDiscretisedDensity<stir::VoxelsOnCartesianGrid<stir::KineticParameters<2, float> > > >’
   Derived * der_ptr = new Derived;
                       ^~~~~~~~~~~
In file included from /home/anderbiguri/SIRF/build/sources/STIR/src/recon_buildblock/recon_buildblock_registries.cxx:60:0:
/home/anderbiguri/SIRF/build/sources/STIR/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.h:58:7: note:   because the following virtual functions are pure within ‘stir::PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData<stir::ParametricDiscretisedDensity<stir::VoxelsOnCartesianGrid<stir::KineticParameters<2, float> > > >’:
 class PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData:
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/anderbiguri/SIRF/build/sources/STIR/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.h:32:0,
                 from /home/anderbiguri/SIRF/build/sources/STIR/src/recon_buildblock/recon_buildblock_registries.cxx:26:
/home/anderbiguri/SIRF/build/sources/STIR/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMean.h:266:5: note: 	stir::Succeeded stir::PoissonLogLikelihoodWithLinearModelForMean<TargetT>::set_up_before_sensitivity(const std::shared_ptr<const _Tp>&) [with TargetT = stir::ParametricDiscretisedDensity<stir::VoxelsOnCartesianGrid<stir::KineticParameters<2, float> > >]
     set_up_before_sensitivity(shared_ptr<const TargetT > const& target_sptr) = 0;
     ^~~~~~~~~~~~~~~~~~~~~~~~~
```

rijobro:
Ok some ideas:

1. Change all instances of `set_up_before_sensitivity`. Probably a lot of changes.
2. Do a naughty const cast (easiest, not good practice)
3. Make the exam info to be a mutable member so that it can be changed for const images
4. Make the time definitions a mutable object so that it can be changed for const exam info
5. Copy the original image, so that it's no longer const. Probably not feasible since the rest of the recon will use the original `target_sptr`.

AnderBiguri:
Thanks @rijobro, those are great suggestions! 

I guess I'll wait to see what @KrisThielemans prefers :)

rijobro:
You could probably just comment it out for now. I'm not sure, but I don't think it'll be using the target image's time defs during the recon. It'll be updating it so that when it gets saved to interfile, it shows that the image is a composite of multiple time points. You'd have to check though, but I imagine that the Patlak analysis isn't currently clever enough to be using the time def inside of the target image.

AnderBiguri:
I was working on Windows for a bit and I moved to Linux now (Windows was acting up a bit) and I get the following error when compiling (after commenting out the problematic line above, as Rich suggested)

```
[ 14%] Linking CXX executable ctac_to_mu_values
/usr/bin/ld: cannot find -llocal_motion_buildblock
collect2: error: ld returned 1 exit status
```

I see that -`llocal_motion_bildblock` points to `-llocal_motion_buildblock ../modelling_buildblock/libmodelling_buildblock.a `, and that that is indeed correctly at (with the superbuild) `./builds/STIR/build/src/modelling_buildblock/libmodelling_buildblock.a`. 
any idea?

KrisThielemans:
@AnderBiguri for your original problem. We can indeed not let this function modify the `*target_sptr`. There is no guarantee that the user will call `set_up()` with the same image, as what they will reconstruct for instance. So, this setting should be converted into a check if the time-frame-defs are as expected. I guess I'd do this as
- if it has no time-frame defs, issue warning
- else if it has the wrong time-frame defs, issue error

There should normally be a `construct_target` member somewhere, which will create the default target (given voxel-sizes etc) for the input data. The `set` code should be moved there.

KrisThielemans:
The `local_motion_buildblock` error could be related to https://github.com/UCL/STIR/pull/240/files#r458742649, but more likely is because you haven't enabled `STIR_ENABLE_EXPERIMENTAL=ON`.

Obviously, that means we cannot merge this PR until we have moved the relevant motion bits.

AnderBiguri:
> * if it has no time-frame defs, issue warning

Again my C++ lacking. How do I check this? by comparing the getter return to an "empty" (non-set) `TimeFrameDefinitions` ?

KrisThielemans:
This doesn't have anything (or at least very little) to do with C++ luckily.
```c++
...->get_time_frame_definitions().get_num_time_frames()
```
or something with a similar name

AnderBiguri:
Apologies, I know how to *get* the time frames. But you said "if it has no time-frame defs, issue warning". The class will have time-frame defs as a member, and such member will be created by its default constructor when ExamInfo default constructor is called. 

My (C++) question is how do I know `...->get_time_frame_definitions().get_num_time_frames()` is "empty" i.e. "it has no time frame defs"

KrisThielemans:
my wrong terminology, I just meant `...->get_time_frame_definitions().get_num_time_frames()==0`

AnderBiguri:
Thanks! I see now! ah, always the easiest solutions after an hour of reading half of stackoverflow.... Thanks :)

rijobro:
You could check more than just the number of time frames (start and stop times of each frame). Not sure if time frame definitions class overloads the `==` and `!=` operators, but that would be pretty easy.

AnderBiguri:
Yes they do! Checking the number of tf is just to see if they are set, but if there is more then zero, then I'm comparing them properly with ==

AnderBiguri:
I managed to make it work, but I get images out of this that are completely zero in value as a recon result. 

I loaded up my own dynamic images that I am using with indirect patlak for my research, and have forward projected them. I generated multiplicative sinos that are of value 1 and additive sinos of value 0. I have created deformation fields filled with value zero too, to test if the code works without motion. 

I've been looking at the pieces of this for a bit now and I have no idea what can make the result all zero. Do you have any idea what to look at that I might have missed? 

I have checked that all sinos have "correct values (sino has sinograms, mult has all 1, add has all 0 etc).

Its quite hard in STIR to make a minimal complete example, because there are many files needed for that, so let me know what extra info I can add here:

Code that generates data, in SIRF. 

```
import numpy as np
import scipy
import scipy.io as sio
import scipy.ndimage as snd
import matplotlib.pyplot as plt
import sirf.STIR as stir
import sirf.Reg as reg

from sirf.Utilities import  show_3D_array
import subprocess
import os
import sys


def save_multi(time,fname,prefix,extension=".hs"):
    data=[""]*(len(time)+3)
    data[0]="Multi :=\n"
    data[1]="  total number of data sets := "+str(len(time)-1)+"\n"
    for ii in range(len(time)-1):
        data[ii+2]="  data set["+ str(ii+1) + "] := ./sinos/"+prefix+str(ii)+extension+"\n"
    data[len(time)+2]="end :="
    with open("./sinos/"+fname+".txt","w") as outputfile:
        outputfile.writelines(data)


def save_fdef(filename,time,bed_frame_duration):
    data=[]
    data.append("0 " +str(time[0])+"\n")
    for i in range(len(time)-1):
        data.append("1 "+str(bed_frame_duration)+"\n")
        data.append("0 "+str(time[i+1]-time[i]-bed_frame_duration)+"\n")
    with open("./sinos/"+filename+".fdef","w") as outputfile:
        outputfile.writelines(data)


# dynamic sim parameters. 

filename_prefix="dynamicXcat_WB"
bed_frame_duration=35
wbd_pass_time=6*bed_frame_duration
start_time=3*wbd_pass_time
end_time=wbd_pass_time*8 #60*70
eps=0.0000001
desired_t=np.arange(start_time,end_time,wbd_pass_time) 

save_fdef("test",desired_t,bed_frame_duration)

# load dynamic images
dynamic_image=[]
for i in range(len(desired_t)-1):
    dynamic_image.append(stir.ImageData("./dynamicXcat_WB/"+filename_prefix+"_"+str(desired_t[i])+"_"+str(desired_t[i]+bed_frame_duration)+".hv"))

#create acq model. 
acq_data_template=stir.AcquisitionData('Discovery MI')
acq_model = stir.AcquisitionModelUsingRayTracingMatrix()
acq_model.set_up(acq_data_template,dynamic_image[0])


# make DVFs
# as I donkt know how to create a DVF from an image template, I register nothing at the correct geometry and then fill it with zero
empty=acq_data_template.create_uniform_image(0)
reg_alg = reg.NiftyF3dSym()
reg_alg.set_initial_affine_transformation(reg.AffineTransformation(np.eye(4, dtype=np.float32)))
reg_alg.set_reference_image(empty)
reg_alg.add_floating_image(empty)
reg_alg.process()
dvf=reg_alg.get_deformation_field_forward()
dvf.fill(0)

for i in range(len(desired_t)-1):
    dvf.write("./sinos/dvf_"+str(i)+".nii")
save_multi(desired_t,"dvf_","dvf_",".nii")
sinos=[]

#make sinograms
for i in range(len(desired_t)-1):
    sinos.append(acq_model.forward(dynamic_image[i]))

# save sino, additive, multiplicative 
for i in range(len(desired_t)-1):
    sinos[i].write("./sinos/sino_"+str(i)+".hs")
    additive=sinos[i].get_uniform_copy(0)
    multiplicative=sinos[i].get_uniform_copy(1)
    additive.write("./sinos/add_"+str(i)+".hs")
    multiplicative.write("./sinos/mult_"+str(i)+".hs")
save_multi(desired_t,"sino_","sino_")
save_multi(desired_t,"add_","add_")
save_multi(desired_t,"mult_","mult_")
```
 Parameter file:

```
OSMAPOSLParameters :=
; example file for running OSEM with interfiltering
; filter specific things are at the end

objective function type:=PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData
PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData Parameters:=

input file := ./sinos/sino_.txt

; if disabled, defaults to maximum segment number in the file
maximum absolute segment number to process := 0
; see Users Guide to see when you need this
zero end planes of segment 0:= 0

;zoom := .15645
;xy output image size (in pixels) := 21
;z output image size (in pixels) := 17
;z offset (in mm) := 0

projector pair type := Matrix
  Projector Pair Using Matrix Parameters :=
  Matrix type := Ray Tracing
  	number of rays in tangential direction to trace for each bin := 10
  End Projector Pair Using Matrix Parameters :=

Dynamic bin normalisation filename := ./sinos/mult_.txt

; we need this for backwards compatibility with the testing script
use subset sensitivities:=1
;sensitivity filename:= sens.img
; if next is set to 1, sensitivity will be recomputed
; and written to file (if "sensitivity filename" is set)
recompute sensitivity := 1
subset sensitivity filenames := subset_sensitivity_%d.hv

; specify additive projection data to handle randoms or so
; see Users Guide for more info
additive sinograms := ./sinos/add_.txt

; patlak related files
Kinetic Model type := Patlak Plot
Patlak Plot Parameters:=

time frame definition filename := ./sinos/test.fdef
starting frame := 1
calibration factor := 1
blood data filename := if_fdg_w_decay.if 
Time Shift := 0
In total counts := 1
In correct scale := 1
end Patlak Plot Parameters:=

Forward displacement field image := ./sinos/dvf_.txt

end PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData Parameters:=

; Number of subsets should be a divisor of num_views/4
number of subsets:= 68
; Use for starting the numbering from something else than 1
start at subiteration number:=1
; Use if you want to start from another subset than 0 (but why?)
start at subset:= 0
number of subiterations:= 7
save estimates at subiteration intervals:= 7
;write update image := 1

;initial estimate := 0
; enable this when you read an initial estimate with negative data
enforce initial positivity condition:=1

inter-update filter subiteration interval:= 0
inter-update filter type := None

inter-iteration filter subiteration interval:= 0

post-filter type := None

output filename prefix := test

END :=
```

AnderBiguri:
Update on the above: the sensitivity is all zeros, and that is what is causing the output to be all zeroes. 

I am trying to follow who does this subset computing, to see what happens, but I am a bit lost of where things are happening.

KrisThielemans:
can you try this without def fields first? If that works, can you make sure that the def fields are written as displacement fields? (then they should just contain zeros).

There could be with writing the dvf to file, and STIR reading them properly (x,y,z quickly starts to loose meaning). It should work of course, but...

KrisThielemans:
Also, does the `recon_test_pack/test_modelling.sh` still work?

AnderBiguri:
> can you try this without def fields first? If that works, can you make sure that the def fields are written as displacement fields? (then they should just contain zeros).
> 

Without them "it works" (I am running 1 subset with 1 iteration, so "it works" means there is a blob in the image). 

With displacements instead of deformations (true, that was my mistake), the result is the same. The only thing I chahnged from the above code is `dvf=reg_alg.get_deformation_field_forward()` -> `dvf=reg_alg.get_displacement_field_forward()`

> Also, does the recon_test_pack/test_modelling.sh still work?

There is no such file. You mean `recon_test_pack/run_test_modelling.sh`?

AnderBiguri:
> Also, does the `recon_test_pack/run_tests_modelling.sh` still work?

Errors!

```
Test the direct POSMAPOSL Patlak Plot reconstruction
./run_tests_modelling.sh: line 187: 22706 Segmentation fault      (core dumped) INPUT=fwd_dyn_from_p0005-p5.hs INIT=indirect_Patlak.hv ${MPIRUN} P${direct} P${direct}.par > P${direct}.log 2>&1
ERROR
```

KrisThielemans:
> With displacements instead of deformations (true, that was my mistake), the result is the same

the same as what? Does it now "work" or still zero sensitivity?

> You mean recon_test_pack/run_tests_modelling.sh?

yes, sorry

> I get errors/sysntax errors running this, investigating a bit.

a bit strange to get syntax errors, as Travis runs it, and it's ok on `master` apparently. Maybe something went wrong with the merge

KrisThielemans:
> (fixed, apparently sh something.sh and ./something.sh are not as equivalent as I thought)

they aren't if `something.sh` syarts with `#! /bin/bash` or anything else that doesn't run `sh`.

> ./run_tests_modelling.sh: line 187: 22706 Segmentation fault 

oh well. Best thing to do is build with `RelWithDebInfo` , and change the script to use `gdb --args /path/to/.../POSMAPOSL whatever.par`. Then run, in `gdb` do `continue`, wait till it crashes, and then `info stack` etc

AnderBiguri:
> > With displacements instead of deformations (true, that was my mistake), the result is the same
> 
> the same as what? Does it now "work" or still zero sensitivity?

Returns zero images. The `subset_sensitivity%d.hv` images (when extracted, they are parametric) are zero. 
The frame sensitivities (stored by adding `    write_to_file("dyn_sensitivity_frame0.hv",dyn_sensitivity[this->_patlak_plot_sptr->get_starting_frame()]);` inside `PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData::add_subset_sensitivity`) is also zero (but interestingly has "number of frames =4" inside the `.hv` header, even if its only 1, I need to remove it to be able to see it in amide). 

> oh well. Best thing to do is build with RelWithDebInfo , and change the script to use gdb --args /path/to/.../POSMAPOSL whatever.par. Then run, in gdb do continue, wait till it crashes, and then info stack etc

Ok, step by step. 
> RelWithDebInfo 

Im in Linux, how do I do this? `-DCMAKE_BUILD_TYPE:STRING="RelWithDebInfo"` ?

> change the script to use `gdb --args /path/to/.../POSMAPOSL whatever.par.`

I have no idea what you mean. Change which script, the `run_tests_modelling`? The line that errors?
After some checking it seems that the line is calling POSMAPSOL in the most obfuscated way possible, you mean just running `gdb` on that particular call?

KrisThielemans:
> The frame sensitivities  ... is also zero

well, that's a nice bug then to look into...

I'll have to let you figure some of this out, but what you say is in the correct direction. See e.g. https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html. Obviously, you don't have to use `gdb`, if you prefer another debugger. And maybe you're using a nice IDE. There's some info on http://stir.sourceforge.net/registered/ amazingly enough.

AnderBiguri:
> > The frame sensitivities  ... is also zero
> 
> well, that's a nice bug then to look into...

Yes! But I don't know where in this code are the calls to the frame sensitivities.... Been looking for a while. Thus the comment above (somewhere)

Plus is strange that without motion they are not (I assume, as it does not return all zero for when I call this without DVFs)

rijobro:
A small disclaimer: this wasn't working before I left, thus why we hadn't merged it. Unfortunately, I can't remember what wasn't working... 

Is there a unit test for running the non-rigid displacement (would have to be using the experimental version)? If so, is it currently being tested with the multi file?

AnderBiguri:
> A small disclaimer: this wasn't working before I left, thus why we hadn't merged it. Unfortunately, I can't remember what wasn't working...
> 

Ah! That makes sense and changes a bit how I am testing it, as I keep trying to find bugs in my code (as its 99.9% of the times, my fault, maybe this is the 0.1%!). Its good to know, thanks! I'll check what happens on those DVFs, as they seem to be causing some issues :)

rijobro:
Apologies, should have flagged that up sooner! But really I can't remember which component(s) were causing issues.

AnderBiguri:
> Apologies, should have flagged that up sooner! But really I can't remember which component(s) were causing issues.

No worries! There were many things that I have being setting wrong until now, so its the right moment to flag that up 👍

KrisThielemans:
Probably the same problem as https://github.com/UCL/STIR/pull/580#issuecomment-649365249

PR #245, opened 2018-09-20
Bed position from interfile

rijobro:
If found in header, set in `ProjDataInfo`, but don't do anything else with it.

### DIFF
--- a/src/IO/InterfileHeader.cxx
+++ b/src/IO/InterfileHeader.cxx
@@ -1350,11 +1350,9 @@ bool InterfilePDFSHeader::post_processing()
     }
   //cerr << data_info_ptr->parameter_info() << endl;
   
-  // If there is a non-zero bed position, set it
-  if (bed_position_horizontal != 0.F)
-    data_info_ptr->set_bed_position_horizontal(bed_position_horizontal);
-  if (bed_position_vertical != 0.F)
-    data_info_ptr->set_bed_position_vertical(bed_position_vertical);
+  // Set the bed position
+  data_info_ptr->set_bed_position_horizontal(bed_position_horizontal);
+  data_info_ptr->set_bed_position_vertical(bed_position_vertical);
 
   return false;
 }
--- a/src/IO/InterfileHeader.cxx
+++ b/src/IO/InterfileHeader.cxx
@@ -600,6 +600,8 @@ InterfilePDFSHeader::InterfilePDFSHeader()
   add_key("applied corrections",
     KeyArgument::LIST_OF_ASCII, &applied_corrections);
 
+  bed_position = 0.F;
+  add_key("start horizontal bed position (mm)", &bed_position);
 }
 
 void InterfilePDFSHeader::resize_segments_and_set()
@@ -1346,6 +1348,10 @@ bool InterfilePDFSHeader::post_processing()
     }
   //cerr << data_info_ptr->parameter_info() << endl;
   
+  // If there is a non-zero bed position, set it
+  if (bed_position != 0.F)
+    data_info_ptr->set_bed_position(bed_position);
+
   return false;
 }
 
--- a/src/IO/InterfileHeader.cxx
+++ b/src/IO/InterfileHeader.cxx
@@ -230,6 +230,11 @@ InterfileHeader::InterfileHeader()
 
   add_key("energy window upper level",
          KeyArgument::FLOAT, &upper_en_window_thres);
+
+  bed_position_horizontal = 0.F;
+  add_key("start horizontal bed position (mm)", &bed_position_horizontal);
+  bed_position_vertical = 0.F;
+  add_key("start vertical bed position (mm)", &bed_position_vertical);
 }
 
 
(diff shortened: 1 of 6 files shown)


rijobro:
Ok for bed_position->bed_position_horizontal and creating bed_position_vertical.

But these are in InterfilePDFSHeader (ie projData related), but the Siemens version is in InterfileRawDataHeaderSiemens, so there is no inheritance. Can I move it to InterfilePDFSHeaderSiemens?

If not, do I create two new variables in InterfileRawDataHeaderSiemens and repeat all the code?

On 21 Sep 2018, at 07:16, Kris Thielemans <notifications@github.com<mailto:notifications@github.com>> wrote:


@KrisThielemans requested changes on this pull request.

bed_position needs to be called horizontal_bed_position. Probably worth adding the vertical at the same time as just copy-paste.

Seems you've missed the Siemens version of it (will have to remove the standard one and insert theirs)

KrisThielemans:
ah. We need it for the siemens hierarchy for the listmode data as well. Duplication is terrible of course.

As usual, this turns out to be harder than we'd hope. 

First of all, `CListModeData` has a private member `proj_data_info_sptr` etc, so will know about bed position. So that's good. Any listmode reader will therefore have to set the bed position. We'll have to create an issue for that.

I think the easiest solution is to move the `*_bed_position` stuff to `InterfileHeader`. That way, it will be in all. It isn't quite appropriate for images, but otherwise we need to create an `InterfileRawDataHeader` in stick that in between. That isn't trivial then as we'd need to adapt the Siemens hierarchy. I guess we can remove the keyword in `InterfileImageHeader` such that it is clear it will be ignored.

By the way, the writing also needs adapting.

rijobro:
Done. What writing needs to be adapted?

rijobro:
Writing of interfile also done.

src/IO/InterfileHeader.cxx
```diff
@@ -1346,6 +1350,12 @@ bool InterfilePDFSHeader::post_processing()
     }
   //cerr << data_info_ptr->parameter_info() << endl;
   
+  // If there is a non-zero bed position, set it
```
KrisThielemans: I don't think we should do the `if`. I'm not sure what the default should be for `ProjDataInfo`. We can argue about this but I think it's fine to say "default is zero".
In any case, using 0 to check for an unset value seems rather tricky as it means it cannot be set to zero...

src/IO/InterfileHeaderSiemens.cxx
```diff
@@ -226,9 +226,9 @@ InterfileRawDataHeaderSiemens::InterfileRawDataHeaderSiemens()
   add_key("tracer activity at time of injection (bq)", KeyArgument::NONE, &KeyParser::do_nothing);
   add_key("injected volume (ml)", KeyArgument::NONE, &KeyParser::do_nothing);
   add_key("horizontal bed translation", KeyArgument::NONE, &KeyParser::do_nothing);
-  add_key("start horizontal bed position (mm)", KeyArgument::NONE, &KeyParser::do_nothing);
+  add_key("start horizontal bed position (mm)", &bed_position_horizontal);
```
KrisThielemans: I'm fairly sure that Siemens keywords are different from the standard Interfile ones (check one of your files). If not, just remove these lines, else remove the normal one, and add the new one (pointing to the same field).

src/IO/InterfileHeaderSiemens.cxx
```diff
@@ -314,6 +314,12 @@ bool InterfileRawDataHeaderSiemens::post_processing()
     //XXX check if order here and segment_table are consistent
   }
 
+  // If there is a non-zero bed position, set it
+  if (bed_position_horizontal != 0.F)
```
KrisThielemans: same here

src/IO/interfile.cxx
```diff
@@ -1385,6 +1385,14 @@ write_basic_interfile_PDFS_header(const string& header_file_name,
     output_header<<"data offset in bytes[1] := "
                  <<pdfs.get_offset_in_stream()<<endl;
     
+  // Write bed position
+  if(fabs(pdfs.get_proj_data_info_sptr()->get_bed_position_vertical()) > 1.e-5F)
```
KrisThielemans: no `if`

src/include/stir/ProjDataInfo.h
```diff
@@ -349,6 +349,20 @@ class ProjDataInfo
   //! Return a string describing the object
   virtual std::string parameter_info() const;
   
+  //! Set horizontal bed position
+  void set_bed_position_horizontal(float &bed_position_horizontal_arg)
```
KrisThielemans: never a reference as argument, unless you want to change the argument, or it's a `const&` for a large object that you don't want to copy

src/include/stir/ProjDataInfo.h
```diff
@@ -349,6 +349,20 @@ class ProjDataInfo
...
+  { bed_position_horizontal = bed_position_horizontal_arg; }
+
+  //! Get horizontal bed position
+  float get_bed_position_horizontal() const { return bed_position_horizontal; }
+
+  //! Set vertical bed position
+  void set_bed_position_vertical(float &bed_position_vertical_arg)
```
KrisThielemans: same here

src/IO/InterfileHeaderSiemens.cxx
rijobro: Looking at a Siemens header they're the same, so I'll delete.

PR #246, opened 2018-09-24
Fix reading of dynamic projdata

rijobro:
Fixes leak caused by using `.get()` with shared pointer.

### DIFF
--- a/src/buildblock/DynamicProjData.cxx
+++ b/src/buildblock/DynamicProjData.cxx
@@ -170,8 +170,8 @@ read_from_file(const string& filename) // The written projection data is read in
         info(boost::format("DynamicProjData::read_from_file trying to read %s as a Multi file.") % filename);
 #endif
 
-      unique_ptr<MultipleProjData> multi_proj_data(MultipleProjData::read_from_file(filename).get());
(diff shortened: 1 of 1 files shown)


src/buildblock/DynamicProjData.cxx
```diff
@@ -170,8 +170,8 @@ read_from_file(const string& filename) // The written projection data is read in
         info(boost::format("DynamicProjData::read_from_file trying to read %s as a Multi file.") % filename);
 #endif
 
-      unique_ptr<MultipleProjData> multi_proj_data(MultipleProjData::read_from_file(filename).get());
-      unique_ptr<DynamicProjData> dynamic_proj_data(new DynamicProjData(*multi_proj_data));
+      const MultipleProjData &temp = *MultipleProjData::read_from_file(filename);
```
KrisThielemans: this makes little sense to me. You never know when the first unique ptr is going to disappear. And now we make yet another copy of the whole thing.

why did we need the `.get()` anyway? Seems to me that this should have worked
```C++
 unique_ptr<MultipleProjData> multi_proj_data(MultipleProjData::read_from_file(filename);
  unique_ptr<DynamicProjData> dynamic_proj_data(new DynamicProjData(*multi_proj_data));
```

Issue #241, opened 2018-09-20
Add ability for DiscretisedDensity to express voxel coordinates in LPS


ashgillman:
- [x] Writer
- [x] Reader (waiting on @rijobro 4D reader to be merged)

KrisThielemans:
all done?

ashgillman:
yes

PR #248, opened 2018-09-27
Add LPS to DiscretisedDensity

ashgillman:
Depends on #247 
Solves #241

### DIFF
--- a/src/include/stir/DiscretisedDensity.inl
+++ b/src/include/stir/DiscretisedDensity.inl
@@ -250,14 +250,21 @@ swap_axes_based_on_orientation(CartesianCoordinate3D<float>& coordinates,
 template<int num_dimensions, typename elemT>
 CartesianCoordinate3D<float>
 DiscretisedDensity<num_dimensions, elemT>::
+get_LPS_coordinates_for_physical_coordinates(const CartesianCoordinate3D<float>& coords) const
+{
+  CartesianCoordinate3D<float> flip_coords = coords;
+  const PatientPosition patient_position = this->get_exam_info().patient_position;
+  swap_axes_based_on_orientation(flip_coords, patient_position);
+  return flip_coords;
+}
+
+template<int num_dimensions, typename elemT>
+CartesianCoordinate3D<float>
+DiscretisedDensity<num_dimensions, elemT>::
 get_LPS_coordinates_for_indices(const BasicCoordinate<num_dimensions, float>& indices) const
 {
-  CartesianCoordinate3D<float> coordinates
-    = this->get_physical_coordinates_for_indices(indices);
-  const PatientPosition patient_position = this->get_exam_info().patient_position;
-
-  swap_axes_based_on_orientation(coordinates, patient_position);
-  return coordinates;
+  return this->get_LPS_coordinates_for_physical_coordinates
+    (this->get_physical_coordinates_for_indices(indices));
 }
 
 template<int num_dimensions, typename elemT>
@@ -269,17 +276,21 @@ get_LPS_coordinates_for_indices(const BasicCoordinate<num_dimensions, int>& indi
 }
 
 template<int num_dimensions, typename elemT>
-BasicCoordinate<num_dimensions, float>
+CartesianCoordinate3D<float>
+DiscretisedDensity<num_dimensions, elemT>::
+get_physical_coordinates_for_LPS_coordinates(const CartesianCoordinate3D<float>& coords) const
+{
+  // operation is symmetric
+  return this->get_LPS_coordinates_for_physical_coordinates(coords);
+}
+
+template<int num_dimensions, typename elemT>
+BasicCoordinate<num_dimensions,float>
 DiscretisedDensity<num_dimensions, elemT>::
 get_index_coordinates_for_LPS_coordinates(const CartesianCoordinate3D<float>& coords) const
 {
-  CartesianCoordinate3D<float> flip_coords = CartesianCoordinate3D<float>(coords);
-  const PatientPosition patient_position = this->get_exam_info().patient_position;
-  swap_axes_based_on_orientation(flip_coords, patient_position);
-
-  CartesianCoordinate3D<float> indices
-    = this->get_index_coordinates_for_physical_coordinates(flip_coords);
-  return indices;
+  return this->get_index_coordinates_for_physical_coordinates
+    (this->get_physical_coordinates_for_LPS_coordinates(coords));
 }
 
 template<int num_dimensions, typename elemT>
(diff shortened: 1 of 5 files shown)


ashgillman:
Rebased and handled comments. But the main change is that the only way to logically resolve alignment was if I assume the internal physical coordinates are actually:
- x running left-right
- y running top-bottom
- z running deep-front

from perspective of standing a bed end, facing gantry.

I took a reconstruction where the subject is at an angle and reconstructed to .hv. Then used `stir_math` to convert to .nii. Then compared with vendor reconstruction.

NB: I think my normalisation is incorrect in the image, but that is unrelated.

image
image
image
image

KrisThielemans:
right. well, I guess that documentation was written about 15 years ago based on experiments with older scanners (either ECAT 953 or GE Advance). It's possible that our conventions are not consistent between those scanners and mMR (and presumably other Siemens scanners). We'll have to put a large caveat on this then.

ashgillman:
Also depends #256 since rebasing

src/include/stir/DiscretisedDensity.inl
```diff
@@ -210,4 +210,100 @@ get_indices_closest_to_physical_coordinates(const CartesianCoordinate3D<float>&
...
+swap_axes_based_on_orientation(CartesianCoordinate3D<float>& coordinates,
+                               const PatientPosition patient_position)
+{
+  // STIR coordinates run:
+  // x: left to right as you face the scanner (as seen from the bed side)
+  // y: top to bottom of the scanner
+  // z: running from deep in the scanner out along the bed (as seen from the bed
+  //    side)
+  // std::cerr << patient_position.get_position_as_string() << std::endl;
+  switch (patient_position.get_position()) {
```
KrisThielemans: coding style: `{` on next line

src/include/stir/DiscretisedDensity.inl
```diff
@@ -210,4 +210,100 @@ get_indices_closest_to_physical_coordinates(const CartesianCoordinate3D<float>&
...
+    throw std::runtime_error("Unsupported patient position, can't convert to LPS.");
+  }
+}
+
+template<int num_dimensions, typename elemT>
+CartesianCoordinate3D<float>
+DiscretisedDensity<num_dimensions, elemT>::
+get_LPS_coordinates_for_physical_coordinates(const CartesianCoordinate3D<float>& coords) const
+{
+  CartesianCoordinate3D<float> flip_coords = coords;
```
KrisThielemans: I sort of prefer the following signature but don't mind too much
```C++
CartesianCoordinate3D<float> flip_coords =swap_axes_based_on_orientation(coords, patient_position);
```

src/test/test_VoxelsOnCartesianGrid.cxx
```diff
@@ -253,6 +253,81 @@ VoxelsOnCartesianGridTests::run_tests()
     }
 
   }
+
+  {
+    cerr << "Test LPS with different orientations:" << std::endl;
+
+    shared_ptr<ExamInfo> hfs_exam_info_sptr = shared_ptr<ExamInfo>(new ExamInfo());
```
KrisThielemans: too complicated :-;
```C++
 shared_ptr<ExamInfo> hfs_exam_info_sptr(new ExamInfo());
```

src/include/stir/DiscretisedDensity.inl
ashgillman: So do I, shall change

ashgillman:
I think all comments should be addressed. See how CI goes

ashgillman:
stir-lps_conversion


I've attached a very professional reference diagram I have produced and have been referring to. It shows a patient in the scanner. On the left are the directions of axes I find it logical that it seems STIR uses given the following outlined experiment. On the right are the axes according to the documentation. I have labelled each both with x, y, z, as well as the patient-based direction.

I encourage anyone interested to look over the relevant functions for an error in my logic.

Here I encode the orientations:
https://github.com/UCL/STIR/blob/1e00b030feb9e97be298a2b6850e4e2d477b5111/src/include/stir/DiscretisedDensity.inl#L215-L253

The only relevant one is HFS as it is all I currently support in IO PR. I flip z in order to convert a LPI (assuming the proposed modified axes) to LPS.

Other orientations are tested here:
https://github.com/UCL/STIR/blob/1e00b030feb9e97be298a2b6850e4e2d477b5111/src/test/test_VoxelsOnCartesianGrid.cxx#L286-L343

based on known relations. They will have to be properly tested with phantoms in the future.

I tested the orientation using the ITK output writer. The relevant code is where the ITK direction matrix is set. Each column in this matrix corresponds to a unit vector in the direction of axes, where x=0, y=1, z=2 (as opposed to STIR dimension order, z=1, y=2, x=3).
https://github.com/UCL/STIR/blob/a526b4e5449944bc42a8b5360fc786abc166b279/src/IO/ITKOutputFileFormat.cxx#L147-L158

I took a STIR reconstruction in interfile, and used stir_math to convert to a NIfTI. I compare the NIfTI image and DICOM image using the MRTrix3 MRView viewer, which I have found to be the best viewer for such tasks. I visually compared the orientation of the images, and confirm that they are the same under the current configuration in the DiscretisedDensity code. I also compared alignment with the vendor DICOM converted to NIfTI using the dcmstack https://github.com/moloney/dcmstack converter, which I have also previously compared with the MEDCON converter. All have been consistent.

The subject in the image is lying tilted, such that left-right symmetry is broken. See below images. The blue image is Vendor DICOM, and the red image is the NIfTI output.

orientation_sag
orientation_cor
orientation_ax1
orientation_ax2


I will be including this in my presentation as the Users' and Developers' Meeting.

Ash

ashgillman:
I should add, that this analysis is also historically accurate. By which I mean, ITK IO has always assumed the patient is oriented in HFS. The ITK input has always reordered voxels before copying into STIR such that the origin is at the right-anterior-superior-most voxel, meaning that the orientation is LPI (yes, ITK reorienter defines the coordinate orientation by the position of the origin rather that the direction of the axes, the reverse of the commonplace terminology).
https://github.com/UCL/STIR/blob/e6c14b7f5d50d43fb27741b1886b88a7d961a754/src/IO/ITKImageInputFileFormat.cxx#L353

Similarly, the output has always defined a direction
    1 0 0
    0 1 0
    0 0 -1
which, defines LPI.
https://github.com/UCL/STIR/blob/e6c14b7f5d50d43fb27741b1886b88a7d961a754/src/IO/ITKOutputFileFormat.cxx#L146-L150

ashgillman:
Both of these parts of code appear to have been added by @alalehrn here
https://github.com/UCL/STIR/commit/4dd09e48a269a6af78b6919b54ee67ad76eda58a

@alalehrn, was this tested just on the mMR data? If so, I guess my appeal to historical accuracy doesn't mean that the current code is not mMR-specific.

Issue #242, opened 2018-09-20
Update ITK Reader/Writer to be orientation-aware


ashgillman:
Depends #241

ashgillman:
I guess also depends on @rijobro's parametric modifications

KrisThielemans:
closed by #257

PR #257, opened 2018-10-04
[ENH] ITK IO to be LPS-aware

ashgillman:
Solution for #242 
Depends #248 

Still one TODO - reading orientation from DICOM. But works for now with NIfTI by assuming HFS (actually it calls it unknown, this is just treated like HFS in #248).

### DIFF
--- a/documentation/release_3.1.htm
+++ b/documentation/release_3.1.htm
@@ -67,7 +67,8 @@ improvements to the documentation.
   and Richard Brown (UCL)
   but with contributions by Jannis Fischer (ETH Zurich) for SAFIR support,
   Elise Emond for IO fixes (UCL), Edoardo Pasca (STFC) for Siemens interfile support,
-  Daniel Deida (Leeds) for the PLS prior, and some code clean-up by David Volgyes. Various bits
+  Daniel Deida (Leeds) for the PLS prior, Ashley Gillman (CSIRO AeHRC) for patient orientation and
+  geometrical support, and some code clean-up by David Volgyes. Various bits
 were tested by (in alphabetical order, all UCL except where indicated):
 Ottavia Bertolli, Matthias Ehrhardt, Pawel Markiewicz, Richard Manber, Alaleh Rashidnasab,
 Benjamin Thomas (CIRC ASTAR and UCL), Yu-jung Tsai, Vesna Cuplov).
@@ -161,7 +162,8 @@ These are only minimally documented at the present stage unfortunately.
 </li>
 <li>
 The orientation of images read/written via ITK has changed. It should now be correct if the patient
-was in HFS position. This is currently not checked (and for many file formats, never can be checked).
+was in HFS position for NIfTI (this can't be verified due to lack of metadata in the format), or if
+in DICOM format for any of HFS, FFS, HFP or FFP if the DICOM containts the correct metadata.
 </li>
 <li><tt>calculate_attenuation_coefficients</tt> now defaults to using the ray-tracing matrix as
   projector. Results should be identical though. This means it will not automatically handle
--- a/documentation/release_3.1.htm
+++ b/documentation/release_3.1.htm
@@ -162,8 +162,9 @@ These are only minimally documented at the present stage unfortunately.
 </li>
 <li>
 The orientation of images read/written via ITK has changed. It should now be correct if the patient
-was in HFS position for NIfTI (this can't be verified due to lack of metadata in the format), or if
-in DICOM format for any of HFS, FFS, HFP or FFP if the DICOM containts the correct metadata.
+was in HFS position. Unfortunately, for most file formats (such as NifTI) this can't be verified due to lack of metadata in the format.
+If the files are in DICOM format, as long as the file contains the correct metadata, other patient positions are supported (currently
+HFS, FFS, HFP or FFP).
 </li>
 <li><tt>calculate_attenuation_coefficients</tt> now defaults to using the ray-tracing matrix as
   projector. Results should be identical though. This means it will not automatically handle
--- a/src/IO/ITKImageInputFileFormat.cxx
+++ b/src/IO/ITKImageInputFileFormat.cxx
@@ -57,13 +57,16 @@ typedef VoxelsOnCartesianGrid<CartesianCoordinate3D<float> > STIRImageMulti;
 // typedef DiscretisedDensity<3, CartesianCoordinate3D<float> > STIRImageMulti;
 // typedef VoxelsOnCartesianGrid<CartesianCoordinate3D<float> > STIRImageMultiConcrete;
 
+/* Convert an ITK image into an internal STIR one.
+   warning: modifies itk_image.
+*/
 static
 STIRImageSingle*
-convert_ITK_to_STIR(const ITKImageSingle::Pointer itk_image);
+convert_ITK_to_STIR(ITKImageSingle::Pointer itk_image);
 
 static
 STIRImageMulti*
-convert_ITK_to_STIR(const ITKImageMulti::Pointer itk_image);
+convert_ITK_to_STIR(ITKImageMulti::Pointer itk_image);
 
 template<typename STIRImageType>
 static
@@ -206,9 +209,8 @@ construct_empty_stir_image_and_orient_itk_image(typename ITKImageType::Pointer i
 
 // Actual conversion function
 STIRImageSingle*
-convert_ITK_to_STIR(const ITKImageSingle::Pointer itk_image_orig)
+convert_ITK_to_STIR(ITKImageSingle::Pointer itk_image)
 {
-  ITKImageSingle::Pointer itk_image = itk_image_orig->Clone();
   STIRImageSingle* image_ptr = construct_empty_stir_image_and_orient_itk_image
     <ITKImageSingle, STIRImageSingle, STIRImageSingleConcrete>(itk_image);
 
@@ -225,9 +227,8 @@ convert_ITK_to_STIR(const ITKImageSingle::Pointer itk_image_orig)
 
 // Actual conversion function
 STIRImageMulti*
-convert_ITK_to_STIR(const ITKImageMulti::Pointer itk_image_orig)
+convert_ITK_to_STIR(ITKImageMulti::Pointer itk_image)
 {
-  ITKImageMulti::Pointer itk_image = itk_image_orig->Clone();
   STIRImageMulti* image_ptr = construct_empty_stir_image_and_orient_itk_image
     <ITKImageMulti, STIRImageMulti, STIRImageMulti>(itk_image);
 
--- a/src/IO/ITKImageInputFileFormat.cxx
+++ b/src/IO/ITKImageInputFileFormat.cxx
@@ -505,6 +505,10 @@ read_file_itk(const std::string &filename)
                 __FILE__, __LINE__);
           return NULL; }
 
+      warning("Only displacement fields are currently supported in STIR (not deformations). "
+              "There is no way of verifying this from the nifti_image metadata, so you need to "
+              "make sure that the image you are supplying is a displacement field image.");
+
       ITKImageMulti::Pointer itk_image = reader->GetOutput();
 
       return convert_ITK_to_STIR<ITKImageMulti, STIRImageMulti>
--- a/src/IO/ITKImageInputFileFormat.cxx
+++ b/src/IO/ITKImageInputFileFormat.cxx
@@ -69,13 +69,13 @@ template<typename STIRImageType>
 static
 STIRImageType *
 read_file_itk(const std::string &filename);
-  
+
+template<typename STIRImageType>
 static
 CartesianCoordinate3D<float>
 ITK_coordinates_to_STIR(const itk::ImageBase<3>::PointType &itk_coord,
-        CartesianCoordinate3D<float> &voxel_size,
-        const BasicCoordinate<3,int> &min_indices,
-        bool is_displacement_field = false);
+                        const STIRImageType stir_image,
+                        bool is_displacement_field = false);
 
 template<typename ITKImageType>
 static
@@ -141,41 +141,39 @@ read_from_file(const std::string& filename) const
 
 template<typename ITKImageType>
 static
-void
-calc_stir_min_max_indices(BasicCoordinate<3, int> &min_indices,
-                          BasicCoordinate<3, int> &max_indices,
-                          const typename ITKImageType::Pointer itk_image)
+IndexRange<3>
+calc_stir_index_range(const typename ITKImageType::Pointer itk_image)
 {
   // find index range in usual STIR convention
   const int z_size = itk_image->GetLargestPossibleRegion().GetSize()[2];
   const int y_size = itk_image->GetLargestPossibleRegion().GetSize()[1];
   const int x_size = itk_image->GetLargestPossibleRegion().GetSize()[0];
-  min_indices = BasicCoordinate<3,int>(make_coordinate(0, -y_size/2, -x_size/2));
-  max_indices = min_indices + make_coordinate(z_size, y_size, x_size) - 1;
+  BasicCoordinate<3, int> min_indices
+    = BasicCoordinate<3,int>(make_coordinate(0, -y_size/2, -x_size/2));
+  BasicCoordinate<3, int> max_indices
+    = min_indices + make_coordinate(z_size, y_size, x_size) - 1;
+  return IndexRange<3>(min_indices, max_indices);
 }
 
 template<typename ITKImageType>
 static
 const CartesianCoordinate3D<float>
 calc_stir_origin(CartesianCoordinate3D<float> voxel_size,
-                 BasicCoordinate<3, int> min_indices, BasicCoordinate<3, int> &max_indices,
+                 IndexRange<3> index_range,
                  const typename ITKImageType::Pointer itk_image)
 {
   // dummy image that has minumum to be able to find ITK -> STIR origin vector
+  CartesianCoordinate3D<float> stir_origin_index(0, 0, 0);
   const VoxelsOnCartesianGrid<float> dummy_image =
-    VoxelsOnCartesianGrid<float>(IndexRange<3>(min_indices, max_indices),
-                                 CartesianCoordinate3D<float>(0, 0, 0),
+    VoxelsOnCartesianGrid<float>(index_range,
+                                 stir_origin_index,
                                  voxel_size);
-  const CartesianCoordinate3D<float> itk_origin
-    = dummy_image.get_physical_coordinates_for_LPS_coordinates
-      (CartesianCoordinate3D<float>(static_cast<float>(itk_image->GetOrigin()[2]),
-                                    static_cast<float>(itk_image->GetOrigin()[1]),
-                                    static_cast<float>(itk_image->GetOrigin()[0])));
-  const CartesianCoordinate3D<float> stir_origin_wrt_itk_origin
-    = -dummy_image.get_relative_coordinates_for_indices(min_indices);
-  const CartesianCoordinate3D<float> stir_origin
-    = stir_origin_wrt_itk_origin + itk_origin;
-  return stir_origin;
+
+  // return ITK_coordinates_to_STIR<VoxelsOnCartesianGrid<float> >
+  //   (itk_origin, dummy_image);
+
+  return ITK_coordinates_to_STIR<VoxelsOnCartesianGrid<float> >
+    (itk_image->GetOrigin(), dummy_image);
 }
 
 // Actual conversion function
@@ -196,14 +194,13 @@ convert_ITK_to_STIR(const ITKImageSingle::Pointer itk_image_orig)
                                           static_cast<float>(itk_image->GetSpacing()[0]));
 
   // find info STIR image geometrical metadata
-  BasicCoordinate<3,int> min_indices, max_indices;
-  calc_stir_min_max_indices<ITKImageSingle>(min_indices, max_indices, itk_image);
+  IndexRange<3> index_range = calc_stir_index_range<ITKImageSingle>(itk_image);
   const CartesianCoordinate3D<float> stir_origin = calc_stir_origin<ITKImageSingle>
-    (voxel_size, min_indices, max_indices, itk_image);
+    (voxel_size, index_range, itk_image);
 
   // create STIR image
   STIRImageSingle* image_ptr = new STIRImageSingleConcrete
-    (exam_info_sptr, IndexRange<3>(min_indices, max_indices), stir_origin, voxel_size);
+    (exam_info_sptr, index_range, stir_origin, voxel_size);
 
   // copy data
   VoxelsOnCartesianGrid<float>::full_iterator stir_iter = image_ptr->begin_all();
@@ -234,14 +231,15 @@ convert_ITK_to_STIR(const ITKImageMulti::Pointer itk_image_orig)
                                           static_cast<float>(itk_image->GetSpacing()[0]));
 
   // find info STIR image geometrical metadata
-  BasicCoordinate<3,int> min_indices, max_indices;
-  calc_stir_min_max_indices<ITKImageMulti>(min_indices, max_indices, itk_image);
+  IndexRange<3> index_range = calc_stir_index_range<ITKImageMulti>(itk_image);
+  BasicCoordinate<3, int> min_indices, max_indices;
+  index_range.get_regular_range(min_indices, max_indices);
   const CartesianCoordinate3D<float> stir_origin = calc_stir_origin<ITKImageMulti>
-    (voxel_size, min_indices, max_indices, itk_image);
+    (voxel_size, index_range, itk_image);
 
   // create STIR image
   STIRImageMulti* image_ptr = new STIRImageMulti
-    (exam_info_sptr, IndexRange<3>(min_indices, max_indices), stir_origin, voxel_size);
+    (exam_info_sptr, index_range, stir_origin, voxel_size);
 
   // copy data
   STIRImageMulti::full_iterator stir_iter = image_ptr->begin_all();
@@ -254,8 +252,8 @@ convert_ITK_to_STIR(const ITKImageMulti::Pointer itk_image_orig)
       itk_coord[1] = -double(it.Get()[1]);
       itk_coord[2] =  double(it.Get()[2]);
 
-      *stir_iter = ITK_coordinates_to_STIR
-        (itk_coord, voxel_size, min_indices, true);
+      *stir_iter = ITK_coordinates_to_STIR<STIRImageMulti>
+        (itk_coord, *image_ptr, true);
   }
   return image_ptr;
 }
@@ -350,7 +348,7 @@ template<>
 STIRImageMulti*
 read_file_itk(const std::string &filename)
 {
-  typedef itk::GDCMImageIO       ImageIOType;
+  //typedef itk::GDCMImageIO       ImageIOType;
 
   try
     {
@@ -457,26 +455,33 @@ struct removePtr<itk::SmartPointer<Type> >
   typedef Type type;
 };
 
+template<typename STIRImageType>
 CartesianCoordinate3D<float>
-ITK_coordinates_to_STIR(
-        const itk::ImageBase<3>::PointType &itk_coord,
-        CartesianCoordinate3D<float> &voxel_size,
-        const BasicCoordinate<3,int> &min_indices,
-        bool is_displacement_field)
+ITK_coordinates_to_STIR(const itk::ImageBase<3>::PointType &itk_coord,
+                        const STIRImageType stir_image,
+                        bool is_displacement_field)
 {
   // find STIR origin
   // Note: need to use - for z-coordinate because of different axis conventions
-  // Only works for HFS!
-  CartesianCoordinate3D<float> stir_coord(static_cast<float>(itk_coord[2]),
-                                          static_cast<float>(itk_coord[1]),
-                                          static_cast<float>(itk_coord[0]));
+  CartesianCoordinate3D<float> stir_coord
+    = stir_image.get_physical_coordinates_for_LPS_coordinates
+      (CartesianCoordinate3D<float>(static_cast<float>(itk_coord[2]),
+                                    static_cast<float>(itk_coord[1]),
+                                    static_cast<float>(itk_coord[0])));
 
   // The following is not required for displacement field images
   if (!is_displacement_field)
   {
-    // make sure that origin is such that
-    // first_pixel_offsets =  min_indices*voxel_size + origin
-    stir_coord -= voxel_size * BasicCoordinate<3,float>(min_indices);
+    // dummy image that has minumum to be able to find ITK -> STIR origin vector
+    CartesianCoordinate3D<float> stir_origin_index(0, 0, 0);
+
+    // assuming we previously oriented the ITK image, min_indices is the
+    // index where ITK origin points to in physical space
+    const CartesianCoordinate3D<float> stir_origin_wrt_itk_origin
+      = stir_image.get_physical_coordinates_for_indices(stir_origin_index)
+      - stir_image.get_physical_coordinates_for_indices(stir_image.get_min_indices());
+
+    stir_coord += stir_origin_wrt_itk_origin;
   }
 
   return stir_coord;
--- a/src/IO/ITKImageInputFileFormat.cxx
+++ b/src/IO/ITKImageInputFileFormat.cxx
@@ -277,7 +277,7 @@ construct_exam_info_from_metadata_dictionary(itk::MetaDataDictionary dictionary)
    This method expects that itk_image is already oriented to be consistent with
    STIR x, y, z axes.
  */
-template<typename ITKImageType, typename STIRImageType, typename STIRConcreteType>
+template<typename ITKImageType, typename STIRImageType>
 STIRImageType*
 construct_empty_stir_image(typename ITKImageType::Pointer itk_image,
                            shared_ptr<ExamInfo> exam_info_sptr)
@@ -294,7 +294,7 @@ construct_empty_stir_image(typename ITKImageType::Pointer itk_image,
     (voxel_size, index_range, itk_image);
 
   // create STIR image
-  STIRImageType* image_ptr = new STIRConcreteType
+  STIRImageType* image_ptr = new STIRImageType
     (exam_info_sptr, index_range, stir_origin, voxel_size);
   return image_ptr;
 }
@@ -370,7 +370,7 @@ orient_ITK_image(const typename ITKImageType::Pointer itk_image_orig,
 }
 
 /* Convert an ITK image into an internal STIR one. */
-template<typename ITKImageType, typename STIRImageType, typename STIRConcreteType>
+template<typename ITKImageType, typename STIRImageType>
 STIRImageType*
 convert_ITK_to_STIR(const typename ITKImageType::Pointer itk_image)
 {
@@ -382,7 +382,7 @@ convert_ITK_to_STIR(const typename ITKImageType::Pointer itk_image)
     = orient_ITK_image<ITKImageType>(itk_image, exam_info_sptr);
   // Make the STIR Image
   STIRImageType* stir_image_ptr = construct_empty_stir_image
-    <ITKImageType, STIRImageType, STIRConcreteType>(reor_itk_image, exam_info_sptr);
+    <ITKImageType, STIRImageType>(reor_itk_image, exam_info_sptr);
   // Copy the ITK image data into the STIR Image
   copy_ITK_data_to_STIR_image<ITKImageType, STIRImageType>
     (reor_itk_image, stir_image_ptr, false);
@@ -409,7 +409,7 @@ read_file_itk(const std::string &filename)
           ITKImageSingle::Pointer itk_image = reader->GetOutput();
 
           return convert_ITK_to_STIR
-            <ITKImageSingle, STIRImageSingle, STIRImageSingleConcrete>
+            <ITKImageSingle, STIRImageSingleConcrete>
             (itk_image);
         }
       else
@@ -469,7 +469,7 @@ read_file_itk(const std::string &filename)
 
           // Finally, convert to STIR!
           return convert_ITK_to_STIR
-            <ITKImageSingle, STIRImageSingle, STIRImageSingleConcrete>
+            <ITKImageSingle, STIRImageSingleConcrete>
             (itk_image);
      
         }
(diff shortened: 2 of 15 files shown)


ashgillman:
@rijobro perhaps you could add this PR to the Image Geometry project for bookkeeping?

KrisThielemans:
@ashgillman could you rebase this on master resolving the conflict that appears after I squash-merged #248?

ashgillman:
Done

ashgillman:
There are some methods here that return image pointers. Should these be changes to be shared pointers?

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -140,76 +139,113 @@ read_from_file(const std::string& filename) const
...
+    <ITKImageMulti, STIRImageMulti, STIRImageMulti>(itk_image);
+
+  // copy data
+  STIRImageMulti::full_iterator stir_iter = image_ptr->begin_all();
+  typedef itk::ImageRegionConstIterator<ITKImageMulti> IteratorType;
+  IteratorType it (itk_image, itk_image->GetLargestPossibleRegion() );
+  for ( it.GoToBegin(); !it.IsAtEnd(); ++it, ++stir_iter) {
+      itk::Point<double,3U> itk_coord;
+
+      itk_coord[0] = -double(it.Get()[0]);
```
KrisThielemans: why the minus signs here?

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -336,46 +372,55 @@ read_file_itk(const std::string &filename)
...
+  typename OrienterType::Pointer orienter = OrienterType::New();
+  orienter->UseImageDirectionOn();
+  orienter->SetInput(itk_image_orig);
+
+  // We need the origin to be in the minimum x, y, z corner. This
+  // depends on the patient position
+  switch (exam_info_sptr->patient_position.get_position()) {
+  case PatientPosition::unknown_position:
+    // If unknown, assume HFS
+    // TODO: warning?
```
KrisThielemans: yes please.

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -400,26 +445,33 @@ struct removePtr<itk::SmartPointer<Type> >
...
 
+template<typename STIRImageType>
 CartesianCoordinate3D<float>
-ITK_coordinates_to_STIR(
-        const itk::ImageBase<3>::PointType &itk_coord,
-        CartesianCoordinate3D<float> &voxel_size,
-        const BasicCoordinate<3,int> &min_indices,
-        bool is_displacement_field)
+ITK_coordinates_to_STIR(const itk::ImageBase<3>::PointType &itk_coord,
+                        const STIRImageType stir_image,
```
KrisThielemans: reference to `STIRImageType`?

ashgillman:
Since rebasing I am getting a numerical exception. Has anyone seen this before? Div by 0?

```
31: Test command: /home/gil2a4/dev/UCL@github.com/STIR/build/src/test/test_IO_DiscretisedDensity "/home/gil2a4/dev/UCL@github.com/STIR/src/test/input/test_ITKNiftiOutputFileFormat.in"
31: Test timeout computed to be: 10000000                                                                                                                  
31:                                                                                                                                                        
31: WARNING: FactoryRegistry:: overwriting previous value of key in registry.
31:      key: None
31: Testing OutputFileFormat parsing function...
31: WARNING: will overwite files called STIRtmp*
31: 
31: About to write the image to disk...
31: OK!
31: 
31: About to read the image back from disk...
5/5 Test #31: test_IO_DiscretisedDensity_test_ITKNiftiOutputFileFormat.in ..........***Exception: Numerical  0.23 sec  
```

ashgillman:
Issue was that itk_image->Clone() returned an empty image with none of the same data nor metadata as itk_image. I opted to allow conversion to modify itk_image, since it is not used elsewhere. Alternatively, this seems to be the way to duplicate an image - rather verbose in classic ITK style.

https://itk.org/Wiki/ITK/Examples/SimpleOperations/ImageDuplicator

src/IO/ITKImageInputFileFormat.cxx
ashgillman: @rijobro I assume this can go now, I'll remove it. Best you double check on your data.

src/IO/ITKImageInputFileFormat.cxx
ashgillman: Actually, best when we set patient_position to unknown I think

src/IO/ITKImageInputFileFormat.cxx
ashgillman: `const STIRImageType& stir_image`? Sure, but  curious why?

KrisThielemans: yes. (potentially) big objects should always be passed by (preferably `const`) reference. Note for stuff like `int`, `float` etc. Really should add this to our developers-guide...

KrisThielemans:
As far as I'm concerned, this can be merged. ok?

Regarding DICOM, I've posted [a question to the ITK Community list](https://itk.org/pipermail/community/2018-October/014012.html)

KrisThielemans:
The ITK list is now replaced by a forum. My question is [here](https://discourse.itk.org/t/get-info-such-as-patient-orientation-timing-from-images/). The answer points to [the section of the manual on DICOM IO](https://itk.org/ITKSoftwareGuide/html/Book2/ITKSoftwareGuide-Book2ch1.html#x7-240001.12). The relevant bit is at the end.

@ashgillman, want to do that before merging, or later?

ashgillman:
Before probably makes sense, I'll have a look into it.

ashgillman:
## Where's the error?
The reconstructions appear correctly oriented, so the error I guess must be in the reader.

## Is the reorienter working?
It looks like it is the reorienter after I read and the reader reorients, I get the Direction and Origin changing as such:
```
1 0 0
0 -1 0
0 0 1
[-359.216, 356.894, -283.957]
1 0 0
0 1 0
0 0 -1
[-359.216, -358.693, 128.387]
```

> Side note
> Why this LAS input axis order? The input NIfTI has affine
> [[  -2.08626008,   -0.        ,    0.        ,  359.21588135],
>        [   0.        ,    2.08626008,    0.        , -356.89422607],
>        [   0.        ,    0.        ,    2.03125   , -283.95703125],
>        [   0.        ,    0.        ,    0.        ,    1.        ]]
> NIfTI affine is defines w.r.t. RAS, so image is therefore in LAS


This is the expected behaviour. Notice though that the axes flipping is the SI and AP directions, the same as I am getting an issue with.

## Does manually messing with the reorienter help?
When I manually change the reorienter for HFS from RAS to RPI, the resulting image out has shifted its origin (an expected consequence), but the axes directions are still incorrect.

## The reorienter works on the meta data, what about voxel data?
So maybe the reorienter isn't actually re-ordering the voxel data. Or maybe the iterator is smart in some way and iterates not in stored order?

So I added this snippet to check the iterator data is actually changing:
```
  std::cerr << itk_image->GetDirection();
  std::cerr << itk_image->GetOrigin() << std::endl;
  typename ITKImageType::Pointer new_itk_image
    = orient_ITK_image<ITKImageType>(exam_info_sptr, itk_image);
  std::cerr << new_itk_image->GetDirection();
  std::cerr << new_itk_image->GetOrigin() << std::endl;

  typedef itk::ImageRegionConstIterator<ITKImageType> IteratorType;
  IteratorType it (itk_image, itk_image->GetLargestPossibleRegion());
  IteratorType nit (new_itk_image, itk_image->GetLargestPossibleRegion());
  int n_different = 0;
  for (it.GoToBegin(); !it.IsAtEnd(); ++it, ++nit)
  {
    n_different += (it.Get() != nit.Get());
  }
  std::cerr << "n different: " << n_different << std::endl;
```

which returns
```
1 0 0                                                                                           
0 -1 0                                                                       
0 0 1                               
[-359.216, 356.894, -283.957]                                                                               
1 0 0                                           
0 1 0                                     
0 0 -1                                                      
[-359.216, -358.693, 128.387]         
n different: 3465846
```
So the reorienter is changing the voxel data...

## Test the test
check that if I craft the reorienter to do nothing that no voxels change (reorient to LAS, the same as this particular input):
```
1 0 0
0 -1 0
0 0 1
[-359.216, 356.894, -283.957]
1 0 0
0 -1 0
0 0 1
[-359.216, 356.894, -283.957]
n different: 0
```

## So...
I must admit I am stumped...

ashgillman:
Fool! The argument to that function is also a return value - so I should have either made it a reference or reset the pointer. ITK smart pointers don't seem to have a reset function, so I pass a smart pointer by reference.

So the issue was that all the metadata was using the correct reoriented image, but when the data was being filled, the original no-reoriented image was being used.

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -294,11 +294,6 @@ read_file_itk(const std::string &filename)
           typedef itk::GDCMSeriesFileNames NamesGeneratorType;
           typename NamesGeneratorType::Pointer nameGenerator = NamesGeneratorType::New();
           nameGenerator->SetUseSeriesDetails( true );
-          // Make sure we read only data from a single frame and gate
-          nameGenerator->AddSeriesRestriction("0008|0022" ); // AcquisitionDate
-          nameGenerator->AddSeriesRestriction("0008|0032" ); // AcquisitionTime
-          nameGenerator->AddSeriesRestriction("0018|1060" ); // TriggerTime
-          nameGenerator->AddSeriesRestriction("0018|1063" ); // FrameTime
```
ashgillman: @KrisThielemans 

I found that when reading DICOM composed-multi-bed-position images, the times are different depending on the slice. But I'd have thought that the default name generator arguments would be sufficient in most cases - although I haven't tested on any dynamic data (which I suspect is the reason for this).

ashgillman: NB: reading composed images is probably not a common use-case so if this conflicts with frames of dynamic images then I can revert this change
NB NB: Perhaps this actually presents an issue if you try to read a dynamic DICOM data @rijobro

ashgillman:
Alright, I think this might be done, pending your review and CI :)

src/IO/ITKImageInputFileFormat.cxx
KrisThielemans: At some point, we need to generalise this class to be able to read dynamic (or gated) data, with correct time frame info being returned. However, at the moment, this function is used for a single time frame (or gate). without the restriction, if the user has multiple gates/frames in the directory, the STIR image is going to be nonsensical (i.e. incorrect).

It's of course tricky what to do with "composed" images (i.e. data reconstructed from multiple-bed positions). The time frame info that we'd return is going to be incorrect in any case.

I feel this is a complication we don't really want to handle right now. I therefore would like to re-instate the restrictions. This way, the user can tell the reader to read a specific time-frame/gate (although in an awkward way).

It's a general problem with the "nice" reading infrastructure that sometimes you'd want to be able to pass options to the reader but we cannot, unless we'd stick options into the filename somehow. But that's for another PR...

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -336,90 +418,83 @@ read_file_itk(const std::string &filename)
...
 
 template<typename ITKImageType>
-void
-orient_ITK_image(typename ITKImageType::Pointer &itk_image,
-                 CartesianCoordinate3D<float> &voxel_size,
-                 BasicCoordinate<3,int> &min_indices,
-                 BasicCoordinate<3,int> &max_indices,
-                 CartesianCoordinate3D<float> &origin,
+typename ITKImageType::Pointer
+orient_ITK_image(shared_ptr<ExamInfo> exam_info_sptr,
```
KrisThielemans:
```suggestion
orient_ITK_image(const shared_ptr<ExamInfo> exam_info_sptr,
```

src/IO/ITKOutputFileFormat.cxx
```diff
@@ -99,7 +102,7 @@ actual_write_to_file(std::string& filename,
     {
       add_extension(filename, this->default_extension);
 
-      const VoxelsOnCartesianGrid<float> image_ptr =
+      const VoxelsOnCartesianGrid<float> image =
```
KrisThielemans:
```suggestion
      const VoxelsOnCartesianGrid<float>& image =
```
avoids creating a copy

src/IO/ITKOutputFileFormat.cxx
```diff
@@ -124,52 +124,86 @@ actual_write_to_file(std::string& filename,
...
+      //Creating the image
+      ImageType::Pointer itk_image = ImageType::New();
+
+      itk_image->SetSpacing(spacing);
+      itk_image->SetRegions( region );
+      itk_image->SetOrigin(origin);
+      itk_image->SetDirection( matrix );
+      itk_image->Allocate();
+
+      // std::cerr << "ITK Origin: "
```
KrisThielemans: let's delete all these

ashgillman:
> However, I'm a bit worried about the passing by reference of the ITK image. I have the impression that a number of functions now modify the ITK image that's first read. That's fine currently because we throw it away, but at a later stage, we want to move the convert_ITK_to_STIR and vice versa functions out, such that they can be used elsewhere, e.g. when calling an ITK filter from within STIR.

Yes it breaks the rule that a function should do one general thing also.

There's a bit of a compromise, to reorient, we need to have pulled out the patient position. And then we will need that again for creating the empty STIR image. Finally, the image needs to be reoriented also when filling in the STIR image data. So I guess we have to repeat something so that it isn't all in one function (not a big deal computation-wise, but allows scope for inconsistencies to creep in).

Maybe the `orient_itk_image()` get's called before (not inside) `construct_empty_stir_image_and_orient_itk_image()` (to be renamed to `construct_empty_stir_image()`), in `convert_ITK_to_STIR()`. Instead of `orient_itk_image()` receiving an `exam_info_sptr()`, it calls `construct_exam_info_from_metadata_dictionary()`. Then you have to rely on the fact that a user will always use these functions together so the metadata is consistent.

ashgillman:
Alright, I think comments have been addressed. I was able to merge the two `convert_ITK_to_STIR()` functions into one, and just have separate data copying functions.

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -51,38 +54,18 @@ START_NAMESPACE_STIR
...
+// typedef VoxelsOnCartesianGrid<CartesianCoordinate3D<float> > STIRImageMultiConcrete;
 
-static
-STIRImageSingle*
-convert_ITK_to_STIR(const ITKImageSingle::Pointer itk_image);
-
-static
-STIRImageMulti*
-convert_ITK_to_STIR(const ITKImageMulti::Pointer itk_image);
+using MetaDataStringType = itk::MetaDataObject< std::string >;
```
KrisThielemans: probably will need to use `typdef` as still not moved to C++-11

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -140,76 +123,263 @@ read_from_file(const std::string& filename) const
...
+template<typename ITKImageType, typename STIRImageType, typename STIRConcreteType>
+STIRImageType*
+convert_ITK_to_STIR(const typename ITKImageType::Pointer itk_image)
+{
+  // Construct extra metadata
+  const shared_ptr<ExamInfo> exam_info_sptr
+    = construct_exam_info_from_metadata_dictionary(itk_image->GetMetaDataDictionary());
+  // Reorient the ITK image to align with STIR axes
+  typename ITKImageType::Pointer reor_itk_image
+    = orient_ITK_image<ITKImageType>(itk_image, exam_info_sptr);
```
KrisThielemans:
```suggestion
    = orient_ITK_image(itk_image, exam_info_sptr);
```
should be fine

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -140,76 +123,263 @@ read_from_file(const std::string& filename) const
...
+    = stir_image.get_physical_coordinates_for_LPS_coordinates
+      (CartesianCoordinate3D<float>(static_cast<float>(itk_coord[2]),
+                                    static_cast<float>(itk_coord[1]),
+                                    static_cast<float>(itk_coord[0])));
+
+  // The following is not required for displacement field images, as
+  // the coordinates are relative.
+  if (!is_displacement_field)
+  {
+    // dummy image that has minumum to be able to find ITK -> STIR origin vector
```
KrisThielemans:
```suggestion
```
obsolete comment

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -140,76 +123,263 @@ read_from_file(const std::string& filename) const
...
+                           shared_ptr<ExamInfo> exam_info_sptr)
+{
+  // find voxel size
+  const CartesianCoordinate3D<float> voxel_size
+    (static_cast<float>(itk_image->GetSpacing()[2]),
+     static_cast<float>(itk_image->GetSpacing()[1]),
+     static_cast<float>(itk_image->GetSpacing()[0]));
+
+  // find info STIR image geometrical metadata
+  IndexRange<3> index_range = calc_stir_index_range<ITKImageType>(itk_image);
```
KrisThielemans:
```suggestion
  IndexRange<3> index_range = calc_stir_index_range(itk_image);
```

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -140,76 +123,263 @@ read_from_file(const std::string& filename) const
...
+  const CartesianCoordinate3D<float> stir_origin = calc_stir_origin<ITKImageType>
```
KrisThielemans:
```suggestion
  const CartesianCoordinate3D<float> stir_origin = calc_stir_origin
```

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -140,76 +123,263 @@ read_from_file(const std::string& filename) const
...
+}
+
+/* Copy a single ITK Image to a single STIR Image.
+   This method expects that itk_image is already oriented to be consistent with
+   STIR x, y, z axes.
+*/
+void copy_ITK_data_to_STIR_image(const ITKImageSingle::Pointer itk_image,
+                                 STIRImageSingle* stir_image_ptr)
+{
+  VoxelsOnCartesianGrid<float>::full_iterator stir_iter = stir_image_ptr->begin_all();
```
KrisThielemans:
```suggestion
  STIRImageSingle::full_iterator stir_iter = stir_image_ptr->begin_all();
```

KrisThielemans:
@rijobro, this PR is about to merged. I'm assuming that you have had a look and put some data through. Ok to go?

ashgillman:
> You could make a single version of copy_ITK_data_to_STIR_image that just has a template for the image-value type if you want, but I don't mind.

Are you sure? I couldn't quite work out how - for single it is just a cast (actually this probably isn't even necessary, should be float anyhow), but for multi we need to convert the ITK variable vector to a Point, then run the conversion routine. Would be nice though to have them in one.

ashgillman:
Removing the templates from the functions wouldn't build for me, perhaps the ITK stuff makes it too difficult for the compiler to auto-determine.

KrisThielemans:
> Removing the templates from the functions wouldn't build for me, perhaps the ITK stuff makes it too difficult for the compiler to auto-determine.

I see. you'd probably have to have templates in terms of the `Pointer`-type. anyway. no point in wasting time on this.

KrisThielemans:
Seems we have lost the capability to differentiate between displacement and deformation fields when reading Nifti.

ashgillman:
I was under the impression we only had displacement, @rijobro ?

KrisThielemans:
I think `copy_ITK_data_to_STIR_image` needs a `is_displacement_field` (or similar) field. Then I think we can unify it with something like this maybe.

* change
```c++
template<typename ITKImageType, typename STIRImageType>
CartesianCoordinate3D<float>
ITK_coordinates_to_STIR_physical_coordinates
(const typename ITKImageType::PointType &itk_coord,
const STIRImageType &stir_image, bool is_displacement_field = false)
```
to
```c++
template<typename ITKCoordType, typename STIRImageType>
CartesianCoordinate3D<float>
ITK_coordinates_to_STIR_physical_coordinates
(const ITKCoordType&itk_coord,
const STIRImageType &stir_image, bool is_displacement_field = false)
```
* introduce something like
```c++
template <class STIRValueT, class ITKValueT, class STIRImageT>
STIRValueT
convert_ITK_value_to_STIR_value(const ITKValueT& itk_value, const STIRImageT& stir_image, const bool is_disp)
```
which would be overloaded such that we can do
```c++
void copy_ITK_data_to_STIR_image(const ITKImage::Pointer itk_image,
                                 STIRImageType* stir_image_ptr)
{
  ...
for (it.GoToBegin(); !it.IsAtEnd(); ++it, ++stir_iter)
  {
    *stir_iter = convert_ITK_value_to_STIR_value(it.Get(), *stir_image_ptr, is_disp);
  }
}
```
Although really I wouldn't pass a `STIRImageType*` but a const ref.

* add specalisations. As I can't see why we still have to go to `ITKImageMulti::PointType` from the original variable length vectors, I think the specialisation can simply be
```c++
template <ITKCoordT>
CartesianCoordinate3D<float>
convert_ITK_value_to_STIR_value(const ITKCoordT& itk_value, const STIRImageMulti& stir_image, const bool is_disp)
{
    return ITK_coordinates_to_STIR_physical_coordinates<ITKCoordT, STIRImageMulti>
(itk_coord, stir_image, is_disp);
```

ashgillman:
Hmm, @rijobro what happened with the flip_axes and permute_axes stuff?

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -140,76 +120,261 @@ read_from_file(const std::string& filename) const
...
-    for ( it.GoToBegin(); !it.IsAtEnd(); ++it, ++stir_iter  ) {
-        *stir_iter = static_cast<float>(it.Get());
-    }
+  // find STIR origin
+  // Note: need to use - for z-coordinate because of different axis conventions
+  CartesianCoordinate3D<float> stir_coord
+    = stir_image.get_physical_coordinates_for_LPS_coordinates
+      (CartesianCoordinate3D<float>(static_cast<float>(itk_coord[2]),
+                                    static_cast<float>(itk_coord[1]),
+                                    static_cast<float>(itk_coord[0])));
```
rijobro: In the comment you mention a minus, but can't see one in the code. I suppose one needs to be updated to match the other?

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -140,76 +120,261 @@ read_from_file(const std::string& filename) const
...
+
+  // The following is not required for displacement field images, as
+  // the coordinates are relative.
+  if (!is_displacement_field)
+  {
+    CartesianCoordinate3D<float> stir_origin_index(0, 0, 0);
```
rijobro: I don't understand this logic. Shouldn't the origins of all images (3D, displacement or deformation) should all be altered?

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -140,76 +120,261 @@ read_from_file(const std::string& filename) const
...
+  default:
+    throw std::runtime_error("Unsupported patient position, not sure how to read.");
+  }
+
+  orienter->Update();
+  return orienter->GetOutput();
+}
+
+/* Convert an ITK image into an internal STIR one. */
+template<typename ITKImageType, typename STIRImageType, typename STIRConcreteType>
```
rijobro: Do we need to template this in terms of `STIRImageType`? Couldn't we just construct a `STIRConcreteType` and return it?

src/IO/ITKImageInputFileFormat.cxx
rijobro: Ah, sure, sorry.

src/IO/ITKImageInputFileFormat.cxx
rijobro: Sure, but why can't we do this in the 3D case:
```
return convert_ITK_to_STIR
        <ITKImageSingle, STIRImageSingleConcrete>
        (itk_image);
```
and this in the 4D case:
```
return convert_ITK_to_STIR<ITKImageMulti, STIRImageMulti>
        (itk_image);
```

rijobro: Just did a PR to this branch. Feel free to ignore if you think there's something wrong with it, but it passes the ctests

rijobro:
I think STIR only has displacement fields, but you might as well chuck in the `is_displacement_field = true` argument into `copy_ITK_data_to_STIR_image` to future-proof it. 

If ever we had to distinguish between deformation and displacement fields, I suppose it would be with an extra argument in `read_from_file`. Alternatively, we could figure it out from the header data (but I have no idea if all packages will update the codes correctly).

src/IO/ITKImageInputFileFormat.cxx
ashgillman: Oh right! Sorry, my brain wasn't working. Yes that makes sense - my solution is unnecessarily similar to the code I refactored, that is better.

ashgillman:
Is this a common thing about the deformation vs displacement fields? I must admit I have never heard of the deformation version before or seen it that way in any software, and I have actually only ever seen the two terms used interchangeably.

If there is some software that uses absolute location for the fields, then we should future proof. If there is no example, I think assume that all nonrigid motion fields give relative vectors.

KrisThielemans:
> I think STIR only has displacement fields

that isn't the only point. The question is what does niftireg have. I thought you told me it had both

rijobro:
NiftyReg has both displacement and deformation fields. But STIR can only handle displacements, so there currently wouldn't be any use in being able to read deformation fields since we can't use them.

KrisThielemans:
we could fix that of course, but for now, we should check and throw an error in the nifti reading code.

ashgillman:
Okay cool. That makes sense then and clears up a little on the intent of the `ITK_coordinates_to_STIR_physical_coordinates()` function.

Offtrack: I guess you wouldn't have to fully support deformation, but could just convert them.

ashgillman:
I think what I just committed is like what you asked for?

Shared copy, with a custom pixel conversion, specialised if a pixel is a number or a vector. I tried to have the first version templated, but it didn't quite work. I think two concrete implementations is adequate.

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -164,6 +159,36 @@ ITK_coordinates_to_STIR_physical_coordinates
   return stir_coord;
 }
 
+/* Convert an ITK Pixel (i.e., float) to a STIR Pixel. */
+typename STIRImageSingle::pixel_type
+ITK_pixel_to_STIR_pixel(typename ITKImageSingle::PixelType itk_pixel,
+                        const STIRImageSingle &stir_image,
+                        bool is_displacement_field=false)
```
KrisThielemans: to avoid a compiler warning, just say `bool` (as the argument isn't used)

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -278,40 +303,21 @@ construct_empty_stir_image(typename ITKImageType::Pointer itk_image,
    This method expects that itk_image is already oriented to be consistent with
    STIR x, y, z axes.
 */
-void copy_ITK_data_to_STIR_image(const ITKImageSingle::Pointer itk_image,
-                                 STIRImageSingle* stir_image_ptr)
+template<typename ITKImageType, typename STIRImageType>
+void copy_ITK_data_to_STIR_image(const typename ITKImageType::Pointer itk_image,
+                                 STIRImageType* stir_image_ptr,
```
KrisThielemans: why not a const ref as opposed to a pointer?

src/include/stir/DiscretisedDensity.h
```diff
@@ -112,6 +112,8 @@ template<int num_dimensions, typename elemT>
   typedef Array<num_dimensions,elemT> base_type;
   typedef DiscretisedDensity<num_dimensions,elemT> self_type;
 public:
+  typedef elemT pixel_type;
```
KrisThielemans: remove. use `full_value_type` which is already there (from `Array`)

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -164,6 +159,36 @@ ITK_coordinates_to_STIR_physical_coordinates
   return stir_coord;
 }
 
+/* Convert an ITK Pixel (i.e., float) to a STIR Pixel. */
+typename STIRImageSingle::pixel_type
```
KrisThielemans: are you sure you need the `typename`? Should normally only be in a template (also below of course).

Could `inline` this. should make it `static`

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -164,6 +159,36 @@ ITK_coordinates_to_STIR_physical_coordinates
...
+ITK_pixel_to_STIR_pixel(typename ITKImageSingle::PixelType itk_pixel,
+                        const STIRImageSingle &stir_image,
+                        bool is_displacement_field=false)
+{
+  return static_cast<typename STIRImageSingle::pixel_type>(itk_pixel);
```
KrisThielemans: same comment about `typename`

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -164,6 +159,36 @@ ITK_coordinates_to_STIR_physical_coordinates
...
+}
+
+/* Specialisation if the pixel is a vector and we want a multi-image */
+typename STIRImageMulti::pixel_type
+ITK_pixel_to_STIR_pixel(typename ITKImageMulti::PixelType itk_pixel,
+                        const STIRImageMulti &stir_image,
+                        bool is_displacement_field=false)
+{
+  // ITK VariableLengthVector to ITK FixedArray
+  // We know it is length 3
```
KrisThielemans: could add `assert`

src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -164,6 +159,36 @@ ITK_coordinates_to_STIR_physical_coordinates
...
+                        bool is_displacement_field=false)
+{
+  return static_cast<typename STIRImageSingle::pixel_type>(itk_pixel);
+}
+
+/* Specialisation if the pixel is a vector and we want a multi-image */
+typename STIRImageMulti::pixel_type
+ITK_pixel_to_STIR_pixel(typename ITKImageMulti::PixelType itk_pixel,
+                        const STIRImageMulti &stir_image,
+                        bool is_displacement_field=false)
```
KrisThielemans: no need to have default value here as it's an internal function

KrisThielemans:
@rijobro, could you add the check and throw an error in the nifti reading code if a deformation field?

KrisThielemans:
@ashgillman could you maybe add some stuff to release_3.1.htm (will be renamed) and STIRUsersGuide.tex (if appropriate) on the ITK. can be brief for now while we're sorting out the other things, but I think it's preferable that the change in the release notes is in the same PR as the code change.

rijobro:
> @rijobro, could you add the check and throw an error in the nifti reading code if a deformation field?

I don't think there is a generic way of distinguishing deformation and displacement nifti images. In `NiftyReg`, they use a field called `intent_p1`, a field that is actually intended for storing nifti codes for codes are for probability distributions. Outside of `NiftyReg`, I don't think we can differentiate deformation from displacement via the metadata. 

It looks like we have to rely on the user supplying the correct type.

KrisThielemans:
yet another limitation of nifti.

Is it possible to detect it's niftyreg output? I guess not.

let's then put a comment in the code and a warning in the user's guide that this is what we're assuming, (explaining terminology)

ashgillman:
Can niftyreg convert between the two? Could always just support the one, especially if the day format wasn't designed to handle it.

On 26 Oct. 2018 19:37, Kris Thielemans <notifications@github.com> wrote:

yet another limitation of nifti.

Is it possible to detect it's niftyreg output? I guess not.

let's then put a comment in the code and a warning in the user's guide that this is what we're assuming, (explaining terminology)

rijobro:
How about:
```      
warning("Only displacement fields are currently supported in STIR (not deformations). "
              "There is no way of verifying this from the nifti_image metadata, so you need to "
              "make sure that the image you are supplying is a displacement field image.");
```

> Can niftyreg convert between the two? Could always just support the one, especially if the day format wasn't designed to handle it.

Yes, `NiftyReg` can convert between them. But `NiftyReg` is an optional dependency of `SIRF`, not `STIR`.

KrisThielemans:
@rijobro, sounds good to me. People might get fed-up seeing the warning, but we have so many repetitive warnings in STIR already that one more isn't going to hurt :-;

rijobro:
@ashgillman, turns out I can make changes directly in your branch. So I did that to add the warning. I'll do it via PR in the future, but this one seemed like a tiny change.

ashgillman:
Alright, I was also able to reduce some of the explicit templating, however it seems many are still required when the output type is assigned to a pointer, as the output type can't be determined.

I guess this is getting close to be merged? Nothing more from me.

src/IO/ITKImageInputFileFormat.cxx
ashgillman: A lot of these are functions, not members. I was just following from what was previously done in this file. So does static make sense?

ashgillman: As for the typename thing, I've now changed it back to how I had originally planned. First function is templated, second is a specialisation. You still need to explicitly define the types when used so that the return type can be inferred.

src/IO/ITKImageInputFileFormat.cxx
ashgillman: It seems I still need this now that the second function is a specialisation so that I can define the default of false.

src/IO/ITKImageInputFileFormat.cxx
ashgillman: Done, but I don't actually have any multi images to test with. @rijobro could you possible run this over some data again sorry?

src/IO/ITKImageInputFileFormat.cxx
ashgillman: I made all the new functions static and inline, but I am curious what static does for non-member functions? I didn't think it would compile. A quick Google didn't show much.

src/include/stir/DiscretisedDensity.h
ashgillman: Thanks, missed that

src/IO/ITKImageInputFileFormat.cxx
ashgillman: With latest changes and the other version being a speicialisation, I think this is now needed again so that I can define the default.

src/IO/ITKImageInputFileFormat.cxx
ashgillman: I've made this a reference, but it can't be const as we update the contents.

src/IO/ITKImageInputFileFormat.cxx
ashgillman: Everywhere inline and static (not static on specialisations, compile error). What exactly does static do for non-member functions? I had a Google, but couldn't find any info.

src/IO/ITKImageInputFileFormat.cxx
ashgillman: Done, @rijobro could you possibly run this on a deformation field again to check? Sorry

src/IO/ITKImageInputFileFormat.cxx
ashgillman: Good point, ignore my previous comment Re just using bool.

src/IO/ITKImageInputFileFormat.cxx
KrisThielemans: > What exactly does static do for non-member functions?

It makes them local to the translation unit, i.e. no chance anyone can ever link to them (otherwise you can, even if there's no .h for them)

ashgillman: Good to know :)

documentation/release_3.1.htm
```diff
@@ -161,7 +162,8 @@ <h3>Changed functionality</h3>
 </li>
 <li>
 The orientation of images read/written via ITK has changed. It should now be correct if the patient
-was in HFS position. This is currently not checked (and for many file formats, never can be checked).
+was in HFS position for NIfTI (this can't be verified due to lack of metadata in the format), or if
```
KrisThielemans: I find this a bit awkward to read. I suggest 

> was in HFS position. Unfortunately, for most file formats (such as NifTI) this can't be verified due to lack of metadata in the format. If the files are in DICOM format, as long as the file contains the correct metadata, other patient positions are supported (currently HFS, FFS, HFP or FFP).
```

File: src/IO/ITKImageInputFileFormat.cxx [hunk 25]
```diff
@@ -140,80 +120,285 @@ read_from_file(const std::string& filename) const
...
+                           shared_ptr<ExamInfo> exam_info_sptr)
+{
+  // find voxel size
+  const CartesianCoordinate3D<float> voxel_size
+    (static_cast<float>(itk_image->GetSpacing()[2]),
+     static_cast<float>(itk_image->GetSpacing()[1]),
+     static_cast<float>(itk_image->GetSpacing()[0]));
+
+  // find info STIR image geometrical metadata
+  IndexRange<3> index_range = calc_stir_index_range(itk_image);
```
KrisThielemans:
```suggestion
  const IndexRange<3> index_range = calc_stir_index_range(itk_image);
```

documentation/release_3.1.htm
ashgillman: Yes was a bit muddled in my head as I tried to get the words out. Much better

ashgillman:
Sorry, should have added the skip ci tag.

rijobro:
Tested by performing non-rigid transformation of nifti image with nifti displacement field. Result was the same as before. Any other tests needed?

rijobro:
Did a small PR. By default, you were setting `is_displacement = false` for both image types. We want it to be true for `STIRImageMulti`.

KrisThielemans:
to me, setting the defaults doesn't make a lot of sense. do we need them? Anyway, whatever works really. I'm eager to press the "squash and merge" button!

rijobro:
I meant to say that the default wasn't being overridden, which means that the displacement image was being treated as if it was a deformation.

Looks good to me now!

ashgillman:
Ah I see, I hadn't gone high enough with the `is_displacement_field`s. Thanks.

I removed the default in the image conversion functions. I kept the default in `ITK_coordinates_to_STIR_physical_coordinates()` and renamed `is_displacement_field` in this funtion to `is_relative_coordinate`, to make it explicit the actual reason. So when converting `ITK_pixel_to_STIR_pixel()` for multi images, we need to know if it `is_displacement_field`, and if so, we convert `ITK_coordinates_to_STIR_physical_coordinates()` such that the vector `is_relative_coordinate`.

KrisThielemans:
done?

ashgillman:
Yep

PR #258, opened 2018-10-04
[FIX] tiny error in assert so only noticed in debug build

rijobro:
Wrong variable name in an assert means that compilation fails in debug mode.


ashgillman:
Ah thanks, I ran into this yesterday but hadn't had a chance to investigate.

PR #261, opened 2018-10-09
Postfilter parametric images

rijobro:
Postfilter parametric images (as well as dynamic and single images). Templated some functions to avoid repetition.


rijobro:
This good to be merged then?

PR #264, opened 2018-10-13
Check each frame of a dynamic image contains 1 time frame

rijobro:
- Also use 1-based index (since `i+1` was used throughout the loop).

### DIFF
--- a/src/IO/interfile.cxx
+++ b/src/IO/interfile.cxx
@@ -174,8 +174,15 @@ read_interfile_image(istream& input,
       (*image_ptr)[i] *= static_cast<float>(hdr.image_scaling_factors[0][i]);
 
   // Check number of time frames
-  if (image_ptr->get_exam_info().get_time_frame_definitions().get_num_frames() != 1)
-      warning(str(boost::format("Discretised density should contain 1 time frame, but this image contains %1%.") % image_ptr->get_exam_info().get_time_frame_definitions().get_num_frames()));
+  if (image_ptr->get_exam_info().get_time_frame_definitions().get_num_frames() > 1) {
+      warning(str(boost::format("Discretised density should contain 1 time frame, but this image contains %1%. "
+                                "Only the first will be kept, and the rest discarded.")
+                  % image_ptr->get_exam_info().get_time_frame_definitions().get_num_frames()));
+      image_ptr->get_exam_info_sptr()->time_frame_definitions.set_num_time_frames(1);
+  }
+  else if (image_ptr->get_exam_info().get_time_frame_definitions().get_num_frames() == 0)
+      warning("DiscretisedDensity does not contain any time frames. This might cause an error.");
+
 
   return image_ptr;
 }
--- a/src/IO/interfile.cxx
+++ b/src/IO/interfile.cxx
@@ -173,6 +173,10 @@ read_interfile_image(istream& input,
     if (hdr.image_scaling_factors[0][i]!= 1)
       (*image_ptr)[i] *= static_cast<float>(hdr.image_scaling_factors[0][i]);
 
+  // Check number of time frames
+  if (image_ptr->get_exam_info().get_time_frame_definitions().get_num_frames() != 1)
+      warning(str(boost::format("Discretised density should contain 1 time frame, but this image contains %1%.") % image_ptr->get_exam_info().get_time_frame_definitions().get_num_frames()));
+
   return image_ptr;
 }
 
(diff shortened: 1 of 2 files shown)


KrisThielemans:
This looks ok, but it seems to the problem in the wrong place. This check really belongs in `read_from_file<DiscretisedDensity<3,float> >`. I think it would be better to move it there somehow, possibly not easy to do it for all registered IO. If we do that, I think we could either call `error` or issue a warning there and say that we'll just keep the first (and adjust the time frame defs).

what do you think?

rijobro:
As you say, would be easy to implement for `DiscretisedDenstiy::read_from_file`, but the only way to get it to work with `read_from_file<DiscretisedDensity>` is to place some sort of check method thjat all registered IO methods would call. Bit of a pain.

I put this in because I had files that had 0 time frames, since the images were older than our timeframedefinition updates. So in those cases, we wouldn't just be able to take the first frame. Possibly: 
```
if num_time_frames == 0
   error
else if num_time_frames > 1
   take first frame
   warning
else
   take first frame
end
```

rijobro:
Thinking about it, we don't necessarily want to throw an error if `num_time_frames == 0`. If we did, we couldn't read any older files, any ITK, etc. 

That's why this check possibly belongs in a place where we <b>require</b> `num_time_frames == 1`.

KrisThielemans:
you're right, but it feels wrong :-;

I guess you're saying that if we read a file via ITK or whatever, the time frame defs says num_time_frames==0? It isn't obvious what we should do with such a case, but I guess this is as good as any choice.

Would it be easy to add the warning in the Interfile reader only? (It's the only supported file format where it could return more than 1 time-frame. It shouldn't happen in ECAT7 (99% sure), nor DICOM (100% sure). If we put that in, then the code in this PR would be fine.

src/include/stir/IO/MultiDynamicDiscretisedDensityInputFileFormat.h
```diff
@@ -89,14 +90,18 @@ public InputFileFormat<DynamicDiscretisedDensity>
...
     dyn_disc_den_ptr->get_exam_info_sptr()->time_frame_definitions.set_num_time_frames(header.get_num_data_sets());
 
-    for (int i=0; i<header.get_num_data_sets(); ++i) {   
-        shared_ptr<DiscretisedDensity<3,float> > t(DiscretisedDensity<3,float>::read_from_file(header.get_filename(i)));
+    for (int i=1; i<=header.get_num_data_sets(); ++i) {
+        shared_ptr<DiscretisedDensity<3,float> > t(DiscretisedDensity<3,float>::read_from_file(header.get_filename(i-1)));
+
+        // Check that there is time frame information
+        if (t->get_exam_info().get_time_frame_definitions().get_num_frames() != 1)
+            error(str(boost::format("The individual components of a dynamic image should contain 1 time frame defintion, but image %1% contains %2% time frames") % i % t->get_exam_info().get_time_frame_definitions().get_num_frames()));
```
KrisThielemans: spelling mistake "defintion", but in any case I'd drop this word.

src/include/stir/IO/MultiDynamicDiscretisedDensityInputFileFormat.h
```diff
@@ -89,14 +90,18 @@ public InputFileFormat<DynamicDiscretisedDensity>
     dyn_disc_den_ptr->set_num_densities(header.get_num_data_sets());
     dyn_disc_den_ptr->get_exam_info_sptr()->time_frame_definitions.set_num_time_frames(header.get_num_data_sets());
 
-    for (int i=0; i<header.get_num_data_sets(); ++i) {   
-        shared_ptr<DiscretisedDensity<3,float> > t(DiscretisedDensity<3,float>::read_from_file(header.get_filename(i)));
+    for (int i=1; i<=header.get_num_data_sets(); ++i) {
+        shared_ptr<DiscretisedDensity<3,float> > t(DiscretisedDensity<3,float>::read_from_file(header.get_filename(i-1)));
```
KrisThielemans: any reason we don't use `unique_ptr` here?

rijobro:
Made all the small changes. Also add the warning to `interfile.cxx` if `num_time_frames != 1`. Squashed into one commit.

src/IO/interfile.cxx
```diff
@@ -173,6 +173,10 @@ read_interfile_image(istream& input,
     if (hdr.image_scaling_factors[0][i]!= 1)
       (*image_ptr)[i] *= static_cast<float>(hdr.image_scaling_factors[0][i]);
 
+  // Check number of time frames
+  if (image_ptr->get_exam_info().get_time_frame_definitions().get_num_frames() != 1)
+      warning(str(boost::format("Discretised density should contain 1 time frame, but this image contains %1%.") % image_ptr->get_exam_info().get_time_frame_definitions().get_num_frames()));
```
KrisThielemans: need to say what will be done (i.e. read the first frame) and remove other time frames from the exam info

ashgillman:
Re: permissions, you may have the same permissions as me - write, but not to branches with special permissions (e.g., master).

PS: Please don't give me write permission to master - with great power...

PR #268, opened 2018-10-23
Extract image executables: name change

rijobro:
I got confused between these two, so best to clear up the names.
extract_dynamic_images -> extract_single_images_from_dynamic_image
get_dynamic_images_from_parametric_images -> extract_dynamic_images_from_parametric_images

### DIFF
--- a/src/modelling_utilities/CMakeLists.txt
+++ b/src/modelling_utilities/CMakeLists.txt
@@ -24,9 +24,9 @@ set(dir_EXE_SOURCES ${dir}_EXE_SOURCES)
 
 set(${dir_EXE_SOURCES}
 	apply_patlak_to_images
-	get_dynamic_images_from_parametric_images
+	extract_dynamic_images_from_parametric_images
 	make_parametric_image_from_components
-	extract_dynamic_images
+	extract_single_images_from_dynamic_images
 	mult_model_with_dyn_images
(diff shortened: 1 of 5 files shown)


KrisThielemans:
Not sure about renaming `get_dynamic_images_from_parametric_images` to `extract_*`. For me, "extracting" is a simple function that gets some stuff out of some other stuff. This function actually computes the dynamic images from the kinetic model.

we could rename it to something better, but if we do, we'd have to check the `test_modelling.sh` script I presume.

rijobro:
We can leave `get_dynamic_images_from_parametric_images` as it is, the main one was `extract_dynamic_images`, which implies that the output is dynamic images (obviously that's my fault since I created it...).

So leave `get_dynamic_images_from_parametric_images` as it is and change `extract_dynamic_images` to `extract_single_images_from_dynamic_image`. Is that ok with you?

KrisThielemans:
sure

rijobro:
done

KrisThielemans:
release notes and users guide? (commit with `[ci skip]` in first line)

KrisThielemans:
sorry. I notice this one. you used `extract_single_images_from_dynamic_images`. To me, `extract_single_images_from_dynamic_image` is more logical...

rijobro:
done.

PR #260, opened 2018-10-08
Refactoring for origin shift

ashgillman:
- [x] Get comments
- [ ] Test scatter
- [x] Rebase away the first commit - it was just for easier building and testing.

This PR Shouldn't actually change any functionality, just using the correct routines where they should be.

NB: I haven't had time to test changes to scatter.

Issue #244, opened 2018-09-20
Use vendor origin and use a bed-based frame of reference


ashgillman:
- [ ] Rebase @rijobro commit with places needing updates onto master and use as base for this PR.
- [ ] Scanner type to be aware of default vendor origin
- [ ] Implement BasicCoord ProjDataInfo->bed_to_gantry_frame_of_reference(BasicCoord point)
- [ ] Implement BasicCoord ProjDataInfo->gantry_to_bed_frame_of_reference(BasicCoord point)
- [ ] Modify PMRT to use ProjDataInfo and DiscretisedDensity for index-to-position conversions.
- [ ] Modify Interpolating Projectors to use ProjDataInfo and DiscretisedDensity for index-to-position conversions.

PR #236, opened 2018-09-12
[ENH] Add `make check` which will compile all tests then run them.

ashgillman:
Currently there is no way (I am aware of?) to compile all of the testing targets. This provides it, with `make check`.

PR #271, opened 2018-10-25
Add "additional origin"

ashgillman:
This is a somewhat cheating method for the change of origin, where the origin isn't actually changed, but I add an additional vendor origin.

Its not really a solution to #244 as you can't read back from a NIfTI, but allows for consistent writing. It is also useful to have as a target to eventually have the STIR origin the same as the vendor origin.

Looks like a lot of commits, but actually only the last two. The rest are #236 #257 #260

### DIFF
--- a/src/IO/InterfileHeader.cxx
+++ b/src/IO/InterfileHeader.cxx
@@ -441,6 +441,8 @@ InterfileImageHeader::InterfileImageHeader()
     
   add_key("first pixel offset (mm)",
 	   KeyArgument::DOUBLE, &first_pixel_offsets);
+  add_key("first pixel vendor offset (mm)",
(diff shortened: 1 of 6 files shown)


ashgillman:
@KrisThielemans I believe a solution more like this is appropriate.

It seems that the vendor origin moves, I have found in our data that it has changed a number of times. I am currently investigating what caused the change, but I guess certain kinds of maintenance - we have had detectors etc. replaced in the last year.

Anyway, this won't do as that means that the origin would be very specific to the scan, and every image loaded would need this information to go back to gantry coordinates.

This solution says lets keep the STIR origin the way it was, gantry based. But, lets define a new field for the image, the vendor origin, which allows vendor-reconstruction-consistent expression of geometry in LPS coordinates.

ashgillman:
NB: #276

PR #273, opened 2018-10-26
Convert image type

rijobro:
Convert any image file format to another supported by STIR. Give output filename, input filename and output file format parameter file.

### DIFF
--- a/src/utilities/CMakeLists.txt
+++ b/src/utilities/CMakeLists.txt
@@ -69,6 +69,7 @@ set(${dir_EXE_SOURCES}
 	find_ML_normfactors3D
 	find_ML_normfactors
 	find_ML_singles_from_delayed
+	convert_image_type
 )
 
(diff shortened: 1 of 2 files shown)


ashgillman:
Doesn't `stir_math` already do this?

KrisThielemans:
Yes, `stir_math` does it, and can also be used for projection data (to get to Interfile only at present).

Can you create an entry in the FAQ instead?

rijobro:
How about for dynamic and parametric images?

KrisThielemans:
`stir_math` uses standard STIR IO for those as well. Its [usage message ](https://github.com/UCL/STIR/blob/e2f0bcb74af6bbf067785e07f58a8430faf3c64d/src/utilities/stir_math.cxx#L264) claims it can do dynamics and parametrics... should work (he says optimistically)

PR #275, opened 2018-10-30
[ENH] Fix #250 - LmToProjData to propagate bed position metadata.

### LINKED ISSUE #250: Listmode readers should set bed position
`CListModeData` has a private member `proj_data_info_sptr` etc, so will know about bed position. Any listmode reader should therefore set the bed position. 

This is discussed [here](https://github.com/UCL/STIR/pull/245#issuecomment-425239353).

ashgillman:
Also [FIX] CListModeDataECAT8 should use the interfile parser to create proj_data
Fixes #250

### DIFF
--- a/src/listmode_buildblock/CListModeDataECAT8_32bit.cxx
+++ b/src/listmode_buildblock/CListModeDataECAT8_32bit.cxx
@@ -49,13 +49,8 @@ CListModeDataECAT8_32bit(const std::string& listmode_filename)
   if (this_scanner_sptr->get_type() == Scanner::Unknown_scanner)
     error(boost::format("Unknown value for originating_system keyword: '%s") % originating_system );
 
-    shared_ptr<ProjDataInfo> tmp( ProjDataInfo::construct_proj_data_info(this_scanner_sptr,
-                                                                         this->interfile_parser.get_axial_compression(),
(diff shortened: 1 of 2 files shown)


ashgillman:
A `ProjDataInfo` is already created by the parser, `InterfileRawDataHeaderSiemens` with the correct meta data.

https://github.com/UCL/STIR/blob/e27e91e07ea4d2eb249cd11e6d18d5f1e09f2f38/src/IO/InterfileHeaderSiemens.cxx#L291-L295

So I modify `CListModeDataECAT8_32bit` to use it instead of creating its own generic one. The parser uses the same arguments except for arc_correction - but I should guess that when parsing a listmode interfile that this value should always be false anyway?

It would be nice to check anyway that arc correction is off, but I wasn't sure how. `dynamic_cast` of the `ProjDataInfo`?

Anyway, then `LmToProjData::post_processing()` then propagates this metadata. Other relevant metadata could be included there.

This should also work for `LmToProjDataBootstrap`

https://github.com/UCL/STIR/blob/e27e91e07ea4d2eb249cd11e6d18d5f1e09f2f38/src/listmode_buildblock/LmToProjDataBootstrap.cxx#L100-L112

PR #277, opened 2018-11-02
Fix scaling issue in inverse fourier

### LINKED ISSUE #221: PSF reconstruction is returning reconstructed images which is scaled by large factor of 10^7.
@ashgillman Can you add the images for describing this issue?

### LINKED ISSUE #269: Separable Cartesian Metz filter
I noticed that my images were decreasing massively in intensity (order of 10^6) when I did Gaussian postfiltering with the Separable Cartesian Metz filter. 

I think the problem originates from PR #205. If I take the commit before (or revert this commit), then the results make sense again. I've tried this on single, dynamic and parametric images (interfile and nifti). 

@NikEfth, as far as I can see there is no `ctest` in place for postfiltering with this filter. How did you validate? It would be simple enough to put a test in, I suppose.

NikEfth:
close #269, 
close #221

### DIFF
--- a/src/include/stir/numerics/fourier.h
+++ b/src/include/stir/numerics/fourier.h
@@ -67,7 +67,7 @@ inline void inverse_fourier(T& c, const int sign=1)
   // disable warning about conversion
   #pragma warning(disable:4244)
 #endif
-  c /= c.size_all();
+  //c /= c.size_all();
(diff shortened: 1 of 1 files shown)


KrisThielemans:
This fix was incorrect. `inverse_fourier` has a scale to make it the real inverse.

PR #285, opened 2018-11-06
[FIX] ECAT8 Norm only checked one of three possible scanner names.

ashgillman:
originating systems accepted are "2008", "Simens mMR" and "mMR"

### DIFF
--- a/src/recon_buildblock/BinNormalisationFromECAT8.cxx
+++ b/src/recon_buildblock/BinNormalisationFromECAT8.cxx
@@ -282,10 +282,13 @@ MatrixFile* mptr = matrix_open(filename.c_str(),  MAT_READ_ONLY, Norm3d);
   s.erase( std::remove_if( s.begin(), s.end(), isspace ), s.end() );
   /*interfile_parser.*/data_file_name=s;
   
-  if (/*interfile_parser.*/originating_system == "2008")
+  if (/*interfile_parser.*/originating_system == "2008"
+      or originating_system == "SiemensmMR" // we stripped the space
+      or originating_system == "mMR")
     this->scanner_ptr.reset(new Scanner(Scanner::Siemens_mMR));
   else
-    error(boost::format("Unknown originating_system '%s', when parsing file '%s'") % /*interfile_parser.*/originating_system % filename );
+    error(boost::format("Incompatible originating_system '%s' for ECAT8, when parsing file '%s'")
+          % /*interfile_parser.*/originating_system % filename );
 
 	char directory_name[max_filename_length];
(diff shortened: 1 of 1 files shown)


src/recon_buildblock/BinNormalisationFromECAT8.cxx
```diff
@@ -282,10 +282,13 @@ MatrixFile* mptr = matrix_open(filename.c_str(),  MAT_READ_ONLY, Norm3d);
   s.erase( std::remove_if( s.begin(), s.end(), isspace ), s.end() );
   /*interfile_parser.*/data_file_name=s;
   
-  if (/*interfile_parser.*/originating_system == "2008")
+  if (/*interfile_parser.*/originating_system == "2008"
+      or originating_system == "SiemensmMR" // we stripped the space
```
KrisThielemans:
```suggestion
      || (originating_system == "SiemensmMR") // we stripped the space
```
C++11? fails on [Appveyor](https://ci.appveyor.com/project/KrisThielemans/stir/builds/20076321/job/kw1tosgov218mcbn)
I feel safer with `()`
similar elsewhere.

KrisThielemans: by the way, why stripped the space?

ashgillman: Haha, ah can you tell what language I've been using all day?

The stripped space was done a few lines prior, apparently to remove trailing newline (specifically \r).

ashgillman:
> Does this actually occur in Siemens files?

Hmm, no. So maybe somewhere was trying to load my norm sinogram as ECAT8 and not catching the error?

KrisThielemans:
ah. if you have a norm sinogram, then this will never work of course. you need the "from projdata" `BinNormalisation`.

This hierarchy doesn't (yet) discover what the type is. you have to specify the class by hand.

KrisThielemans:
If this doesn't occur in the Siemens file, I'm not sure if we should accept this.

ashgillman:
This makes sense - but then why did this change make my script progress (at least to the next bug)? Would SIRF somewhere first try ECAT8, then if it fails, try sinogram, for example when making a sensitivity model?

ashgillman:
I agree

ashgillman:
I'll have another explore

KrisThielemans:
> why did this change make my script progress (at least to the next bug)? Would SIRF somewhere first try ECAT8, then if it fails, try sinogram, for example when making a sensitivity model?

STIR no, SIRF probably...

PR #300, opened 2018-11-29
fix a bug in upsample_and_fit_single_scatter

pkhateri:
In recent updates of normalisation classes, normalisation shared-pointer need to be set_up before usage


KrisThielemans:
thanks!

Seems I've committed to your master branch. Sorry. Always best to create a PR from a separate branch

PR #304, opened 2018-12-06
Time Of Flight reconstruction 

NikEfth:
Introduces TOF reconstuction in STIR


KrisThielemans:
@NikEfth, did you run the recon_test_pack locally? Travis throws up a lot of problems in it. For instance core dumps in the ROOT tests, but also others. it's always difficult to know with Travis of course.

src/IO/interfile.cxx
```diff
@@ -1262,120 +1262,150 @@ write_basic_interfile_PDFS_header(const string& header_file_name,
...
        }
 
        output_header << "maximum ring difference per segment := ";
        {
-	 std::vector<int>::const_iterator seg = segment_sequence.begin();
-	 output_header << "{ " <<proj_data_info_ptr->get_max_ring_difference(*seg);
-	 for (seg++; seg != segment_sequence.end(); seg++)
-	   output_header << "," <<proj_data_info_ptr->get_max_ring_difference(*seg);
-	 output_header << "}\n";
+     std::vector<int>::const_iterator seg = segment_sequence.begin();
```
KrisThielemans: more white space

src/buildblock/ML_norm.cxx
```diff
@@ -768,25 +768,28 @@ void make_fan_data(FanProjData& fan_data,
...
-            int rb = 0, b = 0;
-            
-            proj_data_info_ptr->get_det_pair_for_bin(a, ra, b, rb, bin);
-            
-            fan_data(ra, a, rb, b) =
-	      fan_data(rb, b, ra, a) =
-              (*segment_ptr)[bin.axial_pos_num()][bin.view_num()][bin.tangential_pos_num()];
+	  for (bin.timing_pos_num() = proj_data.get_min_tof_pos_num(); bin.timing_pos_num() <= proj_data.get_max_tof_pos_num(); ++ bin.timing_pos_num())
+	  {
+		segment_ptr.reset(new SegmentBySinogram<float>(proj_data.get_segment_by_sinogram(bin.segment_num(),bin.timing_pos_num())));
```
KrisThielemans: this seems incorrect. fan-sums etc should sum over TOF bins. with current code. the fan data seems to be the one from the last TOF bin.

adaptations in ML_norm seems to need more thought. Can we just error out for now if its TOF?

NikEfth: Sure. Done.

src/IO/interfile.cxx
NikEfth: How would you like me to address this? Revert the file and reply the changes?

KrisThielemans:
current Travis failures
```
The command "./run_test_listmode_recon.sh" exited with 1.
The command "./run_test_zoom_image.sh" exited with 1.
${PYTHON} -m pytest .
collecting 0 items / 1 errors/home/travis/.travis/functions: line 109: 18344 Segmentation fault
```
The `run_test_zoom_image.sh` is entirely weird as it doesn't have anything to do with TOF, so maybe ignore that here.

KrisThielemans:
`run_test_listmode_recon.sh` and the Python test are currently failing

KrisThielemans:
@ALEXJAZZ008008 @NikEfth in my previous commit, I changed `ProjData::get_num_sinograms()` to include the num-TOF-bins. I did this because we call a `Sinogram` as 2D array. It also meant that `size_all()` could be implemented in terms of that, without having to know about TOF. However, this broke `create_array_for_proj_data` in `swig/stir.i` as you'll see.

It's easily fixable, but the question is what `get_num_sinograms` should return. I still think the name implies my change is the best, but feel free to disagree.

This did show that there are currently no Python (or of course MATLAB) tests for the TOF code. Maybe @ALEXJAZZ008008 can contribute some...

KrisThielemans:
oh, maybe the Python test does check it, as I see my comment above that it's failing... Also, listmode recon...

NikEfth:
I think ```get_num_sinograms``` should return the total number of 2D sinograms. That would include tof_poss and segments. Then individual functions will should be able to return the 3D or TOF positions. 

However, now this fees a bit counter-intuitive. Because by defaults the tof positions are the "outer" layer of our ```ProjData```. Not a big deal, though.

NikEfth:
```run_test_zoom_image``` fails, too

File: src/test/test_time_of_flight.cxx [hunk 3]
```diff
@@ -47,10 +47,10 @@ START_NAMESPACE_STIR
 //!
 class cache_index{
 public:
-    cache_index() {
+    cache_index():
```
KrisThielemans: this doesn't seem to change anything at all, but best to use constructor initialisers. I don't see where `cache_index` is used by the way

src/swig/stir.i
```diff
@@ -640,9 +640,7 @@ namespace std {
       //      {
       //        num_sinos += 2*proj_data.get_num_axial_poss(s);
       //      }
-      int num_sinos = proj_data.get_num_sinograms();
-
-      Array<4,float> array(IndexRange4D(proj_data.get_num_tof_poss(),num_sinos, proj_data.get_num_views(), proj_data.get_num_tangential_poss()));
+      Array<4,float> array(IndexRange4D(proj_data.get_num_tof_poss(),proj_data.get_num_segments(), proj_data.get_num_views(), proj_data.get_num_tangential_poss()));
```
KrisThielemans: incorrect. This would have to be
```c++
const int num_non_tof_sinos = proj_data.get_num_sinograms() / proj_data.get_num_tof_poss();
```
and then use that. I'll fix this.

src/test/test_time_of_flight.cxx
NikEfth: I did this for codacy.

KrisThielemans: great. but is this class actually used?(sorry, can't ee it)

KrisThielemans:
python still seg-faulting sadly.

ALEXJAZZ008008:
I thought I would just join in to say that I'm still seeing the same problems too.

I'll copy in what I've said elsewhere about the issue for posterity: "With the TOF branch, after it had the master branch merged into it, where previously after forward projecting a volume you would have a sinogram with dimensions (13, 47, 256, 256) you now have a sinogram with dimensions (13, 611, 256, 256) (this is with the same Python code generating the two results). 611 is 13 * 47 so I believe there must be a for loop somewhere that is iterating over both the TOF dimension and the rings (I think) and causing this dimension to be larger than it should be.

I'm surprised this error doesn't throw any of the CTests as I would have expected one of the CTests to check if the size of a sinogram was as expected given the template."

I'm still seeing the same issue mentioned above after pulling the TOF branch after Kris' most recent push.

NikEfth:
@ALEXJAZZ008008 ok, the sinogram is bigger. 
How about the information inside? 
Are the extra positions empty?

ALEXJAZZ008008:
@NikEfth The last time I viewed the sinograms visually they appeared to be doughnut shaped with values in an ellipse around a centre of NaN values (which I converted to 0.0 to view)

KrisThielemans:
@ALEXJAZZ008008 it cannot be that the size is still the same. if so, your rebuild didn't work. let us know

NikEfth:
@ALEXJAZZ008008  sounds fancy! 
Is the problem persistent ?

src/swig/stir.i
NikEfth: t.b.h. then I do not like this. 
We have to make sure that under any situations the ```get_num_for_poss()``` returns zero. 
Which currently is the default value for the creation of a ```ProjDataInfo```....

ALEXJAZZ008008:
@KrisThielemans @NikEfth I did not visualise the sinogram that I forward projected after having pulled the most recent TOF branch. Because it seems that it's not expected that I would still have the same size sinogram after the most recent commits I'll clean my build folder and rebuild from scratch. I can attach a visualisation of the sinogram that I forward project after I have done this.

KrisThielemans:
i cannot get Python to work at all on my Ubuntu 18.04 VM:
```
ImportError while importing test module '/home/sirfuser/devel/STIR/src/swig/test/python/test_buildblock.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test_buildblock.py:31: in <module>
    from stir import *
../../../../../STIR-install/python/stir.py:17: in <module>
    _stir = swig_import_helper()
../../../../../STIR-install/python/stir.py:16: in swig_import_helper
    return importlib.import_module('_stir')
/usr/lib/python2.7/importlib/__init__.py:37: in import_module
    __import__(name)
E   ImportError: /home/sirfuser/devel/STIR-install/python/_stir.so: undefined symbol: _ZN4stir21ForwardProjectorByBin15forward_projectERNS_3BinERKNS_18DiscretisedDensityILi3EfEE
```
I wonder therefore @ALEXJAZZ008008 if you aren't actually producing anything at all but just looking at previous output.

In any case, the problem is that somewhere in the merge ` void ForwardProjectorByBin::forward_project(Bin&, const DiscretisedDensity<3,float>&)` was dropped from the `.cxx`. Indeed, this has to be written with the "new style" of PR #362.

By the way, I'm a bit amazed that this function appears in this PR. It isn't TOF at all. And the function seems to be needed for listmode recon anyway (I guess - but didn't check - there it's using `ProjMatrixByBin`)

ALEXJAZZ008008:
@KrisThielemans I don't understand how this could be the case, I have entirely deleted my build and install folder. I've then built and installed using my branch of STIR which when I checked the commit history has the most recent commit from tof_sinos_UCL. I've deleted all of the results directories that I had on my laptop so there are no other .hs files on my PC, then I ran a forward project (using JRMoMo) using the version of STIR (through SIRF) that I had just built and I got the same results as I was describing above.
I'll try to attach the .hs and a .png where I've summed across the dimensions of size 13 and 144 of (13, 611, 144, 381)

[sinogram.zip](https://github.com/UCL/STIR/files/4162454/sinogram.zip)

sinogram

Regardless, if the function you're describing above is not present in this branch does it mean that it's not possible for this to forward project at all? (in which case I am very confused about what code I'm running exactly)

NikEfth:
@KrisThielemans 
At your last post, I lost your line of thoughts. Listmode reconstruction uses ```ProjMatrixElemsForOneBin::forward_project(Bin&, DiscretisedDensity<3,float>&)``` for the reconstruction. 

@ALEXJAZZ008008 
That sinogram is full with NaNs ... :/

NikEfth:
Using the executable, I did a fwd projection of a uniform cylinder to a small sinogram(13, 1, 288, 329) using DSTE as a template. 
I cannot spot any NaNs in the sinogram (screenshot across the 13 TOF positions):
Screenshot from 2020-02-06 01-01-38

P.S. 
I found 1-2 duplicate lines, probably from the merging but none of them could have causes NaNs. 
I will look again.

KrisThielemans:
It’s only via python

NikEfth:
@ALEXJAZZ008008 
how is this going? 
If it is still persistent could you check that, if you let Sirf pull and build its own STIR, if does it for the latest commit of this branch. Maybe it pulls a tag or something ...

ALEXJAZZ008008:
@NikEfth You want me to just build vanilla SIRF with the tags set as it would be in the super build if I didn't change them? I assume the version of STIR used by the super build will work otherwise there would be a lot of people who would have noticed this issue there by now.

NikEfth:
@ALEXJAZZ008008 No my question is, whether sirf pulls its own copy of STIR and if that copy is indeed up-to-date.

robbietuk:
Hi all, I thought I would quickly weight in with some results from some testing I did with @ALEXJAZZ008008, I hope it is useful. 

I installed both this branch of the TOF STIR and Alex's modified version of STIR with python support and ran 
`../install/bin/forward_project test my_atten_image.hv tof.hs`
where tof.hs was given to me by @ALEXJAZZ008008,

> !INTERFILE  :=
!imaging modality := PT
name of data file := /dev/zero
originating system := Unknown
!version of keys := STIR3.0
!GENERAL DATA :=
!GENERAL IMAGE DATA :=
!type of data := PET
imagedata byte order := LITTLEENDIAN
!PET STUDY (General) :=
!PET data type := Emission
applied corrections := {None}
!number format := float
!number of bytes per pixel := 4
number of dimensions := 5
matrix axis label [5] := timing positions
!matrix size [5] := 13
matrix axis label [4] := segment
!matrix size [4] := 1
matrix axis label [3] := view
!matrix size [3] := 144
matrix axis label [2] := axial coordinate
!matrix size [2] := { 47 }
matrix axis label [1] := tangential coordinate
!matrix size [1] := 381
%TOF mashing factor := 1
minimum ring difference per segment := { -1 }
maximum ring difference per segment := { 1 }
Scanner parameters:= 
Scanner type := GE Discovery 690
Number of rings                          := 24
Number of detectors per ring             := 576
Inner ring diameter (cm)                 := 81.02
Average depth of interaction (cm)        := 0.94
Distance between rings (cm)              := 0.654
Default bin size (cm)                    := 0.21306
View offset (degrees)                    := -5.021
Maximum number of non-arc-corrected bins := 381
Default number of arc-corrected bins     := 331
Energy resolution := 0
Reference energy (in keV) := 511
Number of TOF time bins := 13
Size of timing bin (ps) := 376.5
Timing resolution (ps) := 375
Number of blocks per bucket in transaxial direction         := 2
Number of blocks per bucket in axial direction              := 4
Number of crystals per block in axial direction             := 6
Number of crystals per block in transaxial direction        := 9
Number of detector layers                                   := 1
Number of crystals per singles unit in axial direction      := 1
Number of crystals per singles unit in transaxial direction := 1
end scanner parameters:=
effective central bin size (cm) := 0.226074
number of time frames := 1
start vertical bed position (mm) := 0
start horizontal bed position (mm) := 0
!END OF INTERFILE :=

This successfully runs and produces a 134MB sinogram. However this is as far as I got because I cannot open the sinogram using python->STIR as 
`stir.ProjData_read_from_file("test.hs")`
errors with 

> 
> WARNING: KeyParser warning: unrecognized keyword: %tof mashing factor
> WARNING: Interfile error: expecting 4D structure 
> WARNING: Interfile error: per-segment information is inconsistent: min_ring_difference
> WARNING: Interfile parsing of PET projection data failed
> ERROR: 
> ProjData::read_from_file could not read projection data test.hs.
> Unsupported file format? Aborting.
> Traceback (most recent call last):
>   File "view_tof_sino.py", line 11, in <module>
>     sino = stir.ProjData_read_from_file("test.hs")
> RuntimeError: 
> ProjData::read_from_file could not read projection data test.hs.
> Unsupported file format? Aborting.

I did give this file to @ALEXJAZZ008008 who I believe opened using SIRF and it had the dimentions of (13, 611, 144, 381).

KrisThielemans:
we realised that there are multiple issues here, causing confusing:
- Travis seg-faults on STIR python test
- @kristhielemans can't get the STIR python module to import as above
- @ALEXJAZZ008008 interfaces to STIR via SIRF and has NaN problems when using attenuation in the SIRF acq_model

KrisThielemans:
I've pushed some (innocent) changes that make the first 2 problems disappear on my VM. Let's see what Travis says. It won't fix the SIRF problem though. It's possible that we don't have enough STIR TOF test currently. Or alternatively, the SIRF TOF PR needs some changes.

src/swig/stir.i
KrisThielemans: I've done this slightly differently (by introducing `get_num_non_tof_sinograms`), but `get_num_tof_poss()` should never be zero of course. I guess we'd need a test for this then.

I don't understand what you mean with "The mashing factor is zero" (but possibly it's no longer relevant)

KrisThielemans:
All green now, with a few extra tests, but probably not enough of them yet.

@ALEXJAZZ008008 , @robbietuk can you provide an update?

robbietuk:
Ran `list_projdata_info` and `extract_segments` on the data produced by the this TOF branch of STIR. From what @ALEXJAZZ008008 and I could tell, it seemed correct. 

I would like to test with @ALEXJAZZ008008 branch of STIR but myriad is down for a few days. If that branch checks out okay, I think that indicates it might be a TOF SIRF problem?

KrisThielemans:
This is now up-to-date with `release_4`. I plan to do that again after releasing 4.1 (i.e. after merging #601), and then merge with `master`. Until that happens, the diff is likely horrible, and there are many other conflicts.

src/include/stir/recon_buildblock/ProjMatrixByBin.inl
```diff
@@ -104,11 +119,85 @@ get_proj_matrix_elems_for_one_bin(
...
+    CartesianCoordinate3D<float> voxel_center;
+    float new_value = 0.f;
+    float low_dist = 0.f;
+    float high_dist = 0.f;
+
+    // The direction can be from 1 -> 2 depending on the bin sign.
+    const CartesianCoordinate3D<float> middle = (point1 + point2)*0.5f;
+    const CartesianCoordinate3D<float> diff = point2 - middle;
+
+    const float lor_length = 1.f / (std::sqrt(diff.x() * diff.x() +
```
NikEfth, line 160: @KrisThielemans  The length should not change with the detector positions, but stay constant, depending on the size of the coincidence window.

NikEfth: The 1/length is to avoid divisions in the for loop. Maybe I should add a comment about it. 

const float lor_length = 1/tof_delta_time_to_mm(-coinc_window/2).

NikEfth (commented):
Modification in the LOR length calculation.

NikEfth (commented):
LOR length calculation

src/include/stir/recon_buildblock/ProjMatrixByBin.inl
KrisThielemans: I can agree with you that your formula gives (1 over) the lor length. However, for me, the name `lor_length` is misleading. The calculation below is
```
d2 = -inner_product(x, diff) * lor_length;
```
which with current formula becomes
```
d2 = -inner_product(x, diff) / norm(diff);
```
or equivalently
```
unit_vector_along_lor = diff/norm(diff);
d2 = -inner_product(x, unit_vector_along_lor );
```
This seems correct to me: `d2` is (up to the sign) the distance between the projection and the middle of the LOR.
```

File: src/include/stir/recon_buildblock/ProjMatrixByBin.inl [hunk 6]
```diff
@@ -104,11 +119,85 @@ get_proj_matrix_elems_for_one_bin(
...
+    for (ProjMatrixElemsForOneBin::iterator element_ptr = probabilities.begin();
+         element_ptr != probabilities.end(); ++element_ptr)
+    {
+        Coordinate3D<int> c(element_ptr->get_coords());
+        symm_ptr->transform_image_coordinates(c);
+
+        voxel_center =
+                image_info_sptr->get_physical_coordinates_for_indices (c);
+
+        project_point_on_a_line(point1, point2, voxel_center);
```
KrisThielemans, line 173: From geometry, I think that this step is not needed. The `inner_product` below will be equal before or after projection (`x1.x2 = norm(x1)norm(x2)cos(theta)`, so we get the `cos(theta)` factor without having to project first).

src/test/test_time_of_flight.cxx
```diff
@@ -0,0 +1,500 @@
...
+                       static_cast<double>(test_proj_data_info_sptr->get_k(bin)), "Error in get_sampling_in_k()");
+    }
+
+    float total_width = test_proj_data_info_sptr->get_k(Bin(0,0,0,0,test_proj_data_info_sptr->get_max_tof_pos_num(),1.f))
+            - test_proj_data_info_sptr->get_k(Bin(0,0,0,0,test_proj_data_info_sptr->get_min_tof_pos_num(),1.f))
+            + test_proj_data_info_sptr->get_sampling_in_k(Bin(0,0,0,0,0,1.f));
+
+    set_tolerance(static_cast<double>(0.005));
+    check_if_equal(static_cast<double>(total_width), static_cast<double>(test_proj_data_info_sptr->get_coincidence_window_width()),
+                   "Coincidence widths don't match.");
```
KrisThielemans, line 225: This puzzles me as `get_coincidence_window_width` is defined in terms of `get_max*` only
https://github.com/UCL/STIR/blob/d88e98cba4e749eb763c61acd2b3cf3c218d687f/src/include/stir/ProjDataInfo.inl#L178-L179
There seems to be a factor 2 wrong in this test (or in `get_coincidence_window_width`)

KrisThielemans: ah no. I misread that. Ignore the previous comment. (I'll delete it)

KrisThielemans:
@ALEXJAZZ008008 the last commit https://github.com/UCL/STIR/pull/304/commits/0f0cccb30c2abe56f6fb2cb2ff6dbf4ade4cb1d4 could influence you. In particular, it affects https://github.com/SyneRBI/SIRF/pull/543 which was going to be wrong. I guess you must have handled this in your local branch of SIRF

NikEfth:
> unfortunate to bury this change in a "cleaning" commit as it seems like a bug fix. oh well.

I am not sure how or why that was there. I think is because it made the initialisation of the LM reconstruction simpler. 
I need to double-check, but I think that there is a confusion with the mashing factors when you run either LM or ProjData. 

But I can not remember now.

KrisThielemans:
sorry. could you create a PR for the change on `overlap_interpolate` on `master` as well? This PR is already very large. I'd like to avoid having non-TOF stuff in the diff as much as we can. (I know it's tiny)

KrisThielemans:
sorry. could you please put the change on `overlap_interpolate` on `master`? This PR is already very large. I'd like to avoid haveing non-TOF stuff in here.

NikEfth:
Then, the branch will start failing.

src/listmode_buildblock/CListRecordROOT.cxx
```diff
@@ -50,6 +51,8 @@ void CListEventROOT::get_detection_position(DetectionPositionPair<>& _det_pos) c
 
     _det_pos.pos1() = det1;
     _det_pos.pos2() = det2;
+    _det_pos.timing_pos() = this->get_uncompressed_proj_data_info_sptr()->get_tof_bin(delta_time);
```
NikEfth, line 54: This is a highly dangerous line. 
I remembered why I had the unmashed version in the previous commit. 
If you set mashing factor in the hroot file, then the get_uncompressed_proj_data_info_sptr() is mashed. 
That is alright if you want to do LM reconstruction with mashing factor.

However, if you do lm_to_projdata you want always you LM data unmashed. 
The line bellow ensures that your LM data will always come unmashed, even if you have forgotten a mashing factor. 

However, if your LM data from a scanner come naturally mashed, then you need to either: (a) have a scanner template that is mashed or (b) use the line above. Otherwise, both LM reconstruction and lm_to_projdata might fail. 

For GATE the best is to always return the unmashed and let the template in the LmToProjdata decide the mashing.

KrisThielemans: I'd have to check this, but the name `get_uncompressed_proj_data_info_sptr()` implies that no mashing should have happened. If it did for `ROOT`, it might be a bug there?

NikEfth (commented):
It is not a bug really. It is a conflict between the TOF mashing applied in this point, which is needed when listmode data are TOF compressed (in comparison to the scanner template). And the TOF mashing applied at ```ProjDataInfoCylindricalNoArcCorr::get_bin_for_det_pos_pair(Bin& bin, const DetectionPositionPair<>& dp)```. These two applications conflict in some scenarios. E.g. The Scanner template doesn't have the TOF mashing factor that has been applied in the listmode data, but the user applies that in the hroot file, which would be alright for listmode reconstruction. However if with the same setup make a sinogram with a template that contains TOF mashing then the final sinogram will be double compressed.

src/listmode_buildblock/CListRecordROOT.cxx
NikEfth: It is not a bug really. It is a conflict between the TOF mashing applied in this point, which is needed when listmode data are TOF compressed (in comparison to the scanner template). And the TOF mashing applied at ```ProjDataInfoCylindricalNoArcCorr::get_bin_for_det_pos_pair(Bin& bin, const DetectionPositionPair<>& dp)```. These two applications conflict in some scenarios. E.g. The Scanner template doesn't have the TOF mashing factor that has been applied in the listmode data, but the user applies that in the hroot file, which would be alright for listmode reconstruction. However if with the same setup make a sinogram with a template that contains TOF mashing then the final sinogram will be double compressed.

src/include/stir/listmode/CListRecordGEHDF5.h
```diff
@@ -100,11 +100,24 @@ namespace RDF_HDF5 {
...
       }
+ inline void get_detection_position(DetectionPositionPair<>& det_pos) const
+  {
+    // TODO 447->get_num_detectors_per_ring()-1
+    det_pos.pos1().tangential_coord() = 447 - loXtalTransAxID;
+    det_pos.pos1().axial_coord() = loXtalAxialID;
+    det_pos.pos2().tangential_coord() = 447 - hiXtalTransAxID;
+//    std::cout << hiXtalTransAxID << " "  << loXtalTransAxID << std::endl;
+    det_pos.pos2().axial_coord() = hiXtalAxialID;
+  }
```
KrisThielemans, line 111: should not be here (it's in the cxx)

KrisThielemans (changes requested):
2 ones related to GE data that I happened to spot

recon_test_pack/run_test_simulate_and_recon_with_motion.sh
```diff
@@ -124,7 +124,7 @@ fi
 echo "===  create template sinogram (DSTE in 3D with max ring diff 2 to save time)"
 template_sino=my_DSTE_3D_rd2_template.hs
 cat > my_input.txt <<EOF
-Discovery STE
+Discovery STE nonTOF
```
KrisThielemans, line 127: there is no TOF STE, so this makes no sense

ashgillman:
I'm merging in master again now - getting ready for the Whitespace fixes following merge of #724. I might need some help with some conflicts

ashgillman:
https://github.com/ashgillman/STIR/commit/a897f197367b0c8bb49ebd7a69fc84b9b7e9ddc0#diff-95d5f59843869aafcb0d6b56dd64f71733d07f3bc61acae6fbb7a0eab3d5d62dR94-R96

vs

https://github.com/ashgillman/STIR/commit/ee9082e5df907aec7aff21ae9b0e0b010a9e48bd#diff-95d5f59843869aafcb0d6b56dd64f71733d07f3bc61acae6fbb7a0eab3d5d62dL93-R96

```c++
template <class coordT>
LORInAxialAndSinogramCoordinates<coordT>::
LORInAxialAndSinogramCoordinates(const coordT z1,
				 const coordT z2,
				 const coordT phi,
				 const coordT s,
				 const coordT radius,
	             const bool swapped)
  :
  LORCylindricalCoordinates_z_and_radius<coordT>(z1, z2, radius),
<<<<<<< HEAD
   _phi(phi), _s(s), _swapped(swapped)
=======
  _phi(to_0_2pi(phi)), _s(s)
>>>>>>> master
{
  if (_phi>=_PI)
    {
      _phi -= coordT(_PI); _s = -_s;
      std::swap(private_base_type::_z1,private_base_type::_z2);
    }
  check_state();
}
```

I assume to:

```c++
template <class coordT>
LORInAxialAndSinogramCoordinates<coordT>::
LORInAxialAndSinogramCoordinates(const coordT z1,
				 const coordT z2,
				 const coordT phi,
				 const coordT s,
				 const coordT radius,
	             const bool swapped)
  :
  LORCylindricalCoordinates_z_and_radius<coordT>(z1, z2, radius),
   _phi(to_0_2pi(phi)), _s(s), _swapped(swapped)
{
  if (_phi>=_PI)
    {
      _phi -= coordT(_PI); _s = -_s;
      std::swap(private_base_type::_z1,private_base_type::_z2);
    }
  check_state();
}
```

ashgillman:
Actually that is all, I think I resolved the rest - I can run tests once I have some help with that one

KrisThielemans:
that looks correct to me (i.e. keep the TOF version, but just add to 0_to_2pi for phi).

ashgillman:
Merged master, one test failing: https://github.com/NikEfth/STIR/pull/18

KrisThielemans:
Looks like we currently ignore Siemens TOF keywords. Some relevant lines
https://github.com/NikEfth/STIR/blob/31c399089876c44538c58b0b691f05edd210d654/src/IO/InterfileHeaderSiemens.cxx#L189 
https://github.com/NikEfth/STIR/blob/31c399089876c44538c58b0b691f05edd210d654/src/IO/InterfileHeaderSiemens.cxx#L323-L324
https://github.com/NikEfth/STIR/blob/31c399089876c44538c58b0b691f05edd210d654/src/IO/InterfileHeaderSiemens.cxx#L423
https://github.com/NikEfth/STIR/blob/31c399089876c44538c58b0b691f05edd210d654/src/IO/InterfileHeaderSiemens.cxx#L512-L513

We do seem to ignore [`num_tof_bins` in general](https://github.com/NikEfth/STIR/blob/31c399089876c44538c58b0b691f05edd210d654/src/IO/InterfileHeader.cxx#L1415-L1419), but presumably this is because it is determined by the scanner and the `tof_mashing_factor`. I guess it would be nice to put a check on consistency somewhere.

KrisThielemans:
`run_test_zoom_image.sh` fails because output is now
```
$ list_projdata_info --sum my_prompts.hs

TOF bin: 0
Data sum: 1.70611e+07
```
We should remove the `TOF bin` line.

KrisThielemans:
All tests work on my local Ubuntu (gcc8) now. However, I cannot get ROOT to work at the moment, so I cannot run `run_test_time_of_flight.sh`

src/include/stir/ProjDataInfoCylindricalNoArcCorr.inl
```diff
@@ -175,12 +194,13 @@ get_det_pos_pair_for_bin(
   dp.pos1().axial_coord()=a1;
   dp.pos2().tangential_coord()=t2;
   dp.pos2().axial_coord()=a2;
+  dp.timing_pos() = std::abs(bin.timing_pos_num())*this->get_tof_mash_factor();
```
ashgillman, line 197: https://github.com/ashgillman/STIR/commit/dff2c9fc1923f1839ad879478eb7fcca5283af07#r59620526

> if `timing_pos_num` was negative, shouldn't it have change the rest of the bin then?

I'm not sure why its necessary, but I think by definition this should already be positive in this convention?

KrisThielemans: confusingly, at present `bin.timing_pos_num` can be negative but `detection_pos_posi.timing_pos` not.

src/include/stir/recon_buildblock/ProjMatrixByBin.inl
KrisThielemans: we could try
```
const float d2_alt = -inner_product(image_info_sptr->get_physical_coordinates_for_indices (c) - middle, diff) * lor_length;
```

src/recon_buildblock/PostsmoothingBackProjectorByBin.cxx
```diff
@@ -65,6 +65,36 @@ PostsmoothingBackProjectorByBin::
   set_defaults();
 }
 
+void PostsmoothingBackProjectorByBin::
+update_filtered_density_image(DiscretisedDensity<3, float> &density)
+{
+//        image_processor_ptr->apply(*filtered_density_sptr);
+//        density += *filtered_density_sptr;
+//        filtered_density_sptr->fill(0.f);
+}
```
KrisThielemans, line 74: what is this? It doesn't do anything

src/recon_buildblock/PostsmoothingBackProjectorByBin.cxx
```diff
@@ -65,6 +65,36 @@ PostsmoothingBackProjectorByBin::
...
+PostsmoothingBackProjectorByBin*
+PostsmoothingBackProjectorByBin::clone() const
+{
+    PostsmoothingBackProjectorByBin* sptr(new PostsmoothingBackProjectorByBin(*this));
+//    sptr->original_back_projector_ptr.reset(this->original_back_projector_ptr->clone());
+    return sptr;
+}
+
+void PostsmoothingBackProjectorByBin::
+init_filtered_density_image(DiscretisedDensity<3, float> &density)
```
KrisThielemans, line 91: where is this used?

src/recon_buildblock/ProjMatrixByBinSPECTUB.cxx
```diff
@@ -672,10 +672,16 @@ set_up(
...
+ProjMatrixByBinSPECTUB::clone() const
+{
+	return new ProjMatrixByBinSPECTUB(*this);
+}
+
 ProjMatrixByBinSPECTUB::
 ~ProjMatrixByBinSPECTUB()
 {
-  delete_UB_SPECT_arrays();
+ // delete_UB_SPECT_arrays();
```
KrisThielemans, line 684: seems to have an enormous leak then. why is this commented out?

src/include/stir/recon_buildblock/BackProjectorByBinUsingProjMatrixByBin.h
```diff
@@ -78,17 +78,25 @@ class BackProjectorByBinUsingProjMatrixByBin:
 
   shared_ptr<ProjMatrixByBin> &
     get_proj_matrix_sptr(){ return proj_matrix_ptr ;} 
-  
+
+  void enable_tof(ProjMatrixElemsForOneBin* );
```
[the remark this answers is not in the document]
robbietuk: The complementary method does not exist in the `ForwardProjectorByBinUsingProjMatrixByBin.h` class.

src/include/stir/ProjDataInfoCylindricalNoArcCorr.inl
KrisThielemans: See https://github.com/NikEfth/STIR/issues/26

KrisThielemans:
Possibly simplification of loops etc (as suggested in conversations with @robbietuk)
- rename class `ViewSegmentNumbers` to `ViewgramIndices` 
- add `const std::vector<ViewgramIndices>& ProjDataInfo::get_all_viewgram_indices() const`
- write loops as
  ```c++
  auto viewgram_indices = proj_data.get_all_viewgram_indices();
  #pragma omp parallel for
  for (unsigned i=0; i<viewgram_indices.size(); ++i)
  {
     // for loops that only go over basic viewgrams
     if (!symmetries_sptr->is_basic(viewgram_indices[i]))
       continue;
     auto viewgram = proj_data.get_viewgram(viewgram_indices[i] );
     ...
  }
  ```
  Note that you'd need very recent OpenMP to be able to do
  ```
  for (auto viewgram_index : viewgram_indices )
  ```
  so we cannot use this yet (VS only supports OpenMP 2.x!). In fact, we might not to have a loop over `int` (I will need to check)

NikEfth:
Hi all, 
Have you tested this branch with root files lately? 
It does not work for me. Are you aware of any problems?

KrisThielemans:
I haven't in a while. It's also a bit out-of-date w.r.t. ROOT, so it's probably time to merge `master` in here again.

What kind of problems are you having?

NikEfth:
The images do not make any sense with TOF, while without (same data) are ok. 
Screenshot 2023-05-06 at 16 44 03

KrisThielemans:
ah. Did you run the ROOT tests? They were designed to catch any problems.

NikEfth:
I was not able to run the consistency test. I don't understand where the test exe is. 
the others checked out alright.

KrisThielemans:
Looks like it should be run automatically when doing `ctest`. It's `recon_test/test_consistency_with_GATE.cxx` and the `CMakeLists.txt` auto-downloads the data.

robbietuk:
> Looks like it should be run automatically when doing `ctest`. It's `recon_test/test_consistency_with_GATE.cxx` and the `CMakeLists.txt` auto-downloads the data.

Pulling the branch and running the `test_consistency_with_GATE` passed for me. @NikEfth, are you reconstructing a point source here?

NikEfth:
@robbietuk No it is a large Derenzo phantom.

robbietuk:
The testing in https://github.com/NikEfth/STIR/pull/30 was pretty extensive to verify max intensity of the LOR kernal overlapped the source, including visualization. 

Are there any `recon_test_pack` tests that verify TOF consistency with previous commits? I think they are currently disabled:
https://github.com/NikEfth/STIR/blob/370d7feb796d6906186724130ec7054ea4b756b9/recon_test_pack/simulate_PET_data_for_tests.sh#L28-L30

KrisThielemans:
There are no tests for consistency with previous commits. `simulate_PET_data_for_tests.sh` would be used to test consistency between simulation and reconstruction, which is another thing.

I'm not aware of any changes, even affecting processing of ROOT data. (We added the consistency tests, but they worked). @NikEfth you can find the GATE macros etc in `examples/ROOT_files/ROOT_STIR_consistency/`

KrisThielemans:
> I'm not aware of any changes, even affecting processing of ROOT data.

well, this depends on how far back you're going I guess.

We did change the detector offset stuff. Looks for `STIR_ROOT_ROTATION_AS_V4`. However, this should only rotate things.

I still have no idea what's going wrong therefore.

KrisThielemans:
@NikEfth you could possibly try the branch from https://github.com/NikEfth/STIR/pull/44. This has merged the listmode caching in (and many many other things). Of course the listmode caching doesn't work with TOF yet, but the non-TOF version works as it should on that branch.

KrisThielemans:
@danieldeidda please do NOT push to this branch in the future. Always work via a PR to this branch. I guess it should be protected really, as it's supposed to be a stable branch that works.

What is https://github.com/UCL/STIR/pull/304/commits/51642dc45b59376bf06a912e85e191dc08de6487? It doesn't look like it's a merge anyway, so this will create lots of havoc.

Also, in general ideally do not push things that do not compile. It's launching about 14 unnecessary jobs that will then all fail.

If @NikEfth agrees, I'd rather remove those last 3 commits from this branch. That will mean a force-push though, which isn't great.

KrisThielemans:
ok. Looks like it was indeed a merge (phew!). A pity of the 2 other commits, but I'm going to leave this.
 Seems that the Actions are not running because `master` has again moved on and we have a new conflict. oh well.

danieldeidda:
so shall I resolve this conflict?

KrisThielemans:
I need to do some other stuff on `master` first. Might as well do that before doing yet another merge.

danieldeidda:
sounds good, I will get going with my local version then since it seems everything is running so far

NikEfth:
> @danieldeidda please do NOT push to this branch in the future. Always work via a PR to this branch. I guess it should be protected really, as it's supposed to be a stable branch that works.
> 
> 
> 
> What is https://github.com/UCL/STIR/pull/304/commits/51642dc45b59376bf06a912e85e191dc08de6487? It doesn't look like it's a merge anyway, so this will create lots of havoc.
> 
> 
> 
> Also, in general ideally do not push things that do not compile. It's launching about 14 unnecessary jobs that will then all fail.
> 
> 
> 
> If @NikEfth agrees, I'd rather remove those last 3 commits from this branch. That will mean a force-push though, which isn't great.
> 
> 

I just saw this. Fee free to do what is needed

danieldeidda:
What is the reason for the following:
"ProjDataInfo: Number of TOF bins should be an odd number. Abort."

mediso scanner has an even number of bins

KrisThielemans:
@danieldeidda we are currently using the (bad) STIR strategy of counting TOF bins from -3 to 3 (for 7 bins), such that TOF bin 0 is in the "centre". This is nice and easy to understand, and also great for applying symmetries. But of course, it no longer works for an even number of bins.

Things that definitely need changing:
- `ProjDataInfo::set_tof_mash_factor` (which is used to initialise all TOF things) as it assumes that it's an odd number.
- the `DataSymmetries` code, certainly `SymmetryOperations_PET_CartesianGrid`

src/include/stir/listmode/CListEventCylindricalScannerWithViewTangRingRingEncoding.inl
```diff
@@ -39,6 +39,7 @@ get_detection_position(DetectionPositionPair<>& det_pos) const
 						 view_num, tangential_pos_num);
   det_pos.pos1().tangential_coord() = det_num_1;
   det_pos.pos2().tangential_coord() = det_num_2;
+  det_pos.timing_pos() = this->get_uncompressed_proj_data_info_sptr()->get_tof_bin(delta_time);
```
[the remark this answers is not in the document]
KrisThielemans: it will be in the next commit

src/include/stir/recon_buildblock/ProjMatrixElemsForOneBin.h
```diff
@@ -113,6 +115,8 @@ class ProjMatrixElemsForOneBin
   inline Bin get_bin() const;
   //! and set the bin coordinates
   inline void set_bin(const Bin&);
+  //! get a ref to the bin
+  inline Bin* get_bin_ptr();
```
KrisThielemans, line 119: cannot retrun a bare pointer. I see this is then because `get_bin` returns a copy, but this is too dangerous.

We could replace `get_bin` to return a `const Bin&`. I see no harm in that.

src/include/stir/recon_buildblock/TrivialDataSymmetriesForBins.h
```diff
@@ -48,7 +48,8 @@ class TrivialDataSymmetriesForBins : public DataSymmetriesForBins
   virtual void
     get_related_bins(std::vector<Bin>&, const Bin& b,
                       const int min_axial_pos_num, const int max_axial_pos_num,
-                      const int min_tangential_pos_num, const int max_tangential_pos_num) const;
+                      const int min_tangential_pos_num, const int max_tangential_pos_num,
+					  const int min_timing_pos_num = 0, const int max_timing_pos_num = 0) const;
```
KrisThielemans, line 52: do we really need these defaults?

src/include/stir/recon_buildblock/distributable.h
```diff
@@ -175,7 +175,8 @@ void distributable_computation(
                                const double start_time_of_frame,
                                const double end_time_of_frame,
                                RPC_process_related_viewgrams_type * RPC_process_related_viewgrams,
-                               DistributedCachingInformation* caching_info_ptr);
+                               DistributedCachingInformation* caching_info_ptr,
+							   int min_timing_pos_num = 0, int max_timing_pos_num = 0);
```
KrisThielemans, line 179: remove defaults

src/include/stir/recon_buildblock/distributableMPICacheEnabled.h
```diff
@@ -65,7 +65,8 @@ void distributable_computation_cache_enabled(
                                              const double start_time_of_frame,
                                              const double end_time_of_frame,
                                              RPC_process_related_viewgrams_type * RPC_process_related_viewgrams, 
-                                             DistributedCachingInformation* caching_info_ptr
+                                             DistributedCachingInformation* caching_info_ptr,
+											 int min_timing_pos_num = 0, int max_timing_pos_num = 0
```
KrisThielemans, line 69: remove defaults

src/listmode_buildblock/CListModeDataECAT.cxx
```diff
@@ -328,6 +328,14 @@ get_num_records() const
 
 #endif
 
+template <class CListRecordT>
+shared_ptr<ProjDataInfo>
```
KrisThielemans, line 332: can we returna `shared_ptr<const ProjDataInfo>`?

src/recon_buildblock/BackProjectorByBinUsingProjMatrixByBin.cxx
```diff
@@ -237,5 +241,41 @@ actual_back_project(DiscretisedDensity<3,float>& image,
...
+    if (proj_matrix_ptr->is_cache_enabled() /*&& !tof_enabled*/)
+    {
+        ProjMatrixElemsForOneBin proj_matrix_row;
+        proj_matrix_ptr->get_proj_matrix_elems_for_one_bin(proj_matrix_row, bin);
+        proj_matrix_row.back_project(image, bin);
+    }
+    else if (proj_matrix_ptr->is_cache_enabled()/* && tof_enabled*/)
+    {
+        tof_row->back_project(image, bin);
+    }
```
KrisThielemans, line 258: this should be reviewed. It is likely obsolete. If we do need it, the interface with `tof_row` etc should be made a lot cleaner (no bare pointers!)

src/recon_buildblock/BinNormalisationFromProjData.cxx
```diff
@@ -93,15 +93,7 @@ set_up(const shared_ptr<const ExamInfo>& exam_info_sptr, const shared_ptr<const
...
       (norm_proj.get_max_tangential_pos_num() ==proj.get_max_tangential_pos_num());
-    
-    for (int segment_num=proj.get_min_segment_num();
-	 ok && segment_num<=proj.get_max_segment_num();
-	 ++segment_num)
-      {
-	ok = 
-	  norm_proj.get_min_axial_pos_num(segment_num) == proj.get_min_axial_pos_num(segment_num) &&
-	  norm_proj.get_max_axial_pos_num(segment_num) == proj.get_max_axial_pos_num(segment_num);
-      }
```
KrisThielemans, line 104: should be reinstated. It's going to take very little time, and only at `set_up`

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeData.cxx
```diff
@@ -372,14 +374,6 @@ set_up_before_sensitivity(shared_ptr <const TargetT > const& target_sptr)
...
     }
 
-  if (this->normalisation_sptr->set_up(
-                this->list_mode_data_sptr->get_exam_info_sptr(), proj_data_info_sptr->create_shared_clone()) == Succeeded::no)
-    {
-      warning("PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin: "
-              "set-up of normalisation failed.");
-      return Succeeded::no;
-    }
-
```
KrisThielemans, line 382: need to review why this was removed

src/recon_buildblock/PresmoothingForwardProjectorByBin.cxx
```diff
@@ -95,6 +95,15 @@ get_symmetries_used() const
...
 }
+
+//void PresmoothingForwardProjectorByBin::
+//update_filtered_density_image(const DiscretisedDensity<3,float>& density)
+//{
+////        filtered_density_sptr.reset(density.get_empty_discretised_density());
+////    image_processor_ptr->apply(*filtered_density_sptr, density);
+////    assert(density.get_index_range() == filtered_density_sptr->get_index_range());
+//}
+
```
KrisThielemans, line 106:
```suggestion
```

src/include/stir/listmode/CListEventCylindricalScannerWithDiscreteDetectors.h
```diff
@@ -10,6 +10,7 @@
 */
 /*
     Copyright (C) 2023, University College London
+    Copyright (C) 2003- 2011, Hammersmith Imanet Ltd
```
KrisThielemans, line 13:
```suggestion
```

src/IO/InputStreamFromROOTFileForCylindricalPET.cxx
```diff
@@ -50,7 +50,7 @@ InputStreamFromROOTFileForCylindricalPET(std::string _filename,
     up_energy_window = _up_energy_window;
     offset_dets = _offset_dets;
 
-    half_block = module_repeater_y * submodule_repeater_y * crystal_repeater_y / 2  - 1;
+    half_block = (module_repeater_y * submodule_repeater_y * crystal_repeater_y) / 2;
```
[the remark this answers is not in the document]
KrisThielemans:
```suggestion
    half_block = module_repeater_y * submodule_repeater_y * crystal_repeater_y / 2 - 1;
```
@NikEfth I am reinstating this to be identical to V4. It is obsolete anyway.

src/include/stir/recon_buildblock/ProjMatrixByBin.h
```diff
@@ -109,12 +116,15 @@ class ProjMatrixByBin :
   
   The implementation is inline as it just gets it in
   terms of the cached_proj_matrix_elems_for_one_bin or 
-  calculate_proj_matrix_elems_for_one_bin.*/
+  calculate_proj_matrix_elems_for_one_bin.
+
+  N.E: Updated to accomondate TOF information.
+*/
```
KrisThielemans, line 122:
```suggestion
  calculate_proj_matrix_elems_for_one_bin.*/
```

src/include/stir/recon_buildblock/ProjMatrixByBin.h
```diff
@@ -223,7 +247,27 @@ class ProjMatrixByBin :
   // KT 15/05/2002 not static anymore as it uses cache_stores_only_basic_bins
   CacheKey cache_key(const Bin& bin) const;
 
-   
+  //! Activates the application of the timing kernel to the LOR
+  //! and performs initial set_up().
+  //! \warning Must be called after set_up()
```
KrisThielemans, line 252:
```suggestion
  //! \warning Must be called during set_up()
```

src/include/stir/recon_buildblock/ProjectorByBinPair.h
```diff
@@ -25,6 +25,7 @@
 #include "stir/recon_buildblock/BackProjectorByBin.h"
 #include "stir/ParsingObject.h"
 #include "stir/shared_ptr.h"
+#include "stir/recon_buildblock/ProjMatrixElemsForOneBin.h"
```
KrisThielemans, line 28:
```suggestion
```

src/recon_buildblock/ProjectorByBinPair.cxx
```diff
@@ -24,6 +24,11 @@
 #include "stir/ProjDataInfo.h"
 #include "stir/DiscretisedDensity.h"
 #include "stir/Succeeded.h"
+#include "stir/recon_buildblock/PresmoothingForwardProjectorByBin.h"
+#include "stir/recon_buildblock/PostsmoothingBackProjectorByBin.h"
+#include "stir/recon_buildblock/ForwardProjectorByBinUsingProjMatrixByBin.h"
+#include "stir/recon_buildblock/BackProjectorByBinUsingProjMatrixByBin.h"
```
KrisThielemans, line 30:
```suggestion
```

KrisThielemans (changes requested):

src/recon_buildblock/GeneralisedObjectiveFunction.cxx
```diff
@@ -47,6 +50,7 @@ GeneralisedObjectiveFunction<TargetT>::
 initialise_keymap()
 {
   this->parser.add_parsing_key("prior type", &prior_sptr);
+  this->parser.add_key("Use TOF information", &use_tof);
```
KrisThielemans, line 53: need to remove. This class knows nothing about the data, so TOF makes no sense here.. The user can always create a non-tof sino (harder with listmode recon of course).

KrisThielemans:
I've now reviewed all files I believe. There are still a number of unresolved comments (easiest to see from the file tab), but not that many anymore.

src/recon_buildblock/GeneralisedObjectiveFunction.cxx
robbietuk: Is there documentation detailing how this is done?

danieldeidda:
I have been looking at the change of Tof_bin index. I think I have changed the relevant bits but the test surprisingly complains on the consistency check of coordinates. I can see this happens because of a series of swapping that I am not sure I understand. For example:
https://github.com/NikEfth/STIR/blob/tof_sino_UCL/src/include/stir/ProjDataInfoCylindricalNoArcCorr.inl#L134-L144

 I find that have a swap between detA and B but not between ring and/or vice versa.

KrisThielemans:
The difficulty here is that bins are "unique" but det_pos is not: there are 2 of the latter for 1 bin.

Obviously, swaps should be such that swapping detA/ringA with detB/ringB and the TOF index should give exactly the same bin. I believe there are tests for that. (There is one if you have ROOT installed).

Not sure if this helped. Maybe you can give some more details then.

danieldeidda:
Hi yes I am using root and those are the test failing. the test on the bin are OK. In the group of line that I referred to in the previous ( actually not the one I wnted to show but I am confused here as well) comment the swap happens when swap is false

danieldeidda:
the function called there is this:
https://github.com/NikEfth/STIR/blob/tof_sino_UCL/src/include/stir/ProjDataInfoCylindricalNoArcCorr.inl#L116-L117

danieldeidda:
A similar thing happens here. Not sure whether it's me who is not grasping what is happening or if this is a consistent bug:
https://github.com/NikEfth/STIR/blob/tof_sino_UCL/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx#L522-L535

NikEfth:
> I find that have a swap between detA and B but not between ring and/or vice versa.
There is a reason. 
These flips are there proactively to flip the detectors according to how the Euler's formula will flip them. 
In Euler's, going from Cartesian to Polar and back does not preserve the orientation. 
So, it would not make sense to flip the rings, as the conversion in not in the ring.

danieldeidda:
> Euler's formula

so should the following only swap detA and B and leave the rings untouched?
https://github.com/NikEfth/STIR/blob/tof_sino_UCL/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx#L522-L535

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeData.cxx
KrisThielemans: It is now in `PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin` as it needs the "sensitivity" projector.

KrisThielemans (commented):
This will have to do for now. Outstanding issues and GE data will have to be handled on `master`.

Thanks to everyone!

PR #306, opened 2019-01-15
Fix CMAKE_MODULE_PATH

rijobro:
We previously overwrote the `CMAKE_MODULE_PATH` value. This meant that packages using `STIRConfig.cmake` (namely SIRF) had their own value overwritten.

With this, we store the original value, and add in ours.


rijobro:
No strong feelings either way.

KrisThielemans:
let's prepend then (as it's closest to previous behaviour)

KrisThielemans:
@rijobro, note: I squash-merged this one

PR #307, opened 2019-01-17
Flags for knowing whether STIR was built with ITK etc or not.

rijobro:
Useful for SIRF, which needs to know


KrisThielemans:
squash-merged this

PR #309, opened 2019-01-23
Prepend stir include files

rijobro:
We had a problem that ITK (and possibly other packages) added certain directories to the beginning of the list. On the cluster we had ITK and an older version of STIR in the same location. We were therefore including this older version of STIR, too, causing problems. 

We first do all the find_packages and then prepend our headers.


KrisThielemans:
one travis failure on OSX due to a time-out getting gcc, so ignoring that.

PR #310, opened 2019-01-24
target_include_directories for boost

rijobro:
This avoids having to use find_package(boost), which should be avoided as much as possible. 

The ITK files were already linked privately, and so the find_package(ITK) can be removed from the STIRConfig.cmake file (which is good because the findITK.cmake prepends all its paths, which is a little naughty).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,6 @@ set(ConfigPackageLocation lib/cmake/)
 #### we need the boost library from boost.org
 set(BOOST_ROOT CACHE PATH "root of Boost")
 find_package( Boost 1.36.0 REQUIRED )
-include_directories(${Boost_INCLUDE_DIRS})
 
 #### optional external libraries. 
 # Listed here such that we know if we should compile extra utilities
--- a/src/cmake/STIRConfig.cmake.in
+++ b/src/cmake/STIRConfig.cmake.in
@@ -46,16 +46,8 @@ set(STIR_REGISTRIES @STIR_REGISTRIES@)
 # to set ITK_DIR etc first before calling find_package
 # to make sure we pick the same version of the external library
 
-if (@BOOST_ROOT@)
-  set(BOOST_ROOT "@BOOST_ROOT@")
-endif()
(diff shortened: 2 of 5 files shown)


rijobro:
When building SIRF, removing the `find_package(ITK)` from `STIRConfig.cmake.in` gives the following errors:
```
[ 29%] Linking CXX executable test4
/opt/rh/devtoolset-3/root/usr/libexec/gcc/x86_64-redhat-linux/4.9.2/ld: cannot find -lITKCommon
/opt/rh/devtoolset-3/root/usr/libexec/gcc/x86_64-redhat-linux/4.9.2/ld: cannot find -litkdouble-conversion
/opt/rh/devtoolset-3/root/usr/libexec/gcc/x86_64-redhat-linux/4.9.2/ld: cannot find -litksys
/opt/rh/devtoolset-3/root/usr/libexec/gcc/x86_64-redhat-linux/4.9.2/ld: cannot find -litkvnl_algo
/opt/rh/devtoolset-3/root/usr/libexec/gcc/x86_64-redhat-linux/4.9.2/ld: cannot find -litkvnl
/opt/rh/devtoolset-3/root/usr/libexec/gcc/x86_64-redhat-linux/4.9.2/ld: cannot find -litkv3p_netlib
```
Should I just put it back?

rijobro:
Using `add_dependencies(${dir} Boost::boost)` gives errors along the lines of:
```
/Users/rich/Documents/Code/STIR/Build/src/include/stir/config.h:55:10: fatal error: 'boost/config.hpp' file not found
#include "boost/config.hpp"
```

rijobro:
`find_package` has both `HINTS` and `PATHS`. What do you think about using either when doing `find_package(ITK)` such that it is more likely to find the version we want it to?

KrisThielemans:
This is fine for me. Squash-merge?

KrisThielemans:
ITK problem:
the linking problem is because we don't do
```cmake
include(${ITK_USE_FILE})
```
still seems weird we'd need it. I suggest we leave this for a different PR

Issue #315, opened 2019-02-19
Compilation Error with Xcode and ROOT


robbietuk:
Error due to unique_ptr implementation in Clang, creating the following error at compilation time:

`[  8%] Building CXX object src/IO/CMakeFiles/IO.dir/InputStreamFromROOTFile.cxx.o
In file included from /Users/RobertTwyman/STIR_bin/STIR/src/IO/InputStreamFromROOTFile.cxx:20:
In file included from /Users/RobertTwyman/STIR_bin/STIR/src/include/stir/IO/InputStreamFromROOTFile.h:37:
In file included from /Users/RobertTwyman/ROOT_bin/install/include/TROOT.h:28:
In file included from /Users/RobertTwyman/ROOT_bin/install/include/TDirectory.h:24:
In file included from /Users/RobertTwyman/ROOT_bin/install/include/TBuffer.h:25:
In file included from /Users/RobertTwyman/ROOT_bin/install/include/TClass.h:31:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/unordered_set:323:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/__hash_table:957:13: error: too many template arguments for
      class template 'auto_ptr'
    typedef unique_ptr<__next_pointer[], __bucket_list_deleter> __bucket_list;
            ^                            ~~~~~~~~~~~~~~~~~~~~~~
/Users/RobertTwyman/STIR_bin/STIR/src/include/stir/unique_ptr.h:84:20: note: expanded from macro 'unique_ptr'
#define unique_ptr auto_ptr
                   ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/memory:2029:28: note: template is declared here
class _LIBCPP_TEMPLATE_VIS auto_ptr
`

To fix this, unordered_set was added to the include list in unique_ptr.h .

However, a new compilation error arises:

`[ 10%] Building CXX object src/IO/CMakeFiles/IO.dir/InputStreamFromROOTFileForCylindricalPET.cxx.o
In file included from /Users/RobertTwyman/STIR_bin/STIR/src/IO/InputStreamFromROOTFile.cxx:20:
In file included from /Users/RobertTwyman/STIR_bin/STIR/src/include/stir/IO/InputStreamFromROOTFile.h:38:
[...]

PR #316, opened 2019-02-21
change checks for std: smart pointers

robbietuk:
Addition of compile tests to the CMakeLists.txt to check if unique_ptr exists. See Issue #315


KrisThielemans:
looks good. let's see what Travis and Appveyor say. However, it now makes sense to do the same for `std::shared_ptr`. Could you give that a go? (One dependency less on Boost.!)

src/CMakeLists.txt
```diff
@@ -181,6 +181,30 @@ include(CheckCXXSymbolExists)
...
+   "
+        STIR_HAVE_UNIQUE_PTR)
+
+if(STIR_HAVE_UNIQUE_PTR)
+    option(STIR_NO_UNIQUE_PTR "Disable use of std::unique_ptr" OFF)
+else()
+    set(STIR_NO_UNIQUE_PTR ON)
+endif()
+
+# Perform test to see if BOOST shared_ptr exists
```
KrisThielemans:
```suggestion
# Perform test to see if std::shared_ptr exists
```

src/CMakeLists.txt
```diff
@@ -181,6 +181,30 @@ include(CheckCXXSymbolExists)
...
+check_cxx_source_compiles("
+        #include <memory>
+        std::shared_ptr<int> f(){int *ptr = new int; return std::shared_ptr<int>(ptr);}
+        int main(){auto p = f(); return 0;}
+    "
+        STIR_USE_BOOST_SHARED_PTR
```
KrisThielemans:
```suggestion
       HAVE_STD_SHARED_PTR
```

src/CMakeLists.txt
```diff
@@ -181,6 +181,30 @@ include(CheckCXXSymbolExists)
...
+        )
+
 # Now we introduce 2 options such that the user can override the default behaviour
 # (not recommended! unless the above failed of course)
 option(STIR_NO_UNIQUE_PTR
```
KrisThielemans: merge this with above code

KrisThielemans: also need to fix STIR_USE_BOOST_SHARED_PTR option below in a similar way as the unique_ptr stuff above.

finally need to move the `mark_as_advanced` statement inside the `if`

robbietuk: I am not sure if the mark_as_advanced stuff is implemented correctly.

KrisThielemans:
Travis is happy on OSX but I'm afraid all Linux builds are failing. See e.g. https://travis-ci.org/UCL/STIR/jobs/496969217

Can you try this on the VM and see what's happening? Sorry.

robbietuk:
Tested on the VM. It appears that both our new checks (the`check_cxx_source_compiles`) are False so indicating that unique_ptr and std::shared_ptr aren't present.

KrisThielemans:
Please check CMakeFiles/CMakeError.log. Hopefully it tells you something. Feel free to post here.

KrisThielemans:
let's add an explicit C++11 case to Travis:
```
 - os: linux
   python: 3
   env: EXTRA_BUILD_FLAGS="-DDISABLE_CERN_ROOT=0 -DSTIR_OPENMP:BOOL=ON" -DCMAKE_CXX_STANDARD=11 CC=gcc-5 CXX=g++-5

robbietuk:
Compiles for me on a VM. Waiting for checks.

PR #318, opened 2019-03-06
more info on image z-spacing and ring spacing if there is a mistmatch

rijobro:
I required this for debugging, and it doesn't seem harmful to add more information for future debuggers.


rijobro:
done.

PR #319, opened 2019-03-18
Two list mode functions to get and set position in list mode data stream

robbietuk:
Added functions: `get_list_mode_position()` and `set_list_mode_position(pos)`

`get_list_mode_position()` returns an unsigned long with the current position in the list mode data

`set_list_mode_position(pos)` set the current position to `pos`

These are two functions I have developed for a _"block"_ subset sampling method for list mode.

### DIFF
--- a/src/include/stir/listmode/CListModeData.h
+++ b/src/include/stir/listmode/CListModeData.h
@@ -214,7 +214,7 @@ public:
   virtual
     Succeeded set_get_position(const SavedPosition&) = 0;
 
-   //! Go to a position in the listmode data
+   //! Set the position for reading to a integer position
   virtual
     Succeeded set_listmode_position(unsigned long pos) = 0;
     //Succeeded set_listmode_position(unsigned long pos) {return Succeeded::no;}//= 0;
--- a/src/include/stir/listmode/CListModeData.h
+++ b/src/include/stir/listmode/CListModeData.h
@@ -215,8 +215,9 @@ public:
     Succeeded set_get_position(const SavedPosition&) = 0;
 
    //! Go to a position in the listmode data
-//  virtual
-//    Succeeded set_listmode_position(unsigned long pos) {return Succeeded::no;}//= 0;
+  virtual
+    Succeeded set_listmode_position(unsigned long pos) = 0;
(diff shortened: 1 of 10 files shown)


robbietuk:
As per @rijobro 's advice, I have modified the old method to `set_get_position()` to depend on `set_list_mode_position()` to remove code duplication

robbietuk:
I might have made a mess of this now. I tried to reset the branch and cherrypick the changes I made.

src/include/stir/listmode/CListModeData.h
```diff
@@ -31,6 +31,7 @@
 #include "stir/ProjDataInfo.h"
 #include "stir/IO/ExamData.h"
 #include "stir/RegisteredParsingObject.h"
+#include "stir/Succeeded.h"
```
rijobro: You shouldn't need this include now (we put it in during debugging)

src/listmode_buildblock/CListModeDataECAT8_32bit.cxx
```diff
@@ -136,5 +136,20 @@ set_get_position(const CListModeDataECAT8_32bit::SavedPosition& pos)
...
+{
+    return current_lm_data_ptr->get_list_mode_position();
+}
+
+Succeeded
+CListModeDataECAT8_32bit::
+set_list_mode_position(const unsigned long pos)
+{
+    return current_lm_data_ptr->set_list_mode_position(pos);
+    //return Succeeded::no;
```
rijobro: can delete this commented line

KrisThielemans:
I really don't see the point of this. What does "position" mean in a generic listmode file? People might be tempted to think it counts records, but it won't. Even for file formats with same size of record, it'll fail in cases where there's a header. 

Once you say "position is just an arbitrary marker to allow be to go back", it's exactly the same as the `save_get_position` in my opinion.

As this is a change that breaks backwards compatibility (classes that are not currently in STIR would need extra stuff), I'm reluctant.

let's talk about it.

robbietuk:
After discussing this with Kris, I agree that this functionality is not generally compatible with general list mode file formats. Possibly `get_list_mode_position()` has some use in debugging but without its set counterpart might be redundant. 

I will keep both methods in my own branch. Shall I close this pull request?

Issue #320, opened 2019-03-19
Number of time frames in wrong position in interfile header


ALEXJAZZ008008:
The line: "number of time frames :=" in Interfile headers is above the line: "!PET data type :=" rather than above the block: "first pixel offset (mm) [1] := \nfirst pixel offset (mm) [2] := \nfirst pixel offset (mm) [3] := "

This is causing issues when trying to open the Interfile images to view or convert to Nifty.

rijobro:
Why does this cause a problem? (haven't look at the code...)
The number of time frames and the offset seem to be unrelated in my mind.

ALEXJAZZ008008:
I'm not certain why it causes a problem, when I opened an Interfile header in Amide before it was entirely blank, I asked Elise about it and she told me the issue was caused by this ordering.
If I manually move the line in my Interfile header it causes Amide to be able to open the file as expected.
I didn't want to manually move the line for every one of my Interfile files and I don't know if it affects converting Interfile to Nifty.

I'm running with my edited branch where this change has been made and it seems fine to me, although I haven't had long to test it.

KrisThielemans:
This problem was originally spotted by @pw351 and @eliseemond. As I stated in #321 this is really a medcon problem. However, changing the order of the keywords in the interfile header in STIR would prevent headaches.

KrisThielemans:
closed by #321

PR #321, opened 2019-03-19
Number of time frames in wrong position in interfile header fix

ALEXJAZZ008008:
Moved the method call "write_interfile_time_frame_definitions(output_header, exam_info);" to be in the correct place when creating interfile headers.

This is to fix issue #320: Number of time frames in wrong position in interfile header

### DIFF
--- a/src/IO/interfile.cxx
+++ b/src/IO/interfile.cxx
@@ -515,7 +515,7 @@ static void interfile_create_filenames(const std::string& filename, std::string&
 
   header_name=filename;
   replace_extension(header_name, ".hv");
-}
+}output_header << "number of time frames := 1\n";
 
 ////// end static functions
 
@@ -593,7 +593,6 @@ write_basic_interfile_image_header(const string& header_file_name,
     {
       output_header << "!PET STUDY (General) :=\n";
     }
-  write_interfile_time_frame_definitions(output_header, exam_info);
   write_interfile_energy_windows(output_header, exam_info);
   write_interfile_image_data_descriptions(output_header, data_type_descriptions);
   if (!is_spect)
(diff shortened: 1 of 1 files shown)


KrisThielemans:
sadly fails with "ERROR: KeyParser: the list corresponding to the keyword "image scaling factor" has to be resized to size 2. This means you have a problem in the keyword values." Not 100% sure why. Would have to compare the interfile header written now with what we had before we changed the location of the time frame defs keyword.

KrisThielemans:
please paste a new .hv header here.

#320 is really due to a problem in medcon (as used by Amide). Your fix might help amide, but it breaks STIR reading of the header.

ALEXJAZZ008008:
!INTERFILE  :=
!imaging modality := PT
!version of keys := STIR3.0
name of data file := ImageData1553019963368335.v
!GENERAL DATA :=
patient orientation := head_in
patient rotation := prone
!GENERAL IMAGE DATA :=
!type of data := PET
imagedata byte order := LITTLEENDIAN
!PET STUDY (General) :=
!PET data type := Image
process status := Reconstructed
!number format := float
!number of bytes per pixel := 4
number of dimensions := 3
matrix axis label [1] := x
!matrix size [1] := 125
scaling factor (mm/pixel) [1] := 3.1088
matrix axis label [2] := y
!matrix size [2] := 125
scaling factor (mm/pixel) [2] := 3.1088
matrix axis label [3] := z
!matrix size [3] := 31
scaling factor (mm/pixel) [3] := 6.75
number of time frames := 1
first pixel offset (mm) [1] := -192.746
first pixel offset (mm) [2] := -192.746
first pixel offset (mm) [3] := 0
!END OF INTERFILE :=

KrisThielemans:
clarification. The error is in `test_IO_ParametricDiscretisedDensity.*`

KrisThielemans:
The reason this fails is actually due to a bug in the Interfile reading code (fix coming up). However, the order of the keywords was a bit strange. I've commited a change to put them all together again.

PR #322, opened 2019-03-20
added set/get pairs for ProjMatrixByBinSPECTUB

KrisThielemans:
This needed some minor changes to set_up as well.

Also made sure that set_up() doesn't leak memory for all the UB arrays if
it's called twice.

### DIFF
--- a/src/include/stir/recon_buildblock/ProjMatrixByBinSPECTUB.h
+++ b/src/include/stir/recon_buildblock/ProjMatrixByBinSPECTUB.h
@@ -142,7 +142,7 @@ class ProjMatrixByBinSPECTUB :
     distance from the collimator.
 
     \verbatim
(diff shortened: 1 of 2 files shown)


KrisThielemans:
@danieldeidda would you be able to have a look? No tests for this yet, but I'll test it via SIRF.

danieldeidda:
> @danieldeidda would you be able to have a look? No tests for this yet, but I'll test it via SIRF.

What should I look? from what I quickly saw this commit introduces changes for attenuation. Shall I check these?

KrisThielemans:
just general check if the code looks ok (I know that the SPECTUBMatrix code is a bit messy). Possibly you could try the new set/get functions from within STIR Python (or Matlab).

SIRF tests seem to work out fine.

PR #325, opened 2019-03-25
SPECT-related fixes

### LINKED ISSUE #323: forward_project with SPECTUBMatrix problem
`forward_project` with a template results in output with twice the z-spacing as the template.

KrisThielemans:
- fixes an issue that the axial spacing in SPECT data was reported twice too large (therefore resulting in wrong information in projdata files written by the forward projector)
- let the `SPECTUBMatrix` `set_up` abort if z-voxel-size and number of slices in the images is not what the code assumes it to be (possibly this condition could be relaxed, but I didn't check)
- introduce new `z zoom` parameter in reconstruction `.par` files such that we can tell STIR to use z-voxel-size compatible with SPECT data. **Warning**: this breaks backwards compatibility.
- remove duplication between many different classes (`AnalyticReconstruction` and all `PoissonLogLikelihood*`) for image keywords.

Note that before this PR
- SPECT reconstruction were fine if you set the number of image slices equal to the number of sinograms and z-spacing equal to the axial spacing when passing an initial image.
- output projection data (e.g., of `forward_project) was written with an incorrect header. This was hidden by 2 bugs (in reading the interfile header, and in ignoring axial spacing by the SPECT UB matrix)
with the result that subsequent reconstructions were still fine.

fixes #323

### DIFF
--- a/documentation/release_3.1.htm
+++ b/documentation/release_3.1.htm
@@ -160,6 +160,11 @@ These are only minimally documented at the present stage unfortunately.
   </li>
 <li>Extract single images from dynamic image with executable <code>extract_dynamic_images</code>.
 </li>
+<li>Reconstruction parameter files now can use a new optional keyword <code>z zoom</code> to llow
+  changing the spacing from the default. However, results of this have not been checked yet.
+  It is currently only recommended for SPECT (where it should be set to <code>0.5</code> due
+  current work-arounds on how to read SPECT dat into STIR).
+  </li>
   </ul>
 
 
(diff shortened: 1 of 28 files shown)


danieldeidda:
Not sure if this is related to this issue but when using stir-math on SPECT projdata the bin size[2]  becomes double the bin size[1]. Note tha I did not try on PET data though.

danieldeidda:
Also when I try to  use a zoom > 1 to have smaller voxels I get this error:
psf length greater than maxszb in calc_psf_bin

although psf is disable

KrisThielemans:
@danieldeidda you said
> stir-math on SPECT projdata the bin size[2] becomes double the bin size[1].

This PR should have fixed that. (It does seem to work in my test). Please check. (after #343 is merged..)

PR #336, opened 2019-05-21
Relax DICOM series reading restrictions

### LINKED ISSUE #332: DICOM series reader restrictions can result in incomplete volumes
Series restrictions can result in incomplete reading of DICOM volumes. This affects CT data - particularly cine CT - and likely MR too.

https://github.com/UCL/STIR/blob/3a277f7a819f35a553a8d6097402ea25cf55a240/src/IO/ITKImageInputFileFormat.cxx#L445

bathomas:
Disables series restrictions for Acquisition, Trigger and Frame times while reading DICOM data. This fixes #332 for data where all slices genuinely belong to a 3D volume, despite having different timing information per slice. However, disabling these restrictions will likely cause issues when reading dynamic PET or CT data from DICOM. One way to avoid this would be to sort the data into sub-folders according to the appropriate tag and read each folder independently. Alternatively, a utility could  be written along these lines.

### DIFF
--- a/src/IO/ITKImageInputFileFormat.cxx
+++ b/src/IO/ITKImageInputFileFormat.cxx
@@ -442,9 +442,9 @@ read_file_itk(const std::string &filename)
           nameGenerator->SetUseSeriesDetails( true );
           // Make sure we read only data from a single frame and gate
           nameGenerator->AddSeriesRestriction("0008|0022" ); // AcquisitionDate
-          nameGenerator->AddSeriesRestriction("0008|0032" ); // AcquisitionTime
-          nameGenerator->AddSeriesRestriction("0018|1060" ); // TriggerTime
-          nameGenerator->AddSeriesRestriction("0018|1063" ); // FrameTime
+          //nameGenerator->AddSeriesRestriction("0008|0032" ); // AcquisitionTime
+          //nameGenerator->AddSeriesRestriction("0018|1060" ); // TriggerTime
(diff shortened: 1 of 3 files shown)


src/IO/ITKImageInputFileFormat.cxx
```diff
@@ -440,11 +440,11 @@ read_file_itk(const std::string &filename)
           typedef itk::GDCMSeriesFileNames NamesGeneratorType;
           typename NamesGeneratorType::Pointer nameGenerator = NamesGeneratorType::New();
           nameGenerator->SetUseSeriesDetails( true );
-          // Make sure we read only data from a single frame and gate
+          // Reads complete series.
           nameGenerator->AddSeriesRestriction("0008|0022" ); // AcquisitionDate
```
ashgillman: If all the others are removed, is this really necessary? The series number should be adequate no?

bathomas: Yes and no. Really it could be removed though my justifications for leaving it were:

- It does no harm.
- Say a participant had two studies on different days, the series number could be identical if the protocol was the same. If for some reason these were both placed in the same folder then there could be the risk of treating them as one. In reality this shouldn't happen, the studyUID *should* prevent it. However, if the UIDs have been re-written, for example during anonymisation, then it is plausible. 

Happy to remove though...

KrisThielemans: definitely don't want to remove that one. It's easy to make different series with the same series number... (even within the same study)

PR #337, opened 2019-06-04
change behaviour of get_scale_factor_per_sinogram

KrisThielemans:
upsample_and_fit_single_scatter previously aborted if there were not enough counts
(after weighting with the masking used for the tail-fit) in the scatter estimate for one sinogram.
Now we just write a warning and set it to 1. For consistency, we now use scale factor 1 if all
sinograms had zero counts (after weighting), while we were setting it 0 before. As this should never occur,
and the scale factor should be irrelevant then, this should not affect any results.


KrisThielemans:
@ludovicabrusaferri this probably doesn't affect you, but maybe you can have a look as well.

PR #345, opened 2019-06-29
[TRAVIS] use different gcc versions

KrisThielemans:
- explicitly install gcc-5 (seems to be needed now)
- use gcc-6, 7 and 8 for some cases


KrisThielemans:
I also attemped to use clang-6.0 on Linux, but this didn't work due to problems with `auto_ptr` and log size. So I've disabled it again.

Issue #351, opened 2019-07-11
add mCT support (allow for virtual crystals)


KrisThielemans:
The mCT has a virtual crystal between every block in axial direction as well. This makes `Scanner::check_consistency` fail as the number of rings is not equal to `num_blocks * num_crystals_per_block`.

In addition, `BinNormalisationFromECAT7` has hard-wired numbers for the mMR.

KrisThielemans:
closed with #352

PR #352, opened 2019-07-11
Siemens mCT and other scanners with virtual crystal additions

KrisThielemans:
Fixes https://github.com/UCL/STIR/issues/351

It should not change results for the mMR.


KrisThielemans:
@olivierverdier Please check if this works for you. (obviously, non-TOF only for now)

KrisThielemans:
Hi

Some doc on how to create a “normalisation” sinogram from the command line:
```bash
export INPUT=bla.s.hdr
export OUTPUT=norm_projdata.hs
export ECATNORM=norm.n.hdr
correct_projdata ~/devel/STIR/examples/Siemens-mMR/correct_projdata.par
```
Note that if you want the “efficiency factors” (which is better as it’s easier to see the gaps which should have zero efficiency), you need to edit the .par file

KrisThielemans:
An important thing to check is how the efficiencies are stored in the norm file. For the mMR, we had to do a strange shift, see https://github.com/UCL/STIR/blob/8942f4e697e9fc724532760fb8da7773593fc004/src/recon_buildblock/BinNormalisationFromECAT8.cxx#L399-L408.

It's likely that this has to be the same for all Siemens scanners I guess, but who knows?

Best way to check is with a high count uniform cylinder and try it both ways. (you could look at the normalised sinograms, or of course the image). Might be a bit subtle.

KrisThielemans:
This could work for the Biograph 16 as well, but would need uncommenting https://github.com/UCL/STIR/blob/8942f4e697e9fc724532760fb8da7773593fc004/src/recon_buildblock/BinNormalisationFromECAT8.cxx#L290.

Note that I changed https://github.com/UCL/STIR/blob/8942f4e697e9fc724532760fb8da7773593fc004/src/buildblock/Scanner.cxx#L191. It previously pretended there was only 1 axial block to make sure that the test didn't fail.

NikEfth:
Good news, I just found some spare data to test this.

KrisThielemans:
Anyone any info on this? I'd like to merge this in a week or so

KrisThielemans:
Listmode code will need some changes as well, in particular https://github.com/UCL/STIR/blob/e2f0bcb74af6bbf067785e07f58a8430faf3c64d/src/listmode_buildblock/CListRecordECAT8_32bit.cxx#L48-L60 Relevant replacement lines are mentioned in
https://github.com/UCL/STIR/pull/355#discussion_r313026647

KrisThielemans:
any feedback please?

KrisThielemans:
I'll close this and put in a PR on the release_4.0.0 branch

KrisThielemans:
I've put in the change for the listmode data suggested above. However, it doesn't fix the problem that the rest of the code is limited to span=1. That needs more work so won't be for this PR.

KrisThielemans:
I'll merge this. However, somebody will have to check if the ECAT8 normalisation code actually work for scanners other than the mMR. Volunteers needed!

PR #361, opened 2019-08-07
fixed axial blocks in ECAT pet

ludovicabrusaferri:
Fixed issue with number of axial blocks in z direction in ECAT PET

### DIFF
--- a/examples/samples/root_headerECAT.hroot
+++ b/examples/samples/root_headerECAT.hroot
@@ -24,7 +24,7 @@ name of input TChain := Coincidences
 Number of blocks Z := 8
 number of blocks Y := 56
 number of crystals X := 1
-number of crystals Y := 8
+number of crystals Y := 9
 number of crystals Z := 8
 
 ;; From GATE's online documentation: 
--- a/examples/samples/root_headerECAT.hroot
+++ b/examples/samples/root_headerECAT.hroot
@@ -1,0 +1,61 @@
+ROOT header := 
+
+originating system := User_defined_scanner
+Number of rings                          := 64
+Number of detectors per ring             := 504
+Inner ring diameter (cm)                 := 65.6
+Average depth of interaction (cm)        := 0.7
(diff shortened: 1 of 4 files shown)


src/include/stir/IO/InputStreamFromROOTFileForECATPET.inl
```diff
@@ -21,29 +21,29 @@ int
 InputStreamFromROOTFileForECATPET::
 get_num_rings() const
 {
-    return this->crystal_repeater_z;
+   return this->block_repeater_y * this->crystal_repeater_z;
```
KrisThielemans: `block_repeater_z`?

ludovicabrusaferri: Yes, i was about to fix it

ludovicabrusaferri: I think there is an example in https://github.com/UCL/STIR/blob/9be73d4615e4cd53a6fbc2abdd6ebf4e8f85e59d/src/include/stir/IO/InputStreamFromROOTFileForECATPET.h#L51-L66

i added the keyword and also corrected the current crystal keywords that were wrong.

KrisThielemans:
I meant like https://github.com/UCL/STIR/blob/e2f0bcb74af6bbf067785e07f58a8430faf3c64d/examples/samples/root_header.hroot. could you create one like that as well? let's call it `root_header_ECAT.hroot` and add a comment in there that this is for an ECAT GATE system

examples/samples/root_headerECAT.hroot
```diff
@@ -0,0 +1,61 @@
+ROOT header := 
+
```
KrisThielemans: add a comment like `This is an example file for a ROOT file generated with GATE for a Siemens mMR-like scanner (but using an 8x9 block) and using the ECAT system as output.`

PR #362, opened 2019-08-13
Back projection can accumulate into same image. 

rijobro:
This breaks the current multithreading in STIR (due to changes to distributable.cxx and DistributedWorker.cxx).

### DIFF
--- a/documentation/release_3.1.htm
+++ b/documentation/release_3.1.htm
@@ -80,11 +80,18 @@ Benjamin Thomas (CIRC ASTAR and UCL), Yu-jung Tsai, Vesna Cuplov).
   <li> minor changes when using Python/MATLAB</li>
   <li> the need for a new <code>z zoom</code> keyword for SPECT reconstructions (see the updated sample .par files)</li>
   <li> changed image orientation when reading/writing images via ITK</li>
-  <li> Images must be set prior to using the forward and back projectors. To tell the back projector 
-    to start projecting into a new image, <tt>start_accumulating_in_new_target</tt> must be used. 
-    The result of the back projection is then returned via <tt>get_output</tt>. Backwards compatibility 
-    can be maintained with the CMake switch, <tt>STIR_PROJECTORS_AS_V3</tt>. The user should not notice 
-  these changes unless they have developed their own projectors.</li>
+  <li> The resulting target must be set prior to using the forward and back projectors. 
+    For the forward projector this is done with <tt>set_input</tt>. 
+    For the back projector, this is done during the <tt>set_up</tt>, 
+    but then<tt>start_accumulating_in_new_target</tt> must be called each time the user
+    wishes to start back projecting into a zero-filled image. 
+    The result of the back projection is then returned via <tt>get_output</tt>. 
+    Backwards compatibility can be maintained with the CMake switch, 
+    <tt>STIR_PROJECTORS_AS_V3</tt>. The user should not notice 
+    these changes unless they have developed their own projectors. 
+    These changes were made so that using data processors (such as motion or 
+    smoothing) prior to the forwards projection or following the back 
+  projection will be more efficient.</li>
 </ul>
 
 It is now highly recommended to specify imaging modality and patient orientation in your Interfile headers. If you don't,
@@ -230,11 +237,6 @@ HFS, FFS, HFP or FFP).
     trigger or frame times in a series is no longer enforced. This allows 3-D DICOM volumes to be read correctly but may cause
     issues for dynamic PET or CT data.
   </li>
-  <li> As mentioned in the breaking of backwards compatibility, images must be set prior to using the forward 
-    and back projectors. To tell the back projector to start projecting into a new image, 
-    <tt>start_accumulating_in_new_target</tt> must be used. The result of the back projection is then 
-    returned via <tt>get_output</tt>. This was done so that using data processors (such as motion or 
-  smoothing) prior to the forwards projection or following the back projection will be more efficient.</li>
 </ul>
 <h4>Python (and MATLAB) interface</h4>
 <ul>
--- a/documentation/release_3.1.htm
+++ b/documentation/release_3.1.htm
@@ -80,6 +80,7 @@ Benjamin Thomas (CIRC ASTAR and UCL), Yu-jung Tsai, Vesna Cuplov).
   <li> minor changes when using Python/MATLAB</li>
   <li> the need for a new <code>z zoom</code> keyword for SPECT reconstructions (see the updated sample .par files)</li>
   <li> changed image orientation when reading/writing images via ITK</li>
+  <li> Images must be set prior to using the forward and back projectors. To tell the back projector to start projecting into a new image, <tt>reset_output</tt> must be used. The result of the back projection is then returned via <tt>get_output</tt>. Backwards compatibility can be maintained with the CMake switch, <tt>STIR_PROJECTORS_AS_V3</tt>. The user should not notice these changes unless they have developed their own projectors.</li>
 </ul>
 
 It is now highly recommended to specify imaging modality and patient orientation in your Interfile headers. If you don't,
@@ -137,7 +138,7 @@ These are only minimally documented at the present stage unfortunately.
 <li><tt>list_image_info</tt> now has options to list exam info, or only part of the information (as <tt>list_projdata_info</tt>).</li>
 <li><tt>list_lm_countrates</tt> is a new list mode utility to output total counts per specified time interval.</li>
 <li>List-mode classes for Siemens ECAT8 (32-bit).</li>
-<li>add Verbosity level (currently only to be set from interactive languages)</li
+<li>add Verbosity level (currently only to be set from interactive languages)</li>
 <li>Added the anatomical parallel level sets (PLS) prior. If an uniform anatomical image is provided it will work as a (smoothed) total variation (TV) prior </li>
 <li>New IO:
   <ul>
@@ -225,6 +226,7 @@ HFS, FFS, HFP or FFP).
     trigger or frame times in a series is no longer enforced. This allows 3-D DICOM volumes to be read correctly but may cause
     issues for dynamic PET or CT data.
   </li>
+  <li> As mentioned in the breaking of backwards compatibility, images must be set prior to using the forward and back projectors. To tell the back projector to start projecting into a new image, <tt>reset_output</tt> must be used. The result of the back projection is then returned via <tt>get_output</tt>. This was done so that using data processors (such as motion or smoothing) prior to the forwards projection or following the back projection will be more efficient.</li>
 </ul>
 <h4>Python (and MATLAB) interface</h4>
 <ul>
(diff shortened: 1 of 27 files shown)


KrisThielemans:
Travis error is
```
/home/travis/build/UCL/STIR/src/recon_buildblock/ForwardProjectorByBin.cxx: In member function ‘void stir::ForwardProjectorByBin::forward_project(stir::ProjData&)’:

/home/travis/build/UCL/STIR/src/recon_buildblock/ForwardProjectorByBin.cxx:216:93: error: ‘stir::ForwardProjectorByBin::_density_sptr’ is not a variable in clause ‘shared’

 #pragma omp parallel for  shared(proj_data, _density_sptr, symmetries_sptr) schedule(runtime)
```

rijobro:
Of the 3 Travis builds with OpenMP enabled, gcc 7 and 8 succeed in build and fail in tests. gcc 5 fails during compilation. 

### gcc 5

> error: ‘stir::ForwardProjectorByBin::_density_sptr’ is not a variable in clause ‘shared’
>  #pragma omp parallel for  shared(proj_data, _density_sptr, symmetries_sptr) schedule(runtime)

Could this come from the fact that we moved `_density_sptr` from `private` to `protected`? Because `proj_data` doesn't throw an error (and is still a private member).

### gcc 7 & 8
Two tests fail:
```
-------- Running OSSPS with a quadratic prior --------
Running OSSPS
---- Comparing output of OSSPS subiter 8 (should be identical up to tolerance)
Running compare_image

Maximum absolute error = 1556.16
Maximum in (1st - 2nd) = 0.176174
Minimum in (1st - 2nd) = -1556.16
Error relative to sup-norm of first image = 696809 %
```
and
```
[0K$ ./run_test_zoom_image.sh
This script should work with STIR version 4.0. If you have
a later version, you might have to update your test pack.
Please check the web site.

Using /home/travis/build/UCL/STIR/install/bin/generate_image
Using /home/travis/build/UCL/STIR/install/bin/zoom_image
===  make emission image
Next shape
===  running zoom_image and compare ROI values
===  make attenuation image

WARNING: image duration not set, so time frame definitions will not be initialised
Next shape
===  create template sinogram (DSTE in 3D with max ring diff 2 to save time)
===  running projections to check if result is the same when using preserve-projections
===  create ACFs
ERROR running calculate_attenuation_coefficients. Check my_create_acfs.log
travis_time:end:0bd75f11:start=1565776945713060331,finish=1565776946828686937,duration=1115626606
[0K[31;1mThe command "./run_test_zoom_image.sh" exited with 1.[0m
```

Have relaunched the gcc 7 & 8 travis but since they both threw the same errors, it seems unlikely that this will solve it.

KrisThielemans:
amazingly, most of the travis jobs worked!

the gcc5 one errors with an easy one
```
/home/travis/build/UCL/STIR/src/recon_buildblock/BackProjectorByBin.cxx:183:83: error: ‘stir::BackProjectorByBin::_local_output_image_sptrs’ is not a variable in clause ‘shared’
```
Then there's a somewhat baffling one that we didn't fix this yet (only 1 Travis job sets `STIR_ENABLE_EXPERIMENTAL=ON`)
```
/Users/travis/build/UCL/STIR/src/experimental/recon_buildblock/PostsmoothingForwardProjectorByBin.cxx:137:35: error: no matching member function for call to 'forward_project'

  original_forward_projector_ptr->forward_project(viewgrams, density,
```
So, nearly there!!!

One parallelisation still to do (#363) but it works without it.

rijobro:
I fixed the projector in `STIR_EXPERIMENTAL`, although I'm not sure whether I did it correctly or not. It's tricky because it's a post-smoothing forward projector, which naturally means that the smoothing has to happen after the forward projection. Pre-smoothing forward projection is easy, as the data processor can be used in the `set_input`, and post-smoothing back projection is easy, as the data processor can be used during the `get_output`. So post-smoothing forward projection seems less clear to me. 

I simply did the smoothing at the end of every call to the new `actual_forward_project`. Is that OK?

documentation/release_3.1.htm
```diff
@@ -225,6 +226,7 @@ <h3>Changed functionality</h3>
     trigger or frame times in a series is no longer enforced. This allows 3-D DICOM volumes to be read correctly but may cause
     issues for dynamic PET or CT data.
   </li>
+  <li> As mentioned in the breaking of backwards compatibility, images must be set prior to using the forward and back projectors. To tell the back projector to start projecting into a new image, <tt>reset_output</tt> must be used. The result of the back projection is then returned via <tt>get_output</tt>. This was done so that using data processors (such as motion or smoothing) prior to the forwards projection or following the back projection will be more efficient.</li>
```
KrisThielemans: add something about `set_input` for the forward projector as well. Also, I'd prefer to have some line breaks (don't matter for html, but easier to read on github!)

KrisThielemans:
I'm not convinced about the name `reset_output`. It implies that you only need to run this after you've already done some computations (i.e., not the first one). I guess we could ensure that by letting `set_up()` call `reset_output()` but that might be non-trivial. A derived class will have to call `base::set_up`, but that implies that `reset_output` will be called before the derived class is completely set-up. Hard to know that will always work. Some other generic name?

By the way, `the `reset_output` (or whatever we call it) function needs to be `virtual` such that it can be overridden.

rijobro:
I've got no more changes to make here. Anything else you want changing? If not, good to be squashed I think!

src/include/stir/recon_buildblock/BackProjectorByBin.h
```diff
@@ -111,14 +112,42 @@ class BackProjectorByBin :
 		   const RelatedViewgrams<float>&,
 		   const int min_axial_pos_num, const int max_axial_pos_num,
 		   const int min_tangential_pos_num, const int max_tangential_pos_num);
+#endif
+ //! project whole proj_data into the volume
+ /*! it overwrites the data already present in the volume */
```
KrisThielemans: is this correct? I think it'd be confusing

src/include/stir/recon_buildblock/BackProjectorByBin.h
```diff
@@ -111,14 +112,42 @@ class BackProjectorByBin :
...
+ virtual void back_project(const ProjData&, int subset_num = 0, int num_subsets = 1);
+
+ /*! \brief projects the viewgrams into the volume
```
KrisThielemans: these usage messages need adjusting

src/include/stir/recon_buildblock/BackProjectorByBin.h
```diff
@@ -111,14 +112,42 @@ class BackProjectorByBin :
...
+void back_project(const RelatedViewgrams<float>&,
+          const int min_axial_pos_num, const int max_axial_pos_num);
+
+ /*! \brief projects the specified range of the viewgrams into the volume
+   it adds to the data already present in the volume.*/
+void back_project(const RelatedViewgrams<float>&,
+          const int min_axial_pos_num, const int max_axial_pos_num,
+          const int min_tangential_pos_num, const int max_tangential_pos_num);
 
+ /*! \brief tell the back projector to start accumulating into a new image*/
```
KrisThielemans:
```suggestion
 /*! \brief tell the back projector to start accumulating into a new target
 
 This function has to be called before any back-projection is initiated. 
 */
```
however, this isn't accurate, as it doesn't apply for the old versions (and in particular for the one old function that we kept).

src/include/stir/recon_buildblock/BackProjectorByBin.h
```diff
@@ -111,14 +112,42 @@ class BackProjectorByBin :
...
+ virtual void start_accumulating_in_new_target();
+
+ /// Get output
```
KrisThielemans:
```suggestion
 /*! \brief Get output
 
   This will overwrite the array-content of the argument with the result of all backprojections since calling `start_accumulating_in_new_target()`. Note that the argument has to have the same characteristics as what was used when calling `set_up`.
 */
```

src/include/stir/recon_buildblock/BackProjectorByBin.h
```diff
@@ -111,14 +112,42 @@ class BackProjectorByBin :
...
 
 protected:
 
   virtual void actual_back_project(DiscretisedDensity<3,float>&,
                                    const RelatedViewgrams<float>&,
 		                   const int min_axial_pos_num, const int max_axial_pos_num,
-		                   const int min_tangential_pos_num, const int max_tangential_pos_num) = 0;
+                           const int min_tangential_pos_num, const int max_tangential_pos_num);
+
+ virtual void actual_back_project(const RelatedViewgrams<float>&,
```
KrisThielemans: we need some documentation on what these 2 do and what should be overriden

src/include/stir/recon_buildblock/BackProjectorByBin.h
```diff
@@ -127,11 +156,11 @@ class BackProjectorByBin :
   virtual void check(const ProjDataInfo& proj_data_info, const DiscretisedDensity<3,float>& density_info) const;
   bool _already_set_up;
 
+  //! The density ptr set with set_up()
```
KrisThielemans: Make clear if this is a new object (created with a clone) or not (I hope so, as otherwise the current (non-OpenMP) code would override its content).

src/include/stir/recon_buildblock/BackProjectorByBin.h
```diff
@@ -140,6 +169,9 @@ class BackProjectorByBin :
 	    const int start_tang_pos_num,const int end_tang_pos_num,
 	    const int start_view, const int end_view);
 
+#ifdef STIR_OPENMP
+  std::vector< shared_ptr<DiscretisedDensity<3,float> > > _local_output_image_sptrs;
```
KrisThielemans: add a one liner doc

KrisThielemans:
are you sure about a squash? That sounds terribly dangerous to me as you've merged this in a few other PRs.

src/include/stir/recon_buildblock/BackProjectorByBin.h
rijobro: How about this:

```
  /*! \brief This actually does the back projection.
   There are two versions of this code to enable backwards compatibility.

   This is the newer version (in which the backprojected image is a member variable).
   In most cases, the new version calls the old version (in which the backprojected image is not a member variable).

   If you are developing your own projector, one of these two needs to be overloaded. It doesn't matter which,
   but it might as well be the new one in case we one day decide to remove the old ones.
  */
```

rijobro:
All doc changes made, feel free to give any corrections. 
Squash will be ok as I'll rebase other branches on top.

KrisThielemans:
Regarding squashing, as you know, I prefer to avoid that and rebasing. If people have merged this PR already somewhere, they'd then be stuck with rebasing all the time (pretty dangerous). let's have a chat :-)

rijobro:
Sure, we can discuss. It was just to have a cleaner commit history.

> If people have merged this PR already somewhere, they'd then be stuck with rebasing all the time

My branches won't be much affected and it seems very unlikely that anyone else would have merged the PR in their work without our knowing!

PR #365, opened 2019-08-20
LMrecon basicbin

danieldeidda:
issue with  "find_basic_bin()" as a condition of if statement, within PoissonLog...ForMeanAndListModeData...ByBin.


KrisThielemans:
thanks. Could you please create an Issue as well describing the problem itself (as opposed to the solution). That'll give it a bit more visibility.

Not sure why you have 6 commits here, so I'll squash-merge this later.

KrisThielemans:
closes #366

PR #367, opened 2019-08-21
Data processor projectors

rijobro:
This PR allows a data processor to be used prior to `ForwardProjectorByBin` and after `BackProjectorByBin`. This is now possible because of PR #362.

Currently untested. I commented out two problematic lines of code, which I intend to put back in once I figure out how, e.g.:
```
forw_projector_ptr->parameter_info();
```
i.e.:
```
ForwardProjectorByBin::parameter_info();
```

We could delete `PresmoothingForwardProjectorByBin` and `PostsmoothingBackProjectorByBin`, but in the interest of keeping backwards compatibility I tried to turn them into empty skeletons (calling the new base class functionality).


rijobro:
The ctest works fine on my machine. The only problem being that I use `Utahscat600k_ca_seg4.hs` and get a lot of the following warning: 
```
4: INFO: Processing view 24 of segment 4
4: 
4: WARNING: ray tracing with equal start and end point. Returning zero
```
Is this anything to worry about? If not, is there anyway to not display it (as we don't want too much output for Travis)?

rijobro:
Slightly confused as to how the parser works for creating new objects. From `bcktest.cxx`
https://github.com/UCL/STIR/blob/41cb3f049aa18c85a713f48cbbd963450652daa2/src/recon_test/bcktest.cxx#L232-L236

I added this new key into `BackProjectorByBin`:
```
parser.add_parsing_key("post data processor", &_post_data_processor_sptr);
```
but I'm not sure how to add it into this scheme (so that it's available from `bcktest.cxx`, for example).

Futher, there seems like there's some code duplication; is this not avoidable? (Just for the back projector, it seems to be in `BackProjectorByBin.cxx`, `bcktest.cxx`, `back_project.cxx` and my new `test_data_processor_projectors.cxx`). Similar for forward projectors.

Would something like this be feasible?
```
shared_ptr<BackProjectorByBin> bck_sptr = KeyParser<BackProjectorByBin>::parse_and_create_object(stream);
```

Apologies if I've misunderstood the parser and gone down a rabbit hole on this one.

KrisThielemans:
> WARNING: ray tracing with equal start and end point. Returning zero

This is not something worrying. It can happen if the image is larger than the FOV for instance. STIR wants to do compute an LOR, but then sees it's of length zero effectively.

It cannot currently be disabled. we could redirect stderr of course

KrisThielemans:
Regarding the parsing:

- `post data processor` will now be automatically available. you don't have to do anything anymore. it'd simply go like
```
 Back Projector parameters:=
    type := Matrix
      Back projector Using Matrix Parameters :=
        Matrix type := Ray Tracing
           data processor type := ...
            ...
        End Back Projector Using Matrix Parameters :=
  End:=
```

- yes, there's duplication in the  creation of the `KeyParser`. We can remove that at some point, but could/should do that in a different PR. It'd need a bit of thought what the easiest way is to do this (it would help for data processors etc as well)

KrisThielemans:
gcc7 and gcc8 jobs fail with 1 ctest test_data_processor_projectors, see  https://travis-ci.org/UCL/STIR/jobs/590412364#L4405-L4405
```
Images don't match
Min im1 / im2 = -0.569623 / -0.569623
Max im1 / im2 = 1092.12 / 1092.12
Sum im1 / im2 = 1.73307e+08 / 1.73307e+08
Tests for pre-data-processor forward projection
```
this just seems to be tolerance?

I guess aside from that, this is good to go?

src/recon_test/test_data_processor_projectors.cxx
```diff
@@ -0,0 +1,279 @@
...
+                    << "x-dir filter Metz power:= .0\n"
+                    << "y-dir filter Metz power:= .0\n"
+                    << "z-dir filter Metz power:=.0\n"
+                    << "END Separable Cartesian Metz Filter Parameters :=\n";
+    data_processor_sptr->parse(parameterstream);
+}
+
+Succeeded compare_images(const VoxelsOnCartesianGrid<float> &im_1, const VoxelsOnCartesianGrid<float> &im_2)
+{
+    if (im_1 == im_2) {
```
KrisThielemans: this doesn't allow for numerical error

src/recon_test/test_data_processor_projectors.cxx
```diff
@@ -0,0 +1,279 @@
...
+                    << "y-dir filter FWHM (in mm):= " << fwhm << "\n"
+                    << "z-dir filter FWHM (in mm):= " << fwhm << "\n"
+                    << "x-dir filter Metz power:= .0\n"
+                    << "y-dir filter Metz power:= .0\n"
+                    << "z-dir filter Metz power:=.0\n"
+                    << "END Separable Cartesian Metz Filter Parameters :=\n";
+    data_processor_sptr->parse(parameterstream);
+}
+
+Succeeded compare_images(const VoxelsOnCartesianGrid<float> &im_1, const VoxelsOnCartesianGrid<float> &im_2)
```
KrisThielemans: I'd make these `DiscretisedDensity<3,float>` and make the method explicitly `static` (i.e. local to this file)

src/recon_test/test_data_processor_projectors.cxx
```diff
@@ -0,0 +1,279 @@
...
+
+    std::cout << "\nImages don't match\n";
+    std::cout << "Min im1 / im2 = " << im_1.find_min() << " / " << im_2.find_min() << "\n";
+    std::cout << "Max im1 / im2 = " << im_1.find_max() << " / " << im_2.find_max() << "\n";
+    std::cout << "Sum im1 / im2 = " << im_1.sum()      << " / " << im_2.sum()      << "\n";
+
+    return Succeeded::no;
+}
+
+Succeeded compare_sinos(const ProjDataInMemory &proj_data_1, const ProjDataInMemory &proj_data_2)
```
KrisThielemans: make these `ProjData` (and the function `static`)

rijobro:
Updated according to your review. If travis comes back happy, I'm happy for a squash merge.

src/include/stir/recon_buildblock/BackProjectorByBin.h
```diff
@@ -77,6 +77,13 @@ class BackProjectorByBin :
     const shared_ptr<DiscretisedDensity<3,float> >& density_info_sptr // TODO should be Info only
     ) =0;
 
+  /// Setup whilst applying a post-back-projection data processor
+  void set_up(
+     const shared_ptr<ProjDataInfo>& proj_data_info_ptr,
+     const shared_ptr<DiscretisedDensity<3,float> >& density_info_sptr, // TODO should be Info only
+     shared_ptr<DataProcessor<DiscretisedDensity<3,float> > > post_data_processor_sptr
+     );
```
KrisThielemans: why do we have this? It's not going to be called anywhere. Seems wrong.

Instead, we should have a `set_post_data_processor_sptr()`.

src/include/stir/recon_buildblock/ForwardProjectorByBin.h
```diff
@@ -78,6 +78,13 @@ virtual void set_up(
     const shared_ptr<DiscretisedDensity<3,float> >& density_info_sptr // TODO should be Info only
     ) =0;
 
+  /// Setup whilst applying a pre-forward-projection data processor
+  void set_up(
+    const shared_ptr<ProjDataInfo>& proj_data_info_ptr,
+    const shared_ptr<DiscretisedDensity<3,float> >& density_info_sptr, // TODO should be Info only
+    shared_ptr<DataProcessor<DiscretisedDensity<3,float> > > pre_data_processor_sptr
+    );
```
KrisThielemans: same problem here

rijobro:
ok, will just normal merge. was going to squash for tidier history.
BTW, should we delete `PreSmoothingForwardProjector` and the `PostSmoothingBackProjector`? or add a deprecated tag

KrisThielemans:
I think we deprecated them in the doc? I believe STIR-exercises uses them, so we'd need to change STIR-exercises before we remove them. Please create an issue for it for our next version.

examples/samples/forward_projector_proj_matrix_ray_tracing_pre_smoothing.par
```diff
@@ -0,0 +1,9 @@
+Forward Projector parameters:=
+; example par file for specifying the forward projector for e.g. fwdtest
+type:=Matrix
+  Forward Projector Using Matrix Parameters :=
+  Matrix type := Ray Tracing
+  Ray tracing matrix parameters :=
+  End Ray tracing matrix parameters :=
+  End Forward Projector Using Matrix Parameters :=
```
KrisThielemans: no pre-smoothing keywords...

examples/samples/back_projector_proj_matrix_ray_tracing_post_smoothing.par
```diff
@@ -0,0 +1,20 @@
...
+  
+  type := Matrix
+  Back projector Using Matrix Parameters :=
+    Matrix type := Ray Tracing
+    Ray tracing matrix parameters :=
+    End Ray tracing matrix parameters :=
+  End Back Projector Using Matrix Parameters :=
+
+  Post Data Processor := Separable Cartesian Metz
+  Separable Cartesian Metz Filter Parameters :=
```
KrisThielemans: can we use the Separable Gaussian one instead? Simpler, and no issues with negatives.

documentation/STIR-UsersGuide.tex
```diff
@@ -3500,6 +3500,13 @@ \subsubsection{
 
 See also the online documentation for the class \textbf{DataProcessor}.
 
+As of STIR version 4.0, it is now possible to use data processors 
+on images prior to forward projection and following back projection
+via with any projector. This supercedes functionality provided by
```
KrisThielemans: delete the 'via'

KrisThielemans: also refer to the examples such that you don't have to repeat it in this file

PR #368, opened 2019-08-22
Object/ParsingObject/RegisteredObject hierarchy

rijobro:
1. `Object` inherits from `ParsingObject`
2. `Object` no longer needs the method `parameter_info()`
3. Classes that inherit from `RegisteredObject` no longer need to inherit from `ParsingObject`, too
4. Modify the default 3rd argument in `RegisteredParsingObject`

The old naming scheme was then misleading. The class, `Object` only exists so that the parser can get the registered name of the class. This is therefore a non-templated version of the `RegisteredObject`, so is now called `RegisteredObjectBase`.

### DIFF
--- a/src/include/stir/RegisteredObjectBase.h
+++ b/src/include/stir/RegisteredObjectBase.h
@@ -28,6 +28,7 @@
 */
 
 #include "stir/common.h"
+#include "stir/ParsingObject.h"
 
 #include <string>
 
@@ -45,20 +46,11 @@ START_NAMESPACE_STIR
   for parsing..
 */
 
-class Object
+class Object : public ParsingObject
 {
 public:
   virtual ~Object() {}
-  /*! \brief return a string describing all parameters of the object
 
-  There is currently no requirement on the format of the returned string.
-  Ideally it would be such that the object can be constructed by parsing
-  the string again.
-
-  \todo Ideally this function would be a const member, but we cannot
-  do this because ParsingObject::parameter_info is not const.
-  */
-  virtual std::string parameter_info()  = 0;
   /*! \brief Returns the name of the type of the object.
 
   Each type that can be parsed has a unique (within its hierarchy) name
--- a/src/include/stir/RegisteredObjectBase.h
+++ b/src/include/stir/RegisteredObjectBase.h
@@ -4,7 +4,7 @@
 
   \file
 
-  \brief Declaration of class stir::Object
+  \brief Declaration of class stir::RegisteredObjectBase
 
   \author Kris Thielemans
   
@@ -27,29 +27,28 @@
     See STIR/LICENSE.txt for details
 */
 
-#include "stir/common.h"
 #include "stir/ParsingObject.h"
 
 #include <string>
 
-#ifndef __stir_Object_H__
-#define __stir_Object_H__
+#ifndef __stir_RegisteredObjectBase_H__
+#define __stir_RegisteredObjectBase_H__
 
 START_NAMESPACE_STIR
 
 /*! \brief Base class for all classes that can parse .par files (and more?) 
   \ingroup buildblock
-  The main reason that this class exists is such that KeyParser can store
+  The only reason that this class exists is such that KeyParser can store
   different types of objects, and get some basic info from it.
 
   \see RegisteredObject for more info on the registries etc that are used
   for parsing..
 */
 
-class Object : public ParsingObject
+class RegisteredObjectBase : public ParsingObject
 {
 public:
-  virtual ~Object() {}
+  virtual ~RegisteredObjectBase() {}
 
   /*! \brief Returns the name of the type of the object.
 
--- a/src/include/stir/RegisteredParsingObject.h
+++ b/src/include/stir/RegisteredParsingObject.h
@@ -97,7 +97,7 @@ class AddParser : public Base, public ParsingObject
   the variables in that file are never referenced, so would not include
   it in the final executable (to try to remove redundant object files).
 */
-template <typename Derived, typename Base, typename Parent = AddParser<Base> >
+template <typename Derived, typename Base, typename Parent = Base>
 class RegisteredParsingObject : public Parent
 {
 public:
--- a/src/include/stir/RegisteredParsingObject.h
+++ b/src/include/stir/RegisteredParsingObject.h
@@ -34,21 +34,6 @@
 #include <string>
 
 START_NAMESPACE_STIR
-
-
-//! Auxiliary class for RegisteredParsingObject
-/*!
-  \ingroup buildblock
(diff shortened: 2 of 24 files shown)


src/include/stir/RegisteredObjectBase.h
```diff
@@ -27,38 +27,29 @@
...
 
   \see RegisteredObject for more info on the registries etc that are used
   for parsing..
 */
 
-class Object
+class RegisteredObjectBase : public ParsingObject
 {
 public:
-  virtual ~Object() {}
```
KrisThielemans: does `ParsingObject` have a virtual destructor? It should now.

src/include/stir/RegisteredParsingObject.h
```diff
@@ -97,7 +97,7 @@ class AddParser : public Base, public ParsingObject
   the variables in that file are never referenced, so would not include
   it in the final executable (to try to remove redundant object files).
 */
-template <typename Derived, typename Base, typename Parent = AddParser<Base> >
+template <typename Derived, typename Base, typename Parent = Base>
```
KrisThielemans: we should get rid of the AddParser class really. It's dangerous to have it now.

rijobro:
Travis fails because of changes required to the python code. Problematic part:
```
%shared_ptr(  stir::AddParser<stir::ForwardProjectorByBin>);
%template (internalAddParserForwardProjectorByBin)
  stir::AddParser<stir::ForwardProjectorByBin>;
%template (internalRPForwardProjectorByBinUsingProjMatrixByBin)  
  stir::RegisteredParsingObject<stir::ForwardProjectorByBinUsingProjMatrixByBin,
     stir::ForwardProjectorByBin>;
%include "stir/recon_buildblock/ForwardProjectorByBinUsingProjMatrixByBin.h"

%shared_ptr(  stir::AddParser<stir::BackProjectorByBin>);
%template (internalAddParserBackProjectorByBin)
  stir::AddParser<stir::BackProjectorByBin>;
%template (internalRPBackProjectorByBinUsingProjMatrixByBin)  
  stir::RegisteredParsingObject<stir::BackProjectorByBinUsingProjMatrixByBin,
     stir::BackProjectorByBin>;
%include "stir/recon_buildblock/BackProjectorByBinUsingProjMatrixByBin.h"
```

src/include/stir/RegisteredObjectBase.h
rijobro: Already virtual

rijobro:
Is this the doxygen in `RegisteredObject` that you want changing? Not sure how to word it clearly: 
```
  In case the construction of the object is done by using ParsingObject,
  nearly all of the necessary functionality can be provided to the
  hierarchy by using RegisteredParsingObject in the hierarchy. In such 
  a case, the hierarchy looks normally as follows:
  \code
  RegisteredObject<Base>
  Base
  ...
  Parent   <- ParsingObject
  RegisteredParsingObject<Derived,Base,Parent>
  Derived
  \endcode

  A <strong>recommended</strong> variation on this is the following:
  \code
  RegisteredObject<Base>
  Base <- ParsingObject
  ...
  RegisteredParsingObject<Derived,Base,Base>
  Derived
  \endcode
  Aside from the fact that this is simpler, it also is more future proof.
  Suppose that at some point you want to add parsing keys that are common to 
  the whole hierarchy. The best/only place to do this is to add them to 
  the \a Base class. But this can only be done if \a Base is derived from 
  ParsingObject, and hence there is can be no other ParsingObject in the 
  hierarchy. So, the recommended variation will need no change at all, while
  the general version would need change to all nodes.
```

KrisThielemans:
SWIG code: just get rid of the `AddParser` bits. The template there should match exactly what's in the .h files, but I think it does.

rijobro:
Gave it a go in the most recent commit, but it doesn't compile. Can you be more specific about what needs to change? The SWIG code is a bit of a black box to me.

src/swig/stir.i
```diff
@@ -1657,17 +1657,17 @@ namespace stir {
 
 %include "stir/recon_buildblock/ProjMatrixByBinUsingRayTracing.h"
 
-%shared_ptr(  stir::AddParser<stir::ForwardProjectorByBin>);
+%shared_ptr(  stir::ForwardProjectorByBin);
 %template (internalAddParserForwardProjectorByBin)
-  stir::AddParser<stir::ForwardProjectorByBin>;
+  stir::ForwardProjectorByBin;
```
KrisThielemans: no `%template` (`ForwardProjectorByBin` isn't one). delete the whole 2 lines

src/swig/stir.i
```diff
@@ -1657,17 +1657,17 @@ namespace stir {
...
 %template (internalRPForwardProjectorByBinUsingProjMatrixByBin)  
   stir::RegisteredParsingObject<stir::ForwardProjectorByBinUsingProjMatrixByBin,
      stir::ForwardProjectorByBin>;
 %include "stir/recon_buildblock/ForwardProjectorByBinUsingProjMatrixByBin.h"
 
-%shared_ptr(  stir::AddParser<stir::BackProjectorByBin>);
+%shared_ptr(  stir::BackProjectorByBin);
 %template (internalAddParserBackProjectorByBin)
-  stir::AddParser<stir::BackProjectorByBin>;
+  stir::BackProjectorByBin;
```
KrisThielemans: as above

src/swig/stir.i
```diff
@@ -1657,17 +1657,17 @@ namespace stir {
 
 %include "stir/recon_buildblock/ProjMatrixByBinUsingRayTracing.h"
 
-%shared_ptr(  stir::AddParser<stir::ForwardProjectorByBin>);
+%shared_ptr(  stir::ForwardProjectorByBin);
```
KrisThielemans: this line is already above. just delete the lines that had `AddParser`

KrisThielemans:
For the doxygen, all mentions of `ParsingObject` are now incorrect obviously. Maybe
```
  We currently assume that the construction of the object is done by using ParsingObject.
  Nearly all of the necessary functionality can be provided to the
  hierarchy by using RegisteredParsingObject in the hierarchy. 
  The hierarchy looks normally as follows:
  \code
  RegisteredObject<Base>
  Base
  ...
  Parent
  RegisteredParsingObject<Derived,Base,Parent>
  Derived
  \endcode

  When there is no intermediate class in hierarchy, this is simplified to:
  \code
  RegisteredObject<Base>
  Base
  RegisteredParsingObject<Derived,Base,Base>
  Derived
  \endcode
```

KrisThielemans:
I've squashed this after all. @rijobro, you will therefore need to rebase your other recent PRs. maybe try and cross-merge different PRs not so quickly :-)

PR #376, opened 2019-09-03
List roi

danieldeidda:
list_ROI_values save image name in a new column


documentation/release_3.1.htm
```diff
@@ -343,14 +343,18 @@ <h3>Major bugs fixed</h3>
 
 <h3>Backward incompatibities</h3>
 <ul>
-  <li>STIR now uses <code>std::unique_ptr</code> as opposed to <code>std::auto_ptr</code>, as recommended for C++-11 and 
+
+<li> <code>list_ROI_values</code> now saves the image name as a new column, and the image name will appear in the same 
+    row as the roi name, mean value and son on.
```
KrisThielemans:
```suggestion
    row as the roi name, mean value and so on. 
```

documentation/release_3.1.htm
```diff
@@ -343,14 +343,18 @@ <h3>Major bugs fixed</h3>
 
 <h3>Backward incompatibities</h3>
 <ul>
-  <li>STIR now uses <code>std::unique_ptr</code> as opposed to <code>std::auto_ptr</code>, as recommended for C++-11 and 
+
+<li> <code>list_ROI_values</code> now saves the image name as a new column, and the image name will appear in the same
```
KrisThielemans:
```suggestion
<li> <tt>list_ROI_values</tt> now saves the image name as a new column, and the image name will appear in the same 
```

src/utilities/list_ROI_values.cxx
```diff
@@ -261,7 +261,8 @@ main(int argc, char *argv[])
 
 	for (int i=min_plane_number;i<=max_plane_number;i++)
 	  {
-	    out << std::setw(15) << *current_name_iter
+        out << std::setw(15) <<input_file
```
KrisThielemans: white spaces seem a bit different. ideally don't use tab, but just spaces and make it align then (but it could be because of existing tabs)

KrisThielemans:
by the way, your commit history is quite complicated. i'll likely squash-rebase this PR. Therefore best to not merge this PR anywhere else.

src/utilities/list_ROI_values.cxx
```diff
@@ -233,13 +233,15 @@ main(int argc, char *argv[])
   if (!is_null_ptr(parameters.filter_ptr))
     parameters.filter_ptr->apply(*image_ptr);
 
-  out << input_file << '\n';
+  out << std::setw(15) << 'ImageName';
```
KrisThielemans:
```suggestion
  out << std::setw(15) << "ImageName";
```

danieldeidda:
> other issues are still outstanding. Don't forget to make this optional (in the code, and in the documentation).

Sorry I did not notice the previous comments. will do all the changes

danieldeidda:
> by the way, your commit history is quite complicated. i'll likely squash-rebase this PR. Therefore best to not merge this PR anywhere else.

Yes, I don't understand what is going on, I can still see that listmode objective function.

danieldeidda:
> thanks! This has resolved the `recon_test_pack` problems as well. One Travis job failure due to networking issues, nothing with your code.
> 
> However, the command line processing seems overly complicated. I'd expect something like this
> 
> ```c++
> while (argc>1 && strncmp(argv[1],"--",2)==0)
> {
>    if (strcmp(argv[1], "--max") == 0)
>      do_max = true;
>    else if (strcmp(argv[1], "--CV") == 0)
>    ...
>    else
>       error(boost::format("Unknown option %s") % argv[1]);
>   --argc; ++argv;
>  }
> ```
> 
> what do you think?

sure, I agree.

PR #380, opened 2019-09-25
Fix to distributable to suppress OpenMP cerr output

robbietuk:
Output was created by a std::cerr which was causing spam output when OpenMP functions are called multiple times. This change will allow verbosity levels to suppress this output.

### DIFF
--- a/src/recon_buildblock/distributable.cxx
+++ b/src/recon_buildblock/distributable.cxx
@@ -423,7 +423,7 @@ void distributable_computation(
 #ifdef STIR_OPENMP
 #pragma omp single
     {
-      info("Starting loop with " + std::to_string(omp_get_num_threads()) + " threads");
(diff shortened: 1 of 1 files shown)


src/recon_buildblock/distributable.cxx
```diff
@@ -423,7 +423,7 @@ void distributable_computation(
 #ifdef STIR_OPENMP
 #pragma omp single
     {
-      std::cerr << "Starting loop with " << omp_get_num_threads() << " threads\n"; 
+      info("Starting loop with " + std::to_string(omp_get_num_threads()) + " threads");
```
KrisThielemans: sorry. will fail as `to_string` is C++-11 onwards, and we currently still support older C++.
Please use `boost::format` as in other places

PR #385, opened 2019-10-16
Fixes LM compute_sub_gradient by reseting gradient_ptr upon call

### LINKED ISSUE #384: LM compute_sub_gradient gradient does not overwrite gradient
Calling `compute_sub_gradient(TargetT& gradient, const TargetT &current_estimate, const int subset_num)` when performing list mode reconstruction will not reset the gradient ptr to zero before computation of the objective function gradient. 

If the gradient-ptr uniform non-zero, the gradient will be added to the non-zero input gradient_ptr.

[GeneralisedObjectiveFunction.h](https://github.com/UCL/STIR/blob/a68340aaf3e78ffd1235c91027042b036f3084cc/src/include/stir/recon_buildblock/GeneralisedObjectiveFunction.h#L108-L123) states that the data should be overwritten, which is true for sinogram reconstructions.

robbietuk:
Fixes #384 

Also corrects documentation

### DIFF
--- a/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.h
+++ b/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.h
@@ -74,7 +74,7 @@ public:
   
   PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin<TargetT>(); 
 
-  //! This should compute the gradient of the objective function at the  current_image_estimate
+  //! This should compute the gradient of the objective function at the current_image_estimate overwritting gradient
   virtual  
   void compute_sub_gradient_without_penalty_plus_sensitivity(TargetT& gradient,  
                                          const TargetT &current_estimate,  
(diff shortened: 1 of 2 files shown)


src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.h
```diff
@@ -74,7 +74,7 @@ typedef RegisteredParsingObject<PoissonLogLikelihoodWithLinearModelForMeanAndLis
   
   PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin<TargetT>(); 
 
-  //! This should compute the gradient of the objective function at the  current_image_estimate
+  //! This should compute the gradient of the objective function at the current_image_estimate overwritting gradient
```
KrisThielemans:
```suggestion
  //! This should compute the gradient of the objective function at the \a current_estimate overwriting \a gradient
```

rijobro:
@robbietuk can you have a look at this?

robbietuk:
@rijobro @KrisThielemans Commited the suggested change.

Issue #395, opened 2019-10-19
STIRConfig.cmake contains local value for ITK_DIR


KrisThielemans:
We attempt to make sure we build with the same ITK here
https://github.com/UCL/STIR/blob/730f21e1655f83ca50cb2a0d6acc1d0bc63e22c3/src/cmake/STIRConfig.cmake.in#L51-L52
but this means that STIR gets installed with a hard-wired location of ITK. This will likely fail with conda etc. Better to just fix the version?

KrisThielemans:
Also the case here https://github.com/UCL/STIR/blob/730f21e1655f83ca50cb2a0d6acc1d0bc63e22c3/src/cmake/STIRConfig.cmake.in#L61

Not 100% sure if this is a problem or not for conda. @paskino, do you know? (not urgent though)

KrisThielemans:
conda takes care of this. I believe we use relocatable packages recommendations, so closing.

PR #398, opened 2019-10-19
install source files for STIR registries

### LINKED ISSUE #211: STIR registries files are not installed (STIR_REGISTRIES points to original sources)
`STIR_REGISTRIES` is a CMake variable that lists all .cxx files that add items to registries, e.g. for IO, see the _Registries of classes_  section in the STIR developers guide.

At present, we do not install these files. However, anyone who uses an installed version of STIR will need them for IO, filtering etc to work. Therefore, at present, an **installed version of STIR will only work on the original developer's machine** (if she doesn't delete the source files).

Installing source files is "not done" however. Where would we put them?

If we'd install them, we'll have to adjust the variable in [STIRConfig.cmake.in](https://github.com/UCL/STIR/blob/4e0a2e89f2dfda87106cb13d14081b70d50dc0d4/src/cmake/STIRConfig.cmake.in#L39)

KrisThielemans:
Fixes #211

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,7 @@ elseif(CYGWIN)
 else()
   set(STIR_DOC_DIR "doc/stir")
 endif()
+set(STIR_DATA_DIR "share/stir")
 set(ConfigPackageLocation lib/cmake/)
 
 ####### External packages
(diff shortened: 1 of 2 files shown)


KrisThielemans:
@paskino this should fix one serious conda problem. There's still #395 to do.

PR #399, opened 2019-10-20
Use ITK_USE_FILE as recommended by ITK

KrisThielemans:


KrisThielemans:
@rijobro, maybe you can try this on OSX. you'd have to see if SIRF is happy with this as STIR doesn't have tests for STIRConfig. not urgent

rijobro:
STIR and SIRF compile with no issues on OSX.

Issue #393, opened 2019-10-19
OSSPS test fails, but silently 


KrisThielemans:
I spotted the following in https://travis-ci.org/UCL/STIR/jobs/599837970
```
$ ./run_test_simulate_and_recon.sh
...
=============================================
Using /home/travis/build/UCL/STIR/install/bin/OSSPS
=============================================
Running OSSPS OSSPS_test_sim_PM_QPweights.par
Input ROI mean: 1
Output ROI mean: 2.84969
DIFFERENCE IN ROI VALUES IS TOO LARGE. CHECK RECONSTRUCTION LOG my_OSSPS_test_sim_PM_QPweights.par.log
=============================================
There were errors. Check  my_OSSPS_test_sim_PM_QPweights.par.log
...
The command "./run_test_simulate_and_recon.sh" exited with 0.
```
2 problems of course:
- the test fails 
- the failure isn't reflected in the return code of `run_test_simulate_and_recon.sh`

I have not investigated since when this is happening. @rijobro maybe you could have a look?

rijobro:
Just checked through master builds on travis and it looks like the first time OSSPS fails is due to PR #362: https://travis-ci.org/UCL/STIR/builds/574564312

rijobro:
I've had a look but I can't see anything that's immediately obvious. I ran the test in debug mode, no errors there either. Presumably, there's a missing `start_accumulating_in_new_target` somewhere or something like that?

KrisThielemans:
The `run_tests.sh` failed on OSSPS as well. Travis didn't know.

KrisThielemans:
seems all `run*.sh` should have `exit 1` when the tests failed. They just return 0 anyway!

KrisThielemans:
Found the problem with OSSPS. The `PoissonLogLikelihoodWithLinearModelForMeanAndProjData<TargetT>::
actual_add_multiplication_with_approximate_sub_Hessian_without_penalty` function is supposed to accumulate the output (that's what the name says), but it overwrites what was there.

PR #401, opened 2019-10-21
return success value of bash scripts

rijobro:
fixes some of #393 (tests failing but travis happy).

The relevant bit of the `.travis.yml` currently looks like this:

```
  - ./run_test_simulate_and_recon.sh
  - ./run_test_listmode_recon.sh
  - ./run_test_simulate_and_recon_with_motion.sh
  - ./run_scatter_tests.sh
  - ./run_tests.sh --nointbp
  - ./run_test_zoom_image.sh
```
@KrisThielemans - would you rather:
```
rtn_run_test_simulate_and_recon=$(./run_test_simulate_and_recon.sh)
rtn_run_test_listmode_recon=$(./run_test_listmode_recon.sh)
...
```
In this way we can run all tests and then check the result, with e.g.:
```
rtn_all_tests=$(( rtn_run_test_simulate_and_recon * rtn_run_test_listmode_recon * ... ))
if [ "$rtn_all_tests" = false ] ; then
    exit 1 or print message, not sure what you're supposed to do with Travis
fi
```

### DIFF
--- a/recon_test_pack/SPECT/run_SPECT_tests.sh
+++ b/recon_test_pack/SPECT/run_SPECT_tests.sh
@@ -150,7 +150,9 @@ echo "============================================="
 if [ -z "${error_log_files}" ]; then
  echo "All tests OK!"
  echo "You can remove all output using \"rm -rf out\""
+ exit 0
 else
  echo "There were errors. Check ${error_log_files}"
+ exit 1
 fi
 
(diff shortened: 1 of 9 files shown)


KrisThielemans:
I think it's best to let all tests run. It gives more diagnostics, although I adiit it's sometimes hard to find which one failed. Anyway, let's leave that for another day.

Issue #292, opened 2018-11-14
replace openmp linker flag with a target


KrisThielemans:
as opposed to use the CXX_FLAGS for linking, [here](https://github.com/UCL/STIR/blob/441bb1f9bb92d020d30faf278d1c20ff169a3c83/src/CMakeLists.txt#L149), we should use an OpenMP target when available, see https://gitlab.kitware.com/cmake/cmake/issues/15392

PR #402, opened 2019-10-21
OSX OpenMP

rijobro:
The following seems to work:
```
brew install libomp
brew install llvm
mkdir build-omp
cd build-omp
export PATH="/usr/local/opt/llvm/bin:$PATH"
cmake ../Source \
	-DCMAKE_C_COMPILER=/usr/local/opt/llvm/bin/clang \
	-DCMAKE_CXX_COMPILER=/usr/local/opt/llvm/bin/clang++ \
	-DOPENMP_LIBRARIES=/usr/local/opt/llvm/lib \
	-DOPENMP_INCLUDES=/usr/local/opt/llvm/include \
	-DSTIR_OPENMP=ON
```
@ludovicabrusaferri @robbietuk could one of you try this? Mine seems to be working with 3 threads (as I have a 4-core machine). Remember, it needs to be in a fresh build directory.


KrisThielemans:
better to do #292 first...

KrisThielemans:
have a look at #403. It should simplify this (at least for recent CMake that knows about clang)

rijobro:
#403 changes behaviour after the `find_package(OpenMP)`. However, it appears that variables are needed before the `find_package` (as per this PR, taken from here: https://stackoverflow.com/a/48216682).

src/CMakeLists.txt
```diff
@@ -144,11 +144,50 @@ if (STIR_MPI)
...
   find_package(OpenMP REQUIRED)  
+
   add_definitions(${OpenMP_CXX_FLAGS})
-  # need to explicitly set his. Definitely for gcc, hopefully also for other systems.
-  # See https://gitlab.kitware.com/cmake/cmake/issues/15392
-  set (OpenMP_EXE_LINKER_FLAGS ${OpenMP_CXX_FLAGS})
+  
+  if (APPLE)
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
```
KrisThielemans: done already in the `add_definitions` line above?

src/CMakeLists.txt
```diff
@@ -144,11 +144,50 @@ if (STIR_MPI)
...
+    set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
```
KrisThielemans: this line should be covered by the old stuff

KrisThielemans:
oh well. This is all going to be CMake version dependent. They're still fixing it now https://github.com/Kitware/CMake/commits/master/Modules/FindOpenMP.cmake

I hope that #403 will allow to simplify this a bit for the linker stuff.

robbietuk:
I ran Richards script above and got the following error. My STIR is master from a few days ago - do I need a newer version

```
CMake Error at /usr/local/Cellar/cmake/3.13.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
   Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES)
 Call Stack (most recent call first):
   /usr/local/Cellar/cmake/3.13.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
   /usr/local/Cellar/cmake/3.13.4/share/cmake/Modules/FindOpenMP.cmake:473 (find_package_handle_standard_args)
   src/CMakeLists.txt:147 (find_package)
```

N.B.? llvm is dependant on swig - tries to: `brew installs swig 4.0.1`. I use STIR with a specific swig (3.0.11) for python that is not overwritten by brew.

rijobro:
Thanks for trying this. Can you `cd` to your build directory and run the two following commands for me?
```
cat CMakeCache.txt | grep llvm
```
and
```
cat CMakeCache.txt | grep openmp
```

rijobro:
Also, did you just run the above script, or did you also check out my branch?

robbietuk:
If I checkout your branch it works. Sorry..I hadnt had my coffee yet.

robbietuk:
Fails to compile when building STIR with root and OpenMP using the above method. STIR detects ROOT and sets CXX FLAG to CXX11. 

STIR produces the error:
```
/usr/local/Cellar/root/6.16.00/include/root/TInterpreter.h:257:85: error: no viable conversion from returned value of type 'int' to function return type 'std::auto_ptr<TInterpreterValue>'
   virtual std::unique_ptr<TInterpreterValue> MakeInterpreterValue() const { return 0; }
```

PR #403, opened 2019-10-21
use the OpenMP::OpenMP_CXX linking target for cmake >= 3.9

### LINKED ISSUE #292: replace openmp linker flag with a target

KrisThielemans:
closes #292


rijobro:
Think you've got the inequality the wrong way round.

src/CMakeLists.txt
```diff
@@ -146,9 +146,14 @@ endif()
 if(STIR_OPENMP)
   find_package(OpenMP REQUIRED)  
   add_definitions(${OpenMP_CXX_FLAGS})
-  # need to explicitly set his. Definitely for gcc, hopefully also for other systems.
-  # See https://gitlab.kitware.com/cmake/cmake/issues/15392
-  set (OpenMP_EXE_LINKER_FLAGS ${OpenMP_CXX_FLAGS})
+  if(${CMAKE_VERSION} VERSION_LESS "3.9.0")
```
rijobro:
```suggestion
  if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.9.0") 
```

KrisThielemans:
oops... sorry!

PR #411, opened 2019-10-29
set verbosity of reconstruction via method

rijobro:
needed for SIRF


KrisThielemans:
I find it confusing that `stir.Verbosity` is a global variable, which we'd set in a class, giving the impression it's only for that object.

we had to put it in the par file, to allow people to set it by parsing, but I hesitate to make it part of the class interface

rijobro:
ok, closing this. 
For STIR, I guess we didn't need to put it inside of the reconstructor, we could have put it inside of `ParsingObject`, so that it is as global as possible.

PR #413, opened 2019-11-13
OSX OpenMP STIRConfig 

rijobro:
We recently helped out OSX users installing STIR with OpenMP. To link against STIR (e.g., SIRF), the same changes are required in `STIRConfig.cmake`.


KrisThielemans:
hmmm. this doesn't like very nice. I think we should set these variables from whatever they were during build time, although I'm not 100% sure. See also #395

PR #44, opened 2016-09-12
Scatter simulation

NikEfth:
So, this is the PR for the scatter simulation. 

New classes ScatterSimulation and SingleScatterSimulation which have taken over the old ScatterEstimation. This is their "minimal" implementation having only the functionality that the older class used to have. I removed, for the time being, the subsampling stuff, in order to first see why/where there is the mistake. Once this has been sorted I will start adding more things bit-by-bit. 

Please don't review the new ScatterEstimation class, yet. I will create a branch, from this one, on which I will work on the ScatterEstimation and let you know when it will be ready for reviewing.

In addition, there are some minor changes in other places, like clone functions in ExamInfo. 
In my laptop all tests succeeded and Travis is doing good, so far. 

ATB, 
Nikos

PR #418, opened 2019-11-16
Fix potential race condition in scatter simulation

### LINKED ISSUE #417: scatter openmp not threadsafe (?)
I believe this line is not thread-safe
https://github.com/UCL/STIR/blob/496674df5a71eef998dd6c07a5e74973f34218e3/src/scatter_buildblock/ScatterEstimationByBin.cxx#L399-L400

(tentative explanation: the processor might be using more bytes for access, i.e. a write operation for a 32bit float might actually get a 64bit, modify the last 32, and then store the whole thing again).

Any multi-threading experts?

KrisThielemans:
on some architectures, assigning to arrays is not threadsafe. so use an `atomic write`.

fixes #417


KrisThielemans:
@NikEfth you will have to merge `master` to #44. not sure if `git` will do the proper thing though with all the renaming. I can check afterwards.

PR #420, opened 2019-11-18
Revert "OSX OpenMP STIRConfig "

KrisThielemans:
Reverts UCL/STIR#413

still work-in-progress by@rijobro


rijobro:
I think this might actually be fine. The problem is that we do the `find_package(STIR)` in `src/xSTIR/`, which means that the other libraries that link against `cSTIR` (e.g., the synergistic stuff) don't know anything about the OpenMP variables. 

Any thoughts as to what we can do?

An obvious one is to put the `find_package(STIR)` at a root level, but that seems less tidy.

Issue #416, opened 2019-11-15
[ENH] Feature Request: KEM with no anatomical prior


ashgillman:
i.e., just the "H" part of HKEM.

CC: @harrymarquis @danieldeidda

Issue #415, opened 2019-11-15
HKEM with a constant anatomical prior causes a NaN image


ashgillman:
Setting the anatomical image to a uniform image (to get no anatomical information) causes the image to be completely filled with NaN values after a single iteration.

@danieldeidda , is the image divided by the standard deviation for normalisation by any chance?

ashgillman:
@harrymarquis

danieldeidda:
> Setting the anatomical image to a uniform image (to get no anatomical information) causes the image to be completely filled with NaN values after a single iteration.
> 
> @danieldeidda , is the image divided by the standard deviation for normalisation by any chance?

Yeah! I didn't think about it. Myabe is worth adding a condition like only_functional
if true set Km=1

danieldeidda:
> > Setting the anatomical image to a uniform image (to get no anatomical information) causes the image to be completely filled with NaN values after a single iteration.
> > @danieldeidda , is the image divided by the standard deviation for normalisation by any chance?
> 
> Yeah! I didn't think about it. Maybe is worth adding a condition like only_functional
> if true set Km=1

or actually even easier we can set the SD=1

markus-jehl:
This looks like it can be closed.

KrisThielemans:
@danieldeidda ?
@markus-jehl did you test it?

danieldeidda:
yes this  was fixed

PR #422, opened 2019-11-19
Only functional hkem

danieldeidda:
solved issue #415 and #416: allowing HKEM to only use the iterative information. an example par file was added for PET and SPECT.
to only use the iterative information the user needs to:

    give the anatomical image (it will be filled with all zeros), or an empty image with the right dimensions and voxel sizes.
    set hybrid:=1
    set only_iterative_info:=1
    @ashgillman @HarryMarquis

### DIFF
--- a/examples/samples/KOSMAPOSL_hkem_spect.par
+++ b/examples/samples/KOSMAPOSL_hkem_spect.par
@@ -6,7 +6,6 @@ KOSMAPOSLParameters :=
 ;the following disables the alpha coefficient output:
 disable output :=1
 
-
 ; here we have the possibility to choose the parameters which define the kernel
 ; matrix and the name of the anatomical image. the following are the defaults values:
 
@@ -14,7 +13,6 @@ disable output :=1
 ; OR
 ; 0 kernel is MR-only
 hybrid:=1
-
 
 ; The following should only be set to 1 if hybrid is set to 1. You still need to load your anatomical
 ; image or an empty image with the right dimensions and voxel size 
@@ -49,10 +47,10 @@ number of neighbours:= 3
 number of non-zero feature elements:=1
 
 ; this are the file names of the anatomical images (at the moment we can accept only one image)
-anatomical image filenames:= {MRbrain.hv}
+anatomical image filenames:= {PET_CL02_003_4hr_head.hv}
 
 ; the following should be 1 if you want to reconstruct 2D data
-only_2D:=1
+only_2D:=0
 
 ; the following is the output prefix of the PET reconstructed image
 kernelised output filename prefix :=KOSMAPOSL
@@ -60,31 +58,26 @@ kernelised output filename prefix :=KOSMAPOSL
 objective function type:= PoissonLogLikelihoodWithLinearModelForMeanAndProjData
 PoissonLogLikelihoodWithLinearModelForMeanAndProjData Parameters:=
 
-input file := my_noisy_prompts_1ax_pos.hs
+input file := input_CL02_003_4hr_head_hkem.hs
 
-; Daniel: here we have the possibility to choose the parameters which define the kernel matrix and the name of the MR image.
-
+;SPECT:
 projector pair type := Matrix
 Projector Pair Using Matrix Parameters :=
-Matrix type := Ray Tracing
-Ray tracing matrix parameters :=
-number of rays in tangential direction to trace for each bin := 10
-do symmetry 90degrees min phi := 1
-do_symmetry_180degrees_min_phi:= 1
-do_symmetry_swap_s:= 1
-do_symmetry_swap_segment:= 1
-do_symmetry_shift_z:= 1
-End Ray tracing matrix parameters :=
+Matrix type := SPECT UB
+Projection Matrix By Bin SPECT UB Parameters:=
+maximum number of sigmas:= 2.0
+psf type:= Geometrical
+psf type:= Geometrical
+attenuation type := Simple
+attenuation map := ${umap}
+mask type := No
+;mask file := mask.hv
+End Projection Matrix By Bin SPECT UB Parameters:=
 End Projector Pair Using Matrix Parameters :=
+;SPECT END
 
 ; additive projection data to handle randoms etc
 additive sinogram := my_additive_sinogram_1ax_pos.hs 
-
-; norm and acf
-Bin Normalisation type := From ProjData
-Bin Normalisation From ProjData :=
-normalisation projdata filename:= my_multfactors_1ax_pos.hs
-End Bin Normalisation From ProjData:=
 
 ; if the next parameters are enabled, 
 ; the sensitivity will be computed and saved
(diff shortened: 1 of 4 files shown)


src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -156,6 +156,8 @@ set_defaults()
   this->only_2D = 0;
   this->kernelised_output_filename_prefix="";
   this->hybrid=0;
+  this->only_iterative_info=0;
```
ashgillman: I think that instead of `only iterative info`, this could be supported if `hybrid := false` and `anatomical image filename := {}`.
This could be implemented at the loop iterating the anatomical images, but requires your PR to support multiple anatomical images first.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -156,6 +156,8 @@ set_defaults()
...
+  this->_disable_output=1;
```
ashgillman: Is this a bugfix or new feature? Should be highlighted separately in CHANGES.md?

examples/samples/KOSMAPOSL_hkem_spect.par
```diff
@@ -0,0 +1,108 @@
+KOSMAPOSLParameters :=
```
ashgillman: Should be highlighted differently in CHANGES.md (i.e., added HKEM SPECT example parameter file)

examples/samples/KOSMAPOSL_hkem_spect.par
```diff
@@ -0,0 +1,108 @@
...
+; Number of neigbouring voxels to compare: (num_neighbours X num_neighbours X num_neighbours)
+; default: 3
+number of neighbours:= 3
+
+; Number of non-zero elements in the feature vectors
+; default: 1
+number of non-zero feature elements:=1
+
+; this are the file names of the anatomical images (at the moment we can accept only one image)
+anatomical image filenames:= {PET_CL02_003_4hr_head.hv}
```
ashgillman: `{${ANATOMICAL}}`

examples/samples/KOSMAPOSL_hkem_spect.par
```diff
@@ -0,0 +1,108 @@
...
+; the following should be 1 if you want to reconstruct 2D data
+only_2D:=0
+
+; the following is the output prefix of the PET reconstructed image
+kernelised output filename prefix :=KOSMAPOSL
+
+objective function type:= PoissonLogLikelihoodWithLinearModelForMeanAndProjData
+PoissonLogLikelihoodWithLinearModelForMeanAndProjData Parameters:=
+
+input file := input_CL02_003_4hr_head_hkem.hs
```
ashgillman: `${INPUT}`

examples/samples/KOSMAPOSL_hkem_spect.par
```diff
@@ -0,0 +1,108 @@
...
+attenuation type := Simple
+attenuation map := ${umap}
+mask type := No
+;mask file := mask.hv
+End Projection Matrix By Bin SPECT UB Parameters:=
+End Projector Pair Using Matrix Parameters :=
+;SPECT END
+
+; additive projection data to handle randoms etc
+additive sinogram := my_additive_sinogram_1ax_pos.hs
```
ashgillman: `${ADDITIVE}`

examples/samples/KOSMAPOSL_hkem_spect.par
```diff
@@ -0,0 +1,108 @@
...
+XY output image size (in pixels):=128
+Z output image size (in pixels):=128 
+subset sensitivity filenames := sens_%d.hv
+recompute sensitivity := 1
+use subset sensitivities := 1
+end PoissonLogLikelihoodWithLinearModelForMeanAndProjData Parameters:=
+
+
+; if you want to continue from a particular image
+; initial estimate:= reconML.hv
```
ashgillman: `${INITIAL_ESTIMATE}`

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -216,22 +219,35 @@ post_processing()
...
       }
 
+  if ((this->only_iterative_info) && (!this->hybrid))
+      error("If you choose to have only the iterative side information hybrid needs to be set as 1");
   if (!this->anatomical_image_filenames.empty()){
       this->anatomical_prior_sptr= (read_from_file<TargetT>(anatomical_image_filenames[0]));
 
-    set_anatomical_prior_sptr (this->anatomical_prior_sptr);
-    info(boost::format("Reading anatomical data '%1%'")
+      set_anatomical_prior_sptr (this->anatomical_prior_sptr);
```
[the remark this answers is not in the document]
ashgillman: Scrap that, see my proposal below to use `anatomical image filenames := {}`.

ashgillman:
I could convince myself it should be `hybrid` True or False. Does hybrid mean that there are multiple sources? Or does hybrid mean that you update each iteration from the current estimate?

I would prefer the notation `sigma_m=0` or `sigma_p=0` to deactivate one of the sources, as it is then unambiguous (and also does away with two other parameters, `hybrid` and `only_iterative_info`). But I might be the only one with this preference.

It would of course break backwards compatibility.
But, if you were doing that, maybe it would make sense to also rename to `sigma_prior` or `sigma_anat` and `sigma_current_estimate` or similar. It got quite confusing when talking with Harry, since all of a suddent the PET was the `m` and the SPECT was the `p`.

KrisThielemans:
I strongly support changing names for `a`, `b`, `p` to something meaningful. I've accepted single letters or symbols (e.g. kappa) in a few cases as it'd made it easier to refer to the paper. However, I agree that this naming will only get more and more confusing in a synergistic scenario.

I'd replace all `anatomical` with something else therefore, e.g. `side_information` (`prior` seems to confusing, as then we can have e.g. the PLS prior with a prior...

> It would of course break backwards compatibility.

true. We haven't actually release 4.0 yet, so in principle could still do this without breaking our own rules :-) However, it'd break SIRF 2.1 then. It seems therefore we would need to provide the "old" `set/get` functions as well. As we really would have to do this for PLS as well, this is probably for our next major release (hopefully in the not too distant future). @ashgillman want to create a new issue for this?

danieldeidda:
> I would prefer the notation `sigma_m=0` or `sigma_p=0` to deactivate one of the sources, as it is then unambiguous (and also does away with two other parameters, `hybrid` and `only_iterative_info`). But I might be the only one with this preference.

I personally don't mind using these instead of hybrid and only_iterative_info

KrisThielemans:
Can this be abandoned?

danieldeidda:
It can but will need to creat a new one to allow only functional HKEM

ashgillman:
@danieldeidda are you sure, I thought it might be already supported with:

hybrid := false
anatomical image filename := {}

since #424

danieldeidda:
> @danieldeidda are you sure, I thought it might be already supported with:
> 
> hybrid := false
> anatomical image filename := {}
> 
> since #424

there is still i little modification to do. If we say hybrid=0 we only use the anatomical image and if there isn't any anatomical image there are going to be issues.
we need a condition where
 if (anatomical_image_filename.empty & hybrid==0)
only use iterative info

ashgillman:
🤦‍♂  Right, I mean `hybrid := true`; `anatomical image filename := {}`.

PS: Hybrid is a bit confusing, perhaps could be proposed to be renamed in #442 to something like `use current estimate`.

Either way, I think we don't need this PR in particular anymore right?

danieldeidda:
> man_facepalming Right, I mean `hybrid := true`; `anatomical image filename := {}`.
> 
> PS: Hybrid is a bit confusing, perhaps could be proposed to be renamed in #442 to something like `use current estimate`.
> 
> Either way, I think we don't need this PR in particular anymore right?

Yes let's close this. I will make a small PR later with the relevant changes

PR #435, opened 2019-12-02
FIX: HKEM didn't respect `start at subiteration number`

ashgillman:
Added test for HKEM that using a neighbourhood of 1 is equivalent to OSEM.

This failed, because the recon_test_pack default test uses `start at subiteration number := 2` (not sure why, but a good test since it picked up an unknown bug).

Turns out KOSMAPOSLReconstructor doesn't use `start_subeteration_num` in it's caluclation - this fixes that.

### DIFF
--- a/recon_test_pack/KOSMAPOSL_test_consistency.par
+++ b/recon_test_pack/KOSMAPOSL_test_consistency.par
@@ -1,0 +1,86 @@
+KOSMAPOSLParameters :=
+
+disable output := 1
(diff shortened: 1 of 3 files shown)


ashgillman:
@danieldeidda

ashgillman:
Woops! We should have commented last night - Daniel and I found a mistake
but couldn't work out how to fix it. I'll make a bugfix PR ASAP today

Issue #429, opened 2019-11-22
`KOSMAPOSLReconstructor` doesn't write output image at end of run.


ashgillman:
OSMAPOSL will write images when iteration % save_interval ==0, and at the end of the program. However, KOSMAPOSL will only write images in the former case. So if the number of iterations is not a multiple of the save interval, you don't get a final image.

@danieldeidda is this because of the extra half iteration required?

danieldeidda:
"OSMAPOSL will write images when iteration % save_interval ==0, and at the end of the program" Is that so? 
This is decided in the parameter file.  So if you say in the parfile that you want to save images every N subiterations, then it makes sense to give a total subiteration number that is a multiple, k, of N. I thought OSMAPOSL would do the same, but I never tried to put a number of subiterations which is not a multiple of the interval. Is there a reason why we would want to write the last?

KrisThielemans:
fixed with #436

PR #436, opened 2019-12-03
FIX: #435 introduced bug where save interval was incorrectly handled.

ashgillman:
REF: Also remove KOSMAPOSLReconstruction ad-hoc iteration counter in lieu of
parent's.
REF: Also use IteratriveReconstruction's formatter for filename.


ashgillman:
Fixes the last bug we introduced, and removes re-implementation of `IterativeReconstructor`'s features.

ashgillman:
Also fixes #429

PR #449, opened 2020-02-06
Parallel KOSMAPOSL

danieldeidda:
made KOSMAPOSL multi-threadable: 
1) apply kernel to sensitivity and gradient using different threads, 
2) make loop inside compute_kernelised_image() parallel

### DIFF
--- a/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
+++ b/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
@@ -252,8 +252,9 @@ const DiscretisedDensityOnCartesianGrid<3,float>* current_anatomical_cast =
 
 
     if(num_non_zero_feat>1){
+         this->kmnorm_sptr.resize(anatomical_image_filenames.size());
          for(int i = 0; i <=anatomical_image_filenames.size()-1; i++){
-           this->kmnorm_sptr.push_back( shared_ptr<TargetT>(this->anatomical_prior_sptr[i]->get_empty_copy ()));
+           this->kmnorm_sptr[i].reset(this->anatomical_prior_sptr[i]->get_empty_copy ());
            this->kmnorm_sptr[i]->resize(IndexRange3D(0,0,0,this->num_voxels-1,0,this->num_elem_neighbourhood-1));
            }
       
--- a/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
+++ b/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
@@ -59,6 +59,12 @@
 
 #include <memory>
 #include <iostream>
+
+#ifdef STIR_OPENMP
+#include <omp.h>
+#endif
+#include "stir/num_threads.h"
+
 #ifdef BOOST_NO_STRINGSTREAM
 #include <strstream.h>
 #else
@@ -355,14 +361,6 @@ KOSMAPOSLReconstruction<TargetT>::
 get_hybrid() const
 { return this->hybrid; }
 
-//template <typename TargetT >
-//shared_ptr<TargetT> &KOSMAPOSLReconstruction<TargetT>::get_kpnorm_sptr()
-//{ return this->kpnorm_sptr; }
-
-//template <typename TargetT >
-//shared_ptr<TargetT> &KOSMAPOSLReconstruction<TargetT>::get_kmnorm_sptr()
-//{ return this->kmnorm_sptr; }
-
 template <typename TargetT>
 
 std::vector<shared_ptr<TargetT> > KOSMAPOSLReconstruction<TargetT>::get_anatomical_prior_sptr()
@@ -509,6 +507,12 @@ calculate_norm_matrix(TargetT &normp,
       const int min_y = emission[z].get_min_index();
       const int max_y = emission[z].get_max_index();
       this->dimy=max_y-min_y+1;
+
+#ifdef STIR_OPENMP
+#pragma omp parallel
+    {
+      #pragma omp  for schedule(dynamic)//schedule(static,1)
+#endif
       for (int y=min_y;y<= max_y;y++)
         {
           const int min_dy = max(distance[0].get_min_index(), min_y-y);
@@ -517,7 +521,6 @@ calculate_norm_matrix(TargetT &normp,
           const int min_x = emission[z][y].get_min_index();
           const int max_x = emission[z][y].get_max_index();
           this->dimx=max_x-min_x+1;
-
 
           for (int x=min_x;x<= max_x;x++)
             {
@@ -555,12 +558,20 @@ calculate_norm_matrix(TargetT &normp,
                     }
             }
         }
-    }
+#ifdef STIR_OPENMP
+}
+#endif
+      }
 
   // the norms of the difference between feature vectors related to the
   // same neighbourhood are calculated now
   int p=0,o=0;
 
+#ifdef STIR_OPENMP
+#pragma omp parallel
+    {
+      #pragma omp  for collapse(3) schedule(dynamic)//schedule(static,1)
+#endif
   for (int q=0; q<=dimf_row-1; ++q){
     for (int n=-(this->num_neighbours-1)/2*(!this->only_2D);
          n<=(this->num_neighbours-1)/2*(!this->only_2D);
@@ -600,6 +611,9 @@ calculate_norm_matrix(TargetT &normp,
               normp[0][q][p] += square(fp[q][i]-fp[o][i]);
             }
       }
+#ifdef STIR_OPENMP
+}
+#endif
 }
 
 template<typename TargetT>
@@ -629,7 +643,9 @@ void KOSMAPOSLReconstruction<TargetT>::estimate_stand_dev_for_anatomical_image(s
         const int max_z = (*anatomical_prior_sptr[i]).get_max_index();
 
         dim_z = max_z -min_z+1;
-
+        #ifdef STIR_OPENMP
+        #pragma omp parallel for reduction(+:kmean,nv)
+        #endif
         for (int z=min_z; z<=max_z; z++)
           {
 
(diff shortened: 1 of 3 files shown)


danieldeidda:
@ashgillman @KrisThielemans Have a look at this small changes. I am quite new to openMP I made few changes following some guideline and I got an acceleration of 62% using 36 threads.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -247,11 +253,11 @@ const DiscretisedDensityOnCartesianGrid<3,float>* current_anatomical_cast =
 
     if(num_non_zero_feat>1){
          for(int i = 0; i <=anatomical_image_filenames.size()-1; i++){
-           this->kmnorm_sptr[i] = shared_ptr<TargetT>(this->anatomical_prior_sptr[i]->get_empty_copy ());
+           this->kmnorm_sptr.push_back( shared_ptr<TargetT>(this->anatomical_prior_sptr[i]->get_empty_copy ()));
```
KrisThielemans: I haven't check this, but the previous version assigned to a specific index, while the new version will just grow. If this new behaviour is correct, it implies that the previous one should have crashed rather dramatically...

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -509,6 +507,12 @@ calculate_norm_matrix(TargetT &normp,
       const int min_y = emission[z].get_min_index();
       const int max_y = emission[z].get_max_index();
       this->dimy=max_y-min_y+1;
+
+#ifdef STIR_OPENMP
+#pragma omp parallel
+    {
+      #pragma omp  for schedule(dynamic) private(l,m)//schedule(static,1)
```
KrisThielemans: syntax ok I believe, but shorter to just say
```c++
#pragma omp parallel for ....
```

KrisThielemans: Also, as you declare `l,m` private, it means that their values outside of the loop will be unpredictable. That makes me wonder if they shouldn't be declared locally below, i.e.
```c++
const int l = (z-min_z)*(max_x-min_x +1)*(max_y-min_y +1)
                + (y-min_y)*(max_x-min_x +1) + (x-min_x);
```
far cleaner...

KrisThielemans: Finally, you can parallelise nested for loops as in https://stackoverflow.com/questions/13357065/how-does-openmp-handle-nested-loops.
that might help with speeding up if you have a lot of cores.

However, I think this is only since openmp 3.1 (do check). Visual Studio is still on OpenMP 2.0... Interestingly, OpenMP versions are encoded via date, see https://stackoverflow.com/questions/1304363/how-to-check-the-version-of-openmp-on-linux/13612520
So, I'm afraid you'll have to do something like
```c++
#ifdef STIR_OPENMP
#  if _OPENMP <201107
#    pragma omp  for schedule(dynamic)
#  else
#    pragma omp  for schedule(dynamic) collapse(2)
#  endif
#endif
```

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -555,12 +558,20 @@ calculate_norm_matrix(TargetT &normp,
...
+      }
 
   // the norms of the difference between feature vectors related to the
   // same neighbourhood are calculated now
   int p=0,o=0;
 
+#ifdef STIR_OPENMP
+#pragma omp parallel
+    {
+      #pragma omp  for collapse(3) schedule(dynamic) private(p,o)//schedule(static,1)
```
KrisThielemans: same for `p,o` of course

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -629,7 +643,9 @@ void KOSMAPOSLReconstruction<TargetT>::estimate_stand_dev_for_anatomical_image(s
         const int max_z = (*anatomical_prior_sptr[i]).get_max_index();
 
         dim_z = max_z -min_z+1;
-
+        #ifdef STIR_OPENMP
+        #pragma omp parallel for reduction(+:kmean,nv)
```
KrisThielemans: sadly not supported by Visual Studio I believe (do check). Easiest might be to check OPENMP version, and only include this pragma if it's recent enough.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -662,7 +678,9 @@ void KOSMAPOSLReconstruction<TargetT>::estimate_stand_dev_for_anatomical_image(s
                 }
             }
                       kmean=kmean / nv;
-
+#ifdef STIR_OPENMP
+                      #pragma omp parallel for reduction(+:kStand_dev)
```
KrisThielemans: as above

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -677,9 +695,7 @@ void KOSMAPOSLReconstruction<TargetT>::estimate_stand_dev_for_anatomical_image(s
 
                                 for (int x=min_x;x<= max_x;x++)
                                   {
-                                    if((*anatomical_prior_sptr[i])[z][y][x]>=0 && (*anatomical_prior_sptr[i])[z][y][x]<=1000000){
-                                        kStand_dev += square((*anatomical_prior_sptr[i])[z][y][x] - kmean);}
-                                    else{continue;}
+                                    kStand_dev += square((*anatomical_prior_sptr[i])[z][y][x] - kmean);
```
KrisThielemans: doesn't seem to be long in this PR. I don't really mind, but it will need to be a different commit (maybe it is already...)

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -920,20 +947,34 @@ update_estimate(TargetT &current_alpha_coefficent_image)
...
+                                                          subset_num);
+  unique_ptr< TargetT > kmultiplicative_update_ptr((*multiplicative_update_image_ptr).get_empty_copy());
+
+  const TargetT& sensitivity =
+    base_type::get_subset_sensitivity(subset_num);
 
+unique_ptr< TargetT > ksens_ptr(sensitivity.get_empty_copy());
+#ifdef STIR_OPENMP
+#pragma omp parallel sections
+ { omp_set_nested(1);
```
KrisThielemans: no idea why you have `omp_set_nested`. It seems deprecated https://www.openmp.org/spec-html/5.0/openmpsu119.html and I guess replaced by `omp_set_max_active_levels`. However, https://www.openmp.org/spec-html/5.0/openmpsu125.html#x162-7300003.2.16 states

    This routine has the described effect only when called from a sequential part of the program. 
    When called from within a parallel or teams region, the effect of this routine is implementation defined. 

So this is probably ill-defined.

KrisThielemans: In any case, this seems like a very bad idea. you're trying to do 2 kernelisations in parallel, but these are already parallelised. That might only be beneficial if you have many more threads then each "kernelisation" can use. But that won't be the case anymore with the collapsed-for loops. I'd remove the sections.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: Yes in fact, the previous version will fail if you use non_zero_element>1. sadly I noticed while running now

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: I had tried to use nested but I get errors! It seems that only works if there is nothing between the consecutive loops, ex:
For(){
   For(){
     For(){}}}

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
KrisThielemans: ok. it's in a separate commit, so that's fine. However, now you have to be sure that `kmnorm_sptr` is empty when you start (maybe that's guaranteed, I don't know)

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
KrisThielemans: ah well... an alternative reason might be that in the code min_x depends on y. that probably makes it too hard for the compiler anyway.
One solution would be to require "rectangular" images, and compute min_z/y/z etc upfront (e.g. with get_regular_range). I suspect that this is required for your code anyway.
you probably would still have to move min_dy etc inside but as that's a near trivial calculation, that's not going to do any harm.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -800,8 +823,12 @@ void KOSMAPOSLReconstruction<TargetT>::compute_kernelised_image(
...
                     }
                   }
-                }
-}
+
+#ifdef STIR_OPENMP
+    }
+    #endif
+    }
+    }
```
ashgillman: This is a bit hard to follow, Its just an extra brace for the OMP pragme?
Would be easier if you could make sure the closing indentation matches the opening.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: mh not sure here. That line is the very first call to kmnorm_sptr. I assumed that would be empty until that point.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: > you probably would still have to move min_dy etc inside but as that's a near trivial calculation, that's not going to do any harm.

Shall I then move min_dz inside so that I can do collapse(3)?

ashgillman:
62% using 36 threads on how many cores? That seems like diminishing returns

danieldeidda:
40 cores. Anyway after 10 threads it doesn't actually get faster. 36 was set automatically by STIR

PR #464, opened 2020-03-24
Demo missing template scanner imaging modality fix

### LINKED ISSUE #463: examples/src forward project errors
Following the README.txt instructions. After running 
   `generate_image generate_image.par` 
calling  
   `forward_project sino.hs image.hv  small.hs` 
results in the following error.

`ERROR: forward_project: Imaging modality should be the same for the image and the projection data
libc++abi.dylib: terminating with uncaught exception of type std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >
Abort trap: 6`

robbietuk:
Fixes #463

This PR fixes an error in the C++ demos when generating data. There was a missing line in the template scanner.


KrisThielemans:
Please add `imaging modality:=PET` to the `generate_image.par` file as well. Ideally also add patient orientation stuff to both, as otherwise it'll break at some point later anyway.

see the one in `examples/samples`.

examples/src/generate_image.par
```diff
@@ -1,4 +1,28 @@
...
+; example below uses Interfile with 16-bit unsigned integers
+output file format type:= Interfile
+  interfile Output File Format Parameters:=
+    number format := unsigned integer
+    number_of_bytes_per_pixel:=2
+    ; fix the scale factor to 1
+    ; comment out next line to let STIR use the full dynamic 
+    ; range of the output type
+    scale_to_write_data:= 1
+  End Interfile Output File Format Parameters:=
```
KrisThielemans:
```suggestion
```
I'd get rid of these.

robbietuk:
Are you happy with the comments?

KrisThielemans:
Does it run?

robbietuk:
Yes. The only warning I could fix is: `WARNING: image duration not set, so time frame definitions will not be initialised`, when `generate_image generate_image.par`

KrisThielemans:
Great. well, let's add that then as well. (add `[ci skip]` in the subject line to save some resources).

fine to squash-merge?

robbietuk:
Ready to squash-merge!

Issue #462, opened 2020-03-23
ProjDataFromStream::get_bin_value and set


KrisThielemans:
We have `get_bin_value`, but no `set`. Seems relatively easy to add.

It would then also make sense to apply the patch
https://github.com/UCL/STIR/blob/5a45c96dd6cabb0c309d63277aefc1d95e752814/src/buildblock/ProjDataInMemory.cxx#L136-L141
which currently sits in the TOF PR #304

KrisThielemans:
@danieldeidda as discussed, maybe you want to give this a go

PR #469, opened 2020-03-25
ProjDataFromStream, adding set_bin_value, fixing #462:

danieldeidda:
added set_bin_value(Bin)

### DIFF
--- a/src/buildblock/ProjDataFromStream.cxx
+++ b/src/buildblock/ProjDataFromStream.cxx
@@ -266,12 +266,12 @@ ProjDataFromStream::set_bin_value(const Bin& this_bin)
 
     const streamoff total_offset = offsets[0];
 
-    sino_stream->seekg(0 , ios::beg); // reset file
-    sino_stream->seekg(total_offset, ios::cur); // start of view within segment
+    sino_stream->seekp(0 , ios::beg); // reset file
+    sino_stream->seekp(total_offset, ios::cur); // start of view within segment
 
     if (! *sino_stream)
     {
-        error("ProjDataFromStream::set_bin_value: error after seekg.");
+        error("ProjDataFromStream::set_bin_value: error after seekp.");
     }
 
     Array< 1,  float>  value(1);
--- a/src/buildblock/ProjDataFromStream.cxx
+++ b/src/buildblock/ProjDataFromStream.cxx
@@ -216,11 +216,11 @@ ProjDataFromStream::get_bin_value(const Bin& this_bin) const
 {
     if (is_null_ptr(sino_stream))
     {
-        error("ProjDataFromStream::get_viewgram: stream ptr is 0\n");
+        error("ProjDataFromStream::get_bin_value: stream ptr is 0\n");
     }
     if (! *sino_stream)
     {
-        error("ProjDataFromStream::get_viewgram: error in stream state before reading\n");
+        error("ProjDataFromStream::get_bin_value: error in stream state before reading\n");
     }
 
     vector<streamoff> offsets = get_offsets_bin(this_bin);
@@ -251,15 +251,15 @@ ProjDataFromStream::get_bin_value(const Bin& this_bin) const
 }
 
 void
-ProjDataFromStream::set_bin_value(const Bin this_bin) 
+ProjDataFromStream::set_bin_value(const Bin& this_bin) 
 {
     if (is_null_ptr(sino_stream))
     {
-        error("ProjDataFromStream::get_viewgram: stream ptr is 0\n");
+        error("ProjDataFromStream::set_bin_value: stream ptr is 0\n");
     }
     if (! *sino_stream)
     {
-        error("ProjDataFromStream::get_viewgram: error in stream state before writing");
+        error("ProjDataFromStream::set_bin_value: error in stream state before writing");
     }
 
     vector<streamoff> offsets = get_offsets_bin(this_bin);
@@ -274,16 +274,15 @@ ProjDataFromStream::set_bin_value(const Bin this_bin)
         error("ProjDataFromStream::set_bin_value: error after seekg.");
     }
 
-    Array< 1,  float>  value(this_bin.get_bin_value());
+    Array< 1,  float>  value(1);
+    value[0]=this_bin.get_bin_value();
     float scale = float(1);
-
     // Now the storage order is not more important. Just read.
     if (write_data(*sino_stream, value, on_disk_data_type, scale, on_disk_byte_order)
             == Succeeded::no)
         error("ProjDataFromStream: error writing data\n");
     if(scale != 1.f)
         error("ProjDataFromStream: error writing data: scale factor returned by write_data should be 1\n");
-
 }
 
 vector<streamoff>
--- a/src/buildblock/ProjDataFromStream.cxx
+++ b/src/buildblock/ProjDataFromStream.cxx
@@ -250,6 +250,41 @@ ProjDataFromStream::get_bin_value(const Bin& this_bin) const
     return value[0];
 }
 
+void
+ProjDataFromStream::set_bin_value(const Bin this_bin) 
+{
+    if (is_null_ptr(sino_stream))
+    {
+        error("ProjDataFromStream::get_viewgram: stream ptr is 0\n");
+    }
+    if (! *sino_stream)
+    {
+        error("ProjDataFromStream::get_viewgram: error in stream state before writing");
+    }
+
+    vector<streamoff> offsets = get_offsets_bin(this_bin);
+
+    const streamoff total_offset = offsets[0];
+
+    sino_stream->seekg(0 , ios::beg); // reset file
+    sino_stream->seekg(total_offset, ios::cur); // start of view within segment
+
+    if (! *sino_stream)
+    {
+        error("ProjDataFromStream::set_bin_value: error after seekg.");
+    }
+
+    Array< 1,  float>  value(this_bin.get_bin_value());
+    float scale = float(1);
+
+    // Now the storage order is not more important. Just read.
+    if (write_data(*sino_stream, value, on_disk_data_type, scale, on_disk_byte_order)
+            == Succeeded::no)
+        error("ProjDataFromStream: error writing data\n");
+    if(scale != 1.f)
+        error("ProjDataFromStream: error writing data: scale factor returned by write_data should be 1\n");
+
+}
 
 vector<streamoff>
 ProjDataFromStream::get_offsets(const int view_num, const int segment_num) const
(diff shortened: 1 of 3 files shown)


src/buildblock/ProjDataFromStream.cxx
```diff
@@ -250,6 +250,41 @@ ProjDataFromStream::get_bin_value(const Bin& this_bin) const
     return value[0];
 }
 
+void
+ProjDataFromStream::set_bin_value(const Bin this_bin) 
+{
+    if (is_null_ptr(sino_stream))
+    {
+        error("ProjDataFromStream::get_viewgram: stream ptr is 0\n");
```
KrisThielemans: change text and in the ones below

src/buildblock/ProjDataFromStream.cxx
```diff
@@ -250,6 +250,41 @@ ProjDataFromStream::get_bin_value(const Bin& this_bin) const
     return value[0];
 }
 
+void
+ProjDataFromStream::set_bin_value(const Bin this_bin)
```
KrisThielemans: let's make it a `const Bin&`

src/include/stir/ProjDataFromStream.h
```diff
@@ -144,6 +144,9 @@ class ProjDataFromStream : public ProjData
 
   //! Get the value of bin.
   float get_bin_value(const Bin& this_bin) const;
+  
+  //! set the value f the bin
+  void set_bin_value(Bin bin);
```
KrisThielemans: needs to match `const Bin&`

KrisThielemans:
I think we should also remove `ProjDataInMemory::get_bin_value` (as it's better in `ProjDataInStream::get_bin_value`)

src/buildblock/ProjDataFromStream.cxx
danieldeidda: with what?

KrisThielemans: it says `get_viewgram` ...

danieldeidda: oh right then should be changed to get_bin as well!

danieldeidda:
> 
> 
> ideally add a simple test in `src/test/test_proj_data_in_memory.cxx`.

a test for for set_bin()?

danieldeidda:
> 
> 
> I think we should also remove `ProjDataInMemory::get_bin_value` (as it's better in `ProjDataInStream::get_bin_value`)

is ProjDataInMemory::get_bin_value called anywere?

KrisThielemans:
> a test for for set_bin()?
yes please. Otherwise we won't know if it works! Not sure if there's one for `get_bin_value` already.

I guess something like this
1. create a `std::vector<float> of size `proj_data.size_all()`
2. fill it with numbers 1... N using `proj_data.fill_from` such that they're all different
3. pick a `Bin` and use `set_bin_value` to anything else
4. use `get_bin_value` to check
5. also use `get_viewgram` or whatever to get access to the same bin, and see if everything's fine

KrisThielemans:
Seems Appveyor is struggling with access to github, but 2 jobs were fine, so let's ignore the 2 others.

Not so sure why Travis isn't running on this. oh well.

KrisThielemans:
> is `ProjDataInMemory::get_bin_value` called anywhere?

sure, but as it's in `ProjDataFromStream` as `public` (I think), it'll still be there. It's just that we forgot to remove its old implementation when adding `get_bin_value` to `ProjDataFromStream`.

KrisThielemans:
By the way, now that we again have access to the `buffer` for the data (since #448), we could/should speed-up `ProjDataInMemory` access for `get/set_bin_value` by directly going to the `buffer`, but let's leave that for another PR....

danieldeidda:
it seems that the value of the bin in  get and set is always 1

KrisThielemans:
not sure what you mean. `get_bin_value` doesn't modify the `Bin.value()`, but returns a number (bad naming maybe?). `set_bin_value()` would have to be used as
```C++
Bin bin(blabla);
bin.set_bin_value(5.F);
proj_data.set_bin_value(bin);
```
a bit awkward of course. You can use `Bin bin(blabla, 5.F)` but it'll become a bit unsafe for the TOF PR. Or we could have an additional member
`proj_data.set_bin_value(bin, 5.F)` which seems clear enough.

danieldeidda:
> 
> 
> not sure what you mean. `get_bin_value` doesn't modify the `Bin.value()`, but returns a number (bad naming maybe?). `set_bin_value()` would have to be used as
> 
> ```c++
> Bin bin(blabla);
> bin.set_bin_value(5.F);
> proj_data.set_bin_value(bin);
> ```
> 
> a bit awkward of course. You can use `Bin bin(blabla, 5.F)` but it'll become a bit unsafe for the TOF PR. Or we could have an additional member
> `proj_data.set_bin_value(bin, 5.F)` which seems clear enough.

yes this is what I'm doing. But what is the float number the it's being returned by get_bin_value()? I thought that would have to be the actual bin value.

KrisThielemans:
sure. that's what it should do. Looking at both the `ProjDataFromStream` and `ProjDataInMemory` implementation, they seem correct. I guess time to push your current code such that we can have a look?

src/test/test_proj_data_in_memory.cxx
```diff
@@ -151,6 +151,39 @@ run_tests()
...
+      
+      Bin old_bin=bin;
+      
+      bin.set_bin_value(42);
+      proj_data.set_bin_value(bin);
+      
+      Viewgram<float> viewgram=proj_data.get_viewgram(
+                  proj_data_info_sptr->get_max_view_num()/2,
+                  0,
+                  false);
```
KrisThielemans:
```suggestion
                  bin.view(), bin.segment());
```
much clearer... Also no need for the `false` these days (it's obsolete usage actually)

src/test/test_proj_data_in_memory.cxx
```diff
@@ -151,6 +151,39 @@ run_tests()
...
+  {
+      std::vector<float> test;
+      test.resize(proj_data.size_all());
+      
+      for(int i=0;i<test.size();i++)
+      test[i]=i;
+      
+      proj_data.fill_from(test.begin());
+      
+      Bin bin(0,proj_data_info_sptr->get_max_view_num()/2,
```
KrisThielemans: seems you can merge usage of `old_bin` and `bin` into 1 variable

src/test/test_proj_data_in_memory.cxx
```diff
@@ -151,6 +151,39 @@ run_tests()
...
+                  0,
+                  false);
+      
+      Bin new_bin=old_bin;
+      new_bin.set_bin_value(viewgram[proj_data_info_sptr->get_max_axial_pos_num(0)/2][0]);
+     
+      
+//      std::cout<<"old "<<proj_data.get_bin_value(bin)<<" new "<<new_bin.get_bin_value()<<std::endl;
+      check(old_bin.get_bin_value()==new_bin.get_bin_value(), "ProjDataFromStream::set_bin_value  did not change the value of the viewgram bin");
+//      proj_data_info_sptr->get
```
KrisThielemans:
```suggestion
check(bin.get_bin_value()==proj_data.get_bin_value(bin), "ProjDataFromStream::set_bin_value/get_bin_value not consistent");
```
or something like this (you didn't check `get_bin_value` here)

danieldeidda:
when using debugging seems that when I set the bin value that value is being written to the stream, however when I use get_viewgram() wchi than call read(stream...) the value that it is reading is the old one

KrisThielemans:
Strange, is that with the version with my suggested changes? 

This might all be a bit easier to debug from Python (or MATLAB). Create a `ProjDataInMemory`, fill with 0, set 1 bin, get the array, check if there's a single value which is 1, etc. (too painful in C++).

A few possible causes: there's a problem in the test code, it's writing somewhere else than what we're reading, it's not writing at all, it's writing to a different stream. I guess it's only the first two which are likely though. 

Interestingly, I don't think we have a check if the various `get_offset` functions in `ProjDataFromStream` are consistent. Ideally, we'd rewrite all of them in terms of `get_offsets_bin`. However, we can only do that once we know that `get_offset_bin` is actually correct (the others are correct with a 99.99% probability, as they have been tested in practice by now).

src/test/test_proj_data_in_memory.cxx
danieldeidda: > Strange, is that with the version with my suggested changes?
yeah but it happens without the changes as well. I thought as well writing and reading is happening trough different data but I am not sure that is what is happening.
 I could check the offset in writing and reading operations

src/buildblock/ProjDataFromStream.cxx
```diff
@@ -250,6 +250,40 @@ ProjDataFromStream::get_bin_value(const Bin& this_bin) const
...
+    {
+        error("ProjDataFromStream::set_bin_value: error in stream state before writing");
+    }
+
+    vector<streamoff> offsets = get_offsets_bin(this_bin);
+
+    const streamoff total_offset = offsets[0];
+
+    sino_stream->seekg(0 , ios::beg); // reset file
+    sino_stream->seekg(total_offset, ios::cur); // start of view within segment
```
KrisThielemans: needs to be `seekp`

src/buildblock/ProjDataFromStream.cxx
```diff
@@ -250,6 +250,40 @@ ProjDataFromStream::get_bin_value(const Bin& this_bin) const
...
+
+    if (! *sino_stream)
+    {
+        error("ProjDataFromStream::set_bin_value: error after seekg.");
```
KrisThielemans: `seekp`

src/buildblock/ProjDataFromStream.cxx
KrisThielemans: also comment says `start of view within segment`

KrisThielemans:
thanks, but could you put in the recommendations from above?e.g.
```c++
check_if_equal(bin.get_bin_value(), viewgram[bin.axial_pos_num()][bin.tangential_pos_num()],
            "ProjDataFromStream::set_bin_value/ get_viewgram not consistent");
```
 sorry.

danieldeidda:
> 
> 
> thanks, but could you put in the recommendations from above?e.g.
> 
> ```c++
> check_if_equal(bin.get_bin_value(), viewgram[bin.axial_pos_num()][bin.tangential_pos_num()],
>             "ProjDataFromStream::set_bin_value/ get_viewgram not consistent");
> ```
> 
> sorry.

sorry I missed part of the comment, sure

danieldeidda:
> 
> 
> thanks, but could you put in the recommendations from above?e.g.

so are saying that we need to different checks?

KrisThielemans:
> so are you saying that we need to different checks?

replace `check(a==b, string)` with `check_if_equal(a,b,string)`, yes (not the same for float comparisons).

Issue #466, opened 2020-03-25
use shared_ptr for data_info_ptr in InterfilePDFSHeader


KrisThielemans:
We're still using a bare pointer https://github.com/UCL/STIR/blob/3f71c65c4c00b1f09efe816a89291bad721b020d/src/include/stir/IO/InterfileHeader.h#L241
This can lead to memory leaks, and does so in SIRF https://github.com/CCPPETMR/SIRF/issues/593

Just using a `shared_ptr` should fix this.

PR #470, opened 2020-03-26
remove uses of raw pointer to ProjData and its children

rijobro:
fixes https://github.com/UCL/STIR/issues/466. Probably much larger PR than expected @KrisThielemans.

### DIFF
--- a/src/analytic/FBP2D/FBP2DReconstruction.cxx
+++ b/src/analytic/FBP2D/FBP2DReconstruction.cxx
@@ -152,7 +152,7 @@ bool FBP2DReconstruction::post_processing_only_FBP2D_parameters()
       const shared_ptr<const ProjDataInfoCylindrical> proj_data_info_cyl_sptr =
 	dynamic_pointer_cast<const ProjDataInfoCylindrical>(proj_data_ptr->get_proj_data_info_sptr());
 
-      if (proj_data_info_cyl_sptr==0)
+      if (proj_data_info_cyl_sptr==nullptr)
         num_segments_to_combine = 1; //cannot SSRB non-cylindrical data yet
       else
 	{
@@ -261,7 +261,7 @@ actual_reconstruct(shared_ptr<DiscretisedDensity<3,float> > const & density_ptr)
   ArcCorrection arc_correction;
   bool do_arc_correction = false;
   if (dynamic_pointer_cast<const ProjDataInfoCylindricalArcCorr>
-      (proj_data_ptr->get_proj_data_info_sptr()) != 0)
+      (proj_data_ptr->get_proj_data_info_sptr()) != nullptr)
     {
       // it's already arc-corrected
       arc_corrected_proj_data_info_sptr =
--- a/src/analytic/FBP2D/FBP2DReconstruction.cxx
+++ b/src/analytic/FBP2D/FBP2DReconstruction.cxx
@@ -152,7 +152,7 @@ bool FBP2DReconstruction::post_processing_only_FBP2D_parameters()
       const shared_ptr<const ProjDataInfoCylindrical> proj_data_info_cyl_sptr =
 	dynamic_pointer_cast<const ProjDataInfoCylindrical>(proj_data_ptr->get_proj_data_info_sptr());
 
-      if (proj_data_info_cyl_sptr==nullptr)
+      if (is_null_ptr(proj_data_info_cyl_sptr))
         num_segments_to_combine = 1; //cannot SSRB non-cylindrical data yet
       else
 	{
@@ -260,8 +260,8 @@ actual_reconstruct(shared_ptr<DiscretisedDensity<3,float> > const & density_ptr)
   // arc-correction if necessary
   ArcCorrection arc_correction;
   bool do_arc_correction = false;
-  if (dynamic_pointer_cast<const ProjDataInfoCylindricalArcCorr>
-      (proj_data_ptr->get_proj_data_info_sptr()) != nullptr)
+  if (!is_null_ptr(dynamic_pointer_cast<const ProjDataInfoCylindricalArcCorr>
+      (proj_data_ptr->get_proj_data_info_sptr())))
     {
       // it's already arc-corrected
       arc_corrected_proj_data_info_sptr =
--- a/src/analytic/FBP2D/FBP2DReconstruction.cxx
+++ b/src/analytic/FBP2D/FBP2DReconstruction.cxx
@@ -149,17 +149,17 @@ bool FBP2DReconstruction::post_processing_only_FBP2D_parameters()
 
     if (num_segments_to_combine==-1)
     {
-      const ProjDataInfoCylindrical * proj_data_info_cyl_ptr =
-	dynamic_cast<const ProjDataInfoCylindrical *>(proj_data_ptr->get_proj_data_info_ptr());
+      const shared_ptr<const ProjDataInfoCylindrical> proj_data_info_cyl_sptr =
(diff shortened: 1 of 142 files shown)


rijobro:
Since I would have deleted the method, `get_proj_data_info_ptr`, we would also have to change this outdated example:

https://github.com/UCL/STIR/blob/3f71c65c4c00b1f09efe816a89291bad721b020d/documentation/STIR-developers-overview.tex#L538-L544

rijobro:
I just tried separating the two out and tackling the first. So just switched `ProjDataInfo*` to `shared_ptr<ProjDataInfo>` in `InterfilePDFSHeader` and then made minimal changes to allow that, see [here](https://github.com/rijobro/STIR/commit/7219f6328ed6e007964b384a4eaa76a507ce5745).

Get a weird error:
```
[ 49%] Linking CXX executable test_BSpline_transformations
Undefined symbols for architecture x86_64:
  "stir::ProjData::get_proj_data_info_ptr() const", referenced from:
      stir::MultipleProjData::get_proj_data_info_ptr() const in libbuildblock.a(MultipleProjData.cxx.o)
```
Not sure what's going on, as I don't see any overlap between `MultipleProjData` and `InterfilePDFSHeader`. Any idea? 

Maybe it's not worth trying to fix. If I continue with this PR and make all `get_proj_data_info_sptr` return shared_ptr<const ProjDataInfo>, would you be happy?

KrisThielemans:
The linking problem is probably due to circular dependencies in our libraries, see #6. You could try the solution mentioned there (horrible as it is)

KrisThielemans:
I think that we'll have this problem anyway when fixing the rest, but maybe not.

> If I continue with this PR and make all get_proj_data_info_sptr return shared_ptr, would you be happy?

I guess you means return `shared_ptr<const ...>`? That'd be a good thing to do, although I have no clue how much work that'll be. Quite a lot I imagine. If so, it's probably best to replace it with references anyway.

I realise we need to have a clear-cut rule on when to use/return `shared_ptr` vs references vs whatever else. That'll be hard to decide on... Certainly not for discussion in this PR!

rijobro:
Well at any rate, I've updated this PR to return `shared_ptr<const ProjDataInfo>`. What do you think?

rijobro:
> I guess you means return shared_ptr<const ...>?

Ah yes, sorry. Surprisingly not much work!

src/analytic/FBP2D/FBP2DReconstruction.cxx
```diff
@@ -149,17 +149,17 @@ bool FBP2DReconstruction::post_processing_only_FBP2D_parameters()
 
     if (num_segments_to_combine==-1)
     {
-      const ProjDataInfoCylindrical * proj_data_info_cyl_ptr =
-	dynamic_cast<const ProjDataInfoCylindrical *>(proj_data_ptr->get_proj_data_info_ptr());
+      const shared_ptr<const ProjDataInfoCylindrical> proj_data_info_cyl_sptr =
+	dynamic_pointer_cast<const ProjDataInfoCylindrical>(proj_data_ptr->get_proj_data_info_sptr());
 
-      if (proj_data_info_cyl_ptr==0)
+      if (proj_data_info_cyl_sptr==0)
```
KrisThielemans: that'lll surely fail on pre-C++11...

src/analytic/FBP3DRP/FBP3DRPReconstruction.cxx
```diff
@@ -501,19 +501,19 @@ actual_reconstruct(shared_ptr<DiscretisedDensity<3,float> > const& target_image_
   // find out if arc-correction if necessary
   // and initialise proj_data_info_with_missing_data_sptr accordingly
   {
-    if (dynamic_cast<const ProjDataInfoCylindricalArcCorr*>
-	(proj_data_ptr->get_proj_data_info_ptr()) != 0)
+    if (dynamic_pointer_cast<const ProjDataInfoCylindricalArcCorr>
+	(proj_data_ptr->get_proj_data_info_sptr()) != 0)
```
KrisThielemans: I can't image this will work on  pre-C++11

src/include/stir/ProjData.h
```diff
@@ -120,10 +120,14 @@ class ProjData : public ExamData
...
   virtual ~ProjData() {}
   //! Get shared pointer to proj data info
+  inline shared_ptr<const ProjDataInfo>
+    get_proj_data_info_sptr() const;
+  //! Get shared pointer to proj data info
   /*! \warning Use with care. If you modify the object in a shared ptr, everything using the same
     shared pointer will be affected. */
   inline shared_ptr<ProjDataInfo>
-    get_proj_data_info_sptr() const;
+    get_proj_data_info_sptr();
```
KrisThielemans: hmmm. But this exposes everything anyway. Moreover, we have no clue which one will be called. I thought the plan was to remove this

rijobro:
I was under the impression that C++ will try to use the `const` version of the method wherever possible.

I therefore left the original signature such that if the object is not `const`, `shared_ptr<ProjDataInfo>` gets returned, but if the object is marked `const` then they will return `shared_ptr<const ProjDataInfo>`. As long as the objects/methods that call `get_proj_data_info_sptr` have been correctly marked `const` where necessary, then this seems safe to me.

KrisThielemans:
nope. C++ will only call a `const` method when the object is `const`, not when it thinks it's safe to do so (as it cannot know!). It's the reason we have a `begin_const` for instance.

We never intentionally want to expose the innards of the object. Cut the unsafe method! :-)

rijobro:
> I don't know what dynamic_pointer_cast returns if the cast fails.

From http://www.cplusplus.com/reference/memory/dynamic_pointer_cast/:
```
template <class T, class U>
  shared_ptr<T> dynamic_pointer_cast (const shared_ptr<U>& sp) noexcept;
Dynamic cast of shared_ptr
Returns a copy of sp of the proper type with its stored pointer casted dynamically from U* to T*.

If sp is not empty, and such a cast would not return a null pointer, the returned object shares ownership over sp's resources, increasing by one the use count.

Otherwise, the returned object is an empty shared_ptr.
```

rijobro:
> We never intentionally want to expose the innards of the object. Cut the unsafe method! :-)

I'd rather not have it obviously, but certain methods pass `shared_ptr<ProjDataInfo> const&`. I don't know if I should modify them or call `ProjDataInfo::create_shared_clone()`.

KrisThielemans:
I guess you meant "certain method take as argument `shared_ptr<ProjDataInfo> const&`"? 

time to do some light reading... https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/

rijobro:
Removed the non-const version and used the `create_shared_clone()` where necessary (4 occasions: https://github.com/UCL/STIR/pull/470/commits/c48f168e218d6cc4ce6c676781ae7b0906951acc). 

Thoughts?

rijobro:
Travis is green although it's not listed in the above checks: https://travis-ci.org/github/UCL/STIR/builds/669237921.

What do you think? 

Overall I think the PR is positive as it removes the `shared_ptr<ProjDataInfo> get_proj_data_info_sptr() const` methods that returned non-const shared pointers.

KrisThielemans:
ok. let's go ahead with this. However, needs an item in the Release notes in "backwards incompatibilities" (for developers). (do a `[ci skip]' commit)

Travis is green. Looks like we'll have to check https://travis-ci.org/github/UCL/STIR/pull_requests until Github/Travis gets sorted out.

KrisThielemans:
did you squash merge this?

rijobro:
Yes, I squash merged.

PR #477, opened 2020-03-30
Demo expansion

### LINKED ISSUE #465: add C++ demo on using objective function hierarchy
We need a demo like the ones in `examples/src` that creates an objective function and compute its value and gradient.

Sadly, it's currently somewhat involved in STIR to create an objective function from C++. We could do it via a reconstruction parameter file, and then use `get_objective_function`.

Code to do all this is in https://github.com/UCL/STIR/blob/master/src/recon_test/test_PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx

robbietuk:
This PR will provide some small modifications and an expansion to the current C++ STIR demos. Fixes #465 

demo4 - a demo to illistrate the objective function computation. [Complete]

demo5 - a demo to illistrate objective function gradient computation [In progress]

demo6 - ???

I thought these should all be added in a single PR.
N.B. We will need to squash merge this, I got confused with my git.


robbietuk:
@rijobro I have been working on this to get use to STIR C++.

examples/src/demo4.cxx
```diff
@@ -0,0 +1,128 @@
...
+
+  /////// load data from file
+  shared_ptr<DiscretisedDensity<3,float> > 
+    density_sptr(read_from_file<DiscretisedDensity<3,float> >(image_filename));
+
+  // Checks density_sptr is loaded correctly.
+  if (is_null_ptr(density_sptr))
+  {
+      error("density_sptr is null");
+  }
```
robbietuk: Is this required? Will the `if (is_null_ptr(density_sptr))  = False `

KrisThielemans: Do you mean "will `read_from_file` throw an exception or return a 0 pointer? I can't recall. Check the documentation... (or the code...), or just try!

if it is required, best to give a more informative message.

examples/src/demo4.cxx
```diff
@@ -0,0 +1,128 @@
...
+
+  /////// setup objective function object
+  objective_function_sptr->set_up(density_sptr);
+
+  /////// Compute objective function value
+  double my_val = objective_function_sptr->compute_objective_function(*density_sptr);
```
KrisThielemans:
```suggestion
  const double my_val = objective_function_sptr->compute_objective_function(*density_sptr);
```

examples/src/demo_obj_fun.par
```diff
@@ -0,0 +1,25 @@
...
+		    disable caching := 0
+		    store only basic bins in cache := 1
+		    restrict to cylindrical fov := 1
+		    number of rays in tangential direction to trace for each bin := 50
+		    use actual detector boundaries := 1
+		    do symmetry 90degrees min phi := 1
+		    do symmetry 180degrees min phi := 1
+		    do symmetry swap segment := 1
+			do symmetry swap s := 1
+	    	do symmetry shift z := 1
```
KrisThielemans: I'd cut all of this for this demo

examples/src/generate_image.par
```diff
@@ -1,14 +1,4 @@
...
-
-; optional: values: PET|nucmed *defaults to PET for backwards compatibility)
-imaging modality:=PET
-; optional patient position keywords (defaulting to "unknown")
-; orientation: allowed values: head_in|feet_in|other|unknown
-patient orientation := head_in
-; rotation: allowed values: prone|supine|other|unknown
-patient rotation :=  supine
-image duration (sec) := 1 ; defaults to -1 (i.e. unknown)
-
```
KrisThielemans: hmmm. you're removing it again?

examples/src/small.hs
```diff
@@ -1,14 +1,6 @@
...
 name of data file := small.s
-!imaging modality := PET
 originating system := RATPET
-; optional: values: PET|nucmed *defaults to PET for backwards compatibility)
-imaging modality:=PET
-; optional patient position keywords (defaulting to "unknown")
-; orientation: allowed values: head_in|feet_in|other|unknown
-patient orientation := head_in
-; rotation: allowed values: prone|supine|other|unknown
-patient rotation :=  supine
```
KrisThielemans: same here

KrisThielemans:
why not combine demo 4 and 5?  99% of the code will be the same

KrisThielemans:
Do you want to merge this on top of the release branch before merging to master? Might be easiest. Don't do the merges yet. first let me know.

KrisThielemans:
@robbietuk this seems ok to merge (to release_4), even without the additional demos. Possibly rename `demo4.cxx` to `demo4_objFun.cxx` first. Or we can postpone, as you prefer,

robbietuk:
Hmm, I dont think this is where we really want it, I think there is a lot more that should be incorporated like gradient computations and a basic gradient ascent (I had problems there due to lack of preconditioner/ normalisation).

KrisThielemans:
Sure. We could merge in pieces, but as you prefer. I'll postpone to 5.0 then

robbietuk:
Okay I accidenrtly force pushed after a rebase onto master... :S Can I reverse, shall I restart?

Regardless, this demo4_obj_fun does gradient ascent, which doesn't work if step size or number of iterations is much larger than the example as it is unconstrained and unpreconditioned). Is this okay? Should I just compute the gradient and save it? The optimisation doesnt really work. Perhaps it could work if a preconditioner (EM?) was incorporated, but I think it makes this demo very complicated and too simiular to OSMAPOSL.

robbietuk:
Im not sure how to make Codacy happy
`The class 'MyStuff' does not have a constructor although it has private member variables.`
Any suggestions would be nice.

examples/src/README.txt
```diff
@@ -36,6 +36,15 @@ demo3.cxx
 	Note that the same functionality could be provided without deriving
 	a new class from ParsingObject. One could have a KeyParser object
 	in main() and fill it in directly.
+
+demo4.cxx
```
KrisThielemans, line 40:
```suggestion
demo4_obj_fun.cxx
```

examples/src/README.txt
```diff
@@ -58,6 +67,11 @@ demoPM.par
 	backprojector that uses a projection matrix (using the ray tracing 
 	model)
 
+demo_obj_fun.par
+	An example parameter file for demo4.cxx, used to
```
KrisThielemans, line 71:
```suggestion
	An example parameter file for demo4_obj_fun.cxx, used to 
```

examples/src/README.txt
```diff
@@ -102,22 +116,27 @@ First you need to create some data.
...
+	# you can display the output using for instance
+	manip_image output.hv
       
-
+# demo2
 	$EXE_LOC/demo2
-# demo2 contains a call stir::display, so you'll see the display immediately
-( at least when on Unix)
-
+	# demo2 contains a call stir::display, so you'll see the display immediately ( at least when on Unix)
```
KrisThielemans, line 127:
```suggestion
	# demo2 contains a call to `stir::display`, so you'll see the display immediately ( at least when on Unix)
```

examples/src/demo4_obj_fun.cxx
```diff
@@ -0,0 +1,153 @@
...
+#include "stir/is_null_ptr.h"
+
+namespace stir {
+
+class MyStuff: public ParsingObject
+{
+public:
+  void set_defaults();
+  void initialise_keymap();
+  bool post_processing();
```
KrisThielemans, line 46: make all these private. Add a default constructor that calls `set_defaults()`

robbietuk: Doing the above, calling `set_defaults()` from constructor, when `set_defaults()` is private/protected wont work. It needs to be public.

robbietuk:
Either way have both pros and cons. I feel doing the gradient ascent is the best, it demonstrates how general iterative optimisation works and gives some basis to how methods like OSMAPOSL work.  It doesn't work if the parameters are too out of scope but it does work with the defaults (ish). It does improve the objective function value alot and results in an okay (ish) image. 
image


I vote to keep it how it is but am easily swayed if you feel it would be better to simply save the gradient image.

KrisThielemans:
fine for me to leave it as gradient ascent. Just write a comment then that this does not take positivity into account ("exercise for the reader")

robbietuk:
Remember this is linked to #465

KrisThielemans:
great. Can you add a line in `release_5.0.htm`? (as we're merging on `master`). Then I'll squash merge.

Issue #479, opened 2020-04-01
#470 broke backwards compatibility too much


KrisThielemans:
Since merging #470, SIRF (and everyone else using STIR) will now get errors like https://travis-ci.org/github/CCPPETMR/SIRF/jobs/669274690#L22194
```
/home/travis/build/CCPPETMR/SIRF/src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h:85:24: error: no matching function for call to stir::ProjDataInterfile::ProjDataInterfile(std::shared_ptr<stir::ExamInfo>, std::shared_ptr<const stir::ProjDataInfo>, const string&, std::_Ios_Openmode)
```
caused by `get_proj_data_info_sptr` return now a `shared_ptr<const ProjDataInfo>`.

I suggest to add a constructor for `ProjDataInterfile`, `ProjDataInMemory` (and how many others?) that accepts this. They presumably will have to create their own clone. We could keep the original version as well, as this will allow people to pass a `shared_ptr<ProjDataInfo>`  and save the extra clone. (This constructor is a bit unsafe, but we do preserve the extra safety introduced by #470).

KrisThielemans:
Doing a search on the use of `shared_ptr<ProjDataInfo>` in our `.*.h` files throws up about 90 places where this should be changed... Creating "double" constructors/`set_up*` functions etc is far too much work and would lead to far too much code duplication.

Seems it would be best to stick to using `shared_ptr<const ProjDataInfo>` in most places, indeed even in the class members. That would be better in any case (no scope for an object to sneakily modify a pointer that then would affect everybody else).

However, I have no idea if doing this wholesale (i.e. not keeping any `shared_ptr<ProjDataInfo>`  members/arguments at all), will be a ton of work, and/or will break too many things. In some sense, if we need to make changes in STIR code that uses the classes, then we know we will have trouble in SIRF/other places as well.

rijobro:
Would you prefer to revert https://github.com/UCL/STIR/pull/470 until we sort this out?

KrisThielemans:
depends a bit how long we're going to take with the suggested approach here. Would you be able to try something this morning? Possibly adding `const` in more places in the class interfaces (only) might not be too hard

rijobro:
> Seems it would be best to stick to using shared_ptr<const ProjDataInfo> in most places, indeed even in the class members.

I agree with this. Can you think of any cases where the `ProjDataInfo` **shouldn't** be `const`?

[...]

PR #480, opened 2020-04-01
shared_ptr<ProjDataInfo> -> shared_ptr<const ProjDataInfo>

rijobro:
Internally use shared_ptr<const ProjDataInfo> where possible

### DIFF
--- a/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx
+++ b/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx
@@ -304,7 +304,7 @@ set_up_before_sensitivity(shared_ptr<TargetT > const& target_sptr)
       return Succeeded::no;
     }
 
-  shared_ptr<ProjDataInfo> proj_data_info_sptr(
+  const shared_ptr<ProjDataInfo> proj_data_info_sptr(
 					       (this->_dyn_proj_data_sptr->get_proj_data_sptr(1))->get_proj_data_info_sptr()->clone());
   proj_data_info_sptr->
     reduce_segment_range(-this->_max_segment_num_to_process,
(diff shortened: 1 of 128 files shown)


rijobro:
fixes #479

KrisThielemans:
This is too much for me :-(

Instead, I'd like to see a PR that only changes .h files, and corresponding internal implementations. No "outside" code should be affected (I hope)

rijobro:
don't think I can do more minimal than the second commit.

rijobro:
> No "outside" code should be affected (I hope)

Outside code will be affected if e.g., `get_proj_data_info` was called, since the result will now be `const`, so the outside code needs to either clone or store it as a const, too.

src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.txx
```diff
@@ -304,7 +304,7 @@ set_up_before_sensitivity(shared_ptr<TargetT > const& target_sptr)
       return Succeeded::no;
     }
 
-  shared_ptr<ProjDataInfo> proj_data_info_sptr(
+  const shared_ptr<ProjDataInfo> proj_data_info_sptr(
```
KrisThielemans, line 307: not really needed, and maybe somewhat confusing. (it's a const pointer to a non-const object), but as you prefer

rijobro:
Ok to (squash) merge this? Travis is happy.

KrisThielemans:
Let's remove the `Use with care` warnings in the proj-data classes (do a search in `include/stir` only, not in subdirs). Then squash merge

KrisThielemans:
ah, you're faster than me! A tiny PR for that then I guess

rijobro:
Did that in the last commit before merge, is this what you wanted? https://github.com/UCL/STIR/pull/480/commits/f92a502d315387ab425a1de93509cb83f7bb3d12

KrisThielemans:
you're even faster than I think you are! great. thanks

PR #481, opened 2020-04-02
Convert most shared_ptr<DiscretisedDensity> arguments to shared_ptr<const...>

KrisThielemans:
This increases safety, similar to #470

### DIFF
--- a/src/Shape_buildblock/DiscretisedShape3D.cxx
+++ b/src/Shape_buildblock/DiscretisedShape3D.cxx
@@ -35,9 +35,9 @@ void
 DiscretisedShape3D::
 set_origin(const CartesianCoordinate3D<float>& new_origin)
 { 
-  assert(this->get_origin() == density_ptr->get_origin());
+  assert(this->get_origin() == density_sptr->get_origin());
   Shape3D::set_origin(new_origin);
-  density_ptr->set_origin(new_origin);
+  density_sptr->set_origin(new_origin);
 }
 
 // TODO check code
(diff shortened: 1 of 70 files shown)


KrisThielemans:
@rijobro this should handle most DiscretisedDensity case. It might need a SIRF change as well. Would you be able to check?

Did you have a look at `exam_info`? If not, probably easiest to add it on top of this one.

KrisThielemans:
Travis is ok

rijobro:
Will have a look this morning. Also need to check that it compiles in Debug mode, to avoid needing a repeat of #483.

rijobro:
Should perhaps think about having a travis in debug mode.

KrisThielemans:
I've done most of the exam-info. Was sadly more work than I hoped. I will push this soon

src/Shape_buildblock/DiscretisedShape3D.cxx
```diff
@@ -127,10 +127,10 @@ DiscretisedShape3D(const VoxelsOnCartesianGrid<float>& image_v)
 
 
 DiscretisedShape3D::
-DiscretisedShape3D(const shared_ptr<DiscretisedDensity<3,float> >& density_ptr_v)
-   : density_ptr(density_ptr_v)
+DiscretisedShape3D(const shared_ptr<const DiscretisedDensity<3,float> >& density_ptr_v)
```
rijobro, line 130: change variable names to sptr?

rijobro:
Ok thanks. So far everything builds on debug, too. In terms of review, didn't see anything out of the ordinary, but there was too much to look at in detail.

rijobro:
From SIRF:
```
auto sptr_im = prior.get_anatomical_image_sptr();
shared_ptr<STIRImageData> sptr_id(new STIRImageData(sptr_im));
return newObjectHandle(sptr_id);
```
Constructor of STIRImageData doesn't take a const DiscDensity as it needs to be able to modify the data if desired. This means lots of copying of data, but unavoidable I think.

KrisThielemans:
SIRF PR is https://github.com/CCPPETMR/SIRF/pull/615

PR #490, opened 2020-04-09
Change treatment of vectorised keys in KeyParser

### LINKED ISSUE #148: KeyParser needs to know about vectored keys
Currently the fact that a key is vectored or not is stored in the keymap, instead it is decided at parsing time (if there's a `[1]` or so, it'll decide it's a vectored key, and therefore assume that the variable it needs to store the value in is a std::vector; and vice-versa). This is unsafe and can lead to seg-faults.

KrisThielemans:
`KeyParser` did not store if a key is vectorised or not, but just assumed
that the stream it parsed was correct. This led to segmentation faults when
the stream used a vectorised key when KeyParser did not expect it, or vice versa.

This commit forces the developer to explicitly say what the level of vectorisation
is for a certain keyword (i.e. key, key[1], or key[1][1]).

This means that the change is backwards incompatible.

Classes using KeyParser have been rewritten to use the "shorter" form
of `add_key`, such that the vectorisation can be found by type-matching.

Also introduced a new `ignore_key()` member to make things more concise.

Fixes #148

### DIFF
--- a/documentation/release_4.0.htm
+++ b/documentation/release_4.0.htm
@@ -440,6 +440,10 @@ from the above):</H2>
         This is because the class only exists for the parser to get information about the class.</li>
   </ul>
 </li>
+ <li><code>KeyParser</code> <a href="https://github.com/UCL/STIR/pull/490">has changed</a> to be safer for
+   parsing of vectorised keys. However, the &quot;simple&quot; <code>add_key</code>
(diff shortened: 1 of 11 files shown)


KrisThielemans:
This is working fine. I need to update the documentation. Will do that with force-push, as current doc isn't 100% correct.

Issue #491, opened 2020-04-11
LPS returns SPL


KrisThielemans:
@ashgillman I find the following confusing:
```c++
CartesianCoordinate3D<float> LPS = image.get_LPS_for_indices();
const float S = LPS[1];
const float L = LPS[3];
```
i.e. I'd expect that LPS coordinates are in order LPS.

I guess you did this because of the usual STIR `z,y,x` convention (where `z` is along the axis of the scanner), but we never call this `get_xyz`.

This seems quite non-trivial to change however. Moreover, it would also have repercussions for SIRF I guess (@rijobro @evgueni-ovtchinnikov, please confirm).

Not sure what to do with this...

ashgillman:
Sorry, I missed this.

Hmm, looks like this is handled here:
https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/include/stir/DiscretisedDensity.inl#L218

I use the CartesianCoordinate3D.{x(),y(),z()} interface as opposed to indexing. So it would seem the issue might be with the use of that data type?

PR #492, opened 2020-04-11
added options to list_image_values

KrisThielemans:
New options for output in LPS, CSV format, and suppress first line.
Default is same behaviour as before (although with an extra space)


KrisThielemans:
Note that I had to follow the convention noted in #491 to get correct results

PR #496, opened 2020-04-15
apply_normfactors3D: 

danieldeidda:
efficiency file was not read successfully into the array

added a for loop to read it


KrisThielemans:
seems you've merged `master` on top. can't do that in this case.

KrisThielemans:
However, I think this change will break things. It seems to me that you are now reading efficiencies from a space-separated file. However, `apply_normfactors3D` is designed to run on output of `find_ML_normfactors3D`. This uses `stream.h` to output a file like `{{1,2,3},{4,5,6}}`. Therefore, the current code will read it correctly.

Possibly you've created your efficiencies yourself?

danieldeidda:
> 
> 
> However, I think this change will break things. It seems to me that you are now reading efficiencies from a space-separated file. However, `apply_normfactors3D` is designed to run on output of `find_ML_normfactors3D`. This uses `stream.h` to output a file like `{{1,2,3},{4,5,6}}`. Therefore, the current code will read it correctly.
> 
> Possibly you've created your efficiencies yourself?

Mh I see. Yes I created it to much what I thought would be the STIR format (I had a look at the mmr eff file). I can actually reshape the sensitivity to match the  `{{1,2,3},{4,5,6}}` (brackets included?)  then we can ignore this pull request.

There is one thing though: it seems apply_normfactors3D uses a projdata member that does not exist anymore

KrisThielemans:
yes, brackets and commas are necessary. EOL and whitespace is ignored. (This is from the days when I was using Mathematica a lot...)

> There is one thing though: it seems apply_normfactors3D uses a projdata member that does not exist anymore

Surprising as Travis is green (both on master and release_4.0.0 branches). can you be more specific?

danieldeidda:
> yes, brackets and commas are necessary. EOL and whitespace is ignored. (This is from the days when I was using Mathematica a lot...)
> 
> > There is one thing though: it seems apply_normfactors3D uses a projdata member that does not exist anymore
> 
> Surprising as Travis is green (both on master and release_4.0.0 branches). can you be more specific?

Sorry, I had this problem because I merged everything with master but apply_normfactors3D.cxx with master, then the get_progdata_ptr() was not updated.
I am gonna close this

PR #506, opened 2020-04-21
additions for Siemens mCT and other scanner with virtual crystals

KrisThielemans:
Replaces #352, but put in the release_4.0.0 branch


KrisThielemans:
figured out how to put #352 on release_4.0.0,so losing this such that we have the comments there

Issue #126, opened 2017-12-10
new clang warning about undefined-var-template


KrisThielemans:
```/home/travis/build/UCL/STIR/src/include/stir/RegisteredParsingObject.h:129:51: warning: instantiation of variable 'stir::QuadraticPrior<float>::registered_name' required here, but no definition is available [-Wundefined-var-template]
```
[Possible solution](https://stackoverflow.com/questions/47230290/xcode-9-only-how-to-resolve-the-wundefined-var-template-warning-caused-by-s) seems to imply we need to define `registered_name` for every type, not with a template.

KrisThielemans:
I'll give a more complete example here (but look at any Travis OSX log)
```
In file included from /Users/travis/build/UCL/STIR/src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h:40:
In file included from /Users/travis/build/UCL/STIR/src/include/stir/RegisteredParsingObject.h:142:
/Users/travis/build/UCL/STIR/src/include/stir/RegisteredParsingObject.inl:34:21: warning: instantiation of variable 'stir::OSMAPOSLReconstruction<stir::DiscretisedDensity<3, float> >::registered_name' required here, but no definition is available [-Wundefined-var-template]
  { return Derived::registered_name; }
                    ^
/Users/travis/build/UCL/STIR/src/include/stir/RegisteredParsingObject.h:86:7: note: in instantiation of member function 'stir::RegisteredParsingObject<stir::OSMAPOSLReconstruction<stir::DiscretisedDensity<3, float> >, stir::Reconstruction<stir::DiscretisedDensity<3, float> >, stir::IterativeReconstruction<stir::DiscretisedDensity<3, float> > >::get_registered_name' requested here
class RegisteredParsingObject : public Parent
      ^
/Users/travis/build/UCL/STIR/src/include/stir/OSMAPOSL/OSMAPOSLReconstruction.h:101:31: note: forward declaration of template entity is here
    static const char * const registered_name;
```
These would be nice to get rid of, as it means it's almost impossible to see what else is going wrong.

Suggested solutions seem to all say you need to include something like
```c++
template <> extern const char * const OSMAPOSLReconstruction<DiscretisedDensity<3,float> >::registered_name;
```
[...]

PR #513, opened 2020-04-23
if clang no undefined template warning

rijobro:
closes https://github.com/UCL/STIR/issues/126.


KrisThielemans:
hardly any warnings left in the OSX builds! I only see some things with SWIG, which will/should disappear on `master` . great.

PR #528, opened 2020-05-08
Modify the computation of the log-likelihood to use doubles.

### LINKED ISSUE #527: Log-Likelihood is not smooth or concave when approaching convergence
In an attempt to compute a converged image estimate using line searches I found that the objective function was non-concave close to the solution. This was discovered by computering a descent direction using the full data and then performing a gradient preconditioned ascent using an array of step sizes. For each step size image, the  log-likelihood, penalty, and objective function values were computed and plotted. See figure.

image

The issue lies with the log-likelihood value computation. Please note, the issue is only relavant when close to convergence. Fix PR to follow.

robbietuk:
Fixes #527

Use doubles in the computation of the log-likelihood value. 

This PR improves the non-smoothness of the likelihood function value in the line search experiments close to convergence, see Fig _(note likelihood y-axis scale is greatly reduced in this PR plot)_.
image

Possible improvement is to precompute a single double(new_estimate) before multiple calls.
i.e. 
```
new_estimate_double = double(new_estimate);
sub_result += projection_data[r][b]*log(new_estimate_double) - new_estimate_double;
```

### DIFF
--- a/documentation/release_4.1.htm
+++ b/documentation/release_4.1.htm
@@ -15,7 +15,7 @@
       <ul>
         <li>Kris Thielemans (UCL) for general overview, fixes, help and maintenance</li>
         <li>Richard Brown (UCL) for wrapping of NiftyPET's GPU functionality for projecting and unlisting</li>
-        <li>Robert Twyman (UCL) for Relative Differnce Prior and compuation log-likelihood modification</li>  
+        <li>Robert Twyman (UCL) for Relative Difference Prior and computation of the log-likelihood modification</li>  
       </ul>
     </p>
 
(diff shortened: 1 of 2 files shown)


robbietuk:
Sorry about the inconsitancy between the positive/negitive RDP plot between this PR and #527, I corrected the plotting. Other plots are unaffected.

KrisThielemans:
interesting! Much better now (although still not amazingly great).  I guess that the only way to avoid this is to do the whole line search with doubles (i.e. double(sino)+alpha*double(grad)). That'd require quite  a lot of changes though.

I cannot imagine that precomputing `new_estimate_double` is going to make any difference. Any decent compiler will optimise that already.

Of course, it makes sense that numerical precision is going to matter when log-lik and penalty almost cancel each-other out.

I think the trick now has to be that a line search cannot take too small steps. Obviously, it'll be hard to know what too small is...

Please add something in the release notes.

KrisThielemans:
@robbietuk ok to merge?

KrisThielemans:
ah. add a line in the release notes first please

robbietuk:
Done. Sorry, forgot [ci -skip]

KrisThielemans:
ok to squash-merge?

robbietuk:
Yes. Sorry for spelling, I am multitasking.

PR #536, opened 2020-05-12
Tew fix 524

danieldeidda:
the window width are assumed to be different and both ar used in the estimation.


KrisThielemans:
@mastergari can you have a look?

KrisThielemans:
Excellent. @danieldeidda can you then add a line in release 4.1.htm? I'll merge then

PR #538, opened 2020-05-13
fix SyneRBI/SIRF#654:

danieldeidda:
moved non-parsing operations in set_up().


KrisThielemans:
sorry. would you mind putting this on release_4.0 branch? I can't face thinking about releasing 4.1 right now...

(obviously, would have to be only this change, not base the whole PR on that branch)

PR #540, opened 2020-05-13
fix SyneRBI/SIRF#654:

danieldeidda:
moved non-parsing operations in set_up().

### DIFF
--- a/src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
+++ b/src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
@@ -210,7 +210,7 @@ public:
   std::vector<shared_ptr<TargetT> > anatomical_prior_sptrs,kmnorm_sptrs;
   shared_ptr<TargetT> kpnorm_sptr;
  //kernel parameters
-  int num_neighbours,num_non_zero_feat,num_elem_neighbourhood,num_voxels,dimz,dimy,dimx;
+  int num_neighbours,num_non_zero_feat,num_elem_neighbourhood,num_voxels,dimz,dimy,dimx, an_size;
   std::vector<double> sigma_m;
   bool only_2D;
   bool hybrid;
@@ -230,7 +230,7 @@ private:
   friend void do_sensitivity(const char * const par_filename);
 
   //! operations prior to the iterations
-  virtual Succeeded set_up(shared_ptr <TargetT > const& target_image_ptr);
+  virtual Succeeded set_up(shared_ptr <TargetT > const& target_image_sptr);
  
   //! the principal operations for updating the image iterates at each iteration
   virtual void update_estimate (TargetT& current_image_estimate);
--- a/src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
+++ b/src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
@@ -173,9 +173,9 @@ public:
    one place, all objects that use the shared pointer will be affected.
   */
  //@{
-  void set_anatomical_prior_sptrs(shared_ptr<TargetT>, int index);
+  void set_anatomical_prior_sptr(shared_ptr<TargetT>, int index);
   //! sets all elements of vector anatomical_prior to the same value
-  void set_anatomical_prior_sptrs(shared_ptr<TargetT> arg);
+  void set_anatomical_prior_sptr(shared_ptr<TargetT> arg);
   void set_anatomical_image_filename(const std::string&, const int index);
   void set_anatomical_image_filename(const std::string&);
 
--- a/src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
+++ b/src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
@@ -210,7 +210,7 @@ public:
   std::vector<shared_ptr<TargetT> > anatomical_prior_sptrs,kmnorm_sptrs;
   shared_ptr<TargetT> kpnorm_sptr;
  //kernel parameters
-  int num_neighbours,num_non_zero_feat,num_elem_neighbourhood,num_voxels,dimz,dimy,dimx, an_size;
+  int num_neighbours,num_non_zero_feat,num_elem_neighbourhood,num_voxels,dimz,dimy,dimx;
   std::vector<double> sigma_m;
   bool only_2D;
   bool hybrid;
--- a/src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
+++ b/src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
@@ -164,7 +164,7 @@ public:
   const bool get_only_2D() const;
   const bool get_hybrid()const;
 
-  std::vector<shared_ptr<TargetT> > get_anatomical_prior_sptr();
+  std::vector<shared_ptr<TargetT> > get_anatomical_prior_sptrs();
 //@}
 
     /*! \name Functions to set parameters
@@ -173,11 +173,11 @@ public:
    one place, all objects that use the shared pointer will be affected.
   */
  //@{
-  void set_anatomical_prior_sptr(shared_ptr<TargetT>, int index);
+  void set_anatomical_prior_sptrs(shared_ptr<TargetT>, int index);
   //! sets all elements of vector anatomical_prior to the same value
-  void set_anatomical_prior_sptr(shared_ptr<TargetT> arg);
-  void set_anatomical_image_filenames(const std::string&, const int index);
-  void set_anatomical_image_filenames(const std::string&);
+  void set_anatomical_prior_sptrs(shared_ptr<TargetT> arg);
+  void set_anatomical_image_filename(const std::string&, const int index);
+  void set_anatomical_image_filename(const std::string&);
 
 
   void set_num_neighbours(const int);
@@ -207,7 +207,7 @@ public:
   //! Anatomical image filename
   std::vector<std::string> anatomical_image_filenames;
 
-  std::vector<shared_ptr<TargetT> > anatomical_prior_sptr,kmnorm_sptr;
+  std::vector<shared_ptr<TargetT> > anatomical_prior_sptrs,kmnorm_sptrs;
   shared_ptr<TargetT> kpnorm_sptr;
  //kernel parameters
   int num_neighbours,num_non_zero_feat,num_elem_neighbourhood,num_voxels,dimz,dimy,dimx;
--- a/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
+++ b/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
@@ -377,12 +377,17 @@ std::vector<shared_ptr<TargetT> > KOSMAPOSLReconstruction<TargetT>::get_anatomic
   set_ functions
 ***************************************************************/
 
+
 template<typename TargetT>
 void
 KOSMAPOSLReconstruction<TargetT>::
 set_anatomical_prior_sptr (shared_ptr<TargetT> arg, int index)
 {
-  this->anatomical_prior_sptr[index] = arg;
+    int n = this->anatomical_prior_sptr.size();
+    if (index < n)
+        this->anatomical_prior_sptr[index] = arg;
+    else
+        this->anatomical_prior_sptr.push_back(arg);
 }
 
 template<typename TargetT>
@@ -688,8 +693,13 @@ void KOSMAPOSLReconstruction<TargetT>::estimate_stand_dev_for_anatomical_image(s
                                   }
                                }
                        }
-
-       SD[i]=((double)sqrt(kStand_dev/(nv-1) ));//std::cout<<"SD :"<<SD[i]<<std::endl;
+                      int nsd = SD.size();
+                      double s = ((double)sqrt(kStand_dev/(nv-1) ));
+                      if (i < nsd)
+                          SD[i] = s;
+                      else
+                          SD.push_back(s);
+      /* SD[i]=((double)sqrt(kStand_dev/(nv-1) ));*///std::cout<<"SD :"<<SD[i]<<std::endl;
     }
 }
 
--- a/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
+++ b/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
@@ -205,7 +205,6 @@ post_processing()
   if (base_type::post_processing())
     return true;
   
-  this->subiteration_counter=0;
   this->anatomical_sd.resize(anatomical_image_filenames.size());
 
   if (this->anatomical_image_filenames.size()!=sigma_m.size()){
@@ -213,12 +212,6 @@ post_processing()
       return false;
   }
 
-
-  if(!this->only_2D){
-     this->num_elem_neighbourhood=this->num_neighbours*this->num_neighbours*this->num_neighbours ;}
-  else{
-     this->num_elem_neighbourhood=this->num_neighbours*this->num_neighbours ;
-      }
 for(int i = 0; i<=anatomical_image_filenames.size()-1; i++)
 {  if (!this->anatomical_image_filenames.empty()){
       this->anatomical_prior_sptr.push_back(shared_ptr<TargetT>(read_from_file<TargetT>(anatomical_image_filenames[i])));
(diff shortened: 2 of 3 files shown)


src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -302,6 +258,51 @@ set_up(shared_ptr <TargetT > const& target_image_ptr)
  if (base_type::set_up(target_image_ptr) == Succeeded::no)
      error("KOSMAPOSL::set_up(): Error building post filter");
```
KrisThielemans, line 259: actually, lots of other things could have gone wrong. let's just say
```suggestion
     error("KOSMAPOSL::set_up(): Error setting-up underlying OSMAPOSLReconstruction object");
```
or similar

danieldeidda:
> 
> 
> looks good to me. I haven't tried it of course. I'm assuming it still works?

yes everything looks the same

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -205,20 +205,13 @@ post_processing()
...
   }
 
-
-  if(!this->only_2D){
-     this->num_elem_neighbourhood=this->num_neighbours*this->num_neighbours*this->num_neighbours ;}
-  else{
-     this->num_elem_neighbourhood=this->num_neighbours*this->num_neighbours ;
-      }
 for(int i = 0; i<=anatomical_image_filenames.size()-1; i++)
 {  if (!this->anatomical_image_filenames.empty()){
```
KrisThielemans, line 216: This test doesn't make sense, as if `anatomical_image_filenames` is `empty()`, it won't go into this loop. Maybe you meant `anatomical_image_filenames[i].empty()`? I think this is what happened in https://github.com/SyneRBI/SIRF/issues/654. `set_defaults` sets it to `""`. It'll try to read that then, hence the error message.

Checking if a string is empty feels somewhat strange  to me, and possibly it'd be better to have `set_defaults` call `anatomical_image_filenames.clear()` (and also for other `vector`s). That seems to lead us somewhat further away though.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -300,8 +256,56 @@ set_up(shared_ptr <TargetT > const& target_image_ptr)
 {
 
  if (base_type::set_up(target_image_ptr) == Succeeded::no)
-     error("KOSMAPOSL::set_up(): Error building post filter");
-
+     error("KOSMAPOSL::set_up(): Error setting-up underlying OSMAPOSLReconstruction object");
+
+ if (this->anatomical_image_filenames.empty())
+     error("KOSMAPOSL::set_up(): anatomical_image_filenames has not been set");
```
KrisThielemans, line 262: now you prevent using this class without filenames at all. Intentional? Or do you want
```suggestion
 if (this->anatomical_prior_sptrs.empty())
     error("KOSMAPOSL::set_up(): no anatomical prior-info set");
```
On the other hand, with @ashgillman we thought we could allow this size to be 0, such that we can do "PET-only kernel" stuff. That might be going too far here though.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -205,20 +205,13 @@ post_processing()
   if (base_type::post_processing())
     return true;
   
-  this->subiteration_counter=0;
   this->anatomical_sd.resize(anatomical_image_filenames.size());
 
   if (this->anatomical_image_filenames.size()!=sigma_m.size()){
```
KrisThielemans, line 210: this test should be repeated in `set_up` but with `anatomical_prior_sptrs.size()`

KrisThielemans:
In fact, `get_anatomical_prior_sptr` has the wrong name, as it does not return a shared-ptr to an anatomical-prior, but to a vector of them. Hence I was writing `get_anatomical_prior_sptrs` above. (Could change the internal member as well, but less important as it's private).

On the other hand, `set_anatomical_image_filenames` functions should be called `set_anatomical_image_filename` as tehy set a single filename.

I don't understand https://github.com/UCL/STIR/blob/37c77b5fb44c7567fa53b498d39da7c48af849d4/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx#L396-L398
either. Are you saying you want to have all anatomical images the same? That looks weird to me. Shouldn't this be
```
this->anatomical_prior_sptr.resize(1);
this->anatomical_prior_sptr[0] = arg;
```

https://github.com/UCL/STIR/blob/37c77b5fb44c7567fa53b498d39da7c48af849d4/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx#L639 should not use the filenames member. also below, e.g. https://github.com/UCL/STIR/blob/37c77b5fb44c7567fa53b498d39da7c48af849d4/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx#L713 https://github.com/UCL/STIR/blob/37c77b5fb44c7567fa53b498d39da7c48af849d4/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx#L797

apologies that I didn't catch these when reviewing the original PR
Same remark for `set_sigma_m(s)` at https://github.com/UCL/STIR/blob/37c77b5fb44c7567fa53b498d39da7c48af849d4/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx#L446.

https://github.com/UCL/STIR/blob/37c77b5fb44c7567fa53b498d39da7c48af849d4/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx#L438 should use `.at()` to avoid seg faults (or better for now, grow like the others)

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: shall I make the check on the anatomical prior sptr?

danieldeidda: like `is_null_ptr(this->anatomical_prior_sptr[0])`

KrisThielemans: I feel that in `post_processing` a check on filenames is appropriate. `set_up` should check indeed on `is_null_ptr(this->anatomical_prior_sptr[i])`

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: mh no I didn't want that actually. maybe the check can be on anatomical_prior==null and hybrid==0

KrisThielemans: I guess `hybrid!=0` then. fine for me (for now certainly)

danieldeidda: > 
> 
> I guess `hybrid!=0` then. fine for me (for now certainly)

theoretically it should be hybrid!=0 but the iterative part of the kernel is calculated only when hybrid==1
maybe is less confusing to check on sigma_m!=0 if that is zero we can ignore the anatomical part

KrisThielemans: I can't say I understand that one :smile:. For clarity I was thinking you'd want to check
```
if (anatomical_prior==null and hybrid!=0)
  error()
```
A separate check should be (in `set_up`) that sizes are equal. In fact, you could move the whole check there as well I guess.

Tricky to get this all correct of course. I do now think that the `clear()` in `set_defaults()` will be necessary, as otherwise you could have size mismatches if you don't parse.

sorry

danieldeidda: > 
> 
> I can't say I understand that one 😄. 

Sorry for the confusion. Let's say we would like to run KOSMAPOSL only with the functional (iterative) side information. These estimates are calculated only if hybrid =1. So if we throw an error when (anatomical=null and hybrid!=0) we are not allowing this option. if anatomical=null and hybrid=0, then we would not have any side information.

KrisThielemans:
@danieldeidda any chance you can address this today?

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: > Tricky to get this all correct of course. I do now think that the `clear()` in `set_defaults()` will be necessary, as otherwise you could have size mismatches if you don't parse.

sorry I don't understand this bit.

KrisThielemans: if `set_defaults()` sets `anatomical_filenames` to a vector of size 1 (albeit with an empty string), but leaves others vectors alone, therefore size 0, it seems that any size checks are going to fail.

Also, if we want to re-use an existing object, one way would be to call `set_defaults` again, so it'd need to get rid of existing arrays. (Although probably we don't expose `set_defaults` at the moment, so this is maybe not anything to worry about).

danieldeidda: So your suggestion to clear anatomical_filenames (and probably sigma_m) at the beginning of set_default()?

KrisThielemans: yes. all `vector`s that you have really. Hopefully it doesn't make things fall over...

danieldeidda: all member vectors?

KrisThielemans: Anything that isn't set it in `set_up`, i.e. anything that you think the user should set.

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -173,11 +173,11 @@ class KOSMAPOSLReconstruction:
    one place, all objects that use the shared pointer will be affected.
   */
  //@{
-  void set_anatomical_prior_sptr(shared_ptr<TargetT>, int index);
+  void set_anatomical_prior_sptrs(shared_ptr<TargetT>, int index);
```
KrisThielemans, line 176: sorry. this doesn't make sense to me. we're setting a single one. original choice was fine

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -173,11 +173,11 @@ class KOSMAPOSLReconstruction:
...
   //! sets all elements of vector anatomical_prior to the same value
-  void set_anatomical_prior_sptr(shared_ptr<TargetT> arg);
-  void set_anatomical_image_filenames(const std::string&, const int index);
-  void set_anatomical_image_filenames(const std::string&);
+  void set_anatomical_prior_sptrs(shared_ptr<TargetT> arg);
```
KrisThielemans, line 178: this should also be singular

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -153,6 +153,9 @@ set_defaults()
 {
   base_type::set_defaults();
 
+  this->sigma_m.clear();
+  this->anatomical_image_filenames.clear();
+
```
KrisThielemans, line 158:
```suggestion
  this->anatomical_prior_sptrs.clear();
  
```
As this can be set by the user (without parsing)

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -258,11 +259,15 @@ set_up(shared_ptr <TargetT > const& target_image_ptr)
  if (base_type::set_up(target_image_ptr) == Succeeded::no)
      error("KOSMAPOSL::set_up(): Error setting-up underlying OSMAPOSLReconstruction object");
 
- if (this->anatomical_image_filenames.empty())
-     error("KOSMAPOSL::set_up(): anatomical_image_filenames has not been set");
+ if (is_null_ptr(this->anatomical_prior_sptrs[0]) && this->sigma_m[0]!=0)
```
KrisThielemans, line 262: nope. this will break if their size is 0.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -300,8 +258,67 @@ set_up(shared_ptr <TargetT > const& target_image_ptr)
 {
 
  if (base_type::set_up(target_image_ptr) == Succeeded::no)
-     error("KOSMAPOSL::set_up(): Error building post filter");
-
+     error("KOSMAPOSL::set_up(): Error setting-up underlying OSMAPOSLReconstruction object");
+
+ if (is_null_ptr(this->anatomical_prior_sptrs[0]) &&
+     this->sigma_m.size()==0 )
```
KrisThielemans, line 264: this one still remains. the rest looks good to me.
```suggestion
if ((this->anatomical_prior_sptrs.size()==0) &&
     (this->hybrid==0))
```

KrisThielemans:
once this is done, do you want a merge or squash-merge?

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: sorry for some reason I must have pressed ctrl+z and changed back.

danieldeidda:
> 
> 
> once this is done, do you want a merge or squash-merge?

I don't mind they are mainly minor modifications

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
KrisThielemans: still not changed the first line to `(this->anatomical_prior_sptrs.size()==0)` I'm afraid. Editor playing up...

KrisThielemans:
> > once this is done, do you want a merge or squash-merge?
> 
> I don't mind they are mainly minor modifications

as long as you haven't merged it anywhere else, I'd prefer to squash-merge then.

danieldeidda:
> 
> 
> > > once this is done, do you want a merge or squash-merge?
> > 
> > 
> > I don't mind they are mainly minor modifications
> 
> as long as you haven't merged it anywhere else, I'd prefer to squash-merge then.

no I haven't

danieldeidda:
> Let me know if you're happy, and have run this on any of your own test cases.

let me check the only-functional case

danieldeidda:
actually there problems if we don't set the anatomical_image_filenames. I am looking into it

KrisThielemans:
> actually there are problems if we don't set the anatomical_image_filenames (with hybrid=1). I am looking into it

when running with a par file? sorry if I've mislead you!

danieldeidda:
`when running with a par file?`
yes, it crashes every time we have a loop from i-0 to anatomical_prior.size()-1. I need to run those operation only if  anatomical_prior.size()!=0

KrisThielemans:
That's weird. `anatomic_prior` should have length 0 then, therefore never get into the loop.

danieldeidda:
there is another problem. Some of the members like distance and kpnorm are set according to the anatomical image because during set_up we don't know have access to current estimate

KrisThielemans:
> Some of the members like distance and kpnorm are set according to the anatomical image because during set_up we don't know have access to current estimate

I'm confused. This was the same in the version on `release_4`?

In any case, you do have `target_image_ptr`, which is guaranteed to have the same dimensions.

Come to think of it, I think you should have a check at
https://github.com/UCL/STIR/blob/cce574f729fe143e35e1e1042c79300672cb0396/src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx#L277
```
   if (!anatomical_prior_sptrs[i]->has_same_characteristics(*target_image_ptr)
       error(...);
```
(should rename `target_image_ptr` to `target_image_sptr` here)

danieldeidda:
> Some of the members like distance and kpnorm are set according to the anatomical image because during set_up we don't know have access to current estimate

> I'm` confused. This was the same in the version on release_4?

The "only_functional" was never really enabled, that's why we were throwing an error when no anatomical image was set. with STIR 4.0 you could only use the kernel with no anatomical information if you set as anatomical image a mask image. I can add this modification to enable it

KrisThielemans:
@evgueni-ovtchinnikov this is all done. Can you please try SIRF with this PR before we merge?

KrisThielemans:
@evgueni-ovtchinnikov sorry. please wait.

danieldeidda:
This works for all my tests

KrisThielemans:
@evgueni-ovtchinnikov we're think we're good to go now.

KrisThielemans:
@danieldeidda I hadn't realised that "only hybrid" wasn't in STIR 4.0. I guess we're introducing a feature in a bug-fix now! Want to add a line to the release notes then?

danieldeidda:
sure

danieldeidda:
> 
> 
> @danieldeidda I hadn't realised that "only hybrid" wasn't in STIR 4.0. I guess we're introducing a feature in a bug-fix now! Want to add a line to the release notes then?

is it always release_4.0?

KrisThielemans:
> > @danieldeidda I hadn't realised that "only hybrid" wasn't in STIR 4.0. I guess we're introducing a feature in a bug-fix now! Want to add a line to the release notes then?
> 
> is it always release_4.0?

This will indeed be merged on the 4.0 branch, therefore in `release_4.0.htm`. It's a bit cheating to add an extra feature in a patch release, but I really don't feel like reverting the feature :smile:

danieldeidda:
> 
> 
> > > @danieldeidda I hadn't realised that "only hybrid" wasn't in STIR 4.0. I guess we're introducing a feature in a bug-fix now! Want to add a line to the release notes then?
> > 
> > 
> > is it always release_4.0?
> 
> This will indeed be merged on the 4.0 branch, therefore in `release_4.0.htm`. It's a bit cheating to add an extra feature in a patch release, but I really don't feel like reverting the feature 😄

right, I actually realised that I didn't the openmp addition for kernel estimation. shall I add this as well?

KrisThielemans:
> I actually realised that I didn't the openmp addition for kernel estimation. shall I add this as well?

in the release notes? sure

evgueni-ovtchinnikov:
I get this error from `hkem_reconstruction.py`:
```
ERROR: The number of sigma_m parameters must be the same as the number of anatomical image filenames
```

KrisThielemans:
@evgueni-ovtchinnikov this is with the latest commit. we fixed something 1-2 commits back.

evgueni-ovtchinnikov:
`git fetch` does nothing, hence I was using the latest `KOSMAPOSL4.0-SIRF654`.

KrisThielemans:
> I get this error from `hkem_reconstruction.py`:
> 
> ```
> ERROR: The number of sigma_m parameters must be the same as the number of anatomical image filenames
> ```

confirmed. Discussion on this at https://github.com/SyneRBI/SIRF/issues/654

PR #547, opened 2020-05-20
generate_image stdcerr filename of saved data

robbietuk:


robbietuk:
For usage in: https://github.com/UCL/STIR-GATE-Connection/pull/9#issuecomment-631372677

KrisThielemans:
it's fragile to rely on `stderr` output (could be an error). I'd rather not do this.

PR #558, opened 2020-05-29
ProjDataInMemory forward iterator

rijobro:
it would be great to have this one too. Gets the buffer, which can be used e.g., in SIRF for dot, norm, etc.

### DIFF
--- a/src/include/stir/ProjDataInMemory.h
+++ b/src/include/stir/ProjDataInMemory.h
@@ -63,9 +63,6 @@ class ProjDataInMemory : public ProjDataFromStream
 {
 public: 
     
-  //! typedef for the data type
-  typedef Array<1,float> DataType;
-    
   //! constructor with only info, but no data
   /*! 
     \param proj_data_info_ptr object specifying all sizes etc.
@@ -95,27 +92,25 @@ public:
   virtual void axpby(const float a, const ProjData& x,
                      const float b, const ProjData& y);
 
+  //! typedefs to simplify iterator
+  typedef Array<1,float>::full_iterator full_iterator;
+  typedef Array<1,float>::const_full_iterator const_full_iterator;
+
   //! start value for iterating through all elements in the array, see full_iterator
-  inline DataType::full_iterator begin_all()
-  { return buffer.begin_all(); }
+  inline full_iterator begin()
+  { return buffer.begin(); }
   //! start value for iterating through all elements in the (const) array, see full_iterator
-  inline DataType::const_full_iterator begin_all() const
-  { return buffer.begin_all(); }
-  //! start value for iterating through all elements in the array, see full_iterator
-  inline DataType::const_full_iterator begin_all_const() const
-  { return buffer.begin_all_const(); }
+  inline const_full_iterator begin() const
+  { return buffer.begin(); }
   //! end value for iterating through all elements in the array, see full_iterator
-  inline DataType::full_iterator end_all()
-  { return buffer.end_all(); }
+  inline full_iterator end()
+  { return buffer.end(); }
   //! end value for iterating through all elements in the (const) array, see full_iterator
-  inline DataType::const_full_iterator end_all() const
-  { return buffer.end_all(); }
-  //! end value for iterating through all elements in the array, see full_iterator
-  inline DataType::const_full_iterator end_all_const() const
-  { return buffer.end_all_const(); }
-    
+  inline const_full_iterator end() const
+  { return buffer.end(); }
+
 private:
-  DataType buffer;
+  Array<1,float> buffer;
   
   size_t get_size_of_buffer_in_bytes() const;
 
--- a/src/include/stir/ProjDataInMemory.h
+++ b/src/include/stir/ProjDataInMemory.h
@@ -92,33 +92,39 @@ public:
   virtual void axpby(const float a, const ProjData& x,
                      const float b, const ProjData& y);
 
-  //! typedefs to simplify iterator
+  /** @name iterator typedefs
+   *  iterator typedefs
+   */
+  ///@{
+  typedef Array<1,float>::iterator iterator;
+  typedef Array<1,float>::const_iterator const_iterator;
   typedef Array<1,float>::full_iterator full_iterator;
   typedef Array<1,float>::const_full_iterator const_full_iterator;
+  ///@}
 
-  //! start value for iterating through all elements in the array, see full_iterator
-  inline full_iterator begin()
+  //! start value for iterating through all elements in the array, see iterator
+  inline iterator begin()
   { return buffer.begin(); }
-  //! start value for iterating through all elements in the (const) array, see full_iterator
-  inline const_full_iterator begin() const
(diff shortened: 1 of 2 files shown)


KrisThielemans:
I can see the attraction, but am wondering about details. We've put in a lot of effort to make STIR access to internals safer, and now we're giving it away.

Should we really give access to the `boost::shared_array<float>`? It ties us down to `boost` (C++17 has `std::shared_ptr<float []>`. We could instead use `float *`. Of course, that implies difficulties with ownership and object life-time. It doesn't seem unreasonable though to say that the underlying object needs to remain alive.

Once we do that, we'd need to guarantee this. A (I believe) safe way to do that is to use the same mechanism as `VectorWithOffset::get_data_ptr` (and `get_const_data_ptr`). Doc is https://github.com/UCL/STIR/blob/8cd126e061fd41d65fb30a006b54b3e6bc897eb4/src/include/stir/VectorWithOffset.inl#L570-L581
If we choose that, the simplest way would be to just use `VectorWithOffset` for the buffer.

A cleaner alternative is to provide iterator access. That way, we hide implementation details, and life-time guarantees will be what people are used to. Sadly, implementing your own iterators isn't as easy as you'd like, see e.g. https://stackoverflow.com/questions/8054273/how-to-implement-an-stl-style-iterator-and-avoid-common-pitfalls. However, in this case, we can just "forward" some typedefs such as
```c++
class ProjDataInMemory
{
private:
  typedef buffer_type boost::shared_array<float>;
public:
  typedef value_type buffer_type::value_type;
  ...
  typedef iterator buffer_type::iterator;
  typedef const_iterator buffer_type::const_iterator;
 
  iterator begin() { return buffer.begin() };
  ...

};
```
This is a pain of course (also C++17 deprecated some `std::iterator` stuff). One way around this would be to use the Boost iterator library (which is what I did in [VectorWithOffset](https://github.com/UCL/STIR/blob/8cd126e061fd41d65fb30a006b54b3e6bc897eb4/src/include/stir/VectorWithOffset.h#L47-L53)). They have an interesting [shared_container](https://www.boost.org/doc/libs/1_73_0/libs/iterator/doc/html/iterator/specialized/shared_container.html) adapter that guarantees lifetime.

All that is still a pain. We could also decide to do the following
```c++
class ProjDataInMemory: public Array<1,float>
```
So now we have iterators, `+=` etc for free. However, it would mean intercepting `grow`, `recycle` and probably a few other things.

not what you signed up for?

rijobro:
I suspected you'd say something along these lines! Obviously I'd much rather have `ProjData` iterators, but doubt I'll have time to implement them.

In the short term, would you accept a `get_data` method, which as you say, returns `float*` and `const float*`. I think it's reasonable to expect the lifetime to match that of the object itself. Plus I'd put it in the doxygen.

rijobro:
By the way, travis will obviously fail until #557 is merged.

KrisThielemans:
Would have to follow same naming as `VectorWithOffset` please.

I see you writing a lot of code that we have though (although you might be writing it in SIRF).

Suggestion: make `buffer` an `Array<1,float>`, add forwarding functions for `get_data_ptr` etc. I strongly suggest to also add forwarding for `operator+=` and whatever (a few lines of code). Add similar forwarding in SIRF and done. If you add forwarding of the typedefs and `begin()` and `end(), you have (good enough) iterators.

KrisThielemans:
#561converts `buffer` to `Array<1,float>`

rijobro:
Will wait for #561 before updating this one.

rijobro:
@KrisThielemans updated to use #561.

KrisThielemans:
pre-C++11 failure. can't use `auto` yet... we need to release 4.1, and move on. sorry

src/include/stir/ProjDataInMemory.h
```diff
@@ -92,20 +95,35 @@ class ProjDataInMemory : public ProjDataFromStream
   virtual void axpby(const float a, const ProjData& x,
                      const float b, const ProjData& y);
 
+  //! start value for iterating through all elements in the array, see full_iterator
+  inline DataType::full_iterator begin_all()
```
KrisThielemans, line 99: no need to expose `begin_all` etc just stick to "normal" iterators I believe? Up to you, but would need both

rijobro: So just `begin()` and `end()` with `const` and non-`const` versions? (That's what I've done in the most recent commit anyway.)

KrisThielemans: I think having only `begin()` is fine, but having `begin_all()` as well is nicer :smile:

rijobro: ok, added too.

KrisThielemans:
want to kill some travis jobs? I can do the Appveyor ones as well as those take ages

KrisThielemans:
@rijobro, sorry, I'm a bit confused by the various commits, so thought best to say it here. We definitely need `begin()` etc.

rijobro:
I kept `begin` (`const` and not), and added `begin_all` (again `const` and not). And the same for `end`. Total of 8 methods. That what you wanted?

KrisThielemans:
yup. thanks. just couldn't see it in the review, but I guess it was outdated.

src/test/test_proj_data_maths.cxx
```diff
@@ -119,10 +119,23 @@ run_tests()
...
+    pd2.ProjData::axpby(a,x2,b,y2);
 
     check_proj_data_are_equal_and_non_zero(pd1,pd2);
+
+    // Check using iterators
+    ProjDataInMemory pd3(pd1);
+    pd3.fill(0.f);
+    ProjDataInMemory::full_iterator pd_iter = pd3.begin();
+    ProjDataInMemory::const_full_iterator x_iter = x1.begin();
+    ProjDataInMemory::const_full_iterator y_iter = y1.begin();
```
KrisThielemans, line 134: confusing to use `full_iterator` here, even though they're identical

src/include/stir/ProjDataInMemory.h
```diff
@@ -92,20 +92,45 @@ class ProjDataInMemory : public ProjDataFromStream
   virtual void axpby(const float a, const ProjData& x,
                      const float b, const ProjData& y);
 
-    
+  //! typedefs to simplify iterator
+  typedef Array<1,float>::full_iterator full_iterator;
+  typedef Array<1,float>::const_full_iterator const_full_iterator;
```
KrisThielemans, line 97: doxygen comment only applies for one line, unless you use groups. See e.g. https://github.com/UCL/STIR/blob/fd32b1f516cd7739c4cf292ddfe16144356643e7/src/include/stir/Array.h#L102-L107

Use the opportunity to include typedefs for `iterator` and `const_iterator`.

KrisThielemans:
thanks!

can you do a quick update of release_4.1.htm? We forgot to put things in there, mention axpby for Array, ProjDataInMemory and the iterators. ci skip of course.

PR #561, opened 2020-05-31
add axpby to Array and related changes in ProjData

### LINKED ISSUE #560: add axpby for Array
https://github.com/UCL/STIR/pull/557 added `axpby` (a*x+b*y) as
```c++
  virtual void ProjData::axpby(const float a, const ProjData& x,
                     const float b, const ProjData& y);
```
modifying `*this`. This would be useful to have for `stir::Array` as well.

However, Intel's [MKL axpby](
https://software.intel.com/content/www/us/en/develop/documentation/mkl-developer-reference-c/top/blas-and-sparse-blas-routines/blas-like-extensions/cblas-axpby.html) and other BLAS libraries (e.g. [ATLAS](http://www.netlib.org/atlas/developer/atlas_contrib/node54.html), [Math-BLAS](https://metacpan.org/pod/Math::BLAS)) modify `y` instead.

Note that `axpby` is not a standard BLAS function, but `axpy` is. I couldn't immediately find the BLAS standard, but [this ref](https://northstar-www.dartmouth.edu/doc/idl/html_6.2/BLAS_AXPY.html) says that `axpy` modifies `y`.

I find that convention a bit weird, but I suppose it's best to stick to BLAS, such that at some point we could use BLAS libraries.

KrisThielemans:
- add axpby to NumericVectorWithOffset and Array
- use Array::axbpy in ProjData::axbpy
- change buffer in ProjDataInMemory to Array<1,float> such that we have axbpy for free (and remove boost dependency)

Fixes #560 although pending decision on signature of axpby

### DIFF
--- a/src/buildblock/ProjData.cxx
+++ b/src/buildblock/ProjData.cxx
@@ -2,7 +2,7 @@
     Copyright (C) 2000 PARAPET partners
     Copyright (C) 2000 - 2010-10-15, Hammersmith Imanet Ltd
     Copyright (C) 2011-07-01 -2013, Kris Thielemans
-    Copyright (C) 2015, University College London
+    Copyright (C) 2015, 2020 University College London
     This file is part of STIR.
 
     This file is free software; you can redistribute it and/or modify
@@ -439,16 +439,7 @@ axpby(const float a, const ProjData& x,
         SegmentBySinogram<float> seg = get_empty_segment_by_sinogram(s);
         const SegmentBySinogram<float> sx = x.get_segment_by_sinogram(s);
         const SegmentBySinogram<float> sy = y.get_segment_by_sinogram(s);
-        SegmentBySinogram<float>::full_iterator seg_iter;
-        SegmentBySinogram<float>::const_full_iterator sx_iter;
-        SegmentBySinogram<float>::const_full_iterator sy_iter;
-        for (seg_iter = seg.begin_all(),
-            sx_iter = sx.begin_all(), sy_iter = sy.begin_all();
-            seg_iter != seg.end_all() &&
-            sx_iter != sx.end_all() && sy_iter != sy.end_all();
-        /*empty*/) {
-            *seg_iter++ = a*(*sx_iter++) + b*(*sy_iter++);
-        }
+        seg.axpby(a, sx, b, sy);
         set_segment(seg);
     }
 }
--- a/src/buildblock/ProjDataInMemory.cxx
+++ b/src/buildblock/ProjDataInMemory.cxx
@@ -9,7 +9,7 @@
 /*
     Copyright (C) 2002 - 2011-02-23, Hammersmith Imanet Ltd
     Copyright (C) 2011, Kris Thielemans
-    Copyright (C) 2019, UCL
+    Copyright (C) 2019, 2020, UCL
     This file is part of STIR.
 
     This file is free software; you can redistribute it and/or modify
@@ -55,7 +55,10 @@ START_NAMESPACE_STIR
 
 ProjDataInMemory::
 ~ProjDataInMemory()
-{}
+{
+  // release such that it can deallocate safely
+  this->buffer.release_data_ptr();
+}
 
 ProjDataInMemory::
 ProjDataInMemory(shared_ptr<ExamInfo> const& exam_info_sptr,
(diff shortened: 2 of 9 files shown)


rijobro:
Looks good. 

1. Do you think `ProjDataInMemory` will lose much performance by switching from `boost::array` to `Array<1,float>`? 
2. I'll create a ctest for this
3. Looking at Travis, the swig compilation isn't happy. I don't think I'll be very useful on that one as I don't know STIR-swig

KrisThielemans:
> Do you think ProjDataInMemory will lose much performance by switching from boost::array to Array<1,float>?

there should be no difference. Behind the scenes `Array<1,float>` is just a `float*` anyway.

> I'll create a ctest for this

I've added a simple test for `Array<4,float>` already in `test_Array.cxx`. That does check the 1D code as well. Of course, it can be improved.

> Looking at Travis, SWIG...

hmmm. example error https://travis-ci.org/github/UCL/STIR/jobs/693170880#L3014
```
/home/travis/build/UCL/STIR/src/include/stir/NumericVectorWithOffset.inl:298:20: error: no match for ‘operator=’ (operand types are ‘stir::Array<2, float>’ and ‘stir::NumericVectorWithOffset<stir::Array<1, float>, float>’)
       *this_iter++ = (*x_iter++) * a + (*y_iter++) * b;
       ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/travis/build/UCL/STIR/src/include/stir/Viewgram.h:38:0,
                 from /home/travis/build/UCL/STIR/build/src/swig/stirPYTHON_wrap.cxx:3841:
/home/travis/build/UCL/STIR/src/include/stir/Array.h:78:7: note: candidate: stir::Array<2, float>& stir::Array<2, float>::operator=(const stir::Array<2, float>&)
 class Array : public NumericVectorWithOffset<Array<num_dimensions-1, elemT>, elemT>
       ^~~~~
/home/travis/build/UCL/STIR/src/include/stir/Array.h:78:7: note:   no known conversion for argument 1 from ‘stir::NumericVectorWithOffset<stir::Array<1, float>, float>’ to ‘const stir::Array<2, float>&’
```
Not sure why this happens or at least why it wasn't already happening in the `test_Array` code. I guess I'll have to see what the generated code looks like (in stirPYTHON_wrap.cxx`), or you could post it here.. Maybe it's a `const` thing.

rijobro:
Sorry, just realised you had already added a ctest. 

I can't build on OSX or linux, both before swig gets involved. Similar to this job: https://travis-ci.org/github/UCL/STIR/jobs/693170887.

KrisThielemans:
aside from this, there's a problem that it fails in `Debug` mode, as currently `check_state` doesn't want the pointer access to be on. This is a case of too restrictive checking. We should check of pointer access is off when reallocating the array. As long as we're just using it, everything is fine. So that needs some modification in `VectorWithOffset` sadly.

alternative is to use `boost::shared_array`, have the `Array` pointing to that memory (but not owning it). That would work after some mods in `Array` (I discover that the relevant [constructor](https://github.com/UCL/STIR/blob/8428898a4f2c3a817d7482144974a3a5408e463f/src/include/stir/VectorWithOffset.h#L147-L150) is only available  in `VectorWithOffset`), but it is somewhat overly complicated.

KrisThielemans:
> I can't build on OSX or linux, both before swig gets involved.

it'd save me time if you copy errors in here directly, and log line numbers (click on the line, grab it from the URL) :wink:

Cause is https://github.com/UCL/STIR/blob/8428898a4f2c3a817d7482144974a3a5408e463f/src/include/stir_experimental/motion/Transform3DObjectImageProcessor.h#L73

which defines an `Array` with non-numerical elements, which could have broken earlier.

This might be resolvable by changing
```c++
  inline virtual void axpby(const elemT a, const Array& x,
                            const elemT b, const Array& y);
```
to a member template
```
 template <typename elemT2>
  inline virtual void axpby(const elemT2 a, const Array& x,
                            const elemT2 b, const Array& y);
```
and same in `NumericVectorWithOffset`. The whole reason for the existence of `NumericVectorWithOffset` is in any case that ages ago, VC wasn't happy with member templates. Otherwise, all the numerical operations could have been written directly in `VectorWithOffset`.

rijobro:
Seems there's also a problem with the STIR experimental code (which is why it didn't get picked up by most Travis builds).

KrisThielemans:
seems to be proliferating a bit...

rijobro:
> to a member template
```
 template <typename elemT2>
  inline virtual void axpby(const elemT2 a, const Array& x,
                            const elemT2 b, const Array& y);
```
I thought you couldn't have templated virtual methods?

KrisThielemans:
> I thought you couldn't have templated virtual methods?

correct.... I guess we don't need `virtual` here?

KrisThielemans:
@rijobro this should be ready to go. Even Appveyor is happy which still tests VC 2013. Have a look.

rijobro:
Compilation and ctests all fine on my machines!

KrisThielemans:
I've removed the #if 0, and moved an `assert`. Decided to force-push. sorry.

I'll merge this, but am still not happy with the signature of this function (see #560 )

KrisThielemans:
One Appveyor job went through. rest still running. I'll merge anyway.

PR #562, opened 2020-06-02
rel4 to master (2nd)

rijobro:
get features from release_4 branch into master


KrisThielemans:
needs updates on the GE stuff for the `proj_data_info_ptr` and probably `shared_ptr<const>`, e.g. 
https://travis-ci.org/github/UCL/STIR/jobs/693903911#L1331

rijobro:
Seems there's a copy constructor missing for `GEHDF5Wrapper`, which means that the `ProjDataGEHDF5` constructors can't take `shared_ptr<const GEHDF5Wrapper>` as argument (only `shared_ptr<GEHDF5Wrapper>`).

I don't know the `GEHDF5Wrapper` class, no idea if it's trivial or not.

KrisThielemans:
Travis all green... So what's the problem with `GEHDF5Wrapper`?

src/include/stir/ProjDataGEHDF5.h
```diff
@@ -0,0 +1,102 @@
...
+    explicit ProjDataGEHDF5(shared_ptr<const ProjDataInfo> input_proj_data_info_sptr,
+                              const std::string& input_filename);
+
+    explicit ProjDataGEHDF5(shared_ptr<const ProjDataInfo> input_proj_data_info_sptr,
+                              shared_ptr<GEHDF5Wrapper> input_hdf5_sptr);
+
+    explicit ProjDataGEHDF5
+      (shared_ptr<const ExamInfo> input_exam_info_sptr,
+                              shared_ptr<const ProjDataInfo> input_proj_data_info_sptr,
+                              shared_ptr<GEHDF5Wrapper> input_hdf5);
```
rijobro, line 67: These constructors take shared pointers for all arguments and then create deep copies of them to ensure that the original object isn't changed elsewhere. It looks like there's no way of copying the `GEHDF5Wrapper`, so that approach can't be taken there. However, it looks like it's opened as read only, so perhaps this isn't a problem. I'm not sure.

KrisThielemans: This mimics`ProjDataFromStream` which takes ` shared_ptr<iostream>`.
https://github.com/UCL/STIR/blob/7dc238f5ec228554f594f1f2228450eb4618472b/src/include/stir/ProjDataFromStream.h#L85-L87
You cannot copy that either. I guess the class interface tells people they have to be careful...

By the way, I'm no sure if we really need these constructors, but will leave them in.

PR #564, opened 2020-06-05
extend GE RDF 9 support

AnderBiguri:
A PR to extend the RDF9 support for the GE Signa PET/MR to other scanners.

File format used by your scanner depends on its software version. If you can get hold of an RDF file or a listmode file (*.BLF), and you can open it in a HDF5 viewer, check `//HeaderData/RDFConfiguration/fileVersion/majorVersion`


AnderBiguri:
Dang, I PR to the wrong branch.

AnderBiguri:
@KrisThielemans While master contains these changes, I can still see all of them here. Have you "squash-merged" the Signa PR? Should I delete and re-PR?

KrisThielemans:
merge `master` in, and they'd disappear. if you haven't done anything yet, it should be a fast-forward merge, so will have the same affect as creating a new PR

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -283,18 +284,29 @@ Succeeded GEHDF5Wrapper::initialise_singles_data(const std::string &path)
...
-            //! \todo Get these numbers from the HDF5 file
-            {
-                m_NX_SUB = 45;    // hyperslab dimensions
-                m_NY_SUB = 448;
-                m_NZ_SUB = 1;
-                m_NX = 45;        // output buffer dimensions
-                m_NY = 448;
-                m_NZ = 1;
-            }
+            m_address = "/Singles/CrystalSingles/sample1"; //AB: todo. '/slms/inm/research/moco/PhantomData_Orsay/R00037/raw/rdf.0' had ./sample1 but ./sample was written here. ?? typo?
```
pw351: The m_address here reads the samples accross the time of scan to collect singles per crystal. The m_address should be:
 m_address = "/Singles/CrystalSingles/sample";

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -349,30 +362,35 @@ Succeeded GEHDF5Wrapper::initialise_geo_factors_data(const std::string& path,
...
+                const int rank = m_dataspace.getSimpleExtentNdims();
+                hsize_t dims[2];
+                m_dataspace.getSimpleExtentDims( dims, NULL);
+                // AB todo : remove all this? seems clutter. 
+            //      int ndims = m_dataspace.getSimpleExtentDims( dims_out, NULL);
+                        std::cout << "rank " << rank << ", dimensions " <<
+                            (unsigned long)(dims[0]) << " x " <<
+                            (unsigned long)(dims[1]) << std::endl;
+    //                m_memspace_ptr = new H5::DataSpace()
+
```
pw351, line 380: MATLAB may have transposed the row and column so I would suggest to make sure if everything is read in correct order.

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -383,31 +401,40 @@ Succeeded GEHDF5Wrapper::initialise_geo_factors_data(const std::string& path,
...
-                m_NX_SUB = 45;    // hyperslab dimensions
-                m_NY_SUB = 448;
-                m_NZ_SUB = 1;
-                m_NX = 45;        // output buffer dimensions
-                m_NY = 448;
-                m_NZ = 1;
-            }
+                     (unsigned long)(dims[0]) << " x " <<
+                     (unsigned long)(dims[1]) << std::endl;
+
```
pw351, line 416: The dimensions of this crystal efficiency in the actual norm3d file is [45 896]
GE repeats the values twice. Whereas, we need to read part of that array :[45 448] 
We need to extract the crystal efficiency factors for ring and detector numbers. We then use the same array to calculate the normalisation factors.

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -283,18 +284,29 @@ Succeeded GEHDF5Wrapper::initialise_singles_data(const std::string &path)
...
+            // Create an array to host the size of the dimensions
+            const int rank = m_dataspace.getSimpleExtentNdims();
+            hsize_t dims[rank];
+            // hsize_t max_dims[dataspace_Ndims]; // AB: Do we want the max_dims?
+            // Read size of dimensions
+            m_dataspace.getSimpleExtentDims( dims, NULL); // AB: Do we want the max_dims?
+
+            // AB todo: I think it returs dataspace_Ndims==2, test.
+            // AB todo: MATLAB reader returns [448 45] yet here it was written NX=45 NY=448. row/colun major issue, or actual difference? test
+
```
pw351, line 303: MATLAB may have transposed the row and column so I would suggest to make sure if everything is read in correct order.

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -283,18 +284,29 @@ Succeeded GEHDF5Wrapper::initialise_singles_data(const std::string &path)
...
-            {
-                m_NX_SUB = 45;    // hyperslab dimensions
-                m_NY_SUB = 448;
-                m_NZ_SUB = 1;
-                m_NX = 45;        // output buffer dimensions
-                m_NY = 448;
-                m_NZ = 1;
-            }
+            m_address = "/Singles/CrystalSingles/sample1"; //AB: todo. '/slms/inm/research/moco/PhantomData_Orsay/R00037/raw/rdf.0' had ./sample1 but ./sample was written here. ?? typo?
+            // Get the DataSpace (metadata) corresponding to the DataSet that we want to read
```
pw351, line 291: The m_address here reads the samples accross the time of scan to collect singles per crystal. The m_address should be:
m_address = "/Singles/CrystalSingles/sample";

src/IO/GEHDF5Wrapper.cxx
pw351, line 416: The dimensions of this crystal efficiency in the actual norm3d file is [45 896]
GE repeats the values twice. Whereas, we need to read part of that array :[45 448]
We need to extract the crystal efficiency factors for ring and detector numbers. We then use the same array to calculate the normalisation factors.

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -383,31 +401,40 @@ Succeeded GEHDF5Wrapper::initialise_geo_factors_data(const std::string& path,
...
-                m_NX = 45;        // output buffer dimensions
-                m_NY = 448;
-                m_NZ = 1;
-            }
+                     (unsigned long)(dims[0]) << " x " <<
+                     (unsigned long)(dims[1]) << std::endl;
+
+            // AB todo: MATLAB reader returns [896 45] yet here it was written NX=45 NY=448. row/colun major issue, or actual difference? test
+            // AB todo: IMPORTANT. Notice the numbers, they are different. the values that were here are different that what I see in norm3d
+            m_NX_SUB = dims[1];    // hyperslab dimensions
```
pw351, line 419: Kindly check if the dimensions are in correct order?

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -84,15 +90,30 @@ GEHDF5Wrapper::open(const std::string& filename)
...
     file.openFile(filename, H5F_ACC_RDONLY);
 
+    //AB  Find out of file is RDF9 or RDF10
+    H5::DataSet str_file_version = file.openDataSet("/HeaderData/RDFConfiguration/fileVersion/majorVersion");
+    str_file_version.read(&rdf_ver, H5::PredType::STD_U32LE); //AB: I have no idea if this is the right type.
+
+    // AB check if the file is compressed, error if that is the case. 
+    if (rdf_ver==9)
+    {
+        unsigned int is_compresed;
```
KrisThielemans, line 100: `is_compressed`

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -84,15 +90,30 @@ GEHDF5Wrapper::open(const std::string& filename)
...
+        H5::DataSet str_file_version = file.openDataSet("/HeaderData/ListHeader/isListCompressed");
+        str_file_version.read(&is_compresed, H5::PredType::STD_U32LE); //AB: I have no idea if this is the right type.
+        if (is_compresed)
+            error("The RDF9 file is compressed, we won't be able to read it. Please uncompress it and retry. Aborting");
```
KrisThielemans, line 104: this check will only work for list-files. You'd first have to check if it is one. (should sit in config I guess). Without it, might fall-over on non-list files.

Note that there will be a similar check for sino files though (not for norm as far as I know)

AnderBiguri:
@KrisThielemans I want to test `GEHDF5Wrapper::initialise_singles_data` and it seem sits only used by the application `construct_randoms_from_GEsingles`, but I can't figure out what to input. The arguments are `out_filename listmode_filename template_projdata` and that last one is resisting me. Any hint of what a valid `template_projdata` is? I did not find any docs relating to this either...

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -168,6 +189,7 @@ Succeeded GEHDF5Wrapper::initialise_scanner_from_HDF5()
     int num_detectors_per_ring = num_transaxial_blocks_per_bucket*num_transaxial_crystals_per_block*radial_modules_per_system;
     int default_num_arccorrected_bins = max_num_non_arccorrected_bins;
     float inner_ring_radius = physical_scanner_radius;
+    //AB TODO check if following is valid
```
KrisThielemans, line 192: hmmm. well-spotted.

@pw351. questions for you:

- we get `physical_scanner_radius` from `/HeaderData/SystemGeometry/sourceRadius` above, but that seems incorrect. (it probably has to do with the good old days when they had a line source inside the scanner, although I'm not sure).
- `inner_ring_diameter` is set from `/HeaderData/SystemGeometry/effectiveRingDiameter`, so this variable has the wrong name.

So, although I agree with the formulas, I wonder if the input is correct. What are the corresponding values?

Note however that STIR will currently only use its `ring_radius+DOI`, resulting in `effective_ring_radius`, which will be correct.

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -246,10 +268,12 @@ Succeeded GEHDF5Wrapper::initialise_listmode_data(const std::string &path)
 {
     if(path.size() == 0)
     {
-        if(is_signa)
+        if(rdf_ver==9) 
         {
             m_address = "/ListData/listData";
             //! \todo Get these numbers from the HDF5 file
+            // AB: todo my file does not have anything in this addres. 
+            // AB only *.BLF file has this address, but not the below variables.
```
KrisThielemans, line 276: I'm confused about these comments. Yes, only BLF file with have this address. what do you mean with "below variables"?

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -303,22 +339,22 @@ Succeeded GEHDF5Wrapper::initialise_singles_data(const std::string &path)
...
+Succeeded GEHDF5Wrapper::initialise_proj_data(const std::string& path,
                                                  const unsigned int view_num)
 {
     if(path.size() == 0)
     {
-        if(is_signa)
+        if(rdf_ver==9) 
         {
             m_address = "/SegmentData/Segment2/3D_TOF_Sinogram/view";
+            // AB: todo norm3d, geo3d, rdf.0 do not have this info. Only rdf.0 has "Segment2/3D_TOF_Sinogram", but no "view" in sight.
```
KrisThielemans, line 350: definitely only appropriate for sinos.

Possibly you have a compressed sino if there's no `view`

src/IO/GEHDF5Wrapper.cxx
AnderBiguri: @KrisThielemans yeah still working on this one. Thing is, quite few "is valid" tests may happen, and I am not sure if they should be here? Should I make a new function to check if the file is valid, and put there all the "if file is TYPE, check this_and_that" ?

KrisThielemans:
> @KrisThielemans @pw351 I want to test `GEHDF5Wrapper::initialise_singles_data` and it seem sits only used by the application `construct_randoms_from_GEsingles` (via `SinglesRatesFromGEHDF5.cxx`), but I can't figure out what to input. The arguments are `out_filename listmode_filename template_projdata` and that last one is resisting me. Any hint of what a valid `template_projdata` is? I did not find any docs relating to this either...

If there isn't one, it'll have to be create with `create_projdata_template`. Have to fill in signa stuff. Use `span=2`. The rest can remain as default.

src/IO/GEHDF5Wrapper.cxx
AnderBiguri: Literally info about the variables below i.e. `m_size_of_record_signature = 6; m_max_size_of_record = 16;`. In MATLAB, If I open ` "/ListData/listData"` I get a list. When I check its contents  get:

```
>> info.Groups(9).Datasets

ans = 

  struct with fields:

          Name: 'listData'
      Datatype: [1×1 struct]
     Dataspace: [1×1 struct]
     ChunkSize: 33554432
     FillValue: 0
       Filters: []
    Attributes: []

>> info.Groups(9).Datasets.Datatype

ans = 

  struct with fields:

          Name: ''
         Class: 'H5T_INTEGER'
          Type: 'H5T_STD_U8LE'
          Size: 1
    Attributes: []

>> info.Groups(9).Datasets.Dataspace

ans = 

  struct with fields:

       Size: 31310208
    MaxSize: Inf
       Type: 'simple'
```

None of this translates obviously to the variables we need to replace in that function. They may be somewhere else, haven't investigated enough of it yet.

src/IO/GEHDF5Wrapper.cxx
KrisThielemans: would make sense. In the end, we will have to be able to call this from elsewhere ("is this a GE list file?"). GEHDF5Wrapper should therefore have some members bool is_list_file() I guess. we'd then check the "signature" (minimum check on HDF5 from GE), check the config dataset on the type, set the variables, and then for each type do a consistency check. The latter probably belongs more in the respective derived classes, but as you prefer.

src/IO/GEHDF5Wrapper.cxx
KrisThielemans: This numbers are not recorded in the HDF5 file (at least not for RDF9). They are part of the (previously confidential) info that we've received from GE. So, leave the code as-is, or add a one liner explanation "info provided by GE" or something.

AnderBiguri: @KrisThielemans  sounds good! There was a `\\todo read from file` there, so it got me confused. Will add a comment.

src/IO/GEHDF5Wrapper.cxx
AnderBiguri: So I can make:

- `GEHDF5Wrapper::is_list_file()` 
- `GEHDF5Wrapper::is_norm_file()` 
- `GEHDF5Wrapper::is_sino_file()` 
- `GEHDF5Wrapper::is_geo_file()` Maybe all these being `check_xxxx_file()`? so its clear they error check?

Then some private attributes so there is no need to rerun

- `is_sino`
- `is_list`
- `is_norm`
- `is_geo`

I do not know what you mean with "config dataset on the type".

And with the derived classes you mean `Scanner::PETMR_Signa`? Not sure which classes inherit from the wrapper. I am happy to put this wherever, just really don't know where.

src/IO/GEHDF5Wrapper.cxx
AnderBiguri: @KrisThielemans hum possibly.... I have `cvt` and then `pwbX` x=[1,15], 16 Datasets in total. Trying to figure out how to know if its compressed or not.

KrisThielemans: checking some Matlab code, it seems that
`HeaderData/Sorter/Segment2` etc should have a field `compDataSegSize`. If it's zero, it is not compressed.

AnderBiguri:
```
 h5read([path 'rdf.0'], '/HeaderData/Sorter/Segment2/compDataSegSize')

ans =
   14849736
```
Hum seems compressed then.

src/IO/GEHDF5Wrapper.cxx
KrisThielemans: I wouldn't make `GEHDF5Wrapper::is_list_file()` etc `static`. Then it has to open the file and read stuff anyway. `GEHDF5Wrapper` is associated to a HDF5 file, so it can tell us what it is.

I thought that `ProjDataGEHDF5` et al were derived from `GEHDF5Wrapper`, but it seems they use it as a member. So that seems to imply the validity checks still have to set in `GEHDF5Wrapper` for all data-types. It's unfortunate, but ok,

Seems to make sense that `check_is_valid_projdata` et al. would throw an error with a diagnostic (use `stir::error`), while the `is_` functions would return a bool (and probably do the check but catch the error). I wouldn't bother putting in the logic to avoid rerunning. This will only be called once or twice anyway. We can always leave that optimisation for others!

> I do not know what you mean with "config dataset on the type".

do you have `HeaderData/RDFConfiguration`? should have `isListFile`. It seems nothing else of value there for this purpose. sino should have `/SegmentData/Segment2`. norm should have whatever other segment number.

> with the derived classes you mean Scanner::PETMR_Signa? 

no, I meant whoever is using `GEHDF5Wrapper`. (ProjDataGEHDF5`, `CListModeData...`)

>  using boost is OK here? to play with filenames (see if file is this type or other type), or should I try to read stuff and figure it out from HDF5 data?

you can never rely on filenames, certainly not when they come from the console. We have to read the content.

src/IO/GEHDF5Wrapper.cxx
pw351: Yes, this part of code is basically not used yet, as far as I remember! 
Looking at the values for :
/HeaderData/SystemGeometry/sourceRadius: 311.3
/HeaderData/SystemGeometry/effectiveRingDiameter: 640.6

Indeed these are incorrect!
I will correct these!

KrisThielemans: `effectiveRingRadius` would therefore be 320.3,  it doesn't make the `sourceRadius` entirely unbelievable to mean "inner ring radius", but just a bit too small.

If the RDF file doesn't give inner ring radius elsewhere, there doesn't seem to be a way for us to figure out what it is. I then suggest it to `effectiveRingRadius - 7mm` using a (somewhat arbitrary) DOI for LYSO (as I believe we have for other GE scanners). Of course, for the Discovery IQ, the crystal is BGO, so therefore its DOI would be smaller. I guess they don't record the scintillator material, so this seems all to lead to some ugly hacks.

pw351: Basically, the physical ring radius must be 311.8 precisely.  The depth of interaction can be calculated using effectiveRingRadius - physicalRingRadius which in my case is 320.3-311.8 which works out to be 8.5. I also need this to be fixed in scanner.cxx. I will do this ASAP.

src/IO/GEHDF5Wrapper.cxx
AnderBiguri: @pw351 indeed MATLAB transposes so it can read in column major order. I am changing all these after testing.

src/IO/GEHDF5Wrapper.cxx
AnderBiguri: @pw351 Do you have any idea if this is the standard for RDF9 or just Signa? say, should I just grab the dims and half its value? because we only want to read half of it?

src/IO/GEHDF5Wrapper.cxx
KrisThielemans: is the 311.8 somewhere in the header? otherwise we have to hard-code DOI somehow (if it's for scanner where we don't know it)

AnderBiguri:
Couple of questions @KrisThielemans  and @pw351 

1. `m_NX_SUB` etc do not seem to be ever used in STIR. I am writing code to fill these in, but are they needed? 

2. Any advice on testing `GEHDF5Wrapper::initialise_proj_data`? Does not seem to be used anywhere.

KrisThielemans:
> Any advice on testing GEHDF5Wrapper::initialise_proj_data? Does not seem to be used anywhere.

This is because `ProjDataGEHDF5` isn't integrated in the rest of STIR as I told @pw351 to remove a specific function. I'm doing this now.

KrisThielemans:
done, but untested.

KrisThielemans:
Forgot to say that with this change, `list_projdata_info --all` should give decent results on an RDF9, `stir_math -s out bla.rdf` should work to copy to Interfile, and (of course) reconstructions can read RDF9 without needing any conversion.

src/IO/GEHDF5Wrapper.cxx
KrisThielemans: now I understand your comment about `static` members. `GEHDF5Wrapper::check_GE_signature` is `static`, which makes sense as it can be used by others to see if it's a file it can read it.

So what about the following then:
1. move most code from `check_GE_signature` to new `static` function `check_current_signature(H5file&)`, such that we can avoid opening the file twice
2. write `check_GE_signature` in terms of that one  (open file, `check_current_signature`, all within `try...catch`)
3. call `check_current_signature` in your new `check_file`
4. remove https://github.com/UCL/STIR/blob/18249ed7f4b10512429cff3a79ff27427c29b4f7/src/IO/GEHDF5Wrapper.cxx#L218-L225 and always use `initialise_scanner_from_HDF5`.
5. add some preliminary logic in `initialise_scanner_from_HDF5` to recognise the Signa

src/IO/GEHDF5Wrapper.cxx
KrisThielemans: let's read half. if scanner is already initialised (should be I guess), add a check that you then get `scanner_sptr->get_num_detectors_per_ring()`

KrisThielemans:
Plan to remove hard-coding on dataspaces etc.:

- remove `count` and `block` arguments from `get_from*_dataset`, could add default values for `stride` and `offset`
- in these functions, 
     - check that `stride` and `offset` are `{1,1}` and `{0,0}` for  now, otherwise `error`
     - either use `m_NX` etc set in `initialise_*`, or move that code here.
     - use `Array::resize` to remove need for preallocation of the output
     - then read data as it does now

- singles:
    - in `GEHDF5Wrapper`, if `list` or `sino`, store `SinglesHeader/numValidSamples` in a new member `num_singles_samples` with public member `get_num_singles_samples()`
    - in SinglesRatesFromGEHDF5 use that as opposed to time-frames
    - rename `get_dataspace` to `get_singles` and add TODO that it should check data-size

src/IO/GEHDF5Wrapper.cxx
AnderBiguri: Doing something slightly different but in the same line. 

- `check_GE_signature()` keeps being `static`
- `check_current_signature()` has no input, assumes file has been opened by `GEHDF5Wrapper::open()`. Returns `int`, instead of `bool`. Hopefully in the future we can add an `enum` class with scanner names, for now returns `1` if Signa. That way @pw351 (or @pwadhwa351 which one to use!) can use it to do step 5 (#583).

KrisThielemans: I can't see how `check_GE_signature(filename)` can call `check_current_signature()` in the way you suggest. It'd have to create a `GEHDF5wrapper` object and then call that, but the constructor would already be called, so there's no point anymore. I think that this solution therefore implies something like
````c++
bool check_GE_signature(filename)
{
   try { GEHDF5Wrapper testobject(filename); return true;} catch (...); return false;
}
```
would work but it constructs a whole object, which might be somewhat expensive.. In my suggestion, I'd say
```c++
bool check_GE_signature(filename)
{
  H5file file;
  // open etc checks
   try { return GEHDF5Wrapper::check_current_signature(file);} catch (...); return false;
}
```
Effect is the same, but it can be more lightweight.

I don't see why `check_current_signature()` needs to return info on scanner. It doesn't make sense with the name either...

We have an enum with scanners, it sits in `Scanner` :wink:. In any case, as long as we make sure that `Scanner::get_scanner_from_name` does the name look-up correctly (by making sure that the name listed in the RDF file is listed in the `string_list` at https://github.com/UCL/STIR/blob/18249ed7f4b10512429cff3a79ff27427c29b4f7/src/buildblock/Scanner.cxx#L347), there's very little else to do.

Summary: I like my solution better :smile:

AnderBiguri: @KrisThielemans ah, I am fast! :) I realized my mistake and deleted my comment few minutes ago!

pwadhwa351:
> @KrisThielemans @pw351 I want to test `GEHDF5Wrapper::initialise_singles_data` and it seem sits only used by the application `construct_randoms_from_GEsingles` (via `SinglesRatesFromGEHDF5.cxx`), but I can't figure out what to input. The arguments are `out_filename listmode_filename template_projdata` and that last one is resisting me. Any hint of what a valid `template_projdata` is? I did not find any docs relating to this either...

Sorry I picked this up now, basically you need the Signa template, which you can create using create_projdata_template for GE PET/MR Signa using default settings apart from span which is 2 and N for arc corrected.

pwadhwa351:
> Couple of questions @KrisThielemans and @pw351
> 
> 1. `m_NX_SUB` etc do not seem to be ever used in STIR. I am writing code to fill these in, but are they needed?

I would suggest to leave them as it sets the dimension of hyperslab.

AnderBiguri:
> I would suggest to leave them as it sets the dimension of hyperslab.

Rigth, but my issue was that neither `m_NX_SUB` nor `m_NX` (or the Y, Z equivalents) are ever used, anywhere in the entire STIR codebase. Since that we found out that they are hard-coded in the other files, so I am working on making use of these.

pwadhwa351:
> > I would suggest to leave them as it sets the dimension of hyperslab.
> 
> Rigth, but my issue was that neither `m_NX_SUB` nor `m_NX` (or the Y, Z equivalents) are ever used, anywhere in the entire STIR codebase. Since that we found out that they are hard-coded in the other files, so I am working on making use of these.

Yeah, this is because the preliminary code that I created initially was translated into GEHDF5Wrapper class later by other contributor and myself. This lead to the pieces of code not really being translated to other files! Sorry!

AnderBiguri:
> Yeah, this is because the preliminary code that I created initially was translated into GEHDF5Wrapper class later by other contributor and myself. This lead to the pieces of code not really being translated to other files! Sorry!

No worries ;) I am working to fix that now :)

AnderBiguri:
@KrisThielemans 

https://github.com/UCL/STIR/blob/11e28b0c66f4335258d357e52a2d899651429feb/src/buildblock/ProjDataGEHDF5.cxx#L75

Is wrong, that its not the right number returned at least. I am searching for the right parameter, but do let me know if you immediately know which one is the correct one.

KrisThielemans:
this needs a bit more explanation. get_max_num_non_arccorrected_bins() has to be on the order of 371 (I don't know the exact number0.

can you point to where you use it?

AnderBiguri:
My problem is in https://github.com/UCL/STIR/blob/1df251c47f3626f2785ec4077b337dbee816b732/src/buildblock/ProjDataGEHDF5.cxx#L106
 the last value is 1981, but should be 357. 

Now, its quite hard for me to figure out where that value is set, because it uses lots of STIR classes in quite a lot of fields (its all `Projdata`). I did just link those 2 because of the comment. 

Now, you use `get_max_num_non_arccorrected_bins()` to set that one up. I have tracked that variable in the wrapper, and in the hdf5 file at the address `"/HeaderData/Sorter/dimension1Size"`. When I read this value from MATLAB, its 1981.

KrisThielemans:
ok. 1981 is `max_num_axial_poss` (effectively the number of sinograms, which for span 2 is `N^2-(N-1)` with `N` the number of rings (45 for the Signa).

This will be related to the "bug" in RDF9, and which took a few months from @pw351's life. The overall buffer size is ok, but the dimensions encoded in the RDF are actually in the wrong order (I guess it'll say 357x27x1981, while actually it is 1981x27x357. So the dimensions of the `tof_data[view] are correct. Reading the data this way was the only way that Palak was able to make sense of it. Unless you see another way to achieve the same result?

Obviously, this scuppers our plan a bit to let `get_from_dataset` figure out the sizes and let it resize the output array by itself...

The only way I see to rescue this is to add an argument `get_from_dataset(...,permute=false)`. If it's `true`, reverse the order of the sizes. It probably also makes sense to change `ProjDataGEHDF5::tof_data` from an `Array<4,uchar>` to `std::vector<Array<3,uchar> >` such that `get_from_dataset` can do the usual resizing. (This might not be necessary as `stir:Array`s can be non-rectangular, but I'm not 100% sure what will happen if we pass `tof_data[view]` as a reference to `get_from_dataset` which will then resize it. Easy to find out of course!)

AnderBiguri:
Hum, not sure if we are in the same page. 

> ok. 1981 is `max_num_axial_poss` (effectively the number of sinograms, which for span 2 is `N^2-(N-1)` with `N` the number of rings (45 for the Signa).

Yes, currently the variable `max_num_axial_poss` is manually set to 1981 which is what we need. However `get_num_tangential_poss()` also is 1981. 
Therefore, the following is false:

> This will be related to the "bug" in RDF9, and which took a few months from @pw351's life. The overall buffer size is ok,

No, as it stands now, the buffer is `unsigned int total_size = get_num_tangential_poss() * num_tof_poss * max_num_axial_poss;` i.e. `1981*27*1981`, so its too big. Causes no problem, but its unnecessary, and definitely not the correct size.

> but the dimensions encoded in the RDF are actually in the wrong order (I guess it'll say 357x27x1981, while actually it is 1981x27x357. So the dimensions of the `tof_data[view] are correct. Reading the data this way was the only way that Palak was able to make sense of it. Unless you see another way to achieve the same result?

This is not true either. I can hardcode the `count` so the `list_projdata_info` actually prints and works.  do this by setting [1981x27x357], which is the same that `m_dataspace.getSimpleExtentDims( dims, NULL);` returns. So its returning the correct dimensions, as far as I can see....

pwadhwa351:
> Hum, not sure if we are in the same page.
> 
> > ok. 1981 is `max_num_axial_poss` (effectively the number of sinograms, which for span 2 is `N^2-(N-1)` with `N` the number of rings (45 for the Signa).
> 
> Yes, currently the variable `max_num_axial_poss` is manually set to 1981 which is what we need. However `get_num_tangential_poss()` also is 1981.

Well this does not make sense! get_num_tangential_poss() comes from projdata info. It can't be 1981 as for this scanner it is 357

AnderBiguri:
> Well this does not make sense! get_num_tangential_poss() comes from projdata info. It can't be 1981 as for this scanner it is 357

`num_tangetial_poss` dies not exist in the Wrapper, so if it is set (as I suspect) with the variable `max_num_non_arccorrected_bins` (as I mentioned above), this variable is read from `/HeaderData/Sorter/dimension1Size` in the wrapper, and you can verify that this value is indeed 1981 in the dataset you sent me.

PD: I have tested/printed all values I mention in these comments. I have no idea how it _should_ be, I am reporting the sizes I see.

KrisThielemans:
> Yes, currently the variable `max_num_axial_poss` is manually set to 1981 which is what we need. However `get_num_tangential_poss()` also is 1981.

ok. This wasn't a problem for @pw351  as she was using the `Scanner.cxx` values. Now you're reading them from the header. This implies that https://github.com/UCL/STIR/blob/1df251c47f3626f2785ec4077b337dbee816b732/src/buildblock/ProjDataGEHDF5.cxx#L145 is wrong. What are the dimensions listed `HeaderData/Sorter`?

> I suppose you mean that you expect different physical ordering of this data maybe? 

yes, if you read the data with the order specified in the header, the result is wrong. `list_projdata_info` will never see this as it just gives globals min/max values etc. It's only once you look at actual sinograms (which you could do from STIR Python or via `extract_segments` (as described in http://stir.sourceforge.net/wiki/index.php/STIR_FAQ#How_to_display_images_generated_by_STIR.3F)

AnderBiguri:
@KrisThielemans hum I got lost there. I have no idea what that line is doing. 

In any case, this are the dims:

```
'/HeaderData/Sorter/dimension1Size' == 1981
'/HeaderData/Sorter/numTOF_bins' == 27
'/HeaderData/Sorter/dimension2Size' == 357
```

KrisThielemans:
oops. copy-paste mistake! I meant https://github.com/UCL/STIR/blob/1df251c47f3626f2785ec4077b337dbee816b732/src/IO/GEHDF5Wrapper.cxx#L145

so, `max_number_of_non_arc_corrected_bins` has to be set from `dimension2`! ideally we'd save numTOF_bins as well (but we're currently grafting this on the non-TOF STIR, so that info isn't in `Scanner`, so we cannot pass it on), maybe just read it and add a TODO comment. We could test the value of `dimension1Size` against `get_num_sinograms()`, after completing #586

AnderBiguri:
@KrisThielemans  I am super stuck with STIR Arrays, I can't find any docs nor even the definitions and I have no idea how they work....

Among other things, this seems to crash, i.e. resizing twice. 

```
    output.resize(IndexRange3D(m_NX_SUB,m_NY_SUB,m_NZ_SUB));
    output.resize(IndexRange3D(m_NX_SUB,m_NY_SUB,m_NZ_SUB));
```

similarly, what does `output.get_data_ptr()` do? is the underlying data 1D? Can I use it as such, or is it a pointer to pointers to pointers?

KrisThielemans:
> I am super stuck with STIR Arrays, I can't find any docs nor even the definitions and I have no idea how they work....

They are documented via doxygen, which means that the doc is inline in the files (`src/include/stir/Array.*`)  Output of doxygen for latest release is on the STIR website. See e.g. http://stir.sourceforge.net/documentation/doxy/html/classstir_1_1Array.html

Of course, that might not tell you what you need to know...

They are implemented recursively, so if `a` is a 3D array, `a[0]` is a 2D array. At this point in time, each level guarantees contiguous memory access, but for all higher dimensions that just gives you access to `Array` objects, which aren't useful for `memcpy` etc.

`get_data_ptr` gives you a ptr to `a[0]` For your purposes therefore only useful for 1D arrays (i.e. it's only `Array<1,float>::get_data_ptr()` that gives you a `float *`. you have to call `release_data_ptr` after it, otherwise it'll raise an error.

Clearly, resizing twice should work. If it doesn't, it's a bug. I'll admit I'd find that rather remarkable (it's ~20 year old code with quite a number of tests, but probably no test for resizing twice... Feel free to add one in `src/test/test_Array.cxx` :smile: ). Probably best to show some code for that.

AnderBiguri:
@KrisThielemans yes I realised my issue is more complex, but no idea yet what it is. 
I have pushed my current stage, I am having problems in 
https://github.com/UCL/STIR/blob/ce01947f3b5037b754503386fee1f9d45933692e/src/buildblock/ProjDataGEHDF5.cxx#L104
The second call segfaults in the resize, but further tests did show that its not because its the second call to resize, but the second call to that function. 
However with what you said, I see that the reading from the H5 should not work with `Array<3,type> `as output, so maybe there is something going on there.

AnderBiguri:
OK, I have made some progress, thanks for the input @KrisThielemans and @pwadhwa351 .

I have the following things missing identified to finish up, and I need some input in some of these:

- [ ] The following values are different than their `_SUB` counterparts. I can figure out where they are coming from, but again, these are never used, so I am not really sure if we really need them and how to fill them/test them.... Should I fill them or just remove this?
https://github.com/UCL/STIR/blob/102ac261ef913f653119998d60740889af3a17e4/src/IO/GEHDF5Wrapper.cxx#L504
- Trying to `resize()` all output buffers inside the data reading utilities for consistency
 
   - [ ] `get_singles` and `get_from_dataspace` (listmode) do not `resize()`, but assume its resized. Should I change this or I leave a TODO ?
   - [ ] For Proj data: I am now forcefully transposing if its RDF9. is this OK?

- [ ] `get_from_dataspace` (the listmode reader) has a very different trace than the rest of readers. Is this for some reason? should I change it? Also does not `resize()`, but can't with the current trace.
https://github.com/UCL/STIR/blob/102ac261ef913f653119998d60740889af3a17e4/src/IO/GEHDF5Wrapper.cxx#L607


Finally in testing: 
- `construct_randoms_from_GEsingles` returns "ERROR: Can only process not arc-corrected data"
- How do I test recon/proj data? From SIRF?

KrisThielemans:
> get_singles and get_from_dataspace (listmode) do not resize(), but assume its resized. Should I change this or I leave a TODO ?

leave. Could insert check on size read if that's easy.

> For Proj data: I am now forcefully transposing if its RDF9. is this OK?

yes, until we know more about RDF10 anyway...

> get_from_dataspace (the listmode reader) has a very different trace than the rest of readers. 

"trace"? Do you  mean "signature"? It has to be like this, as it's going to read only a tiny bit of the whole stream (at the offset given), and advances the offset. All the others read all the data.

Could rename the function to something more explanatory.

> construct_randoms_from_GEsingles returns "ERROR: Can only process not arc-corrected data"

that means you created the wrong template. Have to answer "arc-correction": N. Or edit the .hs file and set `applied corrections := {None}`

> How do I test recon/proj data? From SIRF?

SIRF will need work. I think first step is to see if you get the same output when running the `stir_math` "conversion" and `construct_randoms_from_GEsingles`. (You can use `compare_projdata`).
Then we create a sample script (like for the mMR).

src/buildblock/ProjDataGEHDF5.cxx
```diff
@@ -49,71 +49,61 @@ START_NAMESPACE_STIR
...
+  //this->tof_data.resize(IndexRange4D(get_num_viewgrams, max_num_axial_poss, num_tof_poss, get_num_tangential_poss()));
+  if (!this->tof_data.empty())
+    error("there is already data loaded. Aborting");
 
-Array<1,unsigned char> buffer(total_size);
-
- for (int view_num = get_min_view_num(); view_num <= get_max_view_num(); view_num++)
-{
-    m_input_hdf5_sptr->initialise_proj_data_data("", view_num + 1);
+  tof_data.reserve(get_max_view_num()+1-get_min_view_num());
```
KrisThielemans, line 95:
```suggestion
  tof_data.reserve(get_num_views());
```

src/buildblock/ProjDataGEHDF5.cxx
```diff
@@ -49,71 +49,61 @@ START_NAMESPACE_STIR
...
-    const std::array<unsigned long long int, 3> count  = {max_num_axial_poss, num_tof_poss,static_cast<unsigned long long int>(get_num_tangential_poss())};
-//    {static_cast<unsigned long long int>(get_num_axial_poss(segment_num)), num_tof_poss,
-//     static_cast<unsigned long long int>(get_num_tangential_poss())};
-
-    std::array<unsigned long long int, 3> offset = {0,0,0};
-//    {seg_ax_offset[find_segment_index_in_sequence(segment_num)], 0, 0};
-    std::array<unsigned long long int, 3> block = {1, 1, 1};
+  for (int view_num = get_min_view_num(); view_num <= get_max_view_num(); view_num++)
+  {
+      m_input_hdf5_sptr->initialise_proj_data("", view_num + 1);
```
KrisThielemans, line 100:
```suggestion
      // view numbering for initialise_proj_data starts from 1
      m_input_hdf5_sptr->initialise_proj_data("", view_num - get_min_view_num() + 1);
```

src/include/stir/IO/GEHDF5Wrapper.h
```diff
@@ -40,6 +40,8 @@
 START_NAMESPACE_STIR
 
 namespace GE {
+enum class GE_scanner {not_GE, unknown, Signa};        // Just GE scanner names if needed
+
```
KrisThielemans, line 44: I don't see the need for this. See general review comment

src/include/stir/IO/GEHDF5Wrapper.h
```diff
@@ -54,17 +56,33 @@ class GEHDF5Wrapper
 
     static bool check_GE_signature(const std::string& filename);
 
+    static GE_scanner check_current_signature(H5::H5File& file);
+
     explicit GEHDF5Wrapper();
 
     explicit GEHDF5Wrapper(const std::string& filename);
 
+    Succeeded check_file(); // AB todo, maybe private makes more sense?
```
KrisThielemans, line 65: I agree to make this `private`. It's an internal function called by others.

src/include/stir/IO/GEHDF5Wrapper.h
```diff
@@ -54,17 +56,33 @@ class GEHDF5Wrapper
...
+    bool is_norm_file();
+
     Succeeded open(const std::string& filename);
 
     Succeeded initialise_listmode_data(const std::string& path = "");
 
     Succeeded initialise_singles_data(const std::string& path = "");
 
-    Succeeded initialise_proj_data_data(const std::string& path = "",
+    Succeeded initialise_proj_data(const std::string& path = "",
```
KrisThielemans, line 85: I believe this is now 1-based. Should be documented. No reason to have a default for the `view` (more dangerous than anything else)

src/include/stir/IO/GEHDF5Wrapper.h
```diff
@@ -76,37 +94,32 @@ class GEHDF5Wrapper
     Succeeded get_from_dataspace(std::streampos &current_offset,
                                  char* output);
 
-    Succeeded get_dataspace(const unsigned int current_id,
+    Succeeded get_singles(const unsigned int current_id,
```
KrisThielemans, line 97: actually, can we reorder arguments here? Standard STIR convention is to have output arguments come first.

KrisThielemans: Also, do we actually use both of the `get_singles` functions?

src/include/stir/IO/GEHDF5Wrapper.h
```diff
@@ -157,9 +171,15 @@ class GEHDF5Wrapper
...
 
+    bool is_list = false;
+    bool is_sino = false;
+    bool is_geo  = false;
+    bool is_norm = false;
+
     std::string m_address;
 
-    bool is_signa = false;
+    bool is_signa = false; //AB todo remove (or add more)
```
KrisThielemans, line 181: remove

src/include/stir/data/SinglesRatesFromGEHDF5.h
```diff
@@ -125,7 +125,7 @@ public RegisteredParsingObject<SinglesRatesFromGEHDF5, SinglesRates>
  //
  // The singles rate returned is the rate for a whole singles unit.
  //
- int get_singles_rate(int singles_bin_index, int time_slice) const;
+ int get_singles_rate(unsigned int singles_bin_index,unsigned int time_slice) const;
```
KrisThielemans, line 128: revert. can't do this. It's a `virtual` function

src/include/stir/data/SinglesRatesFromGEHDF5.h
```diff
@@ -135,7 +135,7 @@ public RegisteredParsingObject<SinglesRatesFromGEHDF5, SinglesRates>
  //
  // The singles rate returned is the rate for a whole singles unit.
  //
- void set_singles_rate(int singles_bin_index, int time_slice, int new_rate);
+ void set_singles_rate(unsigned int singles_bin_index,unsigned int time_slice,unsigned int new_rate);
```
KrisThielemans, line 138: revert. can't do this. It's a `virtual` function

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
```diff
@@ -310,13 +310,12 @@ read_norm_data(const string& filename)
...
-    std::array<unsigned long long int, 2> stride = {1, 1};
-    const std::array<unsigned long long int, 2> count  = {45, 448};
-    std::array<unsigned long long int, 2> offset = {0,0};
-    std::array<unsigned long long int, 2> block = {1, 1};
+
     unsigned int total_size = num_rings*num_detectors_per_ring;
     stir::Array<1, float> buffer(0, total_size-1);
-    m_input_hdf5_sptr->get_from_2d_dataset(offset, count, stride, block, buffer);
+    std::array<unsigned long long int, 2> stride = {1, 1};
+    std::array<unsigned long long int, 2> offset = {0, 0};
```
KrisThielemans, line 317: could remove these as default?

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -419,94 +617,144 @@ Succeeded GEHDF5Wrapper::initialise_efficiency_factors(const std::string& path)
...
+    m_memspace_ptr= new H5::DataSpace(2, str_dimsf);
     m_dataset_sptr->read(output.get_data_ptr(), H5::PredType::NATIVE_FLOAT, *m_memspace_ptr, m_dataspace);
     output.release_data_ptr();
 
-    //  // TODO error checking
     return Succeeded::yes;
 }
 
 //! \todo Array read as UINT32 should have an output of std::uint32_t.
 //! \todo Check the H5 data type as it's unlikely that it is NATIVE_UINT32, it may be STD_U32LE.
```
KrisThielemans, line 730: remove this comment (and other similar ones). [Doc on `H5::DataSet::read`](http://davis.lbl.gov/Manuals/HDF5-1.6.5/cpplus_RM/classH5_1_1DataSet.html#a10) says
>void H5::DataSet::read (void * buf, const DataType & mem_type,...)
> This function reads raw data from this dataset into the buffer buf, converting from file datatype and dataspace to memory datatype mem_type and dataspace mem_space.

Therefore, it's correct that we want the memory type specify as a "native uint32"

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -419,94 +617,144 @@ Succeeded GEHDF5Wrapper::initialise_efficiency_factors(const std::string& path)
...
+    m_dataspace.selectHyperslab(H5S_SELECT_SET, str_dimsf, offset.data());
+    m_memspace_ptr= new H5::DataSpace(2, str_dimsf);
     m_dataset_sptr->read(output.get_data_ptr(), H5::PredType::NATIVE_FLOAT, *m_memspace_ptr, m_dataspace);
     output.release_data_ptr();
 
-    //  // TODO error checking
     return Succeeded::yes;
 }
 
 //! \todo Array read as UINT32 should have an output of std::uint32_t.
```
KrisThielemans, line 729: this comment is valid. Let's keep it but just put
```c++
 BOOST_STATIC_ASSERT(sizeof(unsigned int)==4); 
```
in the code below. Then it's checked at compilation time.

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -228,9 +395,13 @@ Succeeded GEHDF5Wrapper::initialise_exam_info()
     unsigned int num_time_slices = 0;
     H5::DataSet timeframe_dataspace = file.openDataSet("/HeaderData/SinglesHeader/numValidSamples");
     timeframe_dataspace.read(&num_time_slices, H5::PredType::NATIVE_UINT32);
+
+    if(is_list_file())
+        m_num_singles_samples=num_time_slices;
```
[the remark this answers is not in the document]
AnderBiguri: `initialise_exam_info` (this function) is called always at `open()`, we can not enforce the file type.

src/include/stir/IO/GEHDF5Wrapper.h
AnderBiguri: So, I have done this a bit due to lack of knowledge of how `Scanner` work sin STIR. But if the idea was "check if Signa, and if so, initialize from Signa scanner, then read Header info" I assumed this may happen in the future with D690, D710 etc. Meaning at some point we would need to know which scanner is in particular the one being used. 

So to future proof, I created that enum class, that way you can also directly check which scanner you are using, if needed. 

If the above is not how we will handle different scanners, then it makes sense to just remove it. Should I then get rid of it?

src/IO/GEHDF5Wrapper.cxx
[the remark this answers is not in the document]
KrisThielemans: ok, then this shouldn't be in this function. Its location a remnant of the erroneous use of `TimeFrameDefinitions`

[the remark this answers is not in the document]
AnderBiguri: I can read it in the listmode initializer. Do you want me to get rid of the rest of the time_frame stuff too?

src/include/stir/IO/GEHDF5Wrapper.h
AnderBiguri: We don't use both no. Want me to get rid of the one we dont?

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
AnderBiguri: Yet error if they are not these values?

AnderBiguri: Ah wait, you mean remove this, and add them as a default?

src/include/stir/IO/GEHDF5Wrapper.h
KrisThielemans: yes, sorry. I thought you had them as default in some of these. If we have them everywhere, the "client" code becomes cleaner.

KrisThielemans: I'd remove it. @pw351 will need to put some code in the `initialise_scanner` function anyway. why duplicate it?

src/include/stir/IO/GEHDF5Wrapper.h
KrisThielemans: sure.

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
KrisThielemans: yes, sorry. I thought you had them as default in some of these. If we have them everywhere, the "client" code becomes cleaner.

src/include/stir/IO/GEHDF5Wrapper.h
AnderBiguri: @KrisThielemans correct, but then path can not have default (only last parameters can have defaults). To be honest, all those paths are unused and I think unnecessary....

src/IO/GEHDF5Wrapper.cxx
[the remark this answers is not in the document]
KrisThielemans: As currently the `num_singles` isn't stored in `exam_info`, it is confusing to me if we read it here.

The time frame definitions constructed here are not correct. They are appropriate for the singles only (and only in listmode I presume). I haven't checked where they are used, but at least for the moment, I'd move it there, until we are sure how it is with non-list files.

Instead, time-frame-defs should be read from `HeaderData/AcqStats`, i.e. .`frameStartTime` and `frameDuration`. So, construct a `vector` with a single `std::pair` with `frameStartTime` and `frameStartTime+frameDuration/1000` I believe.


Also, set 
````
exam_info.start_time_in_secs_since_1970=double(header.sharc_rdf_acq_stats_data.scanStartTime
```

src/include/stir/IO/GEHDF5Wrapper.h
KrisThielemans: if we don't need them, get rid of them. simpler is better.

src/include/stir/IO/GEHDF5Wrapper.h
AnderBiguri: I did this to avoid duplication hehe. https://github.com/UCL/STIR/issues/583#issuecomment-644878246 

But sure I can remove it.

src/IO/GEHDF5Wrapper.cxx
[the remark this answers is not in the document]
AnderBiguri: Haven't really followed this last comment sorry. 

Sure, `num_singles_samples` goes into listmode initializer. 

I can move the code for the time frames somewhere else, but its stored in `exam_info_sptr`.
Do you mean I should move them somewhere else, but also cosntruct them differently, with data from somewhere else (AcqStats)?

Also that last line, I don't know how to implement it. What is `header`?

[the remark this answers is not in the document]
KrisThielemans: the current time-frame code is incorrect. It gives 1s time frames because that is what the singles sampling rate is. @pw351 did that to be able to pass it to the `SinglesRates*` object I guess (I haven't  checked if it's actually used).

the proposed code sets correct time frame info. This new code should remain here.

Final line, I copied some stuff and forgot to edit. I meant
```
exam_info.start_time_in_secs_since_1970=double(scanStartTime);
```
with the latter also read from the `AcqStats` dataset.

If thiiss itoo confusing, let me know and I'll do this myself.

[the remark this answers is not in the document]
AnderBiguri: Ah yes, it was used like that, but I changed it now, and it reads from `num_singles_samples` instead.

I'll give this a go.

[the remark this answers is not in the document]
AnderBiguri: So, some issue, maybe. 

the sino and listmode data do return the same value  for `frameStartTime`, but `norm3d` and `geo3d` do not (a different one each). Is this expected/OK?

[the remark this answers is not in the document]
AnderBiguri: On top of that, it seems that frameStartTime is also since 1970, should I subtract that time? Not sure if the floats are storing these large numbers well.

AnderBiguri:
> SIRF will need work. I think first step is to see if you get the same output when running the `stir_math` "conversion" and `construct_randoms_from_GEsingles`. (You can use `compare_projdata`).
> Then we create a sample script (like for the mMR).

`construct_randoms_from_GEsingles` returns:
WARNING: Interfile warning: inconsistent effective_central_bin_size_in_cm
Value in header is 0.201565 while I expect 0.225309 from the inner ring radius etc
Ignoring value in header
but does not error, as far as I can see. 

Figuring out how to test this is not trivial to me right now, and I need to do some other things now, so either test this yourself, give me some hints of how to test it (how do I call `stir_math`, why do I call it, what should I be expecting, what parts does this test, etc) or I will just pick it up on Tuesday  23/06.

src/IO/GEHDF5Wrapper.cxx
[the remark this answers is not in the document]
KrisThielemans: > the sino and listmode data do return the same value for frameStartTime, but norm3d and geo3d do not (a different one each). Is this expected/OK?

expected. The norm comes from a specific acquisition on a different date.

>  seems that frameStartTime is also since 1970, should I subtract that time? 

ok, please subtract `scanStartTime` then.

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -30,6 +32,7 @@
 */
 
 #include "stir/IO/GEHDF5Wrapper.h"
+#include "stir/IndexRange3D.h" // AB: todo, this should be in .h?
```
KrisThielemans, line 35:
```suggestion
#include "stir/IndexRange3D.h"
```
no. why. it isn't used there.

src/IO/GEHDF5Wrapper.cxx
```diff
@@ -39,43 +42,209 @@ namespace RDF_HDF5 {
...
-    std::string read_str_manufacturer;
+unsigned int GEHDF5Wrapper::get_num_singles_samples()
+{
+    return m_num_singles_samples;
+}
 
-    H5::DataSet dataset = file.openDataSet("/HeaderData/ExamData/scannerDesc");
-    dataset.read(read_str_scanner,vlst);
+
+// AB Cheks if its a GE Scanner file, and returns if it is a GE scanner or not.
```
KrisThielemans, line 69: comments like this have to be in the .h, and in doxygen form

src/include/stir/IO/GEHDF5Wrapper.h
```diff
@@ -53,44 +55,58 @@ class GEHDF5Wrapper
...
 public:
 
+    //! check signature to see if this is a GE RDF9 or higher file
+    /*! calls current_GE_signature(H5::H5file&) */
     static bool check_GE_signature(const std::string& filename);
 
-    static bool check_current_signature(H5::H5File& file);
+    //! check signature to see if this is a GE RDF9 or higher HDF5 file
+    /*! does minimal checks to see if expected datasets are present in the HDF5 object */
+    static bool check_GE_signature(H5::H5File& file);
```
AnderBiguri, line 64: You can not overload static methods, right?

KrisThielemans: sure. why not? You cannot use `virtual` for `static` functions (so you cannot "override" them in a derived class).

AnderBiguri: Ah, google failed me and gave me a StackOverflow question about static vs non-static. My bad.

pwadhwa351:
> @pw351 this seems now ready for testing, although you'll have to merge your changes for `initialise_scanner`. Please do a PR on this one.
> 
> It'd be great if you could do this this week.

I have started a PR: GE Signa Parameters Fix #583 #1
@KrisThielemans May it be possible to have a look to see what objectives are missing from there.

KrisThielemans:
@pw351 I saw your commit on your own pwadhwa351/STIR, but it isn't a PR? It's easier to comment from a PR. The links in your answer above lead to issues.

Please merge current branch into yours first.Then it should be very little work (unless your tests would fail!)

KrisThielemans:
Sinogram reading might be working now. It does recognise the scanner etc. `extract_segments` on the decompressed VQC sinogram and display in `amide` gives
image

However, reading the RDF sinograms is very slow! I'm not sure which bit is causing this. @pw351 was this the case in your version as well?

AnderBiguri:
@KrisThielemans I am permuting them when being read, that is bound to be slow. I think I made the permutation so it maximizes cache for reading, yet it will have a notable impact anyway...

AnderBiguri:
Happens here:

https://github.com/UCL/STIR/blob/b266b949f6b7afa9d23367ce57b59aa77705dc08/src/IO/GEHDF5Wrapper.cxx#L623

KrisThielemans:
no, that's not it. It's actually the HDF5 `read` call itself https://github.com/AnderBiguri/STIR/blob/b266b949f6b7afa9d23367ce57b59aa77705dc08/src/IO/GEHDF5Wrapper.cxx#L614
There doesn't seem to be much that we can do about that!

Some timings on my (slow) laptop within VB. Done using `list_projdata_info some.rdf` or `list_projdata_info --all some.rdf`. The former reads all the data via `ProjDataGEHDF5::initialise_viewgram_buffer`, while the latter in addition call `ProjDataGEHDF5::get_viewgram` for all of the data to get the values. This is currently also slow as it sums over all TOF bins. I then manually commented out bits in `ProjDataGEHDF5::initialise_viewgram_buffer` and computed the difference in timings. All very approximate and a profiler would be much better, but it's clear what the culprit is here.

|task  | real time | user time |
| -----| --------- | -----------|
| minimal construction of data without reading/copying into tof_data | 14s | 9s|
| `H5::read` from shared folder | 190s | 155s |
| `H5::read` from shared folder | 165s | 155s |
| "transpose" | 20s | 20s|
| call all `get_viewgram` + min/max calculations| 20s| 16s|

Of course, maybe this is different on a different system, but it's still quite remarkable.

Oh well.

AnderBiguri:
Hum thats weird indeed. However note that in my (very good) desktop this was taking like 20s in total (unmeasured approximate time), before the transpose. I can make some quick tests tomorrow if you want.

KrisThielemans:
ok, well, maybe I should throw my laptop in the bin. 20s isn't great, but ok. Might be interesting to document the following test:
```
time list_projdata_info --all rdf
time list_projdata_info --all rdf
stir_math -s interfilecopy.hs rdf
time list_projdata_info --all interfilecopy.hs
time list_projdata_info --all interfilecopy.hs
```
but I think we have more urgent things to worry about then.

KrisThielemans:
Travis OSX XCode9 job failed due to
```
Warning: You are using macOS 10.12.
We (and Apple) do not provide support for this old version.
You will encounter build failures with some formulae.
<snip>
Warning: Your Xcode (9.0) is outdated.
Please update to Xcode 9.2 (or delete it).
```
This isn't related to this PR, but just our Travis file.

AnderBiguri:
> ok, well, maybe I should throw my laptop in the bin. 20s isn't great, but ok. Might be interesting to document the following test:
> 
> ```
> time list_projdata_info --all rdf

real	0m43.699s
user	0m42.728s
sys	0m0.904s

(repetitions take the same time, assuming you did not want that though, just the same line without --all)

> time list_projdata_info  rdf

real	0m34.603s
user	0m33.786s
sys	0m0.792s

> stir_math -s interfilecopy.hs rdf

> time list_projdata_info --all interfilecopy.hs
real	0m0.743s
user	0m0.379s
sys	0m0.363s

> time list_projdata_info  interfilecopy.hs
real	0m0.005s
user	0m0.005s
sys	0m0.000s



Huh, that is a difference eh.

KrisThielemans:
> Huh, that is a difference eh.

yup! weird. HDF5 is supposed to be for large data. No clue.

KrisThielemans:
good news on the unlisting! With
```
lm_to_projdata Parameters:=
  input file :=  ${INPUT}
  output filename prefix :=  ${OUTPUT}
  template_projdata := ${TEMPLATE}
End:=
```
and
```
INPUT=LIST0000uncompressed.BLF OUTPUT=unlist TEMPLATE=interfilecopy.hs \
   lm_to_projdata lm_to_projdata.par
compare_projdata  unlist_f1g1d0b0.hs interfilecopy.hs

Maximum absolute error = 1
Maximum in (1st - 2nd) = 1
Minimum in (1st - 2nd) = 0
Error relative to sup-norm of first data-set = 4 %
Projection arrays deemed different
```
I guess this means the listmode file has a longer time interval (I didn't check the times).
@pw351 does that sound reasonable?

@AnderBiguri it does mean you got your transpose correct!

AnderBiguri:
> @AnderBiguri it does mean you got your transpose correct!
ah, its the upteenth time I need to code one of these.... :) Happy that I haven't unlearned it! 
Let me know if I can do something more in the PR. I kind of lost track of where this is now.

KrisThielemans:
unlisting results were after the previous commit

pwadhwa351:
> good news on the unlisting! With
> 
> ```
> lm_to_projdata Parameters:=
>   input file :=  ${INPUT}
>   output filename prefix :=  ${OUTPUT}
>   template_projdata := ${TEMPLATE}
> End:=
> ```
> 
> and
> 
> ```
> INPUT=LIST0000uncompressed.BLF OUTPUT=unlist TEMPLATE=interfilecopy.hs \
>    lm_to_projdata lm_to_projdata.par
> compare_projdata  unlist_f1g1d0b0.hs interfilecopy.hs
> 
> Maximum absolute error = 1
> Maximum in (1st - 2nd) = 1
> Minimum in (1st - 2nd) = 0
> Error relative to sup-norm of first data-set = 4 %
> Projection arrays deemed different
> ```
> 
> I guess this means the listmode file has a longer time interval (I didn't check the times).
> @pw351 does that sound reasonable?

Yes, it must be due to start time and acquisition time difference. It seems fine!

KrisThielemans:
@pw351 would you be able to test the singles? The output is vaguely reasonable, but with this acquisition it cannot really be tested if it's correct. You should be able to compare with what you had before.

If that works, it's a matter of adding a few example files and we should be good to merge.

I still think it'd be best to put this in `release_4` though, so that might need some rebasing (to preserve history of this PR). That'll need some thought.

KrisThielemans:
replaced this PR by #601. That was created by using
```
git checkout release_4
git checkout -b GERDF9
git checkout RDF9_RDF10
git rebase --onto GERDF9 88e7731eb3181d8ff237fd6109e74bce2f02e13d
```

PR #575, opened 2020-06-11
Recon improvements2

KrisThielemans:
continuation of #570


KrisThielemans:
actual duplicate of #570

Issue #105, opened 2017-09-11
remove obsolete function in ProjDatInfoCylindricalNoArcCorr


KrisThielemans:
There's overlap between old and new (safer) functions. We should get rid of the old ones. Let's try to document here what needs to be done.

Functions that handle only rings or only tang_pos/view (these should become either private or protected)
- [ ] `get_det_num_pair_for_view_tangential_pos_num` (replace with `get_det_pos_pair_for_bin`). Used in a few places including `ProjMatrixByBinUsingRayTracing` and some normalisation code.

- [ ] `get_view_tangential_pos_num_for_det_num_pair` (make private). Used in
  - `get_bin_for_det_pair`

- [ ] `get_segment_axial_pos_num_for_ring_pair` (make private?)
  - `CListEventCylindricalScannerWithViewTangRingRingEncoding` (used there to save some computation)
  - `get_bin_for_det_pair`

- [ ] `get_num_ring_pairs_for_segment_axial_pos_num` (use `get_num_det_pos_pairs_for_bin`). Used in
  - `LmToProjData::get_compression_count` 
  - `get_num_det_pos_pairs_for_bin`

- [ ] `get_ring_pair_for_segment_axial_pos_num` (make protected). Used in
  - `get_det_pair_for_bin`

- [ ] `get_all_ring_pairs_for_segment_axial_pos_num` (make protected). Used in
  - `get_all_det_pos_pairs_for_bin`

Functions that use 4 coordinates (these should be removed)
- [ ] `get_bin_for_det_pair` (replace with `get_bin_for_det_pos_pair`, which currently calls the old function). The old function is used in 
  - `get_bin(const LOR<float>& lor)`
  - `find_bin_given_cartesian_coordinates_of_detection` (which is obsolete)
  - `list_detectors_and_bin_info.cxx`

- [ ] `find_bin_given_cartesian_coordinates_of_detection` (replace with `get_bin(LOR&)`). This is not used.

- [ ] `find_scanner_coordinates_given_cartesian_coordinates` (replace with `get_LOR` or `get_LOR_as_2_points`). Used in
  - `find_bin_given_cartesian_coordinates_of_detection`

- [ ] `find_cartesian_coordinates_of_detection` (replace with `get_LOR` or `get_LOR_as_2_points`). Used in
  - ScatterEstimationByBin::find_detectors

- [ ] `find_cartesian_coordinates_given_scanner_coordinates_of_the_front_surface`. This is not used.

[...]

PR #577, opened 2020-06-12
block and generic scanner geometry 

pkhateri:
new classes and changes for accurate scanner geometry


KrisThielemans:
https://travis-ci.org/github/UCL/STIR/jobs/697593231#L3130-3142 says
```
In file included from /home/travis/build/UCL/STIR/src/include/stir/make_array.h:209:0,
                 from /home/travis/build/UCL/STIR/src/include/stir/GeometryBlocksOnCylindrical.h:35,
                 from /home/travis/build/UCL/STIR/src/include/stir/ProjDataInfoBlocksOnCylindricalNoArcCorr.h:32,
                 from /home/travis/build/UCL/STIR/src/include/stir/listmode/CListRecordSAFIR.inl:33,
                 from /home/travis/build/UCL/STIR/src/include/stir/listmode/CListRecordSAFIR.h:261,
                 from /home/travis/build/UCL/STIR/src/include/stir/IO/SAFIRCListmodeInputFileFormat.h:47,
                 from /home/travis/build/UCL/STIR/src/IO/IO_registries.cxx:67:
/home/travis/build/UCL/STIR/src/include/stir/make_array.inl: In instantiation of ‘stir::Array<(num_dimensions + 1), T> 
stir::make_array(const stir::Array<num_dimensions, T>&, const stir::Array<num_dimensions, T>&, const stir::Array<num_dimensions, T>&) [with int num_dimensions = 1; T = float]’:

/home/travis/build/UCL/STIR/src/include/stir/GeometryBlocksOnCylindrical.inl:39:7:   required from here

/home/travis/build/UCL/STIR/src/include/stir/make_array.inl:255:113: error: conversion from ‘stir::NumericVectorWithOffset<stir::Array<1, float>, float>’ to non-scalar type ‘const stir::Array<2, float>’ requested

   const Array<num_dimensions+1,T> a = NumericVectorWithOffset<Array<num_dimensions,T>,T>(make_vector(a0, a1, a2));

make[2]: *** [src/swig/CMakeFiles/_stir.dir/__/IO/IO_registries.cxx.o] Error 1
```
Not sure why all the previous `IO_registries` compilations worked though.

KrisThielemans:
@pkhateri I believe I fixed the problem in #605 . If you merge STIR master in here, that might work.

Potentially I could do it for you, but I'd be pushing to your `master` branch apparently. 9always best to use a branch for a PR in the future).

KrisThielemans:
same problem when compiler `_stir.dir/IO_registries` related to SWIG strangely enough.
https://travis-ci.org/github/UCL/STIR/jobs/707135030#L3168
```
/home/travis/build/UCL/STIR/src/include/stir/make_array.inl: In instantiation of ‘stir::Array<(num_dimensions + 1), T> stir::make_array(const stir::Array<num_dimensions, T>&, const stir::Array<num_dimensions, T>&, const stir::Array<num_dimensions, T>&) [with int num_dimensions = 1; T = float]’:
/home/travis/build/UCL/STIR/src/include/stir/GeometryBlocksOnCylindrical.inl:39:7:   required from here
/home/travis/build/UCL/STIR/src/include/stir/make_array.inl:255:113: error: conversion from ‘stir::NumericVectorWithOffset<stir::Array<1, float>, float>’ to non-scalar type ‘const stir::Array<2, float>’ requested
   const Array<num_dimensions+1,T> a = NumericVectorWithOffset<Array<num_dimensions,T>,T>(make_vector(a0, a1, a2));
```

KrisThielemans:
Example Travis test log https://travis-ci.org/github/UCL/STIR/jobs/707731577

pkhateri:
> @pkhateri I've resolved the SWIG compilation error. However, there are now a lot of failures in the tests. Many of them just state
> 
> ```
> ERROR: Crystal Map not defined!
> ```
> 
> Python tests show
> 
> ```
> RuntimeError: ProjDataInfoGeneric::get_ring_pair_for_segment_axial_pos_num does not work for data with axial compression
> ```
> 
> implying that `ProjDataInfoGeneric` is called for "normal" interfile headers.
> 
> I do not understand your intentions with this PR well enough to solve this. Can you help?

I pushed some fixes to the repository. Now all the tests should pass.
The modifications were originally applied into an old version (from ~3 yrs ago). When I merged that version to this new one, not all changes were copied. So I added them manually now. Hopefully, there's no problem any more. Otherwise please let me know.
I also added some new utilities...

KrisThielemans:
thanks @pkhateri sorry for the delay. `ctest` is indeed happy. great. Unfortunately there were still quite  a lot of errors with the `recon_test_pack`. You could have a look at https://travis-ci.org/github/UCL/STIR/jobs/712819897 but will just have to run the scripts yourself to figure out what they are, as Travis doesn't show us the logs at the moment. Feel free to ask questions of course.

Thanks for the extra utilities. Could you add some brief info on them in the STIR_UsersGuide.tex? Ideally also a bit more documentation in their headers (doxygen) if it isn't immediately obvious what they do. I'm reluctant to add the following though:
- `find_counts_in_cylindrical_ROI.cxx` can be done via `list_ROI_values` I guess
- `convert_projdata_types.cxx` seems unnecessary. I believe it just makes a copy, but you can do that with `stir_math -s copy orig`
- `trim_projdata.cxx` seems unnecessary: `SSRB -t <trim> out in 1` should do the same

Please also update `documentation/release_5.0.htm`.

KrisThielemans:
Codacy also found some small issues. you can see them by clicking on the "details" above. Current direct link is https://app.codacy.com/manual/KrisThielemans/STIR/pullRequest?prid=5688358

pkhateri:
Sorry, I have not updated the recon_test_pack according to the new version. Now, it should work. 
In the block/generic geometry new parameters have been added to the interfile header. So I modified the inputs according to new interfiles.
I will update the documentation soon. 

About those utilities: 
`find_counts_in_cylindrical_ROI.cxx`: It works for projection data too. But I don't have a strong opinion in keeping it. 
`convert_projdata_types.cxx`: true.
`trim_projdata.cxx`: this is different from `SSRB`. It crops the projection data, it reduces number of tangential positions.

KrisThielemans:
thanks!

> `trim_projdata.cxx`: this is different from `SSRB`. It crops the projection data, it reduces number of tangential positions.

SSRB can do that as well: usage
```
SSRB [-t num_tangential_poss_to_trim] \
        output_filename input_projdata_name \
        [num_segments_to_combine \
        [ num_views_to_combine [do_norm [max_in_segment_num_to_process ]]]]
```
so `SSRB -t  50 out in` does the trick

KrisThielemans:
Appveyor is failing with
```
C:\projects\stir\src\include\stir/Scanner.h(130): error C2059: syntax error: ',' 
```

pkhateri:
@KrisThielemans Now, it fails in `lm_to_projdata` for root files. I cannot check where the bug is exactly as I don't have root installed on my computer and don't have the time to install it at the moment. Harry said his student can help but I cannot mention him here: "Harry Tunnicliffe". I cannot mention Harry himself either!

src/include/stir/Scanner.inl
```diff
@@ -384,7 +463,21 @@ Scanner::get_transaxial_singles_unit(int singles_bin_index) const {
   return(singles_bin_index % get_num_transaxial_singles_units());
 }
 
-
+// For retrieving the coordinates / detector, ring id from the scanner
+stir::DetectionPosition<>
+Scanner::get_detpos_given_id(const stir::DetectionPosition<> det_pos) const{
```
KrisThielemans, line 468: making these accept a `const &` could speed it up a tiny-bit

KrisThielemans (changes requested):
I still haven't reviewed this properly but a few tiny things that I happened to notice...

jafische:
Dear @pkhateri, @KrisThielemans ,
I also had a look. The Travis test with ROOT input fails because the scanner equality check in LmToProjData.cxx:259-270 fails. This is because of different defaults. Way to fix is in the attached [diff.txt](https://github.com/UCL/STIR/files/5517922/diff.txt).
Hope that helps a bit. The [simulation test](https://github.com/UCL/STIR/blob/master/examples/PET_simulation/run_simulation.sh) fails, not sure if this is an issue due to this PR. It seems to me that [the input file generation script](https://github.com/UCL/STIR/blob/master/examples/PET_simulation/generate_input_data.sh) depends on the order of questions of create_projdata_template and is not up-to-date.
Kind regards,
Jannis

KrisThielemans:
thanks @jafische defaults should be consistent of course. I am a bit worried though that the default falls back to `Cylindrical` (i.e. the old code) as it should, but is nevertheless affects by new variables that are relevant only for blocks. It also seems confusing to have defaults that are not zero for variables that in the cylindrical case are effectively zero.

maybe I misunderstand this a bit.

jafische:
@KrisThielemans : Sure, I did not think about reasonable default values, but also thought that negative values for distances are not too intuitive. As far as I am concerned, they could be 0.

pkhateri:
Thanks @jafische, I added the fix.

src/IO/interfile.cxx
```diff
@@ -1416,7 +1421,103 @@ write_basic_interfile_PDFS_header(const string& header_file_name,
...
+              		 proj_data_info_sptr->get_ring_spacing(),
+              		 scanner.get_ring_spacing());
+
+              output_header << scanner.parameter_info();
+
+              output_header << "effective central bin size (cm) := "
+              		     << proj_data_info_sptr->get_sampling_in_s(Bin(0,0,0,0))/10. << endl;
+
+            }
+        }  // end generic scanner
```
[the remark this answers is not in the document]
pkhateri: done.

src/include/stir/IO/InterfileHeader.h
```diff
@@ -288,6 +290,18 @@ class InterfilePDFSHeader : public InterfileHeader
...
   //! Reference energy.
   float reference_energy;
+  
+  //! new variables for block geometry
+  std::string scanner_orientation;
+  std::string scanner_geometry;
+  float axial_distance_between_crystals_in_cm;
+  float transaxial_distance_between_crystals_in_cm;
+  float axial_distance_between_blocks_in_cm;
+  float transaxial_distance_between_blocks_in_cm;
```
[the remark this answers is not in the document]
pkhateri: done.

src/include/stir/Scanner.inl
pkhateri: done.

NikEfth:
Hi @KrisThielemans and @pkhateri 
I started moving my multilayer code here as this branch is cleaner than mine. 
If you don't have to accept the PR before Xmas, I think that I will have it ready. 
Even if you merge it, I am sure if my code would be in trouble... just thinking.

src/buildblock/ProjDataInfoBlocksOnCylindricalNoArcCorr.cxx
```diff
@@ -0,0 +1,538 @@
...
+  det_pos2.axial_coord() = Ring_B;
+
+  if (crystal_map->find_cartesian_coordinate_given_detection_position(coord_1, det_pos1)==Succeeded::yes &&
+      crystal_map->find_cartesian_coordinate_given_detection_position(coord_2, det_pos2)==Succeeded::yes)
+  {
+    return;
+  }
+  else
+  {
+    std::cerr<<"WARNING: couldn't find corresponding cartesian coordinates for given detection positions.\n";
```
KrisThielemans, line 453: this should call `error`

src/utilities/find_normfactors_from_cylinder_data.cxx
```diff
@@ -0,0 +1,158 @@
...
+            find_LOR_intersections_with_cylinder => c1 & c2
+            c12 = |c1-c2| = sqrt(Δx^2+Δy^2+Δz^2)
+            N_lor/c12 should be the same for all therefore:
+            NF_lor= <N> / (N_lor/c12)
+        */
+
+        float N_lor = cylinder_viewgram[ax][tang]; //counts seen by this lor
+        c1 = intersection_coords.p1();
+        c2 = intersection_coords.p2();
+        float c12 = sqrt( pow(c1.z()-c2.z(), 2)    // length of intersection of lor with the cylinder
```
KrisThielemans, line 140: it seems that this calculation does not take attenuation into account?

pkhateri: No, it does not. The input data should be pre-corrected for attenuation, scatter and random.

src/buildblock/ProjDataInfoBlocksOnCylindricalNoArcCorr.cxx
pkhateri: fixed

src/IO/InterfileHeader.cxx
```diff
@@ -620,6 +624,36 @@ InterfilePDFSHeader::InterfilePDFSHeader()
...
+  // new keys for block geometry
+  scanner_orientation = "X";
+  add_key("Scanner orientation (X or Y)",
+          KeyArgument::ASCII, &scanner_orientation);
+
+  scanner_geometry = "Cylindrical";
+  add_key("Scanner geometry (BlocksOnCylindrical/Cylindrical/Generic)",
+          KeyArgument::ASCII, &scanner_geometry);
+
+  axial_distance_between_crystals_in_cm = -0.1;
```
KrisThielemans, line 636: these defaults are still negative. To me this doesn't make sense. They should be zero, as that's what our default cylindrical scanner has.

src/buildblock/GeometryBlocksOnCylindrical.cxx
```diff
@@ -0,0 +1,301 @@
...
+	int num_transaxial_buckets = get_scanner_ptr()->get_num_transaxial_blocks()/num_transaxial_blocks_per_bucket;
+	int num_detectors_per_ring = get_scanner_ptr()->get_num_detectors_per_ring();
+	float axial_block_spacing = get_scanner_ptr()->get_axial_block_spacing();
+	float transaxial_block_spacing = get_scanner_ptr()->get_transaxial_block_spacing();
+	float axial_crystal_spacing = get_scanner_ptr()->get_axial_crystal_spacing();
+	float transaxial_crystal_spacing = get_scanner_ptr()->get_transaxial_crystal_spacing();
+	std::string scanner_orientation = get_scanner_ptr()->get_scanner_orientation();
+
+	// check for the scanner orientation
+	if (scanner_orientation=="Y" || num_transaxial_buckets%4==0 )
```
VietAnhDao, line 159: If number of buckets in transaxial direction is a multiple of 4 this causes scanner orientation to alway to Y.

src/include/stir/ProjDataInfoGeneric.inl
```diff
@@ -0,0 +1,273 @@
...
+START_NAMESPACE_STIR
+
+//! find phi from correspoding lor
+float
+ProjDataInfoGeneric::get_phi(const Bin& bin)const
+{
+	LORInAxialAndNoArcCorrSinogramCoordinates<float> lor;
+	get_LOR(lor, bin);
+	if (bin.view_num()==0 && lor.phi()>0.1)
+		return lor.phi()-_PI;
```
KrisThielemans, line 46: this is a very strange line. where does `0.1` come from?

KrisThielemans:
@pkhateri a `ProjDataInfoGenericArcCorr` class doesn't seem to make logical sense, so I guess we could merge `ProjDataInfoGenericNoArcCorr` qinto `ProjDataInfoGeneric`. Am I correct in thinking that you kept the original structure as much as possible (to avoid making changes, and make my life easier)? Or is there another reason to have these split into 2?

(same question for the `BlocksOnCylindrical` of course)

pkhateri (commented):
> @pkhateri a `ProjDataInfoGenericArcCorr` class doesn't seem to make logical sense, so I guess we could merge `ProjDataInfoGenericNoArcCorr` into `ProjDataInfoGeneric`. Am I correct in thinking that you kept the original structure as much as possible (to avoid making changes, and make my life easier)? Or is there another reason to have these split into 2?
> 
> (same question for the `BlocksOnCylindrical` of course)

Yes, that's right. This was originally decided by you, Jannis and me to keep the same structure for the  `BlocksOnCylindrical`. Later, we kept the same thing for the  `Genric` too.

src/buildblock/GeometryBlocksOnCylindrical.cxx
pkhateri: That's true. When  `num_transaxial_buckets%4==0`, there's always two faces of the scanner perpendicular to the y axis.

src/include/stir/ProjDataInfoGeneric.inl
pkhateri: The output of ` lor.phi()` for `view=0` is either close to zero or is close to PI. I put `lor.phi()>0.1` instead of `lor.phi()>0` in the condition to avoid values which are close to zero (but not zero).

pkhateri:
@KrisThielemans I don't get why the CI check failed. The only thing that has changed compared to the previous version (which passed all the checks) is few commented lines added for a new license.

src/buildblock/GeometryBlocksOnCylindrical.cxx
VietAnhDao: not necessarily since we can have y-axis going though two corners of the scanner rather than two faces. Due to this condition it will not allow us to get orientation X and num_transaxial_buckets%4==0.

src/include/stir/ProjDataInfoGeneric.inl
KrisThielemans: we've put in changes in `master` such that `lor.phi()` no longer starts from zero. We'll have to merge those into the block code. So this condition will be unexpected then.

Also, 0.1 is not so small in radians.

KrisThielemans:
> @KrisThielemans I don't get why the CI check failed. The only thing that has changed compared to the previous version (which passed all the checks) is few commented lines added for a new license.

You've merge `UCL/master`. I noticed that some tests were broken, but haven't managed to find the reason yet. There must be some conflict between the changes (probably those related to view-offset that were merged in from #181) and your PR.

src/buildblock/GeometryBlocksOnCylindrical.cxx
pkhateri: You can have infinite number of orientations of the scanner toward x and y axis. Not all of them are defined here. Only two common cases are defined: the polygon face is either perpendicular to x axis or to y axis.
If `num_transaxial_buckets%4==0` is true and the y axis goes through the corners of the scanner the x-axis goes through the corner too. It means it is none of the defined orientations.

src/include/stir/ProjDataInfoGeneric.inl
pkhateri: The changes in the `LORCoordinates` class has been made three years ago. I have updated this branch with more recent versions of the master branch (few months ago).  I get those outputs now when I call `lor.phi()` for `view=0`.

Yes, 0.1 is not too small but small enough not to be close to PI, and large enough not to be 0.

KrisThielemans:
some things that will need fixing
- [ ] find why tests fails
- [ ] large amount of code duplication between `BlocksOnCylindrical` and `Generic` (I believe that this can be fixed by deriving `BlocksOnCylindrical` from `Generic`, which would make logical sense as well)
- [ ]  `BlocksOnCylindrical` should not allow specifying crystal maps
- [ ] remove X and Y choice and keep only 1, as this can now be handled with the view offset.
(note that X and Y also have numbering of blocks in different ways, should always start at beginning of module)
- `get_s()` calculations have errors. Would be better to derive them generically.

For later:
- [ ] random estimation and norm estimation seem to work but need work-arounds with changing headers at the moment
- [ ] `interpolate_projdata` (used by upsampling) would need to be able to work on grids that are non-uniform grids
- [ ] scatter uses `SSRB` but this fails with `BlocksOnCylindrical` (could be resolved with `num_segments_to_combine`)
- [ ] scatter simulation needs some other fixes to be able to use `BlocksOnCylindrical`

src/include/stir/recon_buildblock/DataSymmetriesForBins_PET_CartesianGrid.inl
```diff
@@ -89,11 +97,56 @@ find_transform_z(
...
+
+    }
+      }
+    // generic implementation
+    else
+    {
+        const ProjDataInfoGeneric* proj_data_info_gen_ptr =
+        static_cast<const ProjDataInfoGeneric *>(proj_data_info_ptr.get());
+
+        const float delta = proj_data_info_gen_ptr->get_average_ring_difference(segment_num);
```
danieldeidda, line 133: When running in debug mode  the test_DataSymmetriesForBins_PET_CartesianGrid we get the vectorWithOffset assertion "begin_allocated_memory <= num+start" it seems the min_ring_diff and max_ring_diff indeces go from 0 to 4 while segment_num is 5

KrisThielemans:
@danieldeidda I've merged `master` onto here. I just had conflicts in the `CMakeLists.txt`. However, `test_proj_data_info` failed for me. I guess it'll fail on GHA as well. Unfortunately, I didn't check if it worked before I did the merge.

src/include/stir/ProjDataInfoGeneric.inl
danieldeidda: Yes I agree. We test lor.phi() as well as the other coordinates. If there is a swap I suspect the test would fail?

KrisThielemans: I guess the test should fail, but it will probably only start to fail when we introduce rotations.

danieldeidda: When I run the test with the current parisa:master get_s() gives opposite sign results (for test_projdata_info). The test_blocks_on_cylindrical_projectors crashes with an instance of 'H5::PropListException' which I have no idea what it is, but it happens with HDF5 ON or OFF

KrisThielemans: comment in the right place?

question is if `test_projdata_info` worked at some point on your machine, and if so, what point that was :-)

`H5` stuff is more than likely a conflict between HDF5 libraries (between compilations and running). if it happens with HDF5 OFF, it means you haven't rebuilt STIR. (Re-setting it at the SuperBuild level  might not reset it at STIR level). In any case, `test_blocks_on_cylindrical_projectors` works on GHA (and probably Appveyor)

KrisThielemans:
`Scanner::read_detectormap_from_file` sorts the coordinates. This could be unexpected when comparing detector numbers./coordinates. It seems particularly dangerous when using the SAFIR listmode files as this uses ring/detector numbers. @emikhay. this might affect you.

KrisThielemans:
actually I'm not sure what the sorting does. 

However, I discover that in `examples/SAFIR_genericScanner`, `list_lm_events` reports a negative "effective central bin size". This turns out to be `ProjDataInfo::get_sampling_in_s(bin)` [assumes](https://github.com/pkhateri/STIR/blob/e35385cafa56b4200c2fd6722510facc0739f555/src/buildblock/ProjDataInfo.cxx#L78-L83) that `s` increases with `tang_pos` (and similar `m` with `axial_pos`).
It would be easy to insert `std:abs` in those statements but I have no idea what projectors will do if those assumptions are broken. (and we currently don't check).

By the way, the OSEM in `examples/SAFIR_genericScanner` currently fails with a seg-fault (there is an infinite recursion in the ray-tracing, presumably because of the above)

src/include/stir/ProjDataInfoGeneric.inl
danieldeidda: YEs was not sure where to put the comment. Anyway,get_s() tests were working before I pulled the new version of parisa:master. Now the difference appear when view is zero (phi is actually the same) and the main difference is an opposite sign but occasionally some rounding difference too

KrisThielemans:
great commit with excellent results! Now we should resolve the `ring_radius` problem (I forgot where it was...)

danieldeidda:
> 
> 
> great commit with excellent results! Now we should resolve the `ring_radius` problem (I forgot where it was...)

wasn't that fixed by calculating it using the dot product?

danieldeidda:
> 
> 
> > great commit with excellent results! Now we should resolve the `ring_radius` problem (I forgot where it was...)
> 
> wasn't that fixed by calculating it using the dot product?

`GeometryBlocksOnCylindrical::
build_crystal_maps()
`
start_y uses effective_ring_radius() should that be  the new radius?

KrisThielemans:
> > > great commit with excellent results! Now we should resolve the `ring_radius` problem (I forgot where it was...)
> > 
> > 
> > wasn't that fixed by calculating it using the dot product?
> 
> `GeometryBlocksOnCylindrical:: build_crystal_maps() ` start_y uses effective_ring_radius() should that be the new radius?

there is no new radius? I'm confused. There needs to be one parameter that gives the size of the scanner. It's currently called "radius" which doesn't make too much sense (as it's the half the distance between the orthogonal bucket), but let's just leave that for now (as long as it's documented). It needs to be the "effective" to take DOI into account.

I've had a search for `ring_radius`. I think the only remaining problem that I see is `ProjDataInfoGeneric::get_LOR()` which still uses a `ring_radius` that could be too small. (And I think that it isn't overriden lower in the hierarchy).

As this doesn't seem to be used anywhere, it doesn't matter too much, but at some point it'd bite us of course. To fix it, I believe we can use an arbitrary R, which we can just compute as sqrt(max (x1^2+y1^2, x2^2+y2^2)). another solution would be to let the generic case (i.e. `Scanner` presumably) return a min/max_ring_radius (computed from the crystal map). But that' smore work and I don't see why it'd be necessary

danieldeidda:
I am looking at how to change the numbering to make view=0 correspond to the beginning of the block. I can see to options:
adding a shift in `ProjDataInfoGenericNoArcCorr::
get_view_tangential_pos_num_for_det_num_pair(` or  adding this together with the view offset like it's done for cylindrical
`
  azimuthal_angle_offset = scanner_ptr->get_intrinsic_azimuthal_tilt()+shift_scanner_dependent;`

danieldeidda:
> there is no new radius? I'm confused. 
 sorry my confusion. I was talking about that modification on get_tantheta where we had the radius problem but effectively there is no new radius there

danieldeidda:
> As this doesn't seem to be used anywhere, it doesn't matter too much, but at some point it'd bite us of course. To fix it, I believe we can use an arbitrary R, which we can just compute as sqrt(max (x1^2+y1^2, x2^2+y2^2)). another solution would be to let the generic case (i.e. `Scanner` presumably) return a min/max_ring_radius (computed from the crystal map). But that' smore work and I don't see why it'd be necessary 

OK this sounds good and easy

KrisThielemans:
> I am looking at how to change the numbering to make view=0 correspond to the beginning of the block. I can see to options: adding a shift in `ProjDataInfoGenericNoArcCorr:: get_view_tangential_pos_num_for_det_num_pair(` or adding this together with the view offset like it's done for cylindrical ` azimuthal_angle_offset = scanner_ptr->get_intrinsic_azimuthal_tilt()+shift_scanner_dependent;`
> 
> i think i like the second option better though

seems to me that the first thing to do it create the crystal map with the current numbering. Rotation is a separate issue. To me the easiest way to handle that is just rotate the crystal map. (could be do at generation stage as well).

danieldeidda:
> seems to me that the first thing to do it create the crystal map with the current numbering. Rotation is a separate issue. To me the easiest way to handle that is just rotate the crystal map. (could be do at generation stage as well).

so basically here:
https://github.com/pkhateri/STIR/blob/7ec9546ca5b3ff99abe6845f38f580d7a68d4612/src/buildblock/GeometryBlocksOnCylindrical.cxx#L171

KrisThielemans:
yeah, and in particular the comment at https://github.com/pkhateri/STIR/blob/7ec9546ca5b3ff99abe6845f38f580d7a68d4612/src/buildblock/GeometryBlocksOnCylindrical.cxx#L187
which is the wrong convention. It's not so transparant to me where this convention is used.

good luck!

danieldeidda:
> 
> 
> yeah, and in particular the comment at https://github.com/pkhateri/STIR/blob/7ec9546ca5b3ff99abe6845f38f580d7a68d4612/src/buildblock/GeometryBlocksOnCylindrical.cxx#L187 which is the wrong convention. It's not so transparant to me where this convention is used.
> 
> good luck!

to me this seems to imply that if we set start_x=-R*sin(alpha/2) and start_y=-Rcos(alpha/2)  (where alpha is the angle seen by a bucket) than crystal (0,0,0)  will correspond to that instead of y=-R and x=0

KrisThielemans:
> to me this seems to imply that if we set start_x=-R*sin(alpha/2) and start_y=-Rcos(alpha/2) (where alpha is the angle seen by a bucket) than crystal (0,0,0) will correspond to that instead of y=-R and x=0

not sure, i'd have guessed the loop would run over half a bucket at the moment, but I really haven't checked this.

by the way, I think it's easier to start from the centre, as opposed to computing alpha for the edge. but up to you

KrisThielemans:
oh, could check the X code. @VietAnhDao fixed it at some point

VietAnhDao:
I ended up removing the X orientation for the scanner. I believe we discussed some point of using the view offset, maybe this can be changed by the angle alpha.

KrisThielemans:
> I ended up removing the X orientation for the scanner. I believe we discussed some point of using the view offset, maybe this can be changed by the angle alpha.

I know you did, and it's what we agreed, but I was wondering if at some point you adapted X at some point to remove the problem in the code. If so, maybe you could send that to @danieldeidda for inspiration. If not, that's ok.

VietAnhDao:
If remember right the code generates the 3D position of crystal and the crystal index separately then link them in a dictionary? I did manage to change the crystal index by offsetting ```tangental_coord```  variable at https://github.com/pkhateri/STIR/blob/7ec9546ca5b3ff99abe6845f38f580d7a68d4612/src/buildblock/GeometryBlocksOnCylindrical.cxx#L190 . This will let you change where the first crystal is set at without changing the 3D position of crystal map.

danieldeidda:
> 
> 
> If remember right the code generates the 3D position of crystal(`cart_coord`) and the crystal index (`det_pos`) separately then link them in a dictionary? I did manage to change the crystal index by offsetting `tangental_coord` variable at https://github.com/pkhateri/STIR/blob/7ec9546ca5b3ff99abe6845f38f580d7a68d4612/src/buildblock/GeometryBlocksOnCylindrical.cxx#L190 . This will let you change where the first crystal is set at without changing the 3D position of crystal map.
> 
> Changing the alpha or any of the start_x is a bit risky since the 3D position is required to match the crystal map text file to perform the binning of list mode to projection data.
> 
> Sorry I don't have the code anymore but I hope the comment helps.

Thanks, so are you saying we can just put an offset to tangential position? what about view? I guess that needs to change too

KrisThielemans:
> so are you saying we can just put an offset to tangential position? what about view? I guess that needs to change to

we ONLY change the crystal map please. Where are all the crystals? => Geometric calculation. All the rest is taken care of by existing STIR stuff.

VietAnhDao:
I have tried ```tangential_pos = .... + view_offset``` but I never used it. Me and @emikhay did instead mess with the numbering of the crystal map text file which offset by half a block and this does work. So there are at least 2 ways to do view offset but I wouldn't recommend later method. Given how users can define their own crystal and can manipulate the view offset manually this can cause a problem unless existing STIR can account for this already.

But given no view offset changes then the ```tangential_pos``` is only good for changing the crystal indexed coordinate system without altering physical coordinate of crystal map.

KrisThielemans:
I've tried to check redundancy between "crystal-map" code in 3 places. I found
```
class DetectorCoordinateMapFromFile
{
public:
  //! Reads map from file and stores it.
  void read_detectormap_from_file( const std::string& filename );

  //! Returns a cartesian coordinate given a detection position.
  // uses random displacement
  stir::CartesianCoordinate3D<float> get_detector_coordinate( const stir::DetectionPosition<>& det_pos )
}

class Scanner
{
  // first bit is identical to DetectorCoordinateMapFromFile but in addition sorts coordinates
  void read_detectormap_from_file( const std::string& filename );

  // Get the STIR detection position (det#, ring#, layer#) given the detection position id in the input crystal map
  // used in CListRecordSAFIR.inl for accessing the coordinates
  inline stir::DetectionPosition<> get_detpos_given_id(const stir::DetectionPosition<> & det_pos) const;

  // Get the Cartesian coordinates (x,y,z) given the sorted detection position (det#, ring#, layer#)
  // used in ProjInfoDataGenericNoArcCorr.cxx find_cartesian_coordinates_given_scanner_coordinates
  inline stir::CartesianCoordinate3D<float> get_coords_given_detpos(const stir::DetectionPosition<> det_pos) const;
}

class GeometryBlocksOnCylindrical
{
  //! Get cartesian coordinate for a given detection position
  // conceptually the same as Scanner::get_coords_given_detpos
  inline Succeeded
  find_cartesian_coordinate_given_detection_position(CartesianCoordinate3D<float>& ,
                                                     const DetectionPosition<>&) const;

  //! Get cartesian coordinate for a given detection position
  // function missing in other classes
  Succeeded
  find_detection_position_given_cartesian_coordinate(DetectionPosition<>&,
                                                     const CartesianCoordinate3D<float>&) const;
}
```

KrisThielemans:
`CListEventSAFIR::get_bin` uses 
- Generic: `proj_data_info_gen.get_scanner_ptr()->get_detpos_given_id`
- Cylindrical: `map->get_detector_coordinate` followed by `proj_data_info.get_LOR`
- Blocks: `map->get_detector_coordinate` followed by `proj_data_info_blk.find_scanner_coordinates_given_cartesian_coordinates`

KrisThielemans:
@danieldeidda I've force-pushed unfortunately (I had pushed the SWIG changes to early). You want to `pull` again therefore...

Could you remove the `scanner_orientation` in all places? I've removed those that have external visibility as they broke a python test.

danieldeidda:
> Could you remove the `scanner_orientation` in all places? I've removed those that have external visibility as they broke a python test.

As in remove every instance of scanner orientation?

KrisThielemans:
I have completed the next milestone: the redundancy between the various detector maps is gone after https://github.com/pkhateri/STIR/pull/7. `ProjDataInfoBlocksOnCylindrical` is now very small (could probably be trimmed more). 

We still have 2 maps: one in Scanner (used for projectors etc), and one in SAFIR (used for listmode processing). The latter is because of the "old" functionality of the "virtual scanner". 

The current code is very confusing as it tries to cover the "virtual scanner" for cylindrical and blocks, but for *Generic* it is ignored. Here's what I can see

- `CListModeDataSAFIR` has a `template_proj_data` and a `map`. The former seems never used (except when `LmToProjData` checks if `LmToProjData.template_proj_data.scanner` is identical to the one from the listmode data i.e. in this case `CListModeDataSAFIR:template_proj_data.scanner`)
-  The `map` is passed to `CListEventSAFIR` and used by
    - [get_LOR() ](https://github.com/pkhateri/STIR/blob/de20df3554676eaf28d873700fd1ca472f9e6315/src/include/stir/listmode/CListRecordSAFIR.inl#L49-L50) in all cases
    -  [`get_bin(bin, proj_data_info)`]((https://github.com/pkhateri/STIR/blob/de20df3554676eaf28d873700fd1ca472f9e6315/src/include/stir/listmode/CListRecordSAFIR.inl#L60)) in some cases: if `proj_data_info` is of type 
        - *Generic*: `map` is **ignored**. It goes via `proj_data_info.get_bin_for_det_pos_pair` which is fast, but breaks the "virtual scanner" stuff. (There is no check if the `map` is the same as the scanner's).
        - otherwise, an [LOR is constructed via the map](https://github.com/pkhateri/STIR/blob/de20df3554676eaf28d873700fd1ca472f9e6315/src/include/stir/listmode/CListRecordSAFIR.inl#L94-L95) and
            - *Cylindrical*:  uses `proj_data_info.get_bin(lor)` (as in `master`)
            - *BlocksOnCylindrical*: uses `proj_data_info_blk.find_scanner_coordinates_given_cartesian_coordinates` and `get_bin_for_det_pair`. This will fail at run-time unless the `map` and the `proj_data_info` have the same detector-locations, so it breaks the virtual scanner stuff.
 
This is **far too confusing**, especially as the `template_proj_data` has its own `map`, which is currently ignored (except in the initial check by `LmToProjData`). 


Suggestion:
- if the `map` is present, construct `lor` from `map` and always use `proj_data_info.get_bin(lor)` (as in `master`)
- else use `proj_data_info.get_bin_for_det_pos_pair` if available (i.e. *Generic/Blocks/CylindricalNoArcCorr*), otherwise use `proj_data_info.get_bin(event.get_LOR())` 

Note that case 1 will still fail in *Generic/Blocks* unless the coordinate is in the same place (as they currently do not attempt to find the "nearest" detector), but at least the code will be easier, and there will be no need for a `map` in most cases of interest.

@danieldeidda @jafische @emikhay what do you think?

KrisThielemans:
@danieldeidda I noticed that `GeometryBlocksOnCylindrical::build_maps` does not handle multiple buckets in axial direction. Could you have a look at that?

KrisThielemans:
GHA fails in Python tests. It seems to merge the PR onto UCL/master before testing but then has wrong files in `swig/test/python`. No clue why. I've merged UCL/master onto here again. Hopefully that fixes it.

danieldeidda (approved):

danieldeidda:
It seems now that the first block is always at Y=-Ref, X=0 no matter what offset is. Do we need to put the offset somwere in the map now? I am looking at the following in DetectorCoordinatMap.h for example:

`stir::CartesianCoordinate3D<float> get_coordinate_for_det_pos( const stir::DetectionPosition<>& det_pos ) const`

KrisThielemans:
> It seems now that the first block is always at Y=-Ref, X=0 no matter what offset is. Do we need to put the offset somewhere in the map now?

I don't think I changed any behavior there, certainly not intentional! The offset *is* in the map: https://github.com/pkhateri/STIR/blob/master/src/buildblock/GeometryBlocksOnCylindrical.cxx#L133

Are you sure your Scanner picked up the offset?

KrisThielemans:
I'm not sure why these lines are present
https://github.com/pkhateri/STIR/blob/02c7e4a7c4ccc10c60a47036c724402193d8dae9/src/include/stir/listmode/CListRecordSAFIR.inl#L66-L73
They say there'd be a coincidence between the same crystal (potentially on a different ring). If the list file contains that, we'd be in serious trouble, so I'll cut them

danieldeidda:
> > It seems now that the first block is always at Y=-Ref, X=0 no matter what offset is. Do we need to put the offset somewhere in the map now?
> 
> I don't think I changed any behavior there, certainly not intentional! The offset _is_ in the map: https://github.com/pkhateri/STIR/blob/master/src/buildblock/GeometryBlocksOnCylindrical.cxx#L133
> 
> Are you sure your Scanner picked up the offset?

Yes it is rotating the scanner but the starting point is at x=0 instead of being at the beginning of the bucket

danieldeidda:
> > > It seems now that the first block is always at Y=-Ref, X=0 no matter what offset is. Do we need to put the offset somewhere in the map now?
> > 
> > 
> > I don't think I changed any behavior there, certainly not intentional! The offset _is_ in the map: https://github.com/pkhateri/STIR/blob/master/src/buildblock/GeometryBlocksOnCylindrical.cxx#L133
> > Are you sure your Scanner picked up the offset?
> 
> Yes it is rotating the scanner but the starting point is at x=0 instead of being at the beginning of the bucket

image

KrisThielemans:
I don't think I have changed anything in the crystal map building. try
```
git diff -w 7ab36225a054f4eb3517c5b767d7afa5def4f687 HEAD -- GeometryBlocksOnCylindrical.cxx 
```
you'll see stuff being removed, and `get_scanner_ptr()`->`scanner.` but not much else.

However, it is likely the sorting. https://github.com/pkhateri/STIR/blob/02c7e4a7c4ccc10c60a47036c724402193d8dae9/src/buildblock/DetectorCoordinateMap.cxx#L117. It's going to sort the detectors according to the angle around the ring, and therefore what was the first could now be near the end. Sigh.

You could just comment out the `sort` line to check

KrisThielemans:
shows we are indeed missing a test 😉

danieldeidda:
> shows we are indeed missing a test 😉

funny enough it is with new test that I realised this :D

danieldeidda:
> You could just comment out the `sort` line to check

the sort line seems pretty important :D 

image

KrisThielemans:
I've pushed a change to disable the sorting properly (sorry, I thought it was going to be simpler but clearly was wrong!).

Obviously this likely breaks `Generic` but `Blocks` should work

danieldeidda:
> I've pushed a change to disable the sorting properly (sorry, I thought it was going to be simpler but clearly was wrong!).
> 
> Obviously this likely breaks `Generic` but `Blocks` should work

Yes this seems to work and the new test passes

KrisThielemans:
> this seems to work and the new test passes

great. I suggest that you commit the test, and then make the test with the reordering. It might fail at the moment of course. We will definitely have to add `std::abs` in places like this
https://github.com/pkhateri/STIR/blob/1bf66a765d889702ea4306b4c8769ca4c7559ebe/src/buildblock/ProjDataInfo.cxx#L91-L92

danieldeidda:
is there a reason why Scanner::set_detector_map() is private? I would like to use it in the test

KrisThielemans:
You could make the test class a friend. I think best not to expose to avoid temptation
________________________________

KrisThielemans:
some info on different coordinate systems used for projection data
https://github.com/pkhateri/STIR/blob/master/src/include/stir/ProjDataInfoCylindricalNoArcCorr.h#L257-L260

`find_cartesian_coordinates_of_detection`:
- cylindrical: will use 0 in first ring
https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx#L494-L495
- generic: will use whatever the map says
https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/buildblock/ProjDataInfoGenericNoArcCorr.cxx#L373-L374

`get_LOR`
- cylindrical: uses `get_m()`  so zero in centre of scanner https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/buildblock/ProjDataInfoCylindrical.cxx#L539
- generic: uses same as `find_cartesian_coordinates_of_detection`
https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/buildblock/ProjDataInfoGeneric.cxx#L110

`get_m()`
- cylindrical has zero in centre of scanner
https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/include/stir/ProjDataInfoCylindrical.h#L80
- generic uses `get_LOR`
https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/include/stir/ProjDataInfoGeneric.inl#L63

As we don't want to clean-up the mess in different conventions and functions in this PR (see #105 and #618), we will have to reproduce the confusion in this PR. As @ashgillman's work uses gantry coordinates which are centred around the centre of the scanner, i.e. preserves `get_m()` (correct @ashgillman ?), I think we should
- add a test that `get_m()` and `get_LOR()` give same results with Blocks as Cylindrical (I guess at the moment we only test the `find_` functions). I expect this test to fail currently.
- centre the axial coordinates in `GeometryBlocksOnCylindrical`, this way `Scanner::get_coordinate_for_det_pos` will be centred, @ashgillman should this function be called `get_gantry_coordinate_for_det_pos`?
- add doxygen comments to `Scanner` saying that detector map is expected to be w.r.t. centre of scanner
- add a protected member `CartesianCoordinate<3,float> z_shift` to `ProjDataInfoGeneric` that shifts from centre of gantry to centre of first ring
- add `z_shift` in `ProjDataInfoGenericNoArcCorr::find_cartesian_coordinates_of_detection`
- subtract `z_shift` in `ProjDataInfoGeneric::get_LOR`

Presumably `z_shift` will also need to be used in the reverse functions.

Sorry.

@danieldeidda I think this needs to be done via a PR on this PR...

danieldeidda:
> some info on different coordinate systems used for projection data https://github.com/pkhateri/STIR/blob/master/src/include/stir/ProjDataInfoCylindricalNoArcCorr.h#L257-L260
> 
> `find_cartesian_coordinates_of_detection`:
> 
> * cylindrical: will use 0 in first ring
>   https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx#L494-L495
> * generic: will use whatever the map says
>   https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/buildblock/ProjDataInfoGenericNoArcCorr.cxx#L373-L374
> 
> `get_LOR`
> 
> * cylindrical: uses `get_m()`  so zero in centre of scanner https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/buildblock/ProjDataInfoCylindrical.cxx#L539
> * generic: uses same as `find_cartesian_coordinates_of_detection`
>   https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/buildblock/ProjDataInfoGeneric.cxx#L110
> 
> `get_m()`
> 
> * cylindrical has zero in centre of scanner
>   https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/include/stir/ProjDataInfoCylindrical.h#L80
> * generic uses `get_LOR`
>   https://github.com/pkhateri/STIR/blob/9508ab10552a1b14de5e5690b8975f72d0b1aa80/src/include/stir/ProjDataInfoGeneric.inl#L63
> 
> As we don't want to clean-up the mess in different conventions and functions in this PR (see #105 and #618), we will have to reproduce the confusion in this PR. As @ashgillman's work uses gantry coordinates which are centred around the centre of the scanner, i.e. preserves `get_m()` (correct @ashgillman ?), I think we should
> 
> * add a test that `get_m()` and `get_LOR()` give same results with Blocks as Cylindrical (I guess at the moment we only test the `find_` functions). I expect this test to fail currently.
> * centre the axial coordinates in `GeometryBlocksOnCylindrical`, this way `Scanner::get_coordinate_for_det_pos` will be centred, @ashgillman should this function be called `get_gantry_coordinate_for_det_pos`?
> * add doxygen comments to `Scanner` saying that detector map is expected to be w.r.t. centre of scanner
> * add a protected member `CartesianCoordinate<3,float> z_shift` to `ProjDataInfoGeneric` that shifts from centre of gantry to centre of first ring
> * add `z_shift` in `ProjDataInfoGenericNoArcCorr::find_cartesian_coordinates_of_detection`
> * subtract `z_shift` in `ProjDataInfoGeneric::get_LOR`
> 
> Presumably `z_shift` will also need to be used in the reverse functions.
> 
> Sorry.
> 
> @danieldeidda I think this needs to be done via a PR on this PR...

we do test get_LOR in the following:
https://github.com/pkhateri/STIR/blob/78d611aea36de94817067b265db50ebdfa7ea541/src/test/test_proj_data_info.cxx#L707-L741

KrisThielemans:
> we do test get_LOR

ok. `get_m` not yet.  Possibly `ProjDataInfoCylindricalNoArcCorr::get_LOR()` works in the "centre of first ring" coordinate system then. it doesn't seem documented what it does.

danieldeidda:
> > we do test get_LOR
> 
> ok. `get_m` not yet. Possibly `ProjDataInfoCylindricalNoArcCorr::get_LOR()` works in the "centre of first ring" coordinate system then. it doesn't seem documented what it does.

I guess we can just add the test for get_m in the same function

KrisThielemans:
sure

danieldeidda:
> > we do test get_LOR
> 
> ok. `get_m` not yet. Possibly `ProjDataInfoCylindricalNoArcCorr::get_LOR()` works in the "centre of first ring" coordinate system then. it doesn't seem documented what it does.

this is the get_m test output:
rror : unequal values are 19.0472 and 2.2.  test get_m Cylindrical
Error : unequal values are 18.8878 and 2.2.  test get_m Cylindrical
Error : unequal values are 18.8517 and 2.2.  test get_m Cylindrical
Error : unequal values are 18.7878 and 2.2.  test get_m Cylindrical
Error : unequal values are 18.7473 and 2.2.  test get_m Cylindrical
Error : unequal values are 18.7 and 2.2.  test get_m Cylindrical
Error : unequal values are 18.6556 and 2.2.  test get_m Cylindrical

first is block and second is cylindrical. Intersting that one stays  constant and the other changes (at fixed axial pos). When cyl get_m go higher than 6.6 block get_m jumps from about 23 to about 8.8

Error : unequal values are 22.8852 and 6.6.  test get_m Cylindrical
Error : unequal values are 22.8815 and 6.6.  test get_m Cylindrical
Error : unequal values are 23.0116 and 6.6.  test get_m Cylindrical
Error : unequal values are 8.80044 and -7.7.  test get_m Cylindrical
Error : unequal values are 8.81966 and -7.7.  test get_m Cylindrical
Error : unequal values are 8.88779 and -7.7.  test get_m Cylindrical

KrisThielemans:
> Intersting that one stays constant and the other changes (at fixed axial pos).

that is a bit weird. I'd have thought for a fixed axial position, it has to remain the same for a Block scanner. (for a generic, all bets are off)

danieldeidda:
> > Intersting that one stays constant and the other changes (at fixed axial pos).
> 
> that is a bit weird. I'd have thought for a fixed axial position, it has to remain the same for a Block scanner. (for a generic, all bets are off)

yes I am a bit puzzled. Especially axially I would not expect these differences between blocks and cylinder

danieldeidda:
> * subtract `z_shift` in `ProjDataInfoGeneric::get_LOR`

I am not sure we would need to subtract in get_LOR() we are subtracting in the map and then adding again in find_cartesian_coordinates_of_detection(_p1, _p2, bin) which is called by get_LOR.

KrisThielemans:
@danieldeidda Appveyor builds are all failing on `test_blocks_on_cylindrical` with an exception, see bottom of
https://ci.appveyor.com/project/KrisThielemans/stir/builds/42581277/job/lq3sb3pvh81g31gp

It runs fine on my Visual Studio, so I suspect this it hits a memory limit, as these tests use ~6GB. I think best to use `enable_caching(false)` on the matrices. Also, the compilation above of `test_blocks_on_cylindrical.cxx` shows quite a few warnings. Maybe you can kill a few?

danieldeidda:
> @danieldeidda Appveyor builds are all failing on `test_blocks_on_cylindrical` with an exception, see bottom of https://ci.appveyor.com/project/KrisThielemans/stir/builds/42581277/job/lq3sb3pvh81g31gp
> 
> It runs fine on my Visual Studio, so I suspect this it hits a memory limit, as these tests use ~6GB. I think best to use `enable_caching(false)` on the matrices. Also, the compilation above of `test_blocks_on_cylindrical.cxx` shows quite a few warnings. Maybe you can kill a few?

I think we can remove the following:
WARNING: Disabling all symmetries except for symmtery_z since they are not implemented in block geometry yet.

as the symmetry_z is also disabled

KrisThielemans:
I guess we cannot remove it, but have to rephrase it to "all symmetries disabled".

However, I was talking about compilation warnings (as you can see in the Appveyor log). The MS VC compiler issues far more warnings than gcc...

danieldeidda:
the  all seem to be conversion from 'double' to 'float'


Anyway do you want the scatter changes to be pushed here or we merge to master first?

danieldeidda:
This should be my last commit for this PR I guess

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -75,9 +75,43 @@ class BlocksTests: public RunTests
...
+                                                                                VectorWithOffset<int>& min_ring_diff_v,
+                                                                                VectorWithOffset<int>& max_ring_diff_v,
+                                                                                shared_ptr<Scanner> scanner_ptr);
+  };
+/*! The following is a function to allow a projdata_info blocksONCylindrical to be created from the scanner.
+*/
+shared_ptr<ProjDataInfoBlocksOnCylindricalNoArcCorr>
+BlocksTests::set_blocks_projdata_info(VectorWithOffset<int>& num_axial_pos_per_segment,
+                                      VectorWithOffset<int>& min_ring_diff_v,
+                                      VectorWithOffset<int>& max_ring_diff_v,
```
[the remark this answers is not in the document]
danieldeidda: because I can reusing them when I need to create a projdata_info that is not a BlocksOnCylindrical

src/test/test_proj_data_info.cxx
```diff
@@ -947,7 +947,6 @@ run_lor_get_s_test(){
     int det_id_diff=scannerBlocks_ptr->get_num_detectors_per_ring()*(_PI/2)/(2*_PI);
```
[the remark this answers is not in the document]
danieldeidda: that was just a proportion but you are right it can just be 1/4

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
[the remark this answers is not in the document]
danieldeidda: how can we do it in one? can we do a dynamic_cast? or template?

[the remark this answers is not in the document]
KrisThielemans: ok, a bit complicated. the point was of course to simplify things, not to make the code more complicated.

you could of course add another function without those extra parameters that calls this one (as I think you re-use those parameters only once). This is a detail of course. feel free to leave it as-is.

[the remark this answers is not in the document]
danieldeidda: I kind of like the idea of the template I actually already impleamented it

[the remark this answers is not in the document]
danieldeidda: actually no, it is more difficult than I thought :D

danieldeidda:
OK I think I have done all the corrections

KrisThielemans (changes requested):
more things to do, but we'll merge this now as it's in good shape.

KrisThielemans (approved):
sorry. hit the wrong button!

KrisThielemans:
@pkhateri This is merged now. Thanks a lot! I've sent an email to the list, but don't seem to have an email address for you anymore. I hope that this messages reaches you. All the best.

PR #587, opened 2020-06-18
GatedProjData multi format and read_from_file returns unique_ptr

rijobro:
In https://github.com/UCL/STIR/pull/208, we introduced a new format called `Multi`, which is just a list of files:

```
Multi :=
  total number of data sets := 2
  data set[1] := sino1.hs
  data set[2] := sino2.hs
end :=
```
Here, we extend this format to `GatedProjData`, to mirror `DynamicProjData`. 

Also, `read_from_file` returns `unique_ptr` as opposed to raw pointer (again mirroring changes to `DynamicProjData`).

PR #600, opened 2020-06-22
GatedProjData multi format and read_from_file returns unique_ptr (#587)

rijobro:
GatedProjData multi format and read_from_file returns unique_ptr


rijobro:
Replaces https://github.com/UCL/STIR/pull/587. Merges to `master` as opposed to `release_4`.

Issue #555, opened 2020-05-27
GE RDF HDF5 remaining issues


KrisThielemans:
After #390 there's a few more things to do.
- [ ] various hard-wired numbers for the Signa. After fixing those, we should rename Signa files to something more generic
-  [ ] `SinglesRatesFromGEHDF5` 
    - expose `GEHDF5Wrapper.get_exam_info_sptr()`
    - `construct_randoms_from_GEsingles.cxx` needs to reorder data. This should be moved to this class
- [ ] `GESignaListmodeInputFileFormat` only checks if it's an HDF5 file
- [ ] `ProjData` should call `ProjDataGEHDF5`
- [ ] add tests (needs data)

evgueni-ovtchinnikov:
Another one:
```
hsize_t dims_out[dataset_list_Ndims];
```
is wrong (`dataset_list_Ndims` is not a constant), should be
```
hsize_t* dims_out = new hsize_t[dataset_list_Ndims];
```
(and `delete[] dims_out; ` before return).

AnderBiguri:
>     * [ ]  `SinglesRatesFromGEHDF5`
>       
>       * expose `GEHDF5Wrapper.get_exam_info_sptr()`
You mean  creating `SinglesRatesFromGEHDF5::get_exam_info_sptr()`? `GEHDF5Wrapper.get_exam_info_sptr()` is exposed and read in that class.

>     * [ ]  in `ProjDataGEHDF5` avoid reading all views up-front. could read them when necessary

You mean read them in `get_viewgram()/get_sinogram()` instead of in `initialise_viewgram_buffer`?

KrisThielemans:
> You mean creating SinglesRatesFromGEHDF5::get_exam_info_sptr()? GEHDF5Wrapper.get_exam_info_sptr() is exposed and read in that class.

not sure what I was thinking at the time. Possibly because we were (erronuously) using `TimeFrameDefinitions` then, which sit in `exam_info`. Let's ignore this. (please edit the first box and remove from the list)

> you mean read them in get_viewgram()/get_sinogram() instead of in initialise_viewgram_buffer?

yes. it could be done by making `tof_data `mutable`, and fill it when necessary. However, this will have little impact, and likely will conflict with updates on the TOF branch, so I'd postpone this.

KrisThielemans:
on
> construct_randoms_from_GEsingles.cxx needs to reorder data. This should be moved to this class
this is about
[...]

Issue #70, opened 2017-03-12
listmode files can start with non-zero time-tag


KrisThielemans:
Timing is currently in terms of the value of the time tag. For Siemens files, the first time tag is usually zero, but in ~1% of the cases not. GE files never start with zero. It would therefore be better to have timing relative to the value of the first time tag. Current work-around: use list_lm_events to find first time tag and create initial “skip-frame” in the frame-defs file.

KrisThielemans:
example script:
```sh
first_tag=`list_lm_events --time 1 --coincidence --num-events-to-list 1 $list_filename`
echo "0 $first_tag" > $fdef_filename
# now have the rest of the fdef info, e.g.
1 $frame_duration
```

AnderBiguri:
@KrisThielemans How is the skip frame defined then? will STIR know to ignore a frame that contains zero events in the listmode, or will it actually attempt to recon/process a PET image with all zeroes in the sinos?

KrisThielemans:
an fdef with a line `0 100` will make `LmToProjData` et al skip 100 seconds. In this case, there are actually no data in those first `$first_tag` seconds, so it won't do anything, except make sure that all the rest is ok.

AnderBiguri:
Ah, my question was more: Does it skip it because it skips whatever is in the first line? Skips whatever starts with 0? or checks if there are events, and skips if there are no events? What is the logic?

KrisThielemans:
it advances the start of the next time frame without storing data because of the first `0`
https://github.com/UCL/STIR/blob/aae4c92e9cb63699997573b4e02b78fdb86d373d/src/include/stir/TimeFrameDefinitions.h#L65-L67

the fdef doesn't check events itself at all.
`LmToProjData` and other loops only store data that are within the time-frames asked for (but it assumes that the clock starts at zero, which it doesn't for GE data, so we have to "skip" the first time until the start of the GE clock)

AnderBiguri:
In any case, a more complete example script would be:

```
first_tag=`list_lm_events --time 1 --coincidence 0 --num-events-to-list 1 $list_filename | tr -d '\n' | awk '{print $NF}'`
echo "0 $first_tag" > $fdef_filename
```

AnderBiguri:
[...]

PR #601, opened 2020-06-23
extend GE RDF9 raw data support

KrisThielemans:
replaces #564. The current PR is based on `release_4`. I've cleaned the history a little bit as well, but not a lot.


KrisThielemans:
@pw351 @AnderBiguri this is the new PR.

KrisThielemans:
I seem to be missing the last view while I believe it is read.

KrisThielemans:
@evgueni-ovtchinnikov found another problem, see  https://github.com/UCL/STIR/issues/555#issuecomment-648210880. Thisoccurs many times. It's probably only C++14.

In addition, we have many `m_memspace_ptr` leaks . Probably easiest way is to make that a `shared_ptr`.

KrisThielemans:
> I seem to be missing the last view while I believe it is read.

could be an amide display problem. I don't see anything wrong in the code. should check from python I guess

pwadhwa351:
Testing randoms:
WARNING: GEHDF5Wrapper: energy resolution is not set. This will create trouble for scatter estimation
Assertion failed: (_times.size()!=0), function read_singles_from_listmode_file, file /Users/palak/STIR_Palak/STIR-1/src/data_buildblock/SinglesRatesFromGEHDF5.cxx, line 358.

I am getting this assertion fail!
I will dig to fix this

AnderBiguri:
Hi Kris,
With this rebase I can't contribute anymore. Wanted to do some bug fixes but I can't commit them. You want me to create a new PR for each bugfix/addition that I play with, or you prefer for me to leave this and you'll finish the last bits?

KrisThielemans:
> Hi Kris,
> With this rebase I can't contribute anymore. Wanted to do some bug fixes but I can't commit them. You want me to create a new PR for each bugfix/addition that I play with, or you prefer for me to leave this and you'll finish the last bits?

and now? (after accepting the invitiation)

KrisThielemans:
> > I seem to be missing the last view while I believe it is read.
> 
> could be an amide display problem. I don't see anything wrong in the code. should check from python I guess

I've checked that amide (on Linux) has this problem with other data as well, so false alarm. sorry

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
```diff
@@ -225,9 +225,6 @@ set_up(const shared_ptr<ProjDataInfo>& proj_data_info_ptr_v)
     return Succeeded::no;
   }
 
-  span = 
-    proj_data_info_cyl_ptr->get_max_ring_difference(0) - 
-    proj_data_info_cyl_ptr->get_min_ring_difference(0) + 1;
   // TODO insert check all other segments are the same
```
KrisThielemans, line 228:
```suggestion
```

src/include/stir/recon_buildblock/BinNormalisationFromGEHDF5.h
```diff
@@ -129,6 +129,7 @@ class BinNormalisationFromGEHDF5 :
   float get_dead_time_efficiency ( const DetectionPosition<>& det_pos,
 				  const double start_time, const double end_time) const;
 
+  float get_geometric_factors (int geo_plane_num, Bin uncompressed_bin) const;
```
KrisThielemans, line 132:
```suggestion
  float get_geometric_factors (int geo_plane_num, const Bin& uncompressed_bin) const;
```

`Bin`s are fairly large, so a reference is best. Note that this signature won't be ok, but we can fix that later once we know what to do!

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
```diff
@@ -568,6 +542,13 @@ BinNormalisationFromGEHDF5::get_dead_time_efficiency (const DetectionPosition<>&
   return 1;  
 }
 
+float 
+BinNormalisationFromGEHDF5::get_geometric_factors (int geo_plane_num, Bin uncompressed_bin) const
```
KrisThielemans, line 546:
```suggestion
BinNormalisationFromGEHDF5::get_geometric_factors (int geo_plane_num, const Bin& uncompressed_bin) const
```

src/include/stir/recon_buildblock/BinNormalisationFromGEHDF5.h
AnderBiguri: Yup, makes sense to pass as reference, forgot to add it. Fixing. 

 In this case I think its a single Bin, is that actually big?

KrisThielemans: not BIG, but 4 ints (5 for TOF) + 1 float. so 20 bytes. more to come for energy.

who knows what's fasted, but best to use what we use elsewhere of course.

AnderBiguri: @KrisThielemans ah that makes sense, I was expecting around that size. Anyway, still you are right, always pass by reference when one can.

AnderBiguri:
@KrisThielemans have renamed all `get_from_dataset` style functions from the wrapper to `read_XXX` where `XXX` is their unique reader concept (i.e. `read_singles`). Make sense to me to not have them as `get` to not get confused with getters, as they read from file. 

I know I am refactoring a lot, let me know if you want me to revert this, if it somehow does not meet STIR standards/style.

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
```diff
@@ -230,118 +229,66 @@ set_up(const shared_ptr<ProjDataInfo>& proj_data_info_ptr_v)
...
+
+  // Do the reading using a buffer.
+  unsigned int total_size = scanner_ptr->get_num_rings()*scanner_ptr->get_num_detectors_per_ring();
+  stir::Array<1, float> buffer(0, total_size-1);
+  m_input_hdf5_sptr->read_efficiency_factors(buffer);
+  // Aparently GE stores the normalization factor and not the "efficiency factor", so we just need to invert it. 
+  // Lambda function, this just applies 1/buffer. 
+  std::transform(buffer.begin(), buffer.end(),buffer.begin(), [](const float f) { return 1/f;} );
+  // Copy the buffer data to the properly shaped efficiency_factors variable. 
+  std::copy(buffer.begin(), buffer.end(), efficiency_factors.begin_all());
```
KrisThielemans, line 280: could combine these 2 into one of course
```suggestion
  // Apparently GE stores the normalization factor and not the "efficiency factor", so we just need to invert it. 
  // and store it in the properly shaped efficiency_factors variable. 
  // We use a Lambda function, this just applies 1/buffer. 
  std::transform(buffer.begin(), buffer.end(), efficiency_factors.begin_all()), [](const float f) { return 1/f;} );
```

KrisThielemans:
renaming to `read_*`  is fine of course

examples/GE-Signa-PETMR/howto.md
```diff
@@ -0,0 +1,44 @@
...
+
+Author: Palak Wadwha
+Edits: Kris Thielemans
+
+These instructions use a  template that was created using the `create_projdata_template` utility using the following inputs:
+```
+Enter the name of the scanner : GE PET/MR Signa
+Mashing factor for views : 1
+Is the data arc-corrected?: N
+Number of the tangential positions: 357
```
AnderBiguri, line 11: The tool only accepts numbers in the range [0 128]. 
```

Number of tangential positions[1,128 D:128]: 357

Out of bounds. Try again.
```
Missing instructions?

examples/GE-Signa-PETMR/howto.md
```diff
@@ -0,0 +1,44 @@
...
+Span value: 2
+Max. ring difference acquired: 44
```
AnderBiguri, line 13: The tool only accepts [0 15]

examples/GE-Signa-PETMR/correct_projdata_for_norm.par
```diff
@@ -0,0 +1,19 @@
...
+use data (1) or set to one (0) := 0
+; precorrect data, or undo precorrection
+; default is '1'
+; apply (1) or undo (0) correction :=
+; normalisation (or binwise multiplication, so can contain attenuation factors as well)
+Bin Normalisation type := From GE HDF5
+Bin Normalisation From GE HDF5 :=
+normalisation filename:= norm3d
+End Bin Normalisation From GE HDF5:=
+g
```
AnderBiguri, line 18: g!

AnderBiguri:
Progress report:

Testing #608: 

If I use the template in this PR with the `norm3d` shared by Palak weeks ago (not the ones in zenodo, but the HDF5 ones), or I use the uncompressed `rdf_f1b1.rdf` (converted to `.hs`), I get ` scanner object from proj data is different from the one from the normalisation file`

AnderBiguri:
<`help needed`>!

I still don't know how to fix the issue in the previous comment, the scanner read from GEWrapper and ProjData is different. This could be cause by different ways of reading the scanner data in both sides.... I can investigate further of course. 

However, if I remove the check, the code errors, I suspect not caused by the mismatching scanners (because both have the same numbers for the relevant variables described further.
 
The error reads:
```

ERROR: The tangential_pos range (-990 to 990) for this projection data is too large.
Maximum supported range is from -223 to 224
```
This is caused in `ProjDataInfoCylindricalNoArcCorr.cxx` L195, but actually in the normalization code at [`BinNormalisationFromGEHDF5.cxx` here](https://github.com/UCL/STIR/blob/aeb94e71cff27365a4b0151dbc7540898248e872/src/recon_buildblock/BinNormalisationFromGEHDF5.cxx#L493) and thus [defined here](https://github.com/UCL/STIR/blob/aeb94e71cff27365a4b0151dbc7540898248e872/src/recon_buildblock/BinNormalisationFromGEHDF5.cxx#L105).  The error inside the guts seems like an basic consistency checking code that checks if the tangential positions of the detector, described by the scanner:

```
 const int num_detectors = get_scanner_ptr()->get_num_detectors_per_ring();
 const int min_tang_pos_num = -(num_detectors/2)+1;
 const int max_tang_pos_num = -(num_detectors/2)+num_detectors;
```
match the variables descrived by the projection info in 

```
this->get_min_tangential_pos_num()
this->get_max_tangential_pos_num()
```

This all makes sense, except that in this code, our [ProjInfo is an uncompressed scan](https://github.com/UCL/STIR/blob/aeb94e71cff27365a4b0151dbc7540898248e872/src/recon_buildblock/BinNormalisationFromGEHDF5.cxx#L304) ! And therefore it will not match the parameters described in the scanner pointer. 

So.... How to fix this? We do need the uncompressed ProjInfo for the the normalization.

Ideas:

 1. `proj_data_cyl_uncomp->get_det_num_pair_for_view_tangential_pos_num()` is the issue in the current code, we can just replace this by the code that errors within `ProjDataInfoCylindricalNoArcCorr.cxx` with our own error checks. Cons: replication of code. 


   2. If we have a `proj_data_uncompressed` we can have a `scanner_ptr_uncompressed` with the correct info, thus avoiding the error. Cons: seems a bit ugly, and I dont know how to do it. 

   3. ???

examples/GE-Signa-PETMR/howto.md
KrisThielemans: turned out the name was wrong. It has to be `GE PET/MR Signa`. now corrected.

examples/GE-Signa-PETMR/howto.md
KrisThielemans: see above

examples/GE-Signa-PETMR/correct_projdata_for_norm.par
KrisThielemans: ok. will fix

src/utilities/construct_randoms_from_GEsingles.cxx
```diff
@@ -236,9 +236,10 @@ int main(int argc, char **argv)
 
                     float coincidence_time_window = 0.00000000457f;
                     /*(*segment_ptr)[bin.axial_pos_num()]*/
-                    // TODO 447 needs to be num_crystals_per_ring()-1
+                    // GE stores tangential pos in the opposite direction to STIR
+                    unsigned int num_crystals_per_ring=m_input_sptr->get_scanner_sptr()->get_num_detectors_per_ring();
                     sinogram[bin.view_num()][bin.tangential_pos_num()] +=
-                    num_slices*coincidence_time_window*efficiencies[ra][447-a]*efficiencies[rb][447-b%num_detectors_per_ring];
+                    num_slices*coincidence_time_window*efficiencies[ra][num_crystals_per_ring-a]*efficiencies[rb][num_crystals_per_ring-b%num_detectors_per_ring];
```
KrisThielemans, line 242:
```suggestion
                    num_slices*coincidence_time_window*efficiencies[ra][num_crystals_per_ring-1-a]*efficiencies[rb][num_crystals_per_ring-1 - (b%num_detectors_per_ring)];
```

AnderBiguri:
Welp, the good scanner name has fixed my previous issue. I have no idea how exactly hehe.

AnderBiguri:
So, this is what I found out so far. The reason why the scanners are not equal, and therefore the error I described above is that the scanner parameters are different in the norm and the rdf (sinogram files). 

Particularly, in [`GEHDF5Wrapper::initialise_scanner_from_HDF5()`](https://github.com/UCL/STIR/blob/6955e55e61e31740d63821130b01be8d377083d5/src/IO/GEHDF5Wrapper.cxx#L258), all but 1 parameters read have the same value. The differing one is `/HeaderData/Sorter/dimension2Size`, i.e. `max_number_of_non_arc_corrected_bins`. The rdf file contains 357 (same as when load from template) but the norm file contains 1981

KrisThielemans:
ok, So I guess it's this line
https://github.com/UCL/STIR/blob/6955e55e61e31740d63821130b01be8d377083d5/src/IO/GEHDF5Wrapper.cxx#L294
presumably we need to use `dimension1Size` when reading the norm (which would make more sense, as it is indeed the fast running index)

AnderBiguri:
Yes, was checking exactly that. In fact, to be more specific, we need to read `dimension1Size` for geo, norm, `dimension2Size` for sino, nothing for listmode.

AnderBiguri:
While I managed to get the thing not to error (Hurray!) now it outputs.... bad things. 
To start debugging, I have a question that perhaps @pwadhwa351 can help. When doing `stir_math -s RDF.hs rdf_f1b1.rdf` with the dataset shared bu her, I get a dataset than when loaded in SIRF-python via: 

```
norm=sirf.STIR.AcquisitionData("RDF.hs")
normarr=norm.as_array()
#prints
print(np.unique(normarr))
print(norm.shape)
```
returns 

```
[ 0.  1.  2.  3.  4.  5.  6.  7.  8.  9. 10. 11. 12. 13. 14. 15. 16. 17. 18. 19. 20. 21. 22. 23. 24. 25.]
(1, 1981, 224, 357)
```

very confusing numbers to me. Is the data shared (rdf_f1b1.rdf) empty or should it actually contain something? 

I suspect the second, and therefore bugs in the Wrapper readers. Will search more tomorrow.

KrisThielemans:
some confusion that when reading `RDF.hs` you call the variable `norm` (it isn't a norm, but projection data).

`shape` is alright. (1 TOF bin (as we ignore TOF at the moment), 1981 sinos of size 224x357). `unique` seems alright as well, as It'll be counts, some would only give integer values. apparently max counts was 25... I think you should display it... There's too many sinos to look at, but you could just add all 1981 of them and display the resulting 2D image. should look like a sinogram...

Clearly, for the actual norm, the `unique` should give you a whole load of floating point numbers. You can still display it and see what it does (and compare to @pwadhwa351 's output of course)

AnderBiguri:
> some confusion that when reading `RDF.hs` you call the variable `norm` (it isn't a norm, but projection data).

copy paste edit mistake, sorry.


> `shape` is alright. (1 TOF bin (as we ignore TOF at the moment), 1981 sinos of size 224x357). `unique` seems alright as well, as It'll be counts, some would only give integer values. apparently max counts was 25... I think you should display it... There's too many sinos to look at, but you could just add all 1981 of them and display the resulting 2D image. should look like a sinogram...
> 
> Clearly, for the actual norm, the `unique` should give you a whole load of floating point numbers. You can still display it and see what it does (and compare to @pwadhwa351 's output of course)

How many non-TOF sinos should there be? All this makes sense I guess. The TOF-sinos where very, very sparse, so I was getting worried.

KrisThielemans:
> How many non-TOF sinos should there be?

1981 for the Signa PET/MR 😄 

The VQC data is just 4 point sources.

AnderBiguri:
have no idea why the latests commits are not showing in the PR (but are in fact, in the branch), bu have now validated the norms files vs Palaks code, however I needed to add a very very ugly if condition in the end of the normalization calculation code because when no loop is run, then the function returns zero, which breaks the output causing 1e23 values. If I check strictly equal to zero, and substitute by 1, then it works. Its just super ugly.

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
```diff
@@ -323,14 +322,14 @@ read_norm_data(const string& filename)
     m_input_hdf5_sptr->initialise_efficiency_factors();
 
     // Do the reading using a buffer.
-    unsigned int total_size = scanner_ptr->get_num_rings()*scanner_ptr->get_num_detectors_per_ring();
+    unsigned long total_size = (scanner_ptr->get_num_rings()-1)*(scanner_ptr->get_num_detectors_per_ring()-1);
```
KrisThielemans, line 325: using `long` won't work actually. You'd have to cast first, but in any case the `Array`  constructor takes `signed int`

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
```diff
@@ -323,14 +322,14 @@ read_norm_data(const string& filename)
...
     stir::Array<1, float> buffer(0, total_size-1);
+    buffer.fill(0.0);
     m_input_hdf5_sptr->read_efficiency_factors(buffer);
     // Aparently GE stores the normalization factor and not the "efficiency factor", so we just need to invert it. 
     // Lambda function, this just applies 1/buffer and stores it in efficiency_factors 
-    std::transform(buffer.begin(), buffer.end(),efficiency_factors.begin_all(), [](const float f) { return 1/f;} );
+    std::transform(buffer.begin(), buffer.end(),efficiency_factors.begin_all(), [](const float f) { return 1/(f+0.000001);} );
```
KrisThielemans, line 331: Did you really need this epsilon? I can see a zero efficiency occurring (if a detector is out), but as GE actually stores 1/efficiencies in the buffer, I can *not* see why the buffer would contain any zeroes (as this would correspond to an infinitely efficient detector!)

KrisThielemans:
I'm also confused by the CI situation. Appveyor says it's unmergable, Travis as well. github doesn't.

Probably github is down again. https://www.githubstatus.com/ says they're having problems with webhooks

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
AnderBiguri: Oops some debugging changes that I didn't revert, my bad.

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
AnderBiguri: See avobe. Debugging changes unreversed. My mistake

AnderBiguri:
I can maybe tomorrow show you where it happens. Its just in a couple of diagonal 1 pixel width stripes. If I add that ugly check I get bit equal results to Palak. There may be another way to check that zero, I'll think about it. Its just in the cases where the loops are not even run, so there must be a way to initialize it to one or something.

AnderBiguri:
Found bug on my validation code and in the source code. Fixed, ugly if removed, now difference between Palak's code and this one is `unique(diff)=[-0.00048828 -0.00036621 -0.00024414 -0.00012207  0.  0.00012207 0.00024414  0.00036621  0.00048828]` which is interesting due to some pattern appearing, but may be standard numerical shenanigans as it looks arbitrary in pixel locations. Will have a further look.

mevspalak

KrisThielemans:
.0002? What's the mean value before diff? If it's around 1 or larger, then I'd be very happy!

AnderBiguri:
@KrisThielemans ah you may have missed my edit (sing you replied by mail). The difference array has that "stepwise" pattern in the negative and positive values because each of those steps is 1 least significant bit of change (for floats at the scale of the norm). So, I am getting a maximum of 4 least significant bits of difference between my code and Palaks results, which should be enough to have you happy! 

This is only efficiencies, though. I am having bigger differences with geo factors, but these are harder to debug, because the runtime jumps from 3 seconds to 50 minutes when processing geos. The loading takes longer, but the actual big chunk of time is eaten by `BinNormalisationFromGEHDF5::get_geometric_factors()`, not quite sure why that much...

AnderBiguri:
To clarify further, the diff was computed from a sum of all views, so the actual value was around 2000.

KrisThielemans:
> because the runtime jumps from 3 seconds to 50 minutes when processing geos

interesting. it implies that `get_bin_for_det_pos_pair` is slow. Anyway, let's fix that later!

AnderBiguri:
Sure!

But if you can think of some way to speed this up now let me know as it makes my debugging latency 1h...

AnderBiguri:
Current status:

I _almost_ load the geo data 100% as Palak, but I have a weird bug where I do not load particular views. 


If I save as Interfile the geo Loaded in `BinNormalisationFromGEHDF5::read_norm_data()` (so the main loop of corrections does not execute, I just load, save and manually error to terminate). Then I load the result data as:

```
geo=sirf.STIR.AcquisitionData("geo.hs")
slice_geo=geo[:,:,:,0]   #Only rows in 0 are missing, geo[:,:,:,2,:] is equal to Palak
plt.imshow(slice_geo.squeeze())
```
I get the following plot:

Figure_1

Notice the dark vertical bars. They are 0 in value, which I suspect its because they have not been read. The indices where this happens are `88
173
258
339
420
497
574
647
720
789
858
923
988
1049
1110
1167
1224
1277
1330
1379
1428
1473
1518
1559
1600
1637
1674
1707
1740
1769
1798
1823
1848
1869
1890
1907
1924
1937
1950
1959
1968
1973
1978
1979
1980`

Which I printed hoping to find a pattern and its the first quarter of a sin wave 😕.
Trying to debug, very confused.

AnderBiguri:
Ah, red herring. Found it, was just an index too short reading, but would cause very interesting patters in the data!

AnderBiguri:
> > because the runtime jumps from 3 seconds to 50 minutes when processing geos
> 
> interesting. it implies that `get_bin_for_det_pos_pair` is slow. Anyway, let's fix that later!

mmm no, `ProjDataInMemory::get_bin_value` is the slow one, I tested and timed calls to view normalization with each of those functions removed, and `get_bin_value()` is what makes this all terribly slow. 

I can see its definition is:
```

float 
ProjDataInMemory::get_bin_value(Bin& bin)
{
   Viewgram<float> viewgram = get_viewgram(bin.view_num(),bin.segment_num()); 
    
   return viewgram[bin.axial_pos_num()][bin.tangential_pos_num()]; 

}
```

I suspect this is so slow (comparing to a normal read from array) is because `get_viegram` may return a copy of the viegram, instead of a reference to it? In the Bin normalization code this makes the entire thing incredibly slow. 

I guess there are two things we can do:

 1. Change `ProjDataInMemory::get_bin_value` to be faster. Pros: Faster code for everyone! Cons: may break code somewhere else, but I can't think how. 
 2. Refactor `BinNormalisationFromGEHDF5` to not use `ProjDataInMemory` for geometric factors, but perhaps just `sitr::Array<3,float>`. Pros: easier to do (?) Cons: `get_bin_value` is still slow, we calked around the issue. 
3. Leave as it is. Pros: nothing really, aside from less work to do. Cons: 1h to normalize your sinograms on a 32 core machine it an absurdly long time!

KrisThielemans:
> mmm no, `ProjDataInMemory::get_bin_value` is the slow one,

yup. see  https://github.com/UCL/STIR/issues/638. would you be able to do this as a separate PR?

AnderBiguri:
Sure thing

src/include/stir/IO/GEHDF5Wrapper.h
```diff
@@ -52,61 +55,85 @@ class GEHDF5Wrapper
...
+    Succeeded initialise_geo_factors_data(const unsigned int slice_num);
 
-    Succeeded get_from_dataspace(std::streampos &current_offset,
-                                 char* output);
+    Succeeded initialise_efficiency_factors();
 
-    Succeeded get_dataspace(const unsigned int current_id,
-                            Array<1, unsigned int> &output);
+    //! reads coincidence time window from file
+    float get_coincidence_time_window();
```
KrisThielemans, line 102: add `const` I guess

AnderBiguri:
@KrisThielemans I am a bit confused about bin normalization application. I was trying to figure out why my results are different to Palaks. I found out that when only applying geo factors the code is not doing what I though was doing.

When I read the geo factors, I can inmediately plug them into an interfile. If I compare this to Palaks geo file, they are equal. Now, if instead, I apply geo_efficiencies in the data that is all ones (i.e. using correct_projdata) I get the inverse of this geo file. Interestingly, the entire code talsk about "efficiecies" and everything is multiplied by the value. 

I am clearly missing a location where there is an inverse applied... where can that be? Are the names in the `BinNormalisationFromGEHDF5::
get_bin_efficiency` misleading and they should be `get_bin_normalisation`? or is there something else I can find? Or do you have no idea and I need to figure this out?

KrisThielemans:
ok. looks like then that they store geo-norm factors, and not geo-efficiency factors, which is fine. `BinNormalisation::get_bin_efficiency()` does need to return efficiencies, so we do need to divide. It also only makes sense to add detection-efficiency factors (2 detector-pairs combine in total efficiency of course).

we don't want `sum(prod(geo_eff[detector_pair]))` though but `sum(eff[det1] * eff[det2] geo_eff[detector_pair]))` which I guess is what you meant. Generally obviously in the `sum (eps_i eps_j/g_ij) ` you cannot take `g_ij` out, but in this case, all `g_ij` contributing to a bin should be equal in the "native" span=2 case (by construction), so it should give the same result I believe

AnderBiguri:
> ok. looks like then that they store geo-norm factors, and not geo-efficiency factors, which is fine. 

Hum, just to be clear, no norm/efficiencies where applied in the previous test.  Only geo factors.

> 
> we don't want `sum(prod(geo_eff[detector_pair]))` though but `sum(eff[det1] * eff[det2] geo_eff[detector_pair]))` which I guess is what you meant. 

hum not really. Lets still assume only geo factors are applied. The code as it is now, goes:

```
for uncompressed_bin.view_num
    for uncompressed_bin.segment_num
         eff *=geo[detector_pair];
    total_eff+=eff;
return total_eff;
```

so, some sort of `sum(prod(geo_eff[detector_pair]))`, with indices ignored (I wasn't very clear before). So I can not just do `1/geo_eff` because `sum(prod(geo_eff[detector_pair]))!=sum(prod(1/geo[detector_pair]))` , I hope this with the pseudocode explains it better.

I can `return 1/total_eff` to get the same result though. 

All this only with `geo`, no crystal efficiencies in the equation. Introducing them causes a bit more problem of course.

KrisThielemans:
without efficiencies, `eff` in the initial state should be `. I don't think your pseudo-code is correct. The [lines](https://github.com/KrisThielemans/STIR-1/blob/05ace864fcf24690dc5e8fec535147d18739e0d2/src/recon_buildblock/BinNormalisationFromGEHDF5.cxx#L503-L516) with dead-time say
```
float lor_efficiency= 0
loop all uncompressed bins in the actual bin
      float lor_efficiency_this_pair = 1.F;
      lor_efficiency_this_pair *= get_efficiency_factors(detection_position_pair);
      lor_efficiency_this_pair *=get_geometric_factors(detection_position_pair);
      lor_efficiency += lor_efficiency_this_pair;
```
(actual loop splits this up between views and the rest, but he effect will be the same)
So, it is a sum of total efficiencies.

Let `get_geometric_factors` return 1/current_value. It should work...

AnderBiguri:
I think you are missing the sum of my pseudocode in your pseudocode, there are 2 loops.

Subplot 1 is Palak's code. 

If I run the code currently on the repo, only applying geo factors, with [the geo factors inverted as](https://github.com/KrisThielemans/STIR-1/blob/05ace864fcf24690dc5e8fec535147d18739e0d2/src/recon_buildblock/BinNormalisationFromGEHDF5.cxx#L506)

         lor_efficiency_this_pair /= get_efficiency_factors(detection_position_pair);

I get subplot 2. 

If I run the code currently on the repo, only applying geo factors, with [the return value inverted as](https://github.com/KrisThielemans/STIR-1/blob/05ace864fcf24690dc5e8fec535147d18739e0d2/src/recon_buildblock/BinNormalisationFromGEHDF5.cxx#L523)

         return 1/total_efficiency;

I get subplot 3:


Figure_1-1


3 and 1 are the same.

examples/GE-Signa-PETMR/template.hs
```diff
@@ -0,0 +1,49 @@
+!INTERFILE  :=
+!imaging modality := PT
+name of data file := template.s
+originating system := GE PET/MR Signa
```
AnderBiguri, line 4: originating system := GE Signa PET/MR

examples/GE-Signa-PETMR/template.hs
```diff
@@ -0,0 +1,49 @@
...
+!matrix size [2] := { 1,5,9,13,17,21,25,29,33,37,41,45,49,53,57,61,65,69,73,77,81,85,89,85,81,77,73,69,65,61,57,53,49,45,41,37,33,29,25,21,17,13,9,5,1}
+matrix axis label [1] := tangential coordinate
+!matrix size [1] := 357
+minimum ring difference per segment := { -44,-43,-41,-39,-37,-35,-33,-31,-29,-27,-25,-23,-21,-19,-17,-15,-13,-11,-9,-7,-5,-3,-1,2,4,6,8,10,12,14,16,18,20,22,24,26,28,30,32,34,36,38,40,42,44}
+maximum ring difference per segment := { -44,-42,-40,-38,-36,-34,-32,-30,-28,-26,-24,-22,-20,-18,-16,-14,-12,-10,-8,-6,-4,-2,1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31,33,35,37,39,41,43,44}
+Scanner parameters:= 
+Scanner type := GE PET/MR Signa
+Number of rings                          := 45
+Number of detectors per ring             := 448
+Inner ring diameter (cm)                 := 62.38
```
AnderBiguri, line 30: `list_projdata_info template.hs` says:

WARNING: Interfile warning: 'inner ring diameter (cm)' (62.380000) is expected to be 62.359998.

examples/GE-Signa-PETMR/template.hs
```diff
@@ -0,0 +1,49 @@
...
+Average depth of interaction (cm)        := 0.94
```
AnderBiguri, line 31: WARNING: Interfile warning: 'average depth of interaction (cm)' (0.940000) is expected to be 0.850000.

examples/GE-Signa-PETMR/template.hs
AnderBiguri: Adding these so I remember to change/test them

KrisThielemans:
> I think you are missing the sum of my pseudocode in your pseudocode, there are 2 loops.

 more complete pseudo-code
```
 float total_efficiency = 0 ;
 float view_efficiency = 0.;
 loop views
 {
    float lor_efficiency= 0.;   
    loop segments
    {
      float lor_efficiency_this_pair = 1.F;
      lor_efficiency_this_pair *= get_efficiency_factors(detection_position_pair);
      lor_efficiency_this_pair *=get_geometric_factors(detection_position_pair);
      lor_efficiency += lor_efficiency_this_pair;
    }
    view_efficiency += lor_efficiency;
    total_efficiency += view_efficiency;
}
```
we have 2 redundant variables, as in the end this will (or at least should!) boil down to my simplistic pseudo-code above.

I'm not disputing that version 2 and 3 are giving different results of course. I suspect that they are inverse of each-other (but only for a span=2 template).

There's a bit of a complication that we don't really know for sure what @pwadhwa351 intended with the file she supplied, i.e. was it geo-norm-correction-factors or geo-efficiency factors (which are the inverse...)?

In the end, the question is which version gives the best results on measured data of course 😉 

I did check though with GE code that `norm3d` contains geo-norm-correction-factors, so we need to replace https://github.com/KrisThielemans/STIR-1/blob/05ace864fcf24690dc5e8fec535147d18739e0d2/src/recon_buildblock/BinNormalisationFromGEHDF5.cxx#L549 with
```c++
 return 1/this->geo_norm_factors_sptr->get_bin_value(bin);
```

(Please do not return `1/total_efficiency`. It won't work with detector-efficiencies and non-span=2 data.)

> If I run the code currently on the repo,

how? `correct_projdata` with https://github.com/KrisThielemans/STIR-1/blob/GERDF9/examples/GE-Signa-PETMR/correct_projdata_for_norm.par? It's clearly important to know what it says for ` apply (1) or undo (0) correction :=` (default is 1) as this will mean it's multiplying with norm-factors or dividing by them, (i.e. dividing by detection-efficiencies or multiplying with them...)

pwadhwa351:
> There's a bit of a complication that we don't really know for sure what @pwadhwa351 intended with the file she supplied, i.e. was it geo-norm-correction-factors or geo-efficiency factors (which are the inverse...)?

As far as I remember, it is norm correction factors multiplied with geo

AnderBiguri:
> I am clearly missing a location where there is an inverse applied... where can that be?

This is what I am most confused about, as I agree with @pwadhwa351 that they are multiplicative. Because [when I save the data when loaded](https://github.com/KrisThielemans/STIR-1/blob/05ace864fcf24690dc5e8fec535147d18739e0d2/src/recon_buildblock/BinNormalisationFromGEHDF5.cxx#L394) I get exactly the geo file @pwadhwa351, but when that is applied to the code in a multiplicative manner, I get the inverse of that...


 I sidetracked a bit, and perhaps added some bugs to my code, but I have no managed to get the same results as @pwadhwa351 applying ` return 1/this->geo_norm_factors_sptr->get_bin_value(bin);`. Really no idea what has changed. 

But now my question is as the initial one. Why is there a need to invert the geo factors when reading them in`get_geometric_factors` , to get as output the value of `geo_norm_factors_sptr`? seems super counter intuitive to me and I can't figure it out (and its likely the cause of all the mess I caused in the previous comments here). 
 
> It's clearly important to know what it says for apply (1) or undo (0) correction := (default is 1) as this will mean it's multiplying with norm-factors or dividing by them, (i.e. dividing by detection-efficiencies or multiplying with them...)

I was using default, which should be 1, i.e. apply. I have changed that to be explicit, as it was reading default correctly.

KrisThielemans:
ah. I hadn't seen that you had merged your branch in here, so I squash-merged #639. So now we have your commits on this branch, and the squash-merged #639. My fault. I normally always merge, but in this case I thought having a marginally cleaner history would be harmless enough. Oh well.

By the way, **if** you haven't pushed anything yet, it is probably best to do `git pull --rebase` before pushing (as opposed to `git pull`). That avoids a somewhat superfluous merge commit.

KrisThielemans:
> I have now managed to get the same results as @pwadhwa351 applying ` return 1/this->geo_norm_factors_sptr->get_bin_value(bin)`

great



> Why is there a need to invert the geo factors when reading them in`get_geometric_factors` , to get as output the value of `geo_norm_factors_sptr`? seems super counter intuitive to me

Everything in normalisation is confusing due to naming (and the `apply`/`undo` terminology doesn't help a  lot)

GE stores "correction factors". we want "efficiency factors". Hence we needed to do
https://github.com/KrisThielemans/STIR-1/blob/170d730b9d2924e2e51fae2ff60025ca501145bc/src/recon_buildblock/BinNormalisationFromGEHDF5.cxx#L331
We could do the same when we read them, or do it when we return. as you prefer.

We can easily rename variables accordingly. Probably want to call `get_geometric_factors()` `get_geometric_efficiency_factors()`. I believe current `geo_norm_factors_sptr` is the correct name if we store the original values. If we do store `1/GE_factors`, then we'd need to rename the variable.

examples/GE-Signa-PETMR/scatter_and_recon.sh
```diff
@@ -0,0 +1,88 @@
...
+
+### estimate scatter
+
+echo "Estimating scatter (be patient). Log saved in output/scatter.log"
+
+# filename-prefix for additive sino (i.e. "precorrected" sum of scatter and randoms)
+total_additive_prefix=output/total_additive
+num_scat_iters=3
+scatter_pardir=${pardir}/../samples/scatter_estimation_par_files
+# you might have to change this for a different scanner than the mMR
```
KrisThielemans, line 64:
```suggestion
# you might have to change this for a different scanner than the Signa PET/MR
# (it needs to be a divisor of the number of views)
```

examples/GE-Signa-PETMR/scatter_and_recon.sh
```diff
@@ -0,0 +1,88 @@
...
+scatter_recon_num_subiterations=21
+scatter_recon_num_subsets=21
```
KrisThielemans, line 66:
```suggestion
scatter_recon_num_subiterations=28
scatter_recon_num_subsets=28
```

examples/GE-Signa-PETMR/scatter_and_recon.sh
```diff
@@ -0,0 +1,88 @@
...
+NORM=${norm3d} mask_image=${mask_image} mask_projdata_filename=output/mask2d.hs scatter_prefix=output/scatter \
+    estimate_scatter $scatter_pardir/scatter_estimation.par 2>&1 | tee output/scatter.log 
+# output for recon
+additive_sino=${total_additive_prefix}_${num_scat_iters}.hs
+
+### do an image reconstruction
+
+stir_math -s --mult output/mult_factors_3d.hs  ${acf3d} ${norm3d}
+
+# FBP recon (doesn't work for mMR due to gaps)
```
KrisThielemans, line 79:
```suggestion
# FBP recon (currently commented out, not checked)
```

examples/GE-Signa-PETMR/unlist_and_randoms.sh
```diff
@@ -0,0 +1,20 @@
+#! /bin/sh -e
+# unlists listmode data into span2 and creates MLrandoms estimate from the listmode data
+# Author: Kris Thielemans
+
+# right now randoms stuff only handles single time frame
+
+# directory with some standard .par files
+: ${pardir:=~/SIRF/workspace}
```
KrisThielemans, line 8:
```suggestion
: ${pardir:=~/devel/STIR/examples/GE-Signa-PETMR}
```

KrisThielemans:
Attenuation is a bit of a problem, due to the rotation etc. I guess @pwadhwa351 didn't supply one for the VQC phantom (it probably has very little attenuation anyway). You could just create a dummy image with attenuation=0, and run the test with that.

AnderBiguri:
@KrisThielemans in some examples (haven't checked all) there is this requirement for a `.fdef`. My initial feel from reading the .h files (I have not found docs about it anywhere else and the useguide redirects to class definition for docs) is that its a very simple file and that for the examples in the folders, just a 2 line file would cut it. Is there any reason why an example of these files doe snot exist anywhere? Should we add one?

KrisThielemans:
It's a convenient but fairly lousy file format. I guess I've been too ashamed to expose it too much...

Feel free to add one. if you add one in `examples/samples` I'd take one from the .h file. It's also easy to create it from the script of course. I think for using the first 100 s of a file, we just need
```
1 100
```
However, there's a difficulty that GE listmode files don't start with a timestamp of 0. See #70

AnderBiguri:
This should be 👍  now, I think. Ideally we'd test it with more RDF9 data.

AnderBiguri:
@KrisThielemans while we wait for confirmation from the forces with the knowledge, I scripted something that allows to load the data when its in a different format than the Signa, as discussed by mail. I have tested it with the data I have available that uses this different geo format and it works, and seems correct:

Figure_1

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
```diff
@@ -367,24 +367,49 @@ read_norm_data(const string& filename)
...
-          std::transform(buffer.begin(), buffer.end(),viewgram.begin_all(), [](const float f) { return 1/(f*2.2110049e-4);} );
-
+          switch (m_input_hdf5_sptr->get_geo_type())
+          {
+          case 9:
+          {
+            m_input_hdf5_sptr->initialise_geo_factors_data(modulo(i_view,16)+1);
+
+            // Define which chunk of the data we are reading from. 
+            std::array<unsigned long long int, 2> offset = {segment_axial_position_offset[detail::find_segment_index_in_sequence(segment_sequence,i_seg)], 0};
```
KrisThielemans, line 377: change `unsigned long long int` to the usual `h5size` (or whatever the name is)

AnderBiguri: You mean `hsize_t`?

KrisThielemans (commented):
looks good!

I see we still have the hard-coded 16 in here. It should be equal to the `2*num_crystals_per_block` (actually `num_crystals_per_bucket`) I believe

src/recon_buildblock/BinNormalisationFromGEHDF5.cxx
KrisThielemans: sure

AnderBiguri:
> looks good!

Ah unfortunately it may not. I have found some strange behavior in the efficiency factors section. Investigating. Forever bugs!

> I see we still have the hard-coded 16 in here. It should be equal to the `2*num_crystals_per_block` (actually `num_crystals_per_bucket`) I believe

Where is this info in the code? Both those variables don't exist in the scan info, but `num_axial_crystals_per_block` does. However its value is 9, which is neither 16, nor a integer divisor of it....

KrisThielemans:
have to use `num_transaxial_crystals_per_block` (views are in-plane). I believe there are 2 blocks in a "bucket" for GE scanners, but `Scanner` has https://github.com/UCL/STIR/blob/1df251c47f3626f2785ec4077b337dbee816b732/src/include/stir/Scanner.h#L246

Clearly, it has to be related to block size.

AnderBiguri:
Ah yes, thats the one. 
Transaxial makes sense, I think I got confused because I am repeating the geos factors on the axial dim.

There is a bug in the application of the norms, that looks like this when idexed in python `arr[:,200:,:]` (I think these are in the view-axial plane, if  am not wrong):

Figure_1

Mostly writing it here because I'll otherwise forget tomorrow.

KrisThielemans:
not sure why you think there's a bug. Doesn't look unreasonable. The first 47 lines are seg 0 (which are different from anything else), etc.

AnderBiguri:
@KrisThielemans ah, I think I'm having a dejavu, I have already gone through this. It is because if the ordering of those views. The flickering of the first rows confused me, plus the higher frequency flickering to the right, but this is due to the ordering, right? I need to sit down and learn better what is the ordering here I think. I'm happy with it for now if you say it looks good.

KrisThielemans:
I think there's 2 items left:
- the hard-wired code mentioned in #675 
- renaming  remaining`*Signa` files (only listmode?), I guess to `*GEHDF5` or possibly should be `GERDF9`. @AnderBiguri what do you think?

AnderBiguri:
@KrisThielemans 
Is `span==2` true always for GE data?

https://github.com/UCL/STIR/blob/e92a728536df146a4f9e4a9d6b7e1f6332621c05/src/buildblock/ProjDataGEHDF5.cxx#L71-L77

KrisThielemans:
for all present scanners. yes. (older (up to RDF6?) scanners had "2D PET" mode with inly segment zero but higher span, but that currently doesn't affect STIR).

I guess we could have a check on the total number of sinograms if that's easy (`proj_data_info.get_num_sinograms()` would work)

AnderBiguri:
> for all present scanners. yes. (older (up to RDF6?) scanners had "2D PET" mode with inly segment zero but higher span, but that current y doesn't affect STIR).

Good!

> I guess we could have a check on the total number of sinograms if that's easy (`proj_data_info.get_num_sinograms()` would work)

hum, what is the exact formula that we are looking for? Something with the rings/segments and sinograms, but is it written somewhere so I don't need to use my brain too much (and use it somewhere else)?

KrisThielemans:
> > I guess we could have a check on the total number of sinograms if that's easy (`proj_data_info.get_num_sinograms()` would work)
> 
> hum, what is the exact formula that we are looking for? Something with the rings/segments and sinograms, but is it written somewhere so I don't need to use my brain too much (and use it somewhere else)?

`get_num_sinograms()` does all the thinking for you :-) It gives the total number of sinograms over all segments. The question is if there's an easy RDF9 field corresponding to it (should be in the projdata case, but otherwise not I guess). you probably check this already.

AnderBiguri:
@KrisThielemans  hum I am not following. I was thinking on replacing the `span` hard-coded number. How does `ge_num_sinograms()` give me 2? There are certainly more than 2, aren't there?

KrisThielemans:
don't attempt to replace it as it's correct. I was just saying that we could have a check afterwards. if this is too complicated, leave it

AnderBiguri:
OK, trying to get this done so we can see the flips in #675 in isolation and with this either merged or ready to do so. 

I am getting stuck (again). How do I make `CListRecordGEHDF5.h` know about the scanner? This is required both to change the hard-coded 447 `get_num_detectors_per_ring` and I guess the following line: 

https://github.com/UCL/STIR/blob/2500cca678fd6a8a49c2de813fd20f6622c3be38/src/include/stir/listmode/CListRecordGEHDF5.h#L219

KrisThielemans:
Seems that  the `CListRecordGEHDF5` constructor needs to take a `shared_ptr<const Scanner>`. Then it can be called that way in `CListModeDataGEHDF5::get_empty_record_sptr()`. Then we need to have the same for `CListEventDataGEHDF5`.

KrisThielemans:
while you're doing that, you could also change text in these files to refer to RDF9.

KrisThielemans:
I think we should then merge this PR, as this PR then generalised @pwadhwa351 's code to RDF9. We then sort out #636  (clean-up), and then #675 .

AnderBiguri:
I am quite lost on the listmode stuff, help! The latest commit is my unsuccessful attempt to get somewhere... I seem to not understand few things, I think.

This is never called? Seems that its always using the default constructor instead of mine, but I don't know where that happens...
https://github.com/UCL/STIR/blob/ab0fc96d39ec3cdcea72906bbbdf02ae59b975f0/src/include/stir/listmode/CListRecordGEHDF5.h#L67-L71

Whatever I do that changes this size, i.e. the size if `CListEventDataGEHDF5` (which I need to change if I need to store `num_detectors_per_ring` or `scanner_sptr`) makes the entire thing go bad.

https://github.com/UCL/STIR/blob/ab0fc96d39ec3cdcea72906bbbdf02ae59b975f0/src/include/stir/listmode/CListRecordGEHDF5.h#L339

As an example, simply uncommenting the following line will make `lm_to_projdata` not detect any event:

https://github.com/UCL/STIR/blob/ab0fc96d39ec3cdcea72906bbbdf02ae59b975f0/src/include/stir/listmode/CListRecordGEHDF5.h#L93


I seem to lack a solid combination of STIR and C++ knowledge... @KrisThielemans help pease!

KrisThielemans:
hmmm. right. I guess the class set-up might require that `CListEventDataGEHDF5` is a "plain" class with only members related to the actual listmode event (it's a 6-byte record). If that is required or not, will need a bit of code review. It could be a leftover from other classes where it's read from file (as opposed to from HDF5).

Data will be read by [InputStreamWithRecordsFromHDF5](https://github.com/UCL/STIR/blob/ab0fc96d39ec3cdcea72906bbbdf02ae59b975f0/src/include/stir/IO/InputStreamWithRecordsFromHDF5.h). what exactly is required there, I don't know.

I'll have a look after lunch :-)

KrisThielemans:
The problem is that the listmode reading needs to copy data from the HDF5 buffer to the "record". (It currently does that via an intermediate buffer, but fine). The actual copy is here
https://github.com/UCL/STIR/blob/ab0fc96d39ec3cdcea72906bbbdf02ae59b975f0/src/include/stir/listmode/CListRecordGEHDF5.h#L308
The problem with this kind of thing is that we need to get a buffer into a C++ object, but we need to do it in a portable way. As you have no control over packing etc of structures, and we need access to underlying bits anyway, we use [bit fields](https://en.cppreference.com/w/cpp/language/bit_field), although even there there is no struct guarantee about packing (hence the `BOOST_STATIC_ASSERT`). (I've found [a library](https://github.com/wkaras/C-plus-plus-library-bit-fields/blob/master/Bitfield.pdf) that claims to circumvent some of the compiler dependencies , although still with no strict guarantees).

This is fragile however. You can also not take address of bit-fields. Mixing this with anything else in C++ is bound to cause problems on some system.

Therefore, `CListEventDataGEHDF5` can only contain the bit-fields data, and non-virtual methods.

I'll add some of this to the STIR doc.

Now I have to think how to get round our problem

KrisThielemans:
It seems to me that the best (only?) way to handle this is to do the swap in a more-full-fledged C++ class.  At first glance, there's 2 options:
- create a `CListEventGEHDF5` class  with lots of interesting stuff, including a `CListEventDataGEHDF5`. This is the strategy followed in other listmode hierarches, see. e.g. [CListEventECAT8_32bit](https://github.com/UCL/STIR/blob/ab0fc96d39ec3cdcea72906bbbdf02ae59b975f0/src/include/stir/listmode/CListRecordECAT8_32bit.h#L80)
- try and avoid that, and immediately put it in `CListRecordGEHDF5` here https://github.com/UCL/STIR/blob/ab0fc96d39ec3cdcea72906bbbdf02ae59b975f0/src/include/stir/listmode/CListRecordGEHDF5.h#L242-L243 Sadly, I don't think that can be done due to #680, but also even the `const` version wouldn't be able to modify itself anyway.

I can give a go at the first method, unless @AnderBiguri  sees something else?

AnderBiguri:
@KrisThielemans unable to comment here, I have gotten to "more or less I know whats going on here" with these listmode classes, definitely not in the "I have any idea how to improve it" :) . I trust your assessment. Let me know how can I help.

KrisThielemans:
Seems we don't read energy-window info yet, so the scatter fails. Should be easy to fix.

KrisThielemans:
I guess this is now "ready". I've run it through the Signa VQC data and results seem ok. I haven't tried any other data yet. Hopefully tests work out ok.

I did make some changes, but nothing dramatic, so it should still give the same results (I hope!).

@AnderBiguri maybe you could run the DMI data through this with scatter now.

Obviously, I didn't fix #636 nor #675

KrisThielemans:
The scale factor for the randoms when reading from a sinogram is currently incorrect.

KrisThielemans:
Making up a (probably wrong) scale factor gives me the following images for the DMI phantom (left PET, right CT)
image

getting there...

KrisThielemans:
I might have fixed the randoms (I think they were wrong for listmode data as well). I've also created some additional scripts to make  CTAC/MRAC processing a bit easier.

Current OSEM result after 140 subiters (!7 subsets) and a 3x3x3 median
image

Some interesting artefacts. I wonder how the GE non-TOF recon looks like...

KrisThielemans:
This was obtained as follows
```bash
datadir=../../DMI/dual-phantom-from-GE/exam177_dualPhantom_DMI3R_CT_RAW_RDFv9
RDF=${datadir}/raw_HDF5_decompressed/rdf_f1b1.rdf
for f in ${datadir}/raw_dcm/*.3.img; do nm_extract -i $f -o .; done
ctac_to_mu_values -j ctac_slopes.json -m GENERIC -k 511 -o ctac.hv -i ${datadir}/CTAC/*.1.img

# copy to Interfile to get round a limitation of the current
# RDF reader (it doesn't have get_sinogram(), and is rather slow)
stir_math -s `basename ${RDF%%.rdf}`.hs ${RDF}
RDF=`basename ${RDF%%.rdf}`.hs

randoms=randoms_`basename ${RDF%%.*}`.hs
construct_randoms_from_GEsingles ${randoms} ${RDF}

RDFNORM=`ls *norm.rdf`
export RDFNORM

sino_input=${RDF} randoms3d=${randoms} RDFNORM=${RDFNORM} num_subsets=17 num_subiters=42 $pardir/NAC_recon.sh

$pardir/register_GEAC.sh  mu ctac.hv NAC_image_42.hv

sino_input=${RDF} randoms3d=${randoms} RDFNORM=${RDFNORM} atnimg=mu.nii num_subsets=17 num_subiters=140 $pardir/scatter_and_recon.sh
```
where the `ctac_slopes.json` isn't available yet.

KrisThielemans:
We checked and got good agreement on some DMI phantom data (reconstructed from listmode) on global scale of prompts (identical), randoms (within ~1%) and non-TOF scatter(~5%?) as estimated by GE Duetto. So I've merged this now. thanks @AnderBiguri ! And @pwadhwa351 for your assistance!

KrisThielemans:
here's the json file
```
{"scale": {
  "GENERIC": {
    "type": "bilinear",
    "transform": [
     {
        "kvp": 120,
        "kev": 511,
        "a1": 0.096,
        "b1": 9.6e-5,
        "a2": 0.096,
        "b2": 5.73e-5,
        "break": 0
     }
    ]
    }
  }
}
```

AnderBiguri:
@KrisThielemans Trying to figure out what happens... But the latest commit (and merge) and the tests you say worked, simply segfault on my machine when doing `correct_projdata correct_projdata_for_norm.par` (or any `.par). I can only make it work if I revert back all your latest commits until my last one. 

Trying to figure out now, just noting it here.

KrisThielemans:
No idea. For a segfault, you have to compile either in Debug or RelWithDebInfo, and run in gdb (just `gdb –args blabla`). It’ll tell you where it crashed

AnderBiguri:
@KrisThielemans found where it is. Proj info pointers are null in BINNORMFROMGEHDF5, but no idea why.

Issue #606, opened 2020-06-25
problem with presmoothing forward and postsmoothing back projectors


rijobro:
These projectors contain the original projector, whilst themselves inheriting from the projector class. As far as I can tell, this will result in duplication of objects. For example, when we call `set_up` for the forward projector, we do the following: 

```
void
PresmoothingForwardProjectorByBin::
set_up(const shared_ptr<const ProjDataInfo>& proj_data_info_ptr,
       const shared_ptr<const DiscretisedDensity<3,float> >& image_info_ptr)
{
  ForwardProjectorByBin::set_up(proj_data_info_ptr, image_info_ptr);
  original_forward_projector_ptr->set_up(proj_data_info_ptr, image_info_ptr);
}
```
So now there is a copy of the `DiscretisedDensity` in both the `PresmoothingForwardProjectorByBin` class and also in `original_forward_projector_ptr`. 

Before forward projecting, we call `set_input`. At this point, the data processor will be called on the `DiscretisedDensity` inside of `PresmoothingForwardProjectorByBin`. However, when we forward project, we do the following:
```
void
PresmoothingForwardProjectorByBin::
actual_forward_project(RelatedViewgrams<float>& viewgrams,
                  const int min_axial_pos_num, const int max_axial_pos_num,
                  const int min_tangential_pos_num, const int max_tangential_pos_num)
{
    // No need to do the data processing since it was already done on set_input()
    original_forward_projector_ptr->forward_project(viewgrams,
                                                      min_axial_pos_num, max_axial_pos_num,
                                                      min_tangential_pos_num, max_tangential_pos_num);
}
```
It looks like, here, the `DiscretisedDensity` inside of `original_forward_projector_ptr` gets projected. I don't think the data processor will have been applied to this image, so it will just be a normal forward projection. Further, I can't imagine that `set_input` will modify `DiscretisedDensity` this data, so perhaps the image will never change.

The back projector presumably has a similar problem. Notably, `start_accumulating_in_new_image` won't start accumulating in a new image.

I'm rather confused on this one, and I hope I've made a mistake somewhere. 

If I'm not mistaken, I would vote removing the `PreSmoothing` and `PostSmoothing` forward and back projectors altogether.

rijobro:
Potential problem noted here: https://github.com/UCL/STIR/pull/580.

KrisThielemans:
Seems to me that
[...]

PR #607, opened 2020-06-25
remove presmoothingforward & postsmoothingback projectors

rijobro:
Suggestion fix for https://github.com/UCL/STIR/issues/606. Only implemented copies for `...MatrixByBin` projectors, but would have to be implemented for all projectors. Could easily be extended to projector pairs.


src/recon_buildblock/BackProjectorByBin.cxx
```diff
@@ -55,6 +55,12 @@ BackProjectorByBin::BackProjectorByBin()
     set_defaults();
 }
 
+BackProjectorByBin::BackProjectorByBin(const BackProjectorByBin& to_copy)
+    : _already_set_up(false), _density_sptr(nullptr), _post_data_processor_sptr(nullptr)
```
KrisThielemans, line 59: why set these to `null` etc? I can agree you'd want to force people to call `set_up` again (although I'm not sure it's strictly necessary), but I find it dangerous to remove any `_post_data_processor_sptr`. If there was one, it won't be a clone anymore!

src/recon_buildblock/ForwardProjectorByBin.cxx
```diff
@@ -56,6 +56,12 @@ ForwardProjectorByBin::ForwardProjectorByBin()
   set_defaults();
 }
 
+ForwardProjectorByBin::ForwardProjectorByBin(const ForwardProjectorByBin &to_copy)
+    : _already_set_up(false), _density_sptr(nullptr), _pre_data_processor_sptr(nullptr)
```
KrisThielemans, line 60: see comment for `BackProjectorByBin`

src/recon_test/test_projector_clone.cxx
```diff
@@ -0,0 +1,217 @@
+//
+//
+/*!
+
+  \file
+  \ingroup test
```
KrisThielemans, line 6:
```suggestion
  \ingroup recontest
```

src/recon_buildblock/BackProjectorByBin.cxx
rijobro: I thought it best to set them as null so as not to create a shallow copy (since they don't have their own `create_shared_clone` methods). What if you think you're modifying the data processors parameters for the clone, when in actual fact, you're modifying the parameters of both the clone and the original?

rijobro:
the `unique_ptr` was functionality that I copied from SIRF's `clone` methods for `DataContainer`. Would you rather it were `shared_ptr`? I thought `unique_ptr` would let the user know that the copy wasn't stored elsewhere.

rijobro:
> Do we really need to implement most of these copy constructors? Isn't the default one ok in most cases?

If we use the default, then we'll create a shallow copy of any shared pointers, which seems dangerous to me. Ideally, both `DataProcessor` and `DiscretisedDensity` would have their own `create_shared_clone` methods which I would have used in the base-level projectors.

> I guess we need to still add a few others.

It looks like, currently, this functionality would only be used in https://github.com/UCL/STIR/pull/580 and then https://github.com/UCL/STIR/pull/240 once I get that up to date again. Could we do it on a case-by-case basis?

src/recon_buildblock/BackProjectorByBin.cxx
KrisThielemans: That sounds dangerous indeed. However, the question is if it can happen. We've put in a lot of effort to avoid it in the other stuff.

However, as far as I remember, `_density_sptr` is only set by `BackProjector::set_up()` which would create a new sptr. (not modify an existing one). Indeed, I don't think there's a way to modify to modify the shared object.  Similarly, as long as we don't have a `shared_ptr<DataProcessor> get_post_data_processor_sptr()` or `DataProcessor& get_post_data_processor()`, I cannot see how it would be unsafe to reuse it.

It's possibly a bit confusing that `set_post_data_processor` would override anything that was already there. We replace that then with `add` (and use `ChainedDataProcessor`) and add a `remove_post_data_processor`  like I believe you do in SIRF transformations. However, I don't think that's really necessary for now.

So, if we don't do that work, the question is what is more surprising: a clone which doesn't do the same (the current situation), or a clone where a subsequent `set_post_data_processor` would override whatever was there. I certainly think the former is less desirable.

KrisThielemans: > the `unique_ptr` was functionality that I copied from SIRF's `clone` methods for `DataContainer`. Would you rather it were `shared_ptr`? I thought `unique_ptr` would let the user know that the copy wasn't stored elsewhere.

That's true, but it's weird to have a `create_shared` return a `unique`. It should then be called `create_unique_clone`. I'm fine with calling it the latter. (It's certainly better than implying that its properties are shared)

KrisThielemans:
> > Do we really need to implement most of these copy constructors? Isn't the default one ok in most cases?
> 
> If we use the default, then we'll create a shallow copy of any shared pointers, which seems dangerous to me. Ideally, both `DataProcessor` and `DiscretisedDensity` would have their own `create_shared_clone` methods which I would have used in the base-level projectors.

discussed above. I don't think that's necessary here.

Obviously, the aim of having `shared_ptr` is to share some stuff (as long as it's safe). Otherwise we should have `unique_ptr` all over the place.

src/recon_buildblock/BackProjectorByBin.cxx
KrisThielemans: 2nd thoughts. Are they in fact unique?

For instance, the projmatrix presumably still is shared (I would certainly suggest that it is, as otherwise you're going to run out of memory very quickly). Can anyone call its various `set` functions? Obviously, the original owner could. This is an example where sharing is actually necessary, albeit it dangerous. oh well

This is

rijobro: It's confusing, I tried to make them as independent as possible (e.g., no sharing of data processors), but they do share the projmatrix (for memory). 

My feeling is that we can do what we want in terms of which objects are copied shallowly and deeply, provided we document it well. I already put in the doxygen that the projmatrix would only be shallow copied. I can also do a shallow copy of the data processor if you'd like, and document that, too. 

We could name the method `create_clone`, which is less specific on unique versus shared. This seems appropriate, since we're doing a mix of the two.

What do you think?

PR #612, opened 2020-07-01
ProjDataInMemory speed-up

### LINKED ISSUE #611: optimisations for `ProjDataInMemory`
C++
-----

[`ProjData::fill_from`](https://github.com/UCL/STIR/blob/7ad42836e6d813313b8f2f5a83f8c6d3e04a47fb/src/include/stir/ProjData.h#L205) and [`ProjData::copy_to`](https://github.com/UCL/STIR/blob/7ad42836e6d813313b8f2f5a83f8c6d3e04a47fb/src/include/stir/ProjData.h#L241) use explicit loops with `SegmentByView` etc. This is slow.

For `ProjDataInMemory` we should be able to directly copy via the iterator from/to the buffer. This would currently not work however, as the segments are stored in the wrong order: `fill_from` and `copy_to` use order `{0,1,-1,2,-2,...}`. `ProjDataInMemory` uses the default order of `ProjDataFromStream`
https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/buildblock/ProjDataFromStream.cxx#L114-L122
which is `{....,-2,-1,0,1,2,...}`. It seems to me that we should change the constructors
https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/buildblock/ProjDataInMemory.cxx#L104-L105
and
https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/buildblock/ProjDataInMemory.cxx#L115-L117
to use the `{0,1,-1,...}` sequence. We can do that by either:
- changing the default order for `ProjDataFromStream`. probably not a bad idea, but dangerous to rely on a default
- creating a function that returns a `std::vector` with that sequence (can steal code from https://github.com/UCL/STIR/blob/7dc238f5ec228554f594f1f2228450eb4618472b/src/IO/stir_ecat_common.cxx#L193, but note that this uses `{0,-1,1,...}`).

Doing this should speed-up SWIG's `to_numpy` and SIRF's `as_array` as it calls `copy_to` [here](https://github.com/SyneRBI/SIRF/blob/c147b231e4b239177d58a5b1f8d1d9aabd4077b3/src/xSTIR/cSTIR/cstir.cpp#L635).

SWIG
-----
- We should expose the `ProjDataInMemory` iterators. That should be easy: copy lines from
https://github.com/UCL/STIR/blob/7dc238f5ec228554f594f1f2228450eb4618472b/src/swig/stir.i#L1149-L1153

- `to_numpy` for `ProjData` first creates a `Array<3,float>` https://github.com/UCL/STIR/blob/7dc238f5ec228554f594f1f2228450eb4618472b/src/swig/stirextra.py#L81-82 and then converts to `numpy` array. If we'd add a `shape()` to `ProjDataInMemory`,  I believe that the "normal" lines will work already:
https://github.com/UCL/STIR/blob/7dc238f5ec228554f594f1f2228450eb4618472b/src/swig/stirextra.py#L75-L77
However, probably Python won't do that if it's passed a `ProjData` object (even if it's ` ProjDataInMemory`), so possibly we'd have to do a type-cast somehow.

KrisThielemans:
fixes #611


KrisThielemans:
Too early to review. Nothing happened yet...

KrisThielemans:
A few more things that can be done to use new `stir::fill_from` in constructors I believe, but it works now. See timings from `test_proj_data`.

In addition, the current implementation is a bit fragile as it needs manual overloading from objects which don't have `begin_all`, leading to some ugly repetition. SFINAE could help (using things similar to [type erasure](http://davekilian.com/cpp-type-erasure.html)). Note that I currently don't see how SFINAE would help with recognising that when passed a `ProjData&` it should check if it's actually a `ProjDataInMemory&`, which we currently do by hand:
https://github.com/UCL/STIR/blob/4e9fbe73114c500f97e85ee201ba37e2cb9e3ccb/src/include/stir/copy_fill.h#L70.

Issue #357, opened 2019-08-06
enable use of virtual crystals in ROOT processing


KrisThielemans:
The mMR and other scanners use "virtual" crystals. In GATE you have to simulate the true crystal block, but when reading the ROOT file, it would make sense to insert the virtual crystal to be able to reproduce sinograms like the scanner does.

This was implemented on the GATE_ROOT branch. In particular, in the lines 
https://github.com/UCL/STIR/blob/dd5af80c9ee5709bac8b00e211f1aa5e9006213e/src/utilities/Bin_GATE.cxx#L317-L320
`N_CRY_xy` was set to 9, above while the crystal is computed modulo 8 (also [using 8 for the ring](https://github.com/UCL/STIR/blob/dd5af80c9ee5709bac8b00e211f1aa5e9006213e/src/utilities/Bin_GATE.cxx#L308 ).

Equivalent lines are in https://github.com/UCL/STIR/blob/4064c0dd4a75029540624be2ff219081fc41454f/src/IO/InputStreamFromROOTFileForCylindricalPET.cxx#L93-L109 and https://github.com/UCL/STIR/blob/4064c0dd4a75029540624be2ff219081fc41454f/src/IO/InputStreamFromROOTFileForECATPET.cxx#L89-L99

https://github.com/UCL/STIR/pull/352 introduces `Scanner::get_num_virtual_axial_crystals_per_block()` and `Scanner::get_num_virtual_transaxial_crystals_per_block()`. It seems that we could use these.

KrisThielemans:
It'd presumably need adapting the checks for the scanner.

PR #617, opened 2020-07-06
InputStreamFromROOTFile changes for virtual crystals

KrisThielemans:
Closing #357

### DIFF
--- a/src/IO/InputStreamFromROOTFile.cxx
+++ b/src/IO/InputStreamFromROOTFile.cxx
@@ -74,6 +74,8 @@ InputStreamFromROOTFile::set_defaults()
     crystal_repeater_x = -1;
     crystal_repeater_y = -1;
     crystal_repeater_z = -1;
+    num_virtual_axial_crystals_per_block = 0;
+    num_virtual_transaxial_crystals_per_block = 0;
 }
 
 void
@@ -174,6 +176,18 @@ void
 InputStreamFromROOTFile::set_crystal_repeater_z(int val)
 {
     crystal_repeater_z = val;
+}
+
+void
+InputStreamFromROOTFile::set_num_virtual_axial_crystals_per_block(int val)
+{
+  num_virtual_axial_crystals_per_block = val;
+}
+
+void
+InputStreamFromROOTFile::set_num_virtual_transaxial_crystals_per_block(int val)
+{
+  num_virtual_transaxial_crystals_per_block = val;
 }
 
 END_NAMESPACE_STIR
--- a/src/IO/InputStreamFromROOTFile.cxx
+++ b/src/IO/InputStreamFromROOTFile.cxx
@@ -1,3 +1,11 @@
+/*!
+  \file
+  \ingroup IO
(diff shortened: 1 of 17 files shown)


KrisThielemans:
work in progress

KrisThielemans:
This might be working now for the mCT with some limitations/caveats:
- although you can set the number of virtual crystals via the hroot file, you cannot set it to anything else than what `stir::Scanner` expects (due to a limitation in `Scanner`). Therefore, you have to make sure that the hroot file specifies the scanner name correctly (i.e. either "Siemens mCT" or "mCT").
- although the current [run_root_GATE.sh](https://github.com/UCL/STIR/blob/master/recon_test_pack/run_root_GATE.sh) passes, it does only checks on total counts, so I don't know if I broke anything!

Things to do therefore:
- [ ] modify `stir:Scanner` to allow changing the number of virtual crystals
- [ ] modify `run_root_GATE.sh` to do some tests on actual coincidence decoding
- [ ] test with mCT GATE output

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -56,6 +66,11 @@ InputStreamFromROOTFile::set_defaults()
     low_energy_window = 0.f;
     up_energy_window = 1000.f;
     read_optional_root_fields=false;
+    crystal_repeater_x = -1;
+    crystal_repeater_y = -1;
+    crystal_repeater_z = -1;
```
robbietuk, line 71: Does this really need to be -1? STIR can only handle cases right now where z= 1.

KrisThielemans:
could test ROOT interpretation in more detail by using for instance
```
INPUT_ROOT_FILE=test_PET_GATE.root EXCLUDE_SCATTERED=1 EXCLUDE_RANDOM=0 list_lm_events --coincidence 1 --time 0 --num-events-to-list 100 root_header.hroot 
```
A simple thing to do would be to pipe the output of this to a file with a version that is known to work, add the file to the `recon_test_pack` and in the test, compare the output. We'd have to do the same for the timing events.

src/IO/InputStreamFromROOTFile.cxx
KrisThielemans: I've only moved this code from the original location in the derived classes, so no change there. I didn't check if there's actually any error checking or handling of defaults on those values. I agree it seems more logical to let them default to 1 but it doesn't seem high priority

robbietuk:
I unlisted the same root file from a D690 scanner with `lm_to_projdata` from using master and this PR branch. I get the same number of events in both sinograms and subtracting one data file from the other gives a file of 0's. It appears that this PR does not affect the current unlisting code regarding unlisting root files in number of events or shape.

KrisThielemans:
I've updated the doc and rebased on `master` as the history got a bit convoluted (and there was a conflict). should be good to go.

KrisThielemans:
2 Travis jobs failed, nothing to do with us (networking error and unknown error). Merging.

PR #619, opened 2020-07-08
Adding 3D normalisation codes

tniknejad:


KrisThielemans:
thanks! I'll merge current `master` in here. Can I push to this branch then please?

KrisThielemans:
The diff is currently large because I had to merge the OSX fix #626 in here somehow. It will get smaller again once that gets merge into `master` and that here. sorry about that.

Otherwise this should be as good as finished. Still need to update copyright I guess.

Ideally we'd have a better test case, but maybe @tniknejad you could check if it still does what it's supposed to do?

At present, I haven't enabled the "gaps" processing yet. This will need some updating as the current "virtual" crystals are hard-wired to be 1 in every direction. We can get that from the `Scanner` these days. Maybe we'll do it in a different PR.

KrisThielemans:
@robbietuk you could give this a go on your GATE data now.

Usage:
```
outer_iters=5
eff_iters=6
find_ML_normfactors3D  norm_factors $measured_data $model_data $outer_iters $eff_iters
# create ones
stir_math -s -including-first --times-scalar 0 --add-scalar 1 ones.hs $model_data
# mutiply ones with the norm factors to get a sino
apply_normfactors3D $prog norm_sino.hs norm_factors $model_data 1 $outer_iters $eff_iters
```
where `$model` is a proj-data obtained by forward projecting with STIR (including attenuation if any)

PR #621, opened 2020-07-09
interfile reading fixes for Siemens sinograms and attenuation files

### LINKED ISSUE #620: interfile reader can fail on Siemens sinogram files due to EOF
To determine the type of Interfile, we read it multiple times. After the first time, we reset the "get" pointer back to 0
https://github.com/UCL/STIR/blob/3503d735225ad52eef1f1c5f60a9211d44c66fdd/src/IO/interfile.cxx#L1076

However, if the file was read till EOF, `seekg(0)` appears to fail. According to http://www.cplusplus.com/reference/istream/istream/seekg/, this should only happen in pre-C++11, but that doesn't seem to be the case.

I'm confused why this doesn't happen with SPECT headers, but the fix is to do a `clear` first.

KrisThielemans:
we read the Interfile header multiple times. After the first time, the EOF flag can be set,
causing problems later. Now we clear it explicitly.

Fixes #620

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,10 +89,15 @@ addons:
 before_install:
   - |
     if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then 
-      brew update
+
       # seems to be needed on Travis due to conflicts with gcc otherwise
       # see https://github.com/travis-ci/travis-ci/issues/8826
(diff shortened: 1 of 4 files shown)


KrisThielemans:
All tests work except for OSX python due to #623, so merging

PR #639, opened 2020-07-21
Improve speed of get_bin_value

### LINKED ISSUE #638: optimise ProjDataInMemory::get_bin_value()
Currently, `ProjDataInMemory::get_bin_value()` is terribly slow. It should be a lot faster if we just remove it, or explicitly reimplement it in terms of `ProjDataFromStream` as suggested previously https://github.com/UCL/STIR/issues/462#issue-586411241 but it seems we didn't do that yet.

Even better would be to use something like this
```C++
{
// first find offset in the stream
    vector<streamoff> offsets = get_offsets_bin(this_bin);
    const streamoff total_offset = offsets[0];
// now convert to index in the buffer
    const int index = static_cast<int>(offsets[0]/sizeof(float));
   return buffer[index];
}

Clearly, the same would apply for `set_bin_value`.

AnderBiguri:
Fix #638

### DIFF
--- a/src/buildblock/ProjDataInMemory.cxx
+++ b/src/buildblock/ProjDataInMemory.cxx
@@ -140,10 +140,7 @@ get_size_of_buffer_in_bytes() const
 float 
 ProjDataInMemory::get_bin_value(Bin& bin)
 {
-   Viewgram<float> viewgram = get_viewgram(bin.view_num(),bin.segment_num()); 
-    
-   return viewgram[bin.axial_pos_num()][bin.tangential_pos_num()]; 
-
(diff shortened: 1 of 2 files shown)


KrisThielemans:
this is fine, but would you be able to try the other option? Just make `get_bin_offsets` protected. It's bound to be even faster, with very little effort.

AnderBiguri:
Sure thing! I tend to try no tot fiddle with these things too  much as STIR is a bit too big for me to guess what it will break, was waiting this confirmation :) 

13s vs 50s!

KrisThielemans:
great. please merge `release_4` into #601

PR #671, opened 2020-08-25
move ctac_to_mu_values to a HUToMu data processor

### LINKED ISSUE #660: Add example of how to use JSON files for ctac_to_mu_values
`ctac_to_mu_values` has no docs, which makes it hard to use. 

Add either an example `.json` file or some docs to explain how it works.

KrisThielemans:
I've converted it to a data processor, which makes it much more useful.

Also added some doc.

Fixes #660


KrisThielemans:
I should now have resolved the problem with the float values.

AnderBiguri:
Ah, now I am the reviewer! Well, I don't really understand the full STIR yet, so please bear with me: You mention that `ctac_to_mu_values` is now obsolete and now `HUToMu` is a data processor. Good software-wise!

But if we are talking about documentation (which is why I opened the issue) now I don't know neither how to call the `ctac_to_mu_values` nor how to use `HUToMu` data processor (without reverse engineering...). 
I know I am quite strict (arguably _too strict_) with what I accept as "good documentation", but I generally take C++/python/MATLAB docs as reference... For me this means more or less filling the formula (in this order): 

1. Single liner of what the code does
2. How to use the code (example of code, that can be essentially copy pasted) e.g. `ctac_to_mu_values -o output_filename -i input_volume -j slope_filename -m manufacturer_name [-v kilovoltage_peak] -k target_photon_energy` (which is what runs when you run `ctac_to_mu_values` without arguments). 
3. Detailed explanation of what the code does (as you have added, [here](https://github.com/UCL/STIR/pull/671/files#diff-8d80f9b47d8bb1f5669f6d7784cc7f21R26-R78), which I assume goes to Doxigen docs)
4. Detailed explanation on the parameter possibilities (e.g. multiple piecewise? input/output file formats? etc). This is also more or less either "obvious" for STIR, because it accepts everything, or you have added it to the Doxigen chunk. 

Now, the "data processors" have essentially just "1 or maybe 3" in the STIR-UsersGuide.tex, but currently I have no idea how to use them by reading the entire documentation. I know particulars of their inputs, but not how to call them! Similarly, the docs [do not show at any point how to call `ctac_to_mu_values`](https://github.com/UCL/STIR/blob/4447482fa78cae83289f1c377926a4310c6c0ddb/documentation/STIR-UsersGuide.tex#L2883) or if the information is there, its piecewise split in different places (i.e. now in Doxigen and partially in the UsersGuide etc). 

In general when I write code and docs I like to step back and think of the user and how/why does it interact with the code. For something like `OSMAPOSL`, maybe its worth this hypothetical user to spend a day or two fiddling with different documentation spaces and trying to figure out what each thing is or how to call it, as its a complicated code, but right now I'd need at least 30 minutes of trial and error  and source code reading, just to avoid doing

```
mu(HU< 0)= HU(HU< 0)*1.66e-4+0.16;
mu(HU>= 0 )= HU(HU>= 0)*1.48e-4+0.16;
```
in MATLAB. Which ultimately means that I will just do it in MATLAB, where I can expand the piecewise function easily or try different values. 

Of course!!! software-wise this PR is a great idea, and I have no comments on it, :+1: as usual. 


PD1: This sounds super rude when re-read by me, but please don't take it so! Limitations of non-verbal communication on a non-native language, I just wrote a super verbose answer to what I think its good docs to help us write those (if you agree of course!). 

PD2: Let me try to add some docs to this PR so I can show by example what do I mean, and you can accept them or reject them if you want :)

KrisThielemans:
@AnderBiguri as usual I agree with your view on documentation. STIR is lacking in it. But, it is what it is, and without having a paid person fixing this, or volunteers, I can't see how to fix this properly.

One big problem is repetition of documentation. It very quickly gets out of date. So, currently we have it in the doxygen, the usage message, a example parameter file, the user's guide. There is no obvious way to let these things point to each-other.  Some ideas to make this easier:
- Possibly doxygen could generate the user's guide, and then have auto-linking. 
- example parameter files would be generated from the source like ITK does.
- STIR's keymaps have quite a lot, and could be used to generate most of the doc, and even parameter files, but this would need extra work.

STIR currently expects the user to do some (i.e. a lot of) work. It's a major hurdle. But I have no resources to fix this.

`ctac_to_mu_values` was a bad case. It obviously should be added to the User's Guide as well, but I didn't bother (and won't as from my perspective it is now obsolete).

anyway.

AnderBiguri:
Hum, I agree and I see what you say. 

Definitely doc writing is super boring and an enormous task, and its something that is very easy to just let slip away when our main job is to do research, not to do software. No personal criticism here at all! We are all sinners here (and perhaps I may be the worst one!).

 I guess really the only clear "obvious" solution is to re-define how STIR docs are, which would mean choosing one place/system to have all docs and scrapping all the rest. All this is lots of work that I understand we have not time for. In any case, if now `ctac_to_mu_values` is obsolete and `HUToMu` is something more standard STIR-wise, then it requires no much more docs than as it is (as long as you know how to run a "data processor", I guess (which I will need to learn hehe)).

documentation/STIR-UsersGuide.tex
```diff
@@ -3976,6 +3976,29 @@ \subsubsection{
 Otherwise, all non-positive entries in the data are set to SMALL\_NUM 
 times the minimum strictly positive value.
 
+{ \subsubsubsection{HUToMu}
+}
+
+This data processors is intended to convert a CT image in Hounsfield Units to mu-values.
+It takes a file in JSON format specifying the piecewise-linear for various systems, Xray voltages and gamma energies.
+
```
AnderBiguri, line 3984: Add:

"Currently, it supports 2-segment piecewise linear transformations only."

AnderBiguri: (I am happy to add these myself, just noting them here for later)

AnderBiguri (approved):

KrisThielemans:
Pending the fact that we cannot create the User's guide via doxygen, I have made the (non-documented!) decision that the User's Guide should be short in most places, pointing to other things. The idea if somewhat that the name of the section tells the reader where to find it in doxygen (although I'm sure it isn't consistent).

In any case, the doxygen needs to be the most complete, as this is auto-transported to Python doc (and at some point MATLAB).

KrisThielemans:
@danieldeidda this is ready, but if you could check first if it gives the same results as before, that'd be best as we don't have a test for this at the moment.

danieldeidda:
> 
> 
> @danieldeidda could you try this? `ctac_to_mu_values` should behave exactly the same as bef

> 
> 
> @danieldeidda this is ready, but if you could check first if it gives the same results as before, that'd be best as we don't have a test for this at the moment.

Sorry, I just finished to read all the comments, yes let me run this.

danieldeidda:
> 
> 
> @danieldeidda this is ready, but if you could check first if it gives the same results as before, that'd be best as we don't have a test for this at the moment.

I confirm images are exactly the same:

image

danieldeidda (approved):

Issue #446, opened 2020-01-23
BinNormalisation and calibration factor


KrisThielemans:
We need a common facility to be able to include a calibration factor (and use it or not). my inclination is to put it in `BinNormalisation` as we need it for PET as well obviously. That would have a somewhat weird effect that it'd also end up in `BinNormalisationWithAttenuationImage` (or whatever that class is called), and 3 times in all the `Chained` ones. Seems we need a new `BinNormalisationWithCalibration` or something. this could also have a field for normalising with acquisition duration, or even decay. 

_Originally posted by @KrisThielemans in https://github.com/UCL/STIR/pull/445#issuecomment-577747323_

KrisThielemans:
@danieldeidda I've put this discussion in a new issue

KrisThielemans:
Needs #340 ideally

KrisThielemans:
One question that came to mind is where we're going to get the calibration factor from. Different manufacturers have different conventions. Some will store it in the norm files (GE), some in a separate XML (Mediso), and some might store it in the header of the sinogram/listmode (Siemens?). It might mean that the BinnormalisationWithCalibration shouldn't necessarily be doing the parsing itself. a topic for discussion...

Issue #340, opened 2019-06-09
BinNormalisation::apply should use timing info from ExamInfo


KrisThielemans:
`BinNormalisation::apply` uses explicit timing arguments (only used for dead-time). we should get these from the `ExamInfo` instead.

KrisThielemans:
postpone till #44 done, as there are changes there to the norm classes

KrisThielemans:
seems best to pass `exam_info` to `set_up`

danieldeidda:
> 
> 
> seems best to pass `exam_info` to `set_up`

Shall we have a exam_info member in BinNormalisation?

KrisThielemans:
would then have to be, yes. The point is that we shouldn't have to pass timing explicitly in all the `apply`/`undo` calls. (We'd have to allow it in any initial step, as otherwise we get into backwards compatibility issues and trouble with SIRF and STIR versions).

However, note that this is the longer term solution as we discussed. I think we should only tackle this after merging #564 for instance.

danieldeidda:
with "any initial step" you mean something like set_up(). right?

KrisThielemans:
Depends where we are on the path. I think we can resolve #446 (with decay) first without working on this one.

danieldeidda:
BinNormalisationWithCalibration is ready but it only includes calibration factor and branching ratio. Didn't we say we need to separate decay correction for SPECT and PET?

KrisThielemans:
yes. sure.

danieldeidda:
The part for SPECT is already done shallI leave it there or shall I create something like BinNormalisationDecayRotatingScanners  and BinNormalisationDecayStaticScanner (maybe static is not the best world?) This can directly be derived from BinNormalisation I would think.

PR #672, opened 2020-08-26
 BinNormalisationWithCalibration and set_up(projdataInfo, ExamInfo):

danieldeidda:
1) added class for calibration #446 , decay and branching ration corr need to be derived from there;
2) BinNormalisation::set_up() has a new input, exam_info;
3) all instances where set_up() is used have been updated;
4) BinNormalisation::apply() does not need start_time and end_time as inputs, can be extracted from exam_info #340 . Nedd to modify undo().


KrisThielemans (changes requested):
thanks!

I have a few smallish comments, but then also a worry (which actually affects some of my "inline" comments). First the comments:

- best to use `set_up(shared_ptr<const ExamInfo>& exam_info_sptr, const shared_ptr<const ProjDataInfo>&)`. Order is then the same as in other places, and it's best to used a ptr as at some point `ExamInfo` might become a hierarchy
- there's also some in `experimental` (build with it "on", although not all might be included by default)
- I thought we said we'd store the actual `calibration_factor` and `branching_ratio` in `ExamInfo` (certainly the `branching_ratio` should in the future be set by reading the actual sinogram, not the norm file).
- I see no reason to have the `BinNormalisationWithCalibration::apply(RelatedViewgrams&`) members. `BinNormalisation` will implement it in terms of `get_bin_efficiency` (if there is a reason, we obviously still need to do `undo` as you wrote)
- we could provide a backwards compatible `apply(ProjData, start, end)` just calling the new one, but making it deprecated.
- It is a but strange to remove the `start/end` from `apply(ProjData)` but not the one with `RelatedViewgrams` nor `get_bin_efficiency`. However, I strongly recommend to postpone that to another PR, as really we're mixing 2 things here (i.e. how to set calibration factor, and removing unnecessary start/stop parameters).
- will need `set/get_calibration_factor` and `set/get_branching_ratio` members (although the latter would be superseded by a `set_isotope` presumably)

Notes:
- some of my "inline" comments assumed that `BinNormalisation` was derived from `ParsingObject` already, and therefore I suggested calls to `base_type` functions. However  but it appears it isn't, so I think you can ignore those comments for now (or put in the extra mile and derive `BinNormalisation` from `ParsingObject` after all, as is actually recommended).


Now the worry:

How do we see that this fits in the hierarchy actually? The current set-up seems to imply that if you want to do norm with calibration, you need to set up a "chained" bin normalisation, taking the "calibration" and the "normal" one. This would work, but it would be
- inconvenient for the user as having to set-it up quite verbosely, but more importantly doesn't immediately fit with letting `BinNormalisationFromECAT8` or `...GERDF` to read the calibration factor from their input as opposed to the par file.
- somewhat inefficient (we'd first get a viewgram, normalise it with normal factors, then multiply with a global factor), as opposed to doing it all in one go.

The alternative approach is to have this a class that actually doesn't have `apply/undo` members, but a single `get_global_norm_factor()` (or some similar name) that would multiply `calibration_factor`, `branch_ratio` (and later-on `decay`). `BinNormalisationFromECAT8` etc would then be derived from this one, but have to call `get_global_norm_factor()` themselves in their `get_bin_efficiency` (see an alternative below). It would mean modifying those classes though. This might be ok though, as they need to read the calibration factor anyway.

Side note: for the non-stationary scanner (SPECT) case, the decay factor is going to be bin-dependent, so I guess we cannot have a `global` factor after all...

I think the 2nd approach is better as ultimately easier for the user (but a bit more work for you...).

It might be better to isolate all the factor stuff in this class, and force the derived class to take it into account, as opposed to hoping that they will do it. I think we could have therefore something like this
```
BinNormalisationWithCalibration : BinNormalisation // assuming this is already derived from ParsingObject
{
  protected:
   get_calib_decay_whatever_factor(const Bin&) const; // TODO find a better name
   // needs to be implemented by derived class
   float get_uncalibrated_bin_efficiency(const Bin&) = 0;
   float get_bin_efficiency(const Bin&)
    { return this->get_uncalibrated_bin_efficiency(bin)/get_calib_decay_whatever_factor(bin); }

   void set_calibration_factor(float);
private:
  // provide facility to switch off things?
  // would need to be added to the parsing keywords
  bool use_calibration_factor; // default to true
  bool use_branching_ratio; // default to true
};

BinNormalisationFromECAT8 : 
public RegisteredParsingObject<BinNormalisationFromECAT8, BinNormalisation, BinNormalisationWithCalibration>
{
 // adjust base_type, check that all functions like set_up/set_defaults etc call their base_type version 
 // during parsing, call `set_calibration_factor()`
 // rename get_bin_efficiency to get_uncalibrated_bin_efficiency
  // no other changes required
};
```
In this approach, `BinNormalisationWithCalibration` is not a leaf-class, so wouldn't have its own `add_start/stop_parsing_key` (as I commented).

hmmm.... What do you think?

danieldeidda:
Hi, I will take some time to read this carefully, in the meantime I will make step by step the modification you suggested

danieldeidda:
> there's also some in experimental (build with it "on", although not all might be included by default)
I run this with experimental on but I don't get an error with the new code

danieldeidda:
> some of my "inline" comments assumed that BinNormalisation was derived from ParsingObject already, and therefore I suggested calls to base_type functions. However but it appears it isn't, so I think you can ignore those comments for now (or put in the extra mile and derive BinNormalisation from ParsingObject after all, as is actually recommended).

I think it is derived from ParsingObject trough registeredObjected (which is derived from RegisterObjectBase, which is derived from ParsingObject.

src/recon_buildblock/BinNormalisation.cxx
```diff
@@ -113,9 +114,11 @@ undo(RelatedViewgrams<float>& viewgrams,const double start_time, const double en
 
 void 
 BinNormalisation::
-apply(ProjData& proj_data,const double start_time, const double end_time, 
+apply(ProjData& proj_data,
       shared_ptr<DataSymmetriesForViewSegmentNumbers> symmetries_sptr) const
 {
+    float start_time=exam_info.get_time_frame_definitions().get_start_time();
```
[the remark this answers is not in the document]
danieldeidda: not sure I understand how can we check that two objects are the same "up to rounding" do we check each member? I can see for projdata_info there is a >= but this doesn't work for exam_info

KrisThielemans:
> I think it is derived from ParsingObject trough registeredObjected (which is derived from RegisterObjectBase, which is derived from ParsingObject).

ok

src/recon_buildblock/BinNormalisation.cxx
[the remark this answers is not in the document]
KrisThielemans: it would indeed need an `operator==` in `ExamInfo`, and therefore also `TimeFrameDefinitions`. could leave this for later

danieldeidda:
> will need set/get_calibration_factor and set/get_branching_ratio members (although the latter would be superseded by a set_isotope presumab

In ExamInfo or BinNormalisationWithCalibration

danieldeidda:
@KrisThielemans  Did you submit changes or comments here? I remember you were writing but I don't see them. 

Anyway I seem to have forgotten to take a note on this but I think I am correct when saying that all get_bin_efficiency() functions will now become get_uncalibrated_bin_efficiency()

src/include/stir/ExamInfo.inl
```diff
@@ -50,6 +50,12 @@ ExamInfo::set_high_energy_thres(float new_val)
     up_energy_thres = new_val;
 }
 
+void
+ExamInfo::set_radionuclide(std::string name)
```
KrisThielemans, line 54:
```suggestion
ExamInfo::set_radionuclide(const std::string& name)
```
as in other STIR places

KrisThielemans (changes requested):
forgot to press the "submit" button...

One problem with this PR is that we're going to break SIRF, as it's using the "old" style `apply` (e.g. [here](https://github.com/SyneRBI/SIRF/blob/21eb359f33b75cf44a289ae1d1c569a9df627cf7/src/xSTIR/cSTIR/stir_x.cpp#L486)) and `set_up` ([here](https://github.com/SyneRBI/SIRF/blob/21eb359f33b75cf44a289ae1d1c569a9df627cf7/src/xSTIR/cSTIR/stir_x.cpp#L472)). Can you keep the old functions in place(only in `BinNormalisation`) for backwards compatibility?  Flag them as `\deprecated` in their doxygen)

danieldeidda:
I remember we said to remove registered_name as we don't use parsing, however I get this error:

/home/nmdicom-recon/CalibNorm/STIR/src/include/stir/RegisteredParsingObject.inl:34: error: ‘registered_name’ is not a member of ‘stir::BinNormalisationWithCalibration’{ return Derived::registered_name; }

I think this happens when I say:


   public RegisteredParsingObject<BinNormalisationFromECAT8, BinNormalisation, BinNormalisationWithCalibration>

src/include/stir/recon_buildblock/BinNormalisationWithCalibration.h
```diff
@@ -0,0 +1,65 @@
...
+#include "stir/recon_buildblock/BinNormalisation.h"
+#include "stir/decay_correction_factor.h"
+
+START_NAMESPACE_STIR
+
+/*!
+  \ingroup normalisation
+*/
+class BinNormalisationWithCalibration : 
+        public RegisteredParsingObject<BinNormalisationWithCalibration, BinNormalisation >
```
KrisThielemans, line 34:
```suggestion
        public BinNormalisation
```
only (final) leaf classes should be derived from `RegisteredParsingObject`. Changing this should resolve your error removing `registered_name`

danieldeidda:
> 
> 
> forgot to press the "submit" button...
> 
> One problem with this PR is that we're going to break SIRF, as it's using the "old" style `apply` (e.g. [here](https://github.com/SyneRBI/SIRF/blob/21eb359f33b75cf44a289ae1d1c569a9df627cf7/src/xSTIR/cSTIR/stir_x.cpp#L486)) and `set_up` ([here](https://github.com/SyneRBI/SIRF/blob/21eb359f33b75cf44a289ae1d1c569a9df627cf7/src/xSTIR/cSTIR/stir_x.cpp#L472)). Can you keep the old functions in place(only in `BinNormalisation`) for backwards compatibility? Flag them as `\deprecated` in their doxygen)

Shall we have two functions for set_up() and apply() or shall we make the new one a property of the derived classes?

KrisThielemans:
ah, ok. I didn't think you'd do the ECAT7 etc code now (as ideally we'd then read the calibration factor etc from the header), but fine.

I'm not sure what to do with the `ChainedBinNormalisation` though. In normal practice, one of these will be attenuation, the other "true" normalisation. It's the latter that would be derived from `...BinNormalisationWithCalibration`. I feel therefore that `ChainedBinNormalisation` should not be derived from `BinNormalisationWithCalibration`. That sounds simple enough, but the difficulty becomes if we want to store the calibration factor etc in the image after recon. It seems we need to have `get_calibration_factor()` to be moved up to `BinNormalisation` for that. But it's not clear to me how that would work... sorry.

KrisThielemans:
The restarted Travis jobs are still timing-out. Did you run the `recon_test_pack` locally? This is likely the "infinite recursion" in `operator==` though, so it might be fixed now. let's see...

src/recon_buildblock/BinNormalisationWithCalibration.cxx
```diff
@@ -0,0 +1,91 @@
...
+{
+  return base_type::post_processing();;
+}
+
+
+BinNormalisationWithCalibration::
+BinNormalisationWithCalibration()
+{
+  set_defaults();
+  this->calibration_factor=1;
```
KrisThielemans, line 73:
```suggestion
```
see above

src/recon_buildblock/BinNormalisationWithCalibration.cxx
```diff
@@ -0,0 +1,91 @@
...
+//#include <boost/format.hpp>
+
+START_NAMESPACE_STIR
+
+void 
+BinNormalisationWithCalibration::set_defaults()
+{
+  base_type::set_defaults();/*
+  this->calibration_factor = 1;
+  this->branching_ratio=1;*/
```
KrisThielemans, line 49:
```suggestion
  base_type::set_defaults();
  this->calibration_factor = 1;
 /* this->branching_ratio=1;*/
```
best to let `set_defaults()` set all of them. Otherwise somebody might be calling it, expecting it to reset all values, but it wouldn't.

src/include/stir/recon_buildblock/BinNormalisationWithCalibration.h
```diff
@@ -0,0 +1,68 @@
...
+  virtual bool post_processing();
+
+   float get_calib_decay_branching_ratio_factor(const Bin&) const; // TODO find a better name
+   // needs to be implemented by derived class
+   virtual float get_uncalibrated_bin_efficiency(const Bin&, const double start_time, const double end_time) const  = 0;
+   float get_bin_efficiency(const Bin& bin, const double start_time, const double end_time)
+    { return this->get_uncalibrated_bin_efficiency(bin, start_time, end_time)/get_calib_decay_branching_ratio_factor(bin); }
+
+   void set_calibration_factor(const float);
+   void set_radionuclide(const float);
```
KrisThielemans, line 56: `float`?

KrisThielemans (changes requested):

src/include/stir/recon_buildblock/BinNormalisationWithCalibration.h
danieldeidda: oops! my bad

KrisThielemans:
I've cancelled most Travis and Appveyor jobs, to counter global warming...

KrisThielemans:
travis says that `recon_test_pack/run_simulation.sh` fails.

danieldeidda:
> 
> 
> travis says that `recon_test_pack/run_simulation.sh` fails.

let me have a look

KrisThielemans:
I guess should let it print `parameter_info` if both if this happens. Would make it easier to diagnose.

danieldeidda:
They appear to be the same 👍 
Set_up was with
Modality: PT
Patient position: unknown
Scan start time: 0
Time frame start - end (duration), all in secs: 0 - 0 (0)
number of energy windows:=1
energy window lower level[1] := -1
energy window upper level[1] := -1

Called with
Modality: PT
Patient position: unknown
Scan start time: 0
Time frame start - end (duration), all in secs: 0 - 0 (0)
number of energy windows:=1
energy window lower level[1] := -1
energy window upper level[1] := -1

danieldeidda:
It seems that they appear different although they are not. I checked the value of calibration_factor and that is different. However, if I remove that the check still gives false.

KrisThielemans:
not much you can do then checking all of the conditions... Easiest in a debugger, but always painful

src/include/stir/PatientPosition.h
```diff
@@ -79,6 +79,12 @@ class PatientPosition
     }
 
   explicit PatientPosition(PositionValue position);
+ 
+ bool operator == (const PatientPosition &p1) const { 
+     return  this->get_orientation()==p1.get_orientation() &&
+             this->get_position()==p1.get_position() &&
+             this->get_position_as_string()==p1.get_position_as_string() &&
```
KrisThielemans, line 86: I think we don't need this. If `get_orientation` and `get_position` are ok, the string will be ok

src/include/stir/ExamInfo.h
```diff
@@ -116,10 +118,29 @@ public :
...
+      if(this->low_energy_thres!=p1.low_energy_thres)
+          std::cout<<this->low_energy_thres-p1.low_energy_thres<<std::endl;
+      if(this->calibration_factor!=p1.calibration_factor)
+          std::cout<<this->calibration_factor-p1.calibration_factor<<std::endl;
+      if(this->start_time_in_secs_since_1970!=p1.start_time_in_secs_since_1970)
+          std::cout<<this->start_time_in_secs_since_1970<<" "<<p1.start_time_in_secs_since_1970<<std::endl;
+      if(this->get_radionuclide()!=p1.get_radionuclide())
+          std::cout<<this->get_radionuclide()<<" "<<p1.get_radionuclide()<<std::endl;
+      if(!(this->get_time_frame_definitions()==p1.get_time_frame_definitions()))
+          error("frames");
```
KrisThielemans, line 138: will have to remove this. could put it as an `info` with high verbosity

KrisThielemans:
still need to remove `_calibration_factor` member  from `DynamicDiscretisedDensity` 
https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/include/stir/DynamicDiscretisedDensity.h#L236
as it overlaps with the new one in `ExamInfo`. I think we can keep `DynamicDiscretisedDensity::set/get_calibration_factor`` and just forward to `ExamInfo`.

KrisThielemans:
later, will have to move
https://github.com/UCL/STIR/blob/bc14dea33cd30701416a20d126d479599c623320/src/IO/ECAT7DynamicDiscretisedDensityInputFileFormat.cxx#L99
higher in the ECAT7 code.

KrisThielemans:
Also, remove
https://github.com/UCL/STIR/blob/40d70e5c9ea7e8efc95b3999422756858c27f084/src/include/stir/IO/MultiDynamicDiscretisedDensityInputFileFormat.h#L113

src/include/stir/recon_buildblock/BinNormalisationWithCalibration.h
```diff
@@ -0,0 +1,68 @@
...
+#include "stir/RegisteredParsingObject.h"
+#include "stir/Bin.h"
+#include "stir/shared_ptr.h"
+#include "stir/recon_buildblock/BinNormalisation.h"
+#include "stir/decay_correction_factor.h"
+
+START_NAMESPACE_STIR
+
+/*!
+  \ingroup normalisation
```
KrisThielemans, line 31: some documentation needed. say that when this class is used, normalisation data will also be calibrated. This means that a reconstructed image will have values that are calibrated as well.

KrisThielemans:
Our plan was to set the calibration factor in the `ExamInfo` of the reconstructed image. For iterative recon, I think this needs to happen here][(https://github.com/UCL/STIR/blob/eaa79873e36b7d703841ef3c7f60d0fe5be9c717/src/recon_buildblock/IterativeReconstruction.cxx#L438). something like 
```c++
auto exam_info_sptr = std::make_share<ExamInfo>(input_data_sptr->get_exam_info());
if (auto norm_ptr = dynamic_cast<BinNormalisationWithCalibration const * const>(obf_func->get_normalisation_sptr().get()))
{
  exam_info_sptr->set_calibration_factor(norm_ptr->get_calibration_factor());
  // somehow tell the image that it's calibrated (do we have a way?)
}
else
{
  exam_info_sptr->set_calibration_factor(1.F);
  // somehow tell the image that it's not calibrated (do we have a way?)
}
target_stpr->set_exam_info(*exam_info_sptr);
```

lots of possibly gremlins along the way here...
   target_sptr->

AnderBiguri:
> still need to remove `_calibration_factor` member from `DynamicDiscretisedDensity`
> https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/include/stir/DynamicDiscretisedDensity.h#L236
> 
> 
> as it overlaps with the new one in `ExamInfo`. I think we can keep `` DynamicDiscretisedDensity::set/get_calibration_factor`` and just forward to ``ExamInfo`.
> I think best to leave it in `PatlakPlot` for the moment. @AnderBiguri do you want to comment on this?

Hum my memory is a bit fuzzy no on where this is used, but if I don't recall wrong, this is jut a value that is read from the par file? Maybe we just need to rename it, as yes, its a "calibration", but has not physical meaning other than an input from the user? maybe we can call it `_scale_factor` or something like that?

Maybe Im worng, Ive only used it with the simulated data and my memory is fuzzy

danieldeidda:
> 
> 
> still need to remove `_calibration_factor` member from `DynamicDiscretisedDensity`
> 
> https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/include/stir/DynamicDiscretisedDensity.h#L236
> 
> as it overlaps with the new one in `ExamInfo`. I think we can keep `` DynamicDiscretisedDensity::set/get_calibration_factor`` and just forward to ``ExamInfo`.
> 
> I think best to leave it in `PatlakPlot` for the moment. @AnderBiguri do you want to comment on this?

something like this?

void  DynamicDiscretisedDensity::
set_calibration_factor( float calibration_factor) 
{ this->exam_info_sptr->set_calibration_factor(calibration_factor); }

getting error with this though. "this' argument discards qualifiers"

danieldeidda:
> ```c++
> auto exam_info_sptr = std::make_share<ExamInfo>(input_data_sptr->get_exam_info());
> if (auto norm_ptr = dynamic_cast<BinNormalisationWithCalibration const * const>(obf_func->get_normalisation_sptr().get()))
> {
>   exam_info_sptr->set_calibration_factor(norm_ptr->get_calibration_factor());
>   // somehow tell the image that it's calibrated (do we have a way?)
> }
> else
> {
>   exam_info_sptr->set_calibration_factor(1.F);
>   // somehow tell the image that it's not calibrated (do we have a way?)
> }
> target_stpr->set_exam_info(*exam_info_sptr);
> ```

Do we want this to be printed in the header? Also get_normalisation_sptr() is not defined in GenerelaisedObjFunction

danieldeidda:
> Also get_normalisation_sptr() is not defined in GenerelaisedObjFunction

A solution could be to define a virtual get_binNormalisation() that is defined in the derived class but the would be a problem for objective function with gated data as it would stay pure. Same with Poisson...ListModataWithProjMatrix

KrisThielemans:
> Do we want this to be printed in the header?

not sure what you mean. sorry

src/include/stir/ExamInfo.h
KrisThielemans: resolved by still there?

src/include/stir/TimeFrameDefinitions.h
```diff
@@ -107,6 +107,8 @@ class TimeFrameDefinitions
 
     //! Set time frame
     void set_time_frame(const int frame_num, const double start, const double end);
+    
+    bool operator == (const TimeFrameDefinitions &t) const { return this->frame_times==t.frame_times;}
```
KrisThielemans, line 111: probably best to do a float/safe comparison here. It's going to trip us up otherwise. Best to put it in the .cxx then

src/include/stir/recon_buildblock/BinNormalisationFromECAT8.h
```diff
@@ -112,6 +113,7 @@ class BinNormalisationFromECAT8 :
   int num_transaxial_crystals_per_block;
   // TODO move to Scanner
   int num_axial_blocks_per_singles_unit;
+  shared_ptr<const ExamInfo> exam_info_sptr;
```
KrisThielemans, line 116: remove, as overlaps with member in base class

src/recon_buildblock/BinNormalisationFromECAT8.cxx
```diff
@@ -555,7 +556,7 @@ get_bin_efficiency(const Bin& bin, const double start_time, const double end_tim
   Scatter has a higher efficiency than trues when the axial angle is 0 (direct
   planes)
   As the axial angle increase the difference in efficiencies between trues and
-  scatter become closer
+  scatter become closer=
```
KrisThielemans, line 559:
```suggestion
  scatter become closer
```

src/recon_buildblock/BinNormalisationWithCalibration.cxx
```diff
@@ -0,0 +1,89 @@
...
+#include "stir/recon_buildblock/BinNormalisationWithCalibration.h"
+//#include "stir/recon_buildblock/TrivialDataSymmetriesForBins.h"
+//#include "stir/recon_buildblock/find_basic_vs_nums_in_subsets.h"
+//#include "stir/RelatedViewgrams.h"
+//#include "stir/Bin.h"
+//#include "stir/ProjData.h"
+//#include "stir/is_null_ptr.h"
+//#include "stir/Succeeded.h"
+//#include "stir/error.h"
+//#include <boost/format.hpp>
```
KrisThielemans, line 40:
```suggestion
```

src/recon_buildblock/BinNormalisationWithCalibration.cxx
```diff
@@ -0,0 +1,89 @@
...
+
+START_NAMESPACE_STIR
+
+void 
+BinNormalisationWithCalibration::set_defaults()
+{
+  base_type::set_defaults();
+  this->calibration_factor = 1;
```
KrisThielemans, line 48: have to move this to `set_defaults()`

KrisThielemans (changes requested):
some small ones, but I believe we should move `get_uncalibrated_bin_efficiency` to `BinNormalisationWithCalibration`, i.e. keep `get_bin_efficiency` pure virtual in `BinNormalisation`, and implement it in `BinNormalisationWithCalibration`.

This is more logical, and also means fewer changes in classes that do not derive from `BinNormalisationWithCalibration`

KrisThielemans:
> something like this?
> 
> ```
> void  DynamicDiscretisedDensity::
> set_calibration_factor(const  float calibration_factor) 
> { this->exam_info_sptr->set_calibration_factor(calibration_factor); }
> ```
> 
> getting error with this though. "this' argument discards qualifiers"

yup. it's a `const` so need to create a new object first. a bit of a pain, but the only way we can make this safe
```c++
{
  auto new_exam_info_sptr = std::make_shared<ExamInfo>(this->get_exam_info());
  this->exam_info_sptr->set_calibration_factor(calibration_factor); 
  this->set_exam_info(*exam_info_sptr);
}
```
it's a bit wasteful as `set_exam_info` will make another copy, but it looks like we don't have a `set_exam_info_sptr` in `ExamData` yet. (could add it...)

KrisThielemans:
> > I think best to leave calibration_factor in `PatlakPlot` for the moment. @AnderBiguri do you want to comment on this?
> 
> Hum my memory is a bit fuzzy  on where this is used, but if I don't recall wrong, this is jut a value that is read from the par file? Maybe we just need to rename it, as yes, its a "calibration", but has not physical meaning other than an input from the user? maybe we can call it `_scale_factor` or something like that?

I believe it was introduced as a cludge, because we didn't have calibrated images then. Presumably it should become obsolete. However, let's not do this in this PR and just leave `PatlakPlot` alone until we understand that better

KrisThielemans:
> > Also get_normalisation_sptr() is not defined in GenerelaisedObjFunction
> 
> A solution could be to define a virtual get_binNormalisation() that is defined in the derived class but the would be a problem for objective function with gated data as it would stay pure. Same with Poisson...ListModataWithProjMatrix

hmmm. This is becoming messy. The problem of course is that both `GeneralisedObjectiveFunction` and `IterativeReconstruction` are very general, and we somehow want to know something about the actual acquisition model, and data. For instance, it some point we'd want to be able to use STIR for Lucy-Richardson deconvolution. what would calibration mean? (Arguably, even the presence of a calibration factor in `ExamInfo` is too specific (wouldn't mean a lot for MRI...), but let's leave that.)

I have the feeling that this means we're trying to put it in the wrong place. If we look at [IterativeReconstruction::get_initial_data_ptr](https://github.com/UCL/STIR/blob/eaa79873e36b7d703841ef3c7f60d0fe5be9c717/src/recon_buildblock/IterativeReconstruction.cxx#L375), it appears the problem would be solved for un-initialised reconstruction if we let `GeneralisedObjectiveFunction::construct_target_ptr` handle this. (it can, as it's a `virtual` function), but not when we read the initial image from file, nor for  [`IterativeReconstruction::reconstruct(target)`](https://github.com/UCL/STIR/blob/eaa79873e36b7d703841ef3c7f60d0fe5be9c717/src/recon_buildblock/IterativeReconstruction.cxx#L427). In fact, that currently probably has a problem anyway as `ExamInfo` of the `target` isn't set from the input_data.

The only way around this that I see is that we introduce a new 
```c++
virtual shared_ptr<ExamInfo> GeneralisedObjectiveFunction::get_exam_info_for_target() const
{
   return this->get_input_data()->get_exam_info_sptr();
}
```
(and probably call it in its `construct_target_ptr`). Then we can overload that for objective functions that have `BinNormalisation`. Finally we can call that in `IterativeReconstruction::reconstruct(target)`

What do you think?

danieldeidda:
> 
> 
> > Do we want this to be printed in the header?
> 
> not sure what you mean. sorry

I mean do we want the calibration  value saved in the image header?

KrisThielemans:
> I mean do we want the calibration value saved in the image header?

ah, you mean the Interfile header. yes, I guess so. otherwise setting it in the object won't have a lot of real-life consequences...

danieldeidda:
> (and probably call it in its `construct_target_ptr`).

Mh I am a bit confused. This is pure virtual at the moment do you mean something like this:
```
virtual TargetT *
    construct_target_ptr() const
{
this->exam_info_sptr->this->get_exam_info_for_target()
}
although there is no exam_info_sptr member for GeneralisedObjectiveFunction. Also, I assume we need to return something of the type of TargetT
```

KrisThielemans:
> This is pure virtual at the moment do you mean something like this:
> 
> ```
> virtual TargetT *
>     construct_target_ptr() const
> {
> this->exam_info_sptr->this->get_exam_info_for_target()
> }
> ```
> 
> although there is no exam_info_sptr member for GeneralisedObjectiveFunction. Also, I assume we need to return something of the type of TargetT

ah well, if it's pure virtual, we cannot put it there. I guess we'll have to call it in the derived classes then. (I obviously meant to call `this->get_exam_info_for_target()`, and now think we should call that function `get_exam_info_sptr_for_target` to be consistent with other namings.

danieldeidda:
Sorry still slightly confused:
 if the following doesn't work because get_bin_normalisation() is not defined:
`(auto norm_ptr = dynamic_cast<BinNormalisationWithCalibration const * const>(get_objective_function().get_normalisation_sptr().get()`

how can we say whether the image is calibrated or not? We need to put this where we have access to binNormalisation. The objective functions?

KrisThielemans:
yes, that's where it has to be, so for instance we'd have
```c++
virtual shared_ptr<ExamInfo> GeneralisedObjectiveFunction::get_exam_info_sptr_for_target() const
{
   return this->get_input_data()->get_exam_info_sptr();
}
virtual shared_ptr<ExamInfo> PoissonLogLikelihoodBlaBla::get_exam_info_sptr_for_target() const
{
   auto exam_info_sptr = this->get_exam_info_sptr_for_target();
   if ((auto norm_ptr = dynamic_cast<BinNormalisationWithCalibration const * const>(this->get_normalisation_sptr().get())
     {
        ...
     }
   return exam_info_sptr;
}
```

danieldeidda:
There is something wrong  with 
`exam_info_sptr->set_calibration_factor()`

the function get_exam_info_sptr_for_target() is const and it expects set_calibration_factor() to be const. However a function that modifies a class member cannot be const

KrisThielemans:
not sure what you mean. In the above, `get_exam_info_sptr_for_target` can be `const` as it does not modify anything from its object (`*this`), but it creates a new `exam_info_sptr`. no?

danieldeidda:
> 
> 
> not sure what you mean. In the above, `get_exam_info_sptr_for_target` can be `const` as it does not modify anything from its object (`*this`), but it creates a new `exam_info_sptr`. no?

yes but then when I call set_calibration_factor(const float) in the following:

`auto exam_info_sptr = this->get_exam_info_sptr_for_target();
     if (auto norm_ptr = dynamic_cast<BinNormalisationWithCalibration const * const>(get_normalisation_sptr().get()))
     {
       exam_info_sptr->set_calibration_factor(norm_ptr->get_calibration_factor());
       // somehow tell the image that it's calibrated (do we have a way?)
     }`
I get 

`/home/nmdicom-recon/CalibNorm/STIR/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx:410: error: passing ‘std::__shared_ptr_access<const stir::ExamInfo, (__gnu_cxx::_Lock_policy)2, false, false>::element_type {aka const stir::ExamInfo}’ as ‘this’ argument discards qualifiers [-fpermissive]
        exam_info_sptr->set_calibration_factor(norm_ptr->get_calibration_factor());
        ^~~~~~~~~~~~~~`
which makes me think that it wants set_calibration_factor() to be const

KrisThielemans:
I don't understand. Is there an `exam_info_sptr` in the class itself? That could create a conflict (although it shouldn't). Does your `GeneralisedObjectiveFunction::get_exam_info_sptr_for_target()` return a `shared_ptr<const ExamInfo`? It shouldn't. In fact, now that I think about it, it should return a `std::unique_ptr<ExamInfo>` to clearly indicate to the user that it's a new object (not a shared one).

If this doesn't help, I guess you could just push your current code.

danieldeidda:
> 
> 
> I don't understand. Is there an `exam_info_sptr` in the class itself? That could create a conflict (although it shouldn't). Does your `GeneralisedObjectiveFunction::get_exam_info_sptr_for_target()` return a `shared_ptr<const ExamInfo`? It shouldn't. In fact, now that I think about it, it should return a `std::unique_ptr<ExamInfo>` to clearly indicate to the user that it's a new object (not a shared one).
> 
> If this doesn't help, I guess you could just push your current code.

yes that was the solution. I assume we can't use make_unique yet so I used new to make the get_exam_info_sptr() unique

KrisThielemans:
> I assume we can't use make_unique yet so I used new to make the get_exam_info_sptr() unique

indeed. great.

src/include/stir/TimeFrameDefinitions.h
danieldeidda: you mean like a comparison of each element of the vector?

danieldeidda:
Mh Travis is giving error: 

/home/travis/build/danieldeidda/STIR/build/src/swig/stirPYTHON_wrap.cxx:92324:52: error: call to deleted constructor of 'std::unique_ptr<stir::ExamInfo>'

  resultobj = SWIG_Python_NewPointerObj(self, (new std::unique_ptr< stir::ExamInfo >(static_cast< const std::unique_ptr< stir::ExamInfo >& >(result))), swig_types[61], 0x1 | 0);

                                                   ^                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

/usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/bits/unique_ptr.h:397:7: note: 'unique_ptr' has been explicitly marked deleted here

      unique_ptr(const unique_ptr&) = delete;

I am not totally sure why swig_python is having a problem with unique_ptr

danieldeidda:
> 
> 
> Mh Travis is giving error:
> 
> ```
> /home/travis/build/danieldeidda/STIR/build/src/swig/stirPYTHON_wrap.cxx:92324:52: error: call to deleted constructor of std::unique_ptr<stir::ExamInfo>
> 
>   resultobj = SWIG_Python_NewPointerObj(self, (new std::unique_ptr< stir::ExamInfo >(static_cast< const std::unique_ptr< stir::ExamInfo >& >(result))), swig_types[61], 0x1 | 0);
> 
> /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/bits/unique_ptr.h:397:7: note: 'unique_ptr' has been explicitly marked deleted here
> 
>  unique_ptr(const unique_ptr&) = delete;
> ```
> 
> I am not totally sure why swig_python is having a problem with unique_ptr

it seems that unique_ptr goes out of scope if defined as const.  WOULD Making "get_exam_info_sptr_for_target() " non-const solve the issue?

KrisThielemans:
> I  am not totally sure why swig_python is having a problem with unique_ptr

ah. SWIG doesn't really support `std::unique_ptr`. Somehow we've always managed to avoid problems with it, but maybe you're the first to hit it... The wrapped code looks pretty weird though. Which function is this wrapping?

There is a SWIG PR https://github.com/swig/swig/pull/1722 that adds more support to SWIG, but the SWIG maintainer doesn't like it.

danieldeidda:
> 
> 
> > I  am not totally sure why swig_python is having a problem with unique_ptr
> 
> ah. SWIG doesn't really support `std::unique_ptr`. Somehow we've always managed to avoid problems with it, but maybe you're the first to hit it... The wrapped code looks pretty weird though. Which function is this wrapping?
> 
> There is a SWIG PR [swig/swig#1722](https://github.com/swig/swig/pull/1722) that adds more support to SWIG, but the SWIG maintainer doesn't like it.

the error is here:
wrap_GeneralisedObjectiveFunction3DFloat_get_exam_info_sptr_for_target()

and here:  SwigValueWrapper& operator=(const T& t) { SwigMovePointer tmp(new T(t)); pointer = tmp; return *this; } 

in the structure called swigmovepointer

KrisThielemans:
ok. it's wrapping
```c++
 virtual std::unique_ptr<ExamInfo>  get_exam_info_sptr_for_target()  const;
```

A (terrible) way to get around this problem is that put
```
%ignore *::get_exam_info_sptr_for_target
```
in our interface file ... Another is to let it return `shared_ptr` as we had initially, which I don't like too much. A third way is to include SWIG typemaps for `unique_ptr` in STIR. All options are ugly, but I vote for the first, a it's the easiest and gets us going on the C++ front (it's not a function that you'd probably need from Python anyway).

Note that name is wrong, as it's not returning a shared ptr anymore.

danieldeidda:
> %ignore *::get_exam_info_sptr_for_target

where should this go? the stirPython_wrap file is massive and for some reason it's taking a lot of memory to open it (very slow to go through it).

KrisThielemans:
> where should this go?

in the SWIG interface file. I guess just before [here](https://github.com/UCL/STIR/blob/b5ed62bfa581177e9f9d8fc355cf1317ee9e509f/src/swig/stir.i#L1535)

danieldeidda:
Reading the calibration factor from interfile header should be indipenndent by the scanner (for the images). So I can put it in MinimalInterfileHeader::MinimalInterfileHeader()?

KrisThielemans:
Possibly.  I think you can put the `calibration_factor` member at the very base, but parsing is different for Siemens. For instance, for their norm, the mMR has
```
%GLOBAL SCANNER CALIBRATION FACTOR:=
%scanner quantification factor (Bq*s/ECAT counts):=1.98459e+007
%calibration date (yyyy:mm:dd):=2017:08:09
%calibration time (hh:mm:ss GMT+00:00):=07:23:33
%cross calibration factor:=0.96
```
they don't have anything like this in the list/sino data. I guess they don't have Interfile images, except when using e7tools. I haven't checked what the headers are then.

We can also remove keys in derived classes, but it's a bit ugly. Better to add them where they belong (although to be avoided if it implies duplication).

KrisThielemans:
Scatter test is failing, see https://travis-ci.org/github/UCL/STIR/jobs/744895355#L4506
```
 Set_up was with
Modality: PT
Calibration Factor: 6.03034e-38
Radionuclide: 
Patient position: unknown
Scan start time: 0
Time frame start - end (duration), all in secs: 0 - 20 (20)
number of energy windows:=1
energy window lower level[1] := 350
energy window upper level[1] := 650
Called with
Modality: PT
Calibration Factor: 0
Radionuclide: 
Patient position: unknown
Scan start time: 0
Time frame start - end (duration), all in secs: 0 - 20 (20)
number of energy windows:=1
energy window lower level[1] := 350
energy window upper level[1] := 650
```
this seems to say that the calib-factor is not initialised. I guess we have to choose between defaulting it to 1, or defaulting it to -1 (i.e. telling us it isn't set). I'd prefer the latter. It'd imply though that when we use it, we should check if it's >0. It's probably not obvious what to do if it isn't set though. Throw an error? In any case, probably good to add `ExamInfo::is_calibrated` (or whatever naming convention we have).

KrisThielemans (changes requested):
- I guess the Interfile reading/writing should also be done for isotope and branching ratio?
- Still need to revert `BinNormalisation` to what it was before, and put `get_uncalibrated_bin_efficiency` in `BinNormalisationWithCalibration`.

KrisThielemans:
I have had another look. I probably didn't express myself clearly before. sorry.

I thought we were going for the 2nd approach discussed above https://github.com/UCL/STIR/pull/672#pullrequestreview-477360752. I'd prefer to have no changes in `BinNormalisation` (i.e. that has pure virtual `get_bin_efficiency`), and then have `BinNormalisationWithCalibration` implemented that in terms of the pure virtual `get_uncalibrated_bin_efficiency`. I think this only needs changes in these 2 classes, and can be used to some changes elsewhere.

https://github.com/UCL/STIR/pull/672#issuecomment-707216880 asked what to do with `ChainedBinNormalisation` in this implementation. At present, we check in places if it's a `BinNormalisationWithCalibration` to see if there's a calibration factor. That would fail unless we derive `ChainedBinNormalisation` `BinNormalisationWithCalibration` (and let it get the calibration factor from one of the 2 members, and throw an error if both have one I guess). That looks a bit dodgy to me, but I have no better solution.

In addition, I think at the moment, the class will not do anything at all, as the `BinNormalisation::apply` was changed to use `get_uncalibrated_bin_efficiency`, which will give the previous values. Instead, it should still use `get_bin_efficiency`.

danieldeidda:
> 
> 
> Scatter test is failing, see https://travis-ci.org/github/UCL/STIR/jobs/744895355#L4506
> 
> ```
>  Set_up was with
> Modality: PT
> Calibration Factor: 6.03034e-38
> Radionuclide: 
> Patient position: unknown
> Scan start time: 0
> Time frame start - end (duration), all in secs: 0 - 20 (20)
> number of energy windows:=1
> energy window lower level[1] := 350
> energy window upper level[1] := 650
> Called with
> Modality: PT
> Calibration Factor: 0
> Radionuclide: 
> Patient position: unknown
> Scan start time: 0
> Time frame start - end (duration), all in secs: 0 - 20 (20)
> number of energy windows:=1
> energy window lower level[1] := 350
> energy window upper level[1] := 650
> ```
> 
> this seems to say that the calib-factor is not initialised. I guess we have to choose between defaulting it to 1, or defaulting it to -1 (i.e. telling us it isn't set). I'd prefer the latter. It'd imply though that when we use it, we should check if it's >0. It's probably not obvious what to do if it isn't set though. Throw an error? In any case, probably good to add `ExamInfo::is_calibrated` (or whatever naming convention we have).

this reminds me of the issue with calculateAttenuationCoefficients. Is it possible that the information is being taken from the image instead of the data and compare with the exam info of the projdata?

src/include/stir/ExamInfo.h
```diff
@@ -57,18 +57,23 @@ public :
 
   ExamInfo()
     : start_time_in_secs_since_1970(0.),
+
+    calibration_factor(1.F),
```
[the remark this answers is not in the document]
danieldeidda: > 
> 
>     * I guess the Interfile reading/writing should also be done for isotope and branching ratio?
> 

isotope yes, branching ratio i feel like should be extracted from the radionuclide "database"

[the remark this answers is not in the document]
danieldeidda: although I can see that siemens interfile has a branching factor key word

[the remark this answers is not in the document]
KrisThielemans: I agree with you though that looking this up is a better strategy. Obviously, it implies our database would be complete, and have a fixed name for isotopes etc, but let's just assume that for now.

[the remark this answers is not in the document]
danieldeidda: shall we use "isotope name" as keyword. Would this depend on the scanner? or is it an interfile standard?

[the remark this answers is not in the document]
KrisThielemans: amazingly, it seems that Interfile doesn't have isotope info. DICOM however has it standardised of course. For PET: [radiopharmaceutical info](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.9.2.html) and [isotope info](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_4020.html)

I guess we stick to `isotope name`. Might be worth checking with Harry Marquis and Dale Bailey what they do.

danieldeidda:
> [#672 (comment)](https://github.com/UCL/STIR/pull/672#issuecomment-707216880) asked what to do with `ChainedBinNormalisation` in this implementation. At present, we check in places if it's a `BinNormalisationWithCalibration` to see if there's a calibration factor. That would fail unless we derive `ChainedBinNormalisation` `BinNormalisationWithCalibration` (and let it get the calibration factor from one of the 2 members, and throw an error if both have one I guess). That looks a bit dodgy to me, but I have no better solution.
> 

where would you do this check? which function? something like the following?

```
if (apply_first.get_calibration_factor()>0.F & apply_first.get_calibration_factor()>0.F)
error()
```

danieldeidda:
> 
> 
> > [#672 (comment)](https://github.com/UCL/STIR/pull/672#issuecomment-707216880) asked what to do with `ChainedBinNormalisation` in this implementation. At present, we check in places if it's a `BinNormalisationWithCalibration` to see if there's a calibration factor. That would fail unless we derive `ChainedBinNormalisation` `BinNormalisationWithCalibration` (and let it get the calibration factor from one of the 2 members, and throw an error if both have one I guess). That looks a bit dodgy to me, but I have no better solution.
> 
> where would you do this check? which function? something like the following?
> 
> ```
> if (apply_first.get_calibration_factor()>0.F & apply_first.get_calibration_factor()>0.F)
> error()
> ```

Also, if chained... is derived from BinNorm..Calibration then we need to define get_uncalibrated_bin_eff...

danieldeidda:
> 
> 
> > > [#672 (comment)](https://github.com/UCL/STIR/pull/672#issuecomment-707216880) asked what to do with `ChainedBinNormalisation` in this implementation. At present, we check in places if it's a `BinNormalisationWithCalibration` to see if there's a calibration factor. That would fail unless we derive `ChainedBinNormalisation` `BinNormalisationWithCalibration` (and let it get the calibration factor from one of the 2 members, and throw an error if both have one I guess). That looks a bit dodgy to me, but I have no better solution.
> > 
> > 
> > where would you do this check? which function? something like the following?
> > ```
> > if (apply_first.get_calibration_factor()>0.F & apply_first.get_calibration_factor()>0.F)
> > error()
> > ```
> 
> Also, if chained... is derived from BinNorm..Calibration then we need to define get_uncalibrated_bin_eff...

I also wondere whether the current shared_ptr<BinNormalisation> apply_first, apply_second; should be pointers of Bin..WithCalibration

danieldeidda:
> I also wonder whether the current shared_ptr apply_first, apply_second; should be pointers of Bin..WithCalibration

As a result every class calling ChainedBinNormalisation(binNorm1, binNorm2) will require to define binorm1/2 as Bin...WithCalibration

KrisThielemans (commented):
I'm wondering what will happen with the ECAT7/ECAT8/GERDF versions after this PR. They claim to be calibrated, but the calibration factor is not yet initialised. I'm afraid that this will therefore give garbage reconstructions. 3 alternatives:
- set them to 1 for the moment. 
- set them to their proper value (I guess easy enough for ECAT8 after the info found, a bit of work for ECAT7 but not hard ([these lines](https://github.com/UCL/STIR/blob/29f0027429c0ba7f42a8e877eeff3f957768f5b8/src/buildblock/DynamicDiscretisedDensity.cxx#L168) imply it sits in the ECAT7 main header, but for GERDF a lot of work as it's not stored in the RDF)
- comment out your changes for calibration in those files.

Obviously, you could mix these approaches. what do you think?

danieldeidda:
> 
> 
> I'm wondering what will happen with the ECAT7/ECAT8/GERDF versions after this PR. They claim to be calibrated, but the calibration factor is not yet initialised. I'm afraid that this will therefore give garbage reconstructions. 3 alternatives:
> 
>     * set them to 1 for the moment.
> 
>     * set them to their proper value (I guess easy enough for ECAT8 after the info found, a bit of work for ECAT7 but not hard ([these lines](https://github.com/UCL/STIR/blob/29f0027429c0ba7f42a8e877eeff3f957768f5b8/src/buildblock/DynamicDiscretisedDensity.cxx#L168) imply it sits in the ECAT7 main header, but for GERDF a lot of work as it's not stored in the RDF)
> 
>     * comment out your changes for calibration in those files.
> 
> 
> Obviously, you could mix these approaches. what do you think?

so it seems the easiest is to use the actual value for ECAT8-7 and set to 1 GERDF

src/IO/interfile.cxx
```diff
@@ -594,6 +594,13 @@ write_basic_interfile_image_header(const string& header_file_name,
      ? "LITTLEENDIAN"
      : "BIGENDIAN")
 		<< endl;
+  
+  if (exam_info.get_calibration_factor()>0.F)
+  output_header << "calibration factor := "  
+                <<exam_info.get_calibration_factor() << endl;
+  
+  output_header << "isotope name := "
```
KrisThielemans, line 602: let's put this in an `if (!exam_info.get_radionuclide().empty())`

src/include/stir/IO/MultiDynamicDiscretisedDensityInputFileFormat.h
```diff
@@ -110,7 +110,6 @@ public InputFileFormat<DynamicDiscretisedDensity>
         dyn_disc_den_ptr->set_density(*t,i);
     }
     // Hard wire some stuff for now (TODO?)
-    dyn_disc_den_ptr->set_calibration_factor(1.);
     dyn_disc_den_ptr->set_if_decay_corrected(1.);
     dyn_disc_den_ptr->set_isotope_halflife(6586.2F);
```
KrisThielemans, line 114: should probably check on isotope name, could just throw an error if not F18

src/include/stir/TimeFrameDefinitions.h
KrisThielemans: yes, with some accounting for rounding error.

src/include/stir/recon_buildblock/BinNormalisationWithCalibration.h
```diff
@@ -0,0 +1,79 @@
...
+  // parsing stuff
+  virtual void set_defaults();
+  virtual void initialise_keymap();
+  virtual bool post_processing();
+
+
+private:
+  // provide facility to switch off things?
+  //  need to be added to the parsing keywords
+  bool use_calibration_factor; // default to true
```
KrisThielemans, line 70: jsut comment out for now

KrisThielemans (changes requested):

src/include/stir/TimeFrameDefinitions.h
danieldeidda: > 
> 
> yes, with some accounting for rounding error.

I was thinking something like that:

`bool TimeFrameDefinitions::operator == (const TimeFrameDefinitions &t) const {
    bool is_identical=0;
    
    for (int frame=0;frame<frame_times.size(); frame++){
        
        is_identical=((frame_times.at(frame).first == t.frame_times.at(frame).first) &&
                      (frame_times.at(frame).second == t.frame_times.at(frame).second));
    }
    return is_identical;
}

however I guess this does not allow  for rounding errors. Peraphs I should compare the differences with and arbitrary small number, like 10^-6?

danieldeidda:
> 
> 
> Possibly. I think you can put the `calibration_factor` member at the very base, but parsing is different for Siemens. For instance, for their norm, the mMR has
> 
> ```
> %GLOBAL SCANNER CALIBRATION FACTOR:=
> %scanner quantification factor (Bq*s/ECAT counts):=1.98459e+007
> %calibration date (yyyy:mm:dd):=2017:08:09
> %calibration time (hh:mm:ss GMT+00:00):=07:23:33
> %cross calibration factor:=0.96
> ```
> 
> they don't have anything like this in the list/sino data. I guess they don't have Interfile images, except when using e7tools. I haven't checked what the headers are then.
> 
> We can also remove keys in derived classes, but it's a bit ugly. Better to add them where they belong (although to be avoided if it implies duplication).

So this is in the normalisation header?  it seems that we need to multiply for the quantification factor. Don't we?

KrisThielemans:
> So this is in the normalisation header?

yes

> it seems that we need to multiply for the quantification factor. Don't we?

It seems to me that indeed, their "scanner quantification factor" is our "calibration factor".  Of course, there is no guarantee that their "ECAT counts" are the same as our "STIR counts". In fact, it used to be that this was only the case for zoom=1 (as long as we took duration/decay into account of course). This is something that we will need to check for different Siemens scanners, but let's leave that for a separate PR.

danieldeidda:
> 
> 
> > So this is in the normalisation header?
> 
> yes
> 
> > it seems that we need to multiply for the quantification factor. Don't we?
> 
> It seems to me that indeed, their "scanner quantification factor" is our "calibration factor". Of course, there is no guarantee that their "ECAT counts" are the same as our "STIR counts". In fact, it used to be that this was only the case for zoom=1 (as long as we took duration/decay into account of course). This is something that we will need to check for different Siemens scanners, but let's leave that for a separate PR.

I am thinking we need to read the cross calibration factor as well and set our calibration factor to quantification*cross_calibration. In the data I have cross calibration is 1, but in the example you put in one of your comments was different

KrisThielemans:
>  I am thinking we need to read the cross calibration factor as well and set our calibration factor to quantification*cross_calibration


I feel that their header implies that we don't need the `cross_calibration` (which should have to do with well-counter to scanner), but there is essentially no way of knowing this without either asking Siemens, or running e7tools with different headers and seeing what it does.

Given all this uncertainty, I suggest we don't include use the factor yet (i.e. return 1), but just read it, and create an issue for it. Otherwise, people's images are now going to change, and with the next PR, they will change again.

All this is of course the reason I originally suggested to postpone putting the calibration factor in for `ECAT8` etc  😉

danieldeidda:
> 
> 
> > I am thinking we need to read the cross calibration factor as well and set our calibration factor to quantification*cross_calibration
> 
> I feel that their header implies that we don't need the `cross_calibration` (which should have to do with well-counter to scanner), but there is essentially no way of knowing this without either asking Siemens, or running e7tools with different headers and seeing what it does.
> 
> Given all this uncertainty, I suggest we don't include use the factor yet (i.e. return 1), but just read it, and create an issue for it. Otherwise, people's images are now going to change, and with the next PR, they will change again.
> 
> All this is of course the reason I originally suggested to postpone putting the calibration factor in for `ECAT8` etc 😉

seems the same for ECAT7:

```
// now fill in what we can
  mhead.calibration_factor = 1.F;
  mhead.well_counter_factor=1.F;
```
setting it to 1

danieldeidda:
is this going to release 4.1?

KrisThielemans:
No. As it brings backwards compatibility on the code level

danieldeidda:
I am thinking that BinNormFromProjData also may need to be derived from BinNormWitCalib or else in case like the Mediso scanner where the normalisation comes from projData the factor will not be applied. I do not see any drawback of doing this but maybe I am wrong.

danieldeidda:
> 
> 
> No. As it brings backwards compatibility on the code level

OK. shall we add something in the userguide? I think we said that this should go togheter with the normalisation section.

KrisThielemans:
> shall we add something in the userguide? I think we said that this should go togheter with the normalisation section.

We do need to add something, but I'm not 100% sure where. We probably need a new section on image units, which right now is going to be a bit sketchy. 
- proportional to counts, proportionality factor fixed in terms of sum of along horizontal direction of voxel values (therefore scaling with voxel size))
- if normalisation supports calib factor, then will be taken into account (is it mult or divide?). value reported in header for file formats that support it (currently only Interfile)

KrisThielemans:
> I am thinking that BinNormFromProjData also may need to be derived from BinNormWitCalib or else in case like the Mediso scanner where the normalisation comes from projData the factor will not be applied. I do not see any drawback of doing this but maybe I am wrong.

I'm a bit reluctant to do that as it is also used by many to store ACFs. some remarks:

- how do we know which is which? I guess could default calib factor to -1?

- I would not multiply with it in the bin-efficiency calculation (i.e. assume it has been done already). just use it for recording. this is how it would come out in the correct_projdata with an ECAT8 norm anyway.

- it seems to imply you then need to store the calib factor in the "norm" projection data (as where are we going to get it from otherwise)? I suppose this is possible as you could have it for all projdata to signify if they have been calibrated. However, this seems to lead us quite far with quite a lot of other work.

I don't think I'd do this in this PR therefore (or at all).

Any reason you cannot create a `BinNormalisation` class for your Mediso data?

danieldeidda:
> 
> 
> > I am thinking that BinNormFromProjData also may need to be derived from BinNormWitCalib or else in case like the Mediso scanner where the normalisation comes from projData the factor will not be applied. I do not see any drawback of doing this but maybe I am wrong.
> 
> I'm a bit reluctant to do that as it is also used by many to store ACFs. some remarks:
> 
>     * how do we know which is which? I guess could default calib factor to -1?
> 
>     * I would not multiply with it in the bin-efficiency calculation (i.e. assume it has been done already). just use it for recording. this is how it would come out in the correct_projdata with an ECAT8 norm anyway.
> 
>     * it seems to imply you then need to store the calib factor in the "norm" projection data (as where are we going to get it from otherwise)? I suppose this is possible as you could have it for all projdata to signify if they have been calibrated. However, this seems to lead us quite far with quite a lot of other work.
> 
> 
> I don't think I'd do this in this PR therefore (or at all).
> 
> Any reason you cannot create a `BinNormalisation` class for your Mediso data?

OK, no I suppose I could create a BinNormalisationFromMediso

danieldeidda:
> 
> 
> > shall we add something in the userguide? I think we said that this should go togheter with the normalisation section.
> 
> We do need to add something, but I'm not 100% sure where. We probably need a new section on image units, which right now is going to be a bit sketchy.
> 
>     * proportional to counts, proportionality factor fixed in terms of sum of along horizontal direction of voxel values (therefore scaling with voxel size))
> 
>     * if normalisation supports calib factor, then will be taken into account (is it mult or divide?). value reported in header for file formats that support it (currently only Interfile)

So, shall we do it in another PR?

KrisThielemans:
> shall we do it in another PR?

fine for me. Do you have something in the release notes now? Probably important for people to know this has happened. (could even say in the release notes that documentation is pending, and we edit that in the next PR).

danieldeidda:
> 
> 
> > shall we do it in another PR?
> 
> fine for me. Do you have something in the release notes now? Probably important for people to know this has happened. (could even say in the release notes that documentation is pending, and we edit that in the next PR).

I was thinking to add it but not sure in which version, 5.0?

KrisThielemans:
> I was thinking to add it but not sure in which version, 5.0?

yes please

danieldeidda:
I think this is ready to go

src/buildblock/TimeFrameDefinitions.cxx
```diff
@@ -275,4 +275,15 @@ set_time_frame(const int frame_num, const double start, const double end)
...
 
+bool TimeFrameDefinitions::operator == (const TimeFrameDefinitions &t) const {
+    bool is_identical=0;
+    
+    for (int frame=0;frame<frame_times.size(); frame++){
+        
+        is_identical=(abs(frame_times.at(frame).first - t.frame_times.at(frame).first) <= 10e-6) &&
+                      (abs(frame_times.at(frame).second - t.frame_times.at(frame).second) <= 10e-6);
+    }
+    return is_identical;
```
KrisThielemans, line 286:
```suggestion
    for (int frame=0;frame<frame_times.size(); frame++){
        
        const bool is_identical=(abs(frame_times.at(frame).first - t.frame_times.at(frame).first) <= 10e-5) &&
                      (abs(frame_times.at(frame).second - t.frame_times.at(frame).second) <= 10e-5);
        if (!is_identical)
          return false;
    }
    return true;
```
This was effectively only testing for the last frame. Now should be ok I believe.

Also relaxed the numerical error a bit.

KrisThielemans (changes requested):

danieldeidda (commented):
cool

PR #674, opened 2020-09-03
Modelling refresh

KrisThielemans:
various updates to the kinetic modelling code, including independence of ECAT7


KrisThielemans:
The `read_from_file` changes are rather surprising in this PR. They should probably be moved in a separate PR. I might do that later once it all works.

git history is therefore likely to change. Don't merge this anywhere else therefore!

KrisThielemans:
This all works now, including the (previously very outdated) test script.

KrisThielemans:
I'm not going to create another PR for the `read_from_file` change. Life is too short. I've created an issue for it though. #677

KrisThielemans:
Codacy "issues" are because it isn't smart enough and see that we increment `argv` and decrement `argc`

PR #519, opened 2020-04-26
adding tests for priors

KrisThielemans:
currently failing...

Issue #153, opened 2018-02-19
need test for priors


KrisThielemans:
we need an automatic test to see if the gradient/value calculation is consistent for the priors.  We currently only have one for the Poisson-loglikelihood.

PR #695, opened 2020-09-24
Add test priors 2

robbietuk:
This PR is to replace and expand upon that of #519. 

Addresses #153

- [x] add Logcosh #716 
- [ ] consider adding Hessian test
- [ ] consider adding surrogate test

### DIFF
--- a/src/recon_test/CMakeLists.txt
+++ b/src/recon_test/CMakeLists.txt
@@ -27,6 +27,7 @@ set(${dir_SIMPLE_TEST_EXE_SOURCES}
         test_FBP2D
         test_FBP3DRP
         test_OSMAPOSL
+        test_priors
 )
 
 
@@ -35,7 +36,6 @@ set(${dir_INVOLVED_TEST_EXE_SOURCES}
         bcktest
         recontest
         test_data_processor_projectors
-        test_priors
 )
 
 include(stir_test_exe_targets)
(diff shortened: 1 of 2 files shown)


robbietuk:
PLS will fail because of edge artefacts. Here is a difference image. PLS uses a uniform 1's anatomical image and the same randomised densitity as QP and RDP
image

robbietuk:
I believe PLS is failing at the (max) edge because of continue statements in the directional gradient computation https://github.com/UCL/STIR/blob/7b4c27470ad175a80f29e9aabe945b6831a4f159/src/recon_buildblock/PLSPrior.cxx#L318-L333
The solution to this is not obvious.

KrisThielemans:
@robbietuk, what about disabling the PLS test for now, adding logcosh and merging this PR?

robbietuk:
Yes, I think we also expand the testing for RDP and logcosh priors to have more than one evaluation with different `gamma`(RDP)/`scalar`(logcosh) values. We didn't spot the logcosh error regarding the value compuation when the input is large for a couple of weeks.

Ill get on this shortly, did you want it for release4.1?

robbietuk:
`test_priors` passes on my machine (waiting on Travis to confirm). This PR needs to be followed up with a further issue relating to additional tests for other methods but these should wait until the mess with the priors code has been cleaned up.

KrisThielemans (approved):

PR #696, opened 2020-09-24
Fix an issue with the relative difference prior

robbietuk:
The inclusion of the epsilon term in the denominator square

### DIFF
--- a/examples/samples/OSMAPOSL_RDP.par
+++ b/examples/samples/OSMAPOSL_RDP.par
@@ -1,0 +1,59 @@
+OSMAPOSLParameters :=
+; example file for using a relative difference prior (with One Step Late and subsets)
+; this is a minimal file. check other sample files and doc for more options
+
+objective function type:= PoissonLogLikelihoodWithLinearModelForMeanAndProjData
+PoissonLogLikelihoodWithLinearModelForMeanAndProjData Parameters:=
+
+input file := test.hs
+; if disabled, defaults to maximum segment number in the file
+maximum absolute segment number to process := 4
+; see User's Guide to see when you need this
+zero end planes of segment 0:= 1
+
+; if the next parameter is disabled, 
+; the sensitivity will be computed
+; use %d where you want the subset-number (a la printf)
+subset sensitivity filenames:= sens_%d.hv
+; specify additive projection data to handle randoms or so
+; see User's Guide for more info
+additive sinogram := 0
+
+
+; here comes the prior stuff
+
+
+prior type := Relative Difference Prior
+  Relative Difference Prior Parameters:=
+  penalisation factor := 1.0
+  ; next can be used to set weights explicitly. Needs to be a 3D array (of floats). value of only_2D is ignored
+  ; following example uses 2D 'nearest neighbour' penalty
+  ; weights:={{{0,1,0},{1,0,1},{0,1,0}}}
+  ; use next parameter to specify an image with penalisation factors (a la Fessler)
+  ; see class documentation for more info
+  ; kappa filename:=
+  ; gamma value is a RDP specific parameter and controls the edge preservation
+  ; gamma value := 
+  ; epsilon value - a small non-negitve float
+  ; epsilon value :=
+END Relative Difference Prior Parameters:=
+
+end PoissonLogLikelihoodWithLinearModelForMeanAndProjData Parameters:=
+
+initial estimate:= some_image
+; enable this when you read an initial estimate with negative data
+enforce initial positivity condition:=0
+
+output filename prefix := test_RDP
+number of subsets:= 12
+number of subiterations:= 24
+save estimates at subiteration intervals:= 12
+
+; This is the default verbosity, just putting it here as an example
+; lower value means less output, 0 being the minimum
+verbosity := 2
+
+; enable this for multiplicative form of OSMAPOSL (see User's Guide)
+;MAP model := multiplicative
+
+END :=
--- a/examples/samples/OSMAPOSL_RDP.par
+++ b/examples/samples/OSMAPOSL_RDP.par
@@ -32,9 +32,9 @@ prior type := Relative Difference Prior
   ; use next parameter to specify an image with penalisation factors (a la Fessler)
   ; see class documentation for more info
   ; kappa filename:=
-  ; gamma value is a RDP specific parameter and controls the edge preservation
+  ; gamma value is a RDP specific parameter and controls the edge preservation. Default value is 2
   ; gamma value := 
-  ; epsilon value - a small non-negitve float
+  ; epsilon value - a small non-negative float to smooth the prior at very small values. Default value is 0
   ; epsilon value :=
 END Relative Difference Prior Parameters:=
 
--- a/src/include/stir/recon_buildblock/RelativeDifferencePrior.h
+++ b/src/include/stir/recon_buildblock/RelativeDifferencePrior.h
@@ -60,7 +60,8 @@ START_NAMESPACE_STIR
   where \f$\lambda\f$ is the image and \f$r\f$ and \f$dr\f$ are indices and the sum
   is over the neighbourhood where the weights \f$w_{dr}\f$ are non-zero. \f$\gamma\f$ is
   a smoothing scalar term and the \f$\epsilon\f$ is a small non-negative value included to prevent division by zero.
-  For more details, see: <em> J. Nuyts, D. Bequ&eacute;, P. Dupont, and L. Mortelmans,
+  Please note that the RDP is only well defined for non-negative voxel values.
+  For more details, see: <em> J. Nuyts, D. Beque, P. Dupont, and L. Mortelmans,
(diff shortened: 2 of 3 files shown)


robbietuk:
From my notes, we have to consider the RDP with the addition of the epsilon term to prevent divide by zero.
image

AnderBiguri:
Hum, you just changed the location of the epsilon right? if its only there to avoid division by zero, then its value must be much smaller than the expected values for the rest of the numerator and denominator (otherwise it would have a numerical impact, and we just want to avoid division by zero). 

If that is the case, it does not really matter for it to be mathematically correct on its derivative. You just need to add epsilon to any division, and thats it. In theory, it will only have an impact when everything else is zero, so the error you are introducing by not being mathematically correct is neglectable (and if its not, then your epsilon is too big!)

Is that the case here? I am missing a but the context, so maybe I am talking out of place!

robbietuk:
Yes, the purpose is to prevent the divide by zeros and it also exists in the computation of the function https://github.com/UCL/STIR/blob/ef435cc8a50d716962da9b72a77e69ff1ee2aaed/src/recon_buildblock/RelativeDifferencePrior.cxx#L316-L319
In general I would agree with you as generally the function and gradient are slightly decoupled. However, I am trying to be mathmatically correct as I am using line searches to find the MAP solution and therefore discrepancies between the function and its gradient can lead to problems (which is what I think I am finding). 

I could consider discarding the epsilon and instead use `if x == 0 and y == 0: return 0`. This would let it be both mathmatically correct and handle divide by zeros.

AnderBiguri:
@robbietuk fair, fair. However, maybe you are just using an epsilon too large? As I said in the teams chat, if the value you are using as "epsilon" is large enough to have influence in your maths, either it is too large on value, or you have other serious issues on the math itself, as its too sensitive to numerical accuracy.

robbietuk:
I agree, I feel that epsilon is causing a large issue right now. I want to try and replace it with the if statement.

AnderBiguri:
Yes, that works too! But I guess my point is that if those microscopic epsilons are causing a problem, I don't expect your problems to go away...   If the epsilon you are imputing is indeed small enough to be on the "noise" fraction of your data, and that impacts the numerical method, then any other "noise" (i.e. floating point error) that may arise just naturally on your computations will cause the same error!

robbietuk:
Sorry, I dont think I phrased the above correctly. I think the epsilon I am using right now is too large for normalised data. That and the "decoupled" mathmatical function and gradient and there is a problem. https://github.com/UCL/STIR/blob/ef435cc8a50d716962da9b72a77e69ff1ee2aaed/src/recon_buildblock/RelativeDifferencePrior.cxx#L142

Yes there is an computational problem of using very small epsilons and I think we are both in agreement that moving away from it is the correct thing?

AnderBiguri:
No! That is not what I meant. 

I think there is absolutely no issue with using very small epsilons, if their purpose is only avoiding by zero. They just need to be marginally bigger than zero, but just that. so 1e-7 should also work. In fact, you want epsilon to be as small as possible, in such a way that it can be considered noise for your purposes. Yo only want it so when our "incomplete" numerical system of using floating point arithmetic fails because it does not know how to divide by zero (and give you Inf), it doesn't, and gives you a "practical" Inf, something very large.  Or, if the numerator is also zero, then for your division to give something that is essentially zero, but not, and avoid NaN. 

In some sense, it has nothing to do with math, `epsilon` is not there to do something mathematically smart, its just there to help us overcome a limitation of using computers. This is why I was surprised when you used it in the gradient, I don't think its necessary. You just want to replace your divisions of `a/b` by `a/(b+eps)`, not change your equations. 

So, in short, before changing anything and going crazy with equations, I suggest you just do `this->epsilon = 0.00000001;` and try again.

robbietuk:
Yes, as `epsilon > 0` tends to 0, the function and gradient behave like their non-`epsilon` counterparts (except when voxels are 0). The trouble is, what is the correct value of `epsilon` for the general case? 

Removing `epsilon`, adding a check on the two voxel values being 0 (then return 0) becomes equiviant to an optimal `epsion` implementation, and we dont have to answer that question.

KrisThielemans:
I strongly recommend making the gradient consistent "formula-wise" with the objective function. I think that should be the case for whatever value of epsilon you use.

I don't think checking with `if` is going to work. The division will likely explode in other cases as well.

We could check with Johan Nuyts et al. if you like

AnderBiguri:
@robbietuk 
>  The trouble is, what is the correct value of epsilon for the general case?
As close as zero without being zero. For any case. You just want a small enough perturbation to avoid division by zero. 
@KrisThielemans 
>I strongly recommend making the gradient consistent "formula-wise" with the objective function. 
While you are correct, and doing this will have no negative impact, in all cases in the past where I could do this but instead I just replaced divisions, there was no measurable difference with the simple "non-formula" way. In fact, one could argue that if there is a difference when using the formula, then your epsilon is too big in value. Conceptually you just want to introduce a value in the noise range of your computation, and by definition that should mean that it has no impact on your results. Its the same as adding doubles vs singles for all this. You are adding more numbers in the noise level of your method, thus it does not matter. 

In fact, I would argue that if you do the "proper formula way" vs the "just adding it in the division", and there is a measurable impact in your algorithm, then your epsilon is wrong. 


But of course, there should not be an issue with adding the proper way.

robbietuk:
Okay I did some numerical test on the RDP. The trouble with using an `epsilon` in the denominator is that the equation and its gradient are significantly sensitive to poorly selected values. Unlike a function, such as the QP `= (x - y)^2` , the RDP is  sensitive to differences between voxels relatvie to the voxel values. I guess the clue is in the name.

I wrote some python to investigate this. One set of functions (`RDP_e` and `grad_e` have an `epsilon = 1e-10`) and the formulee are equivilant to that of my first comment (https://github.com/UCL/STIR/pull/696#issuecomment-698439145).  

The second set of RDP tests of this use an `epsilon = 0.0` but have the addition of an if statement` if x == 0.0 and y == 0.0 and epsilon == 0.0: return 0.0` for both function and its gradient computation. This is only to prevent the evaluation of `0.0/0.0`, which I have assumed to equal 0.

`x` and `y` are a set of values given by the following python code `[0.0] + [math.e ** i for i in range(-30, 5)]` (i.e. `[0.0, 9e-14, 2e-13,..., 54]`) so voxel values both above and below the epsilon values are compared for this experiment. On the below images, the vertical and horizontal axis values are equivilant. 

I will compare these two sets of functions and show that the if statement version is equivilant to an optimally selected epsilon value of infintesimally small but positive value and therefore I believe we should use the if statements and do away with `epsilon`.


Function
--------
The `RDP` and `RDP_e` are visually comparable (`RDP` shown below). 
image
However subtracting one from the other (`RDP - RDP_e`) we can see the discrepancies, even if they are very small. 
image
As the `epsilon` value tend to 0 (e.g. `epsilon=1e-40`) then this difference image becomes uniform 0's.

Gradient
---------
The gradient is another matter entirely. The gradient of the RDP (`grad`) is given. 
image
Only once is the if statement True, at `x=0 and y=0`. As is clear, the RDP at low voxel values is sensitive. 

Now compare to `grad_e`, with the inclusion of `epsion=1e-10` and we see that the sensitivity at lower values (`<epsilon`) has been reduced as the gradient tends to 0. 
image
 
Lowering the `epsilon` value to `1e-40`, << the smallest non-zero `x` and `y`, leads to virtually equivilant behaviours between `grad` and `grad_e`.

robbietuk:
I guess then the answer to my problem is that the current epsilon **_is_** too large for normalised data and that is why I am seeing weird behaviour in my line searches. I have have not seen this before with the RDP as I did not use normalised data, thus the voxel values were of the order of magnetude `1e3`.

AnderBiguri:
Very nice analysis!

robbietuk:
Drafted the removal of epsilon, if we dont choose to use it. We can just revert the commit.

robbietuk:
I can't see any more issues, I am happy to merge.

robbietuk:
After discussion with @KrisThielemans, we decided to go for a comprimise. We will keep the `epsilon` but by default set it to 0. The `if` statements will handle the case when `epsilon=0`, `x=0`, and `y=0` by returning 0. This allows a user to set the value of epsilon.

robbietuk:
As discussed via email, it is not obvious how to extend the RPD allow for negitive number and a reliable concave function. In the presence of a non-negitiviety constraint, the RDP will now work as expected, even with `epsilon =0`. 

One idea is to incorporate an `assert(x>=0)` and `assert(y>=0)` onto every voxel evaluation. However, I don't know how well this would work in practice (it may be simplified by `assert(min(volume)>=0)`). Do we need to provide this hard boundary? Does STIR ever allow negitive voxel values in its reconstruction algoriothms. Should this assert really be a if statement that outputs a warning?

KrisThielemans (approved):
looks good. would be nice to add an example like \examples\samples\OSMAPOSL_QuadraticPrior.par. if you do, please do it with `[ci skip]` in the first commit line

PR #699, opened 2020-10-05
error check IF file 

AnderBiguri:
Just making this a bit more Ander-proof. 

Checks if first line of .if file is a positive int, errors otherwise.

### DIFF
--- a/src/include/stir/modelling/PlasmaData.inl
+++ b/src/include/stir/modelling/PlasmaData.inl
@@ -60,18 +60,21 @@ void  PlasmaData::read_plasma_data(const std::string input_string)
     std::string first_line;
     if (std::getline(data_stream, first_line))
     {
+      // replace leading/trailing whitespace 
+      first_line.erase(std::find_if(first_line.rbegin(), first_line.rend(), std::bind1st(std::not_equal_to<char>(), ' ')).base(), first_line.end());
+      first_line.erase(first_line.begin(), std::find_if(first_line.begin(), first_line.end(), std::bind1st(std::not_equal_to<char>(), ' ')));
       // now first, check if the first line is a single character. 
       // this is best done in C style, cleaner than iterating over chars
       char* p;
       long converted = strtol(first_line.c_str(), &p, 10);
       if (*p)
-        error("First line of .if file is not number of samples");
+        error("First line of input function file ("+ input_string + ") is not number of samples");
       else
         _sample_size=converted;
     }
     else
     {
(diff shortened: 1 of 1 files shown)


src/include/stir/modelling/PlasmaData.inl
```diff
@@ -55,8 +55,25 @@ void  PlasmaData::read_plasma_data(const std::string input_string)
...
+  { 
+    // Get the first line, which should be the number of samples
+    std::string first_line;
+    if (std::getline(data_stream, first_line))
+    {
+      // now first, check if the first line is a single character. 
+      // this is best done in C style, cleaner than iterating over chars
+      char* p;
+      long converted = strtol(first_line.c_str(), &p, 10);
+      if (*p)
```
KrisThielemans, line 67: would this not fail if there's trailing whitespace?

src/include/stir/modelling/PlasmaData.inl
```diff
@@ -55,8 +55,25 @@ void  PlasmaData::read_plasma_data(const std::string input_string)
...
+        error("First line of .if file is not number of samples");
```
KrisThielemans, line 68: could as well list filename, now that you're cleaning-it up...

src/include/stir/modelling/PlasmaData.inl
```diff
@@ -55,8 +55,25 @@ void  PlasmaData::read_plasma_data(const std::string input_string)
...
+      else
+        _sample_size=converted;
+    }
+    else
+    {
+      error("Input function file is empty");
```
KrisThielemans, line 74: and add input-filename here as well.

KrisThielemans (changes requested):

src/include/stir/modelling/PlasmaData.inl
AnderBiguri: Yes, I didn't want to add too much checks, but I can certainly improve it if you think its worth it.

AnderBiguri: Well, I'd argue that better to risk having a message saying `error("First line of .if file is not number of samples, check trailing whitespace");` than a segmentation fault in case you did not add the first number as integer. 

I'll add the check.

KrisThielemans:
looks fine, but you've used C++-11 `std::tostring` for `release_4`. Either you have to replace with `boost::format` stuff, or base on master`, which would be my preferred solution. We have to stop updating `release_4` and release it!

PR #708, opened 2020-10-08
SSRB can take an template sinogram

robbietuk:
The SSRB utility currently only allows for axial compression of needs to be an odd number (`ERROR: SSRB: num_segments_to_combine (2) needs to be odd`). However, this does not allow us to create `span-2` sinograms. This PR allows for the `--template` flag to be used on the command line instead of providing other the standard arguments.

- [x] Does this work? 
I can convert span-1 sinograms to span-2
- [x] Documentation and usage update
- [x] Cleanup the previous argument methodology. This worked before but it probably isnt very clean now. I think these two input methods should be different functions and handle the input arguments differently.

Additional based upon comments.

- [x] Add `static void print_usage_and_exit(const std::string& prog_name)`
- [x] Add doxygen header
- [x] Add line to `release_5.htm`
- [x] Rename `span` to `num_segments_to_combine`. Span indicates the input is `span-1`, which isn't required to be the case.
- [x] `--template` option should be first argument.

### DIFF
--- a/src/utilities/SSRB.cxx
+++ b/src/utilities/SSRB.cxx
@@ -109,7 +109,7 @@ static void print_usage_and_exit(const std::string& prog_name)
 
 void classic_SSRB(int argc, char **argv)
 {
-  // All this needs checking for standard usage
+  // Does the standard method of SSRB based upon numerical inputs
   int num_tangential_poss_to_trim = 0;
   if (argc>1 && strcmp(argv[1], "-t")==0)
   {
@@ -135,7 +135,7 @@ void classic_SSRB(int argc, char **argv)
 
 void template_based_SSRB(int argc, char **argv)
 {
-
+  // Does the template based SSRB whereby the target template is given as an argument
   shared_ptr<ProjData> template_projdata_ptr = ProjData::read_from_file(argv[2]);
   const string output_filename = argv[3];
   shared_ptr<ProjData> in_projdata_ptr = ProjData::read_from_file(argv[4]);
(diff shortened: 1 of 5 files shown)


src/utilities/SSRB.cxx
```diff
@@ -64,6 +64,7 @@
 #include "stir/SSRB.h"
 #include <string>
 #include <string.h>
+#include <stir/ProjDataInterfile.h>
```
KrisThielemans, line 67:
```suggestion
#include "stir/ProjDataInterfile.h"
```
it isn't a system file.

src/utilities/SSRB.cxx
```diff
@@ -96,19 +91,56 @@ int main(int argc, char **argv)
...
-  SSRB(output_filename,
-       *in_projdata_ptr,
-       span,
-       num_views_to_combine,
-       num_tangential_poss_to_trim,
-       do_norm,
-       max_segment_num_to_process
-       );
+
+  if (strcmp(argv[3], "--template")==0)
```
KrisThielemans, line 105: please follow same conventions as `zoom_image`, i.e. all options upfront, and hence `SSRB --template bla output input`.

Also, will need to check if `argc` is appropriate for this signature.

KrisThielemans (changes requested):
a few suggestions. Once we're happy, we need to update the Usage message (best to put it in a `static void print_usage_and_exit(const std::string& prog_name)` as in many other STIR utilities including `zoom_image`), doxygen header, and add something to `release_5.htm`.

By the way, I'd take the opportunity to rename `span` to `num_segments_to_combine`, as that's what it is (input doesn't have to be `span=1`)

robbietuk:
`main` was getting a little busy, there was little overlap between the new `template_based_SSRB` and `classical_SSRB`, so I split them into two functions to make it clearer. I am not set on the function names, feel free to suggest others. 

I have tested this on D690 sinograms, converting span-1 to span-2. Previously this would have been impossible as `num_segments_to_combine=2` would have errored.

KrisThielemans (changes requested):
tiny doc changes

robbietuk:
Ready for squash merge

Issue #707, opened 2020-10-08
STIR-python priors are GeneralisedPrior objects rather than child class priors


robbietuk:
OSMAPOSL reconstruction object is defined and setup in python as `self.recon`. This setup process has a penalty defined (e.g. QP). The call `self.recon.get_objective_function().get_prior()` returns
`<Swig Object of type 'stir::shared_ptr< stir::GeneralisedPrior< stir::DiscretisedDensity< 3,float > > > *' at 0x125b172f0>`. This GeneralisedPrior object does not contain any methods specific to the QP/RDP and therefore evaluations of things like `parabolic_surrogate_curvature()` (QP), or `get_gamma()`/`set_gamma()` (RDP) cannot be performed.

Can `get_prior()` return the child class prior?, e.g. QP or RDP.

KrisThielemans:
see #662 and #284 

feel like trying what's in https://github.com/UCL/STIR/issues/284#issuecomment-677635171?

robbietuk:
I think I will. Unfortunately I don't think it can be put off for much longer.

KrisThielemans:
could be very easy, but we'll only find out by trying. The SWIG list is quite helpful though. Let me know

robbietuk:
I shall do. Let me learn SWIG first 😄

KrisThielemans:
That could easily take up the rest if you PhD!

(edit src/swig/stir.i, add a single `%factory` line, see what happens...)

robbietuk:
Is there any way we could do something similar to https://github.com/UCL/STIR/blob/d1cd05cf53db8332924991c142cc42969472885f/src/swig/stir.i#L1620-L1626

The python call would be something like `stir.ToQP(self.recon.get_objective_function().get_prior())`, which would return class type QP, with all the methods avalible.

I am guessing this would require an individual stir method for each prior (e.g. `stir.ToRDP()` and `stir.ToPLS()`)and is not ideal for code clarity.

KrisThielemans:
this is easy enough. We probably only need it for `PriorWithParabolicSurrogate` at the moment?

robbietuk:
Ill give it a go for `PriorWithParabolicSurrogate` if not, Ill just do each prior individually 😄 I dont think think would be a perminent fix though. If it works, should we PR it to STIR master?

KrisThielemans:
for what we need, this should work.

Regarding a PR, depends if we can get the SWIG people to help out first or not!

[...]

Issue #662, opened 2020-08-20
Unable to set/get prior kappa image from STIR-python (matlab)


robbietuk:
Performing image reconstruction using STIR-python. After one iteration of OSEM, I would like to use the latest image as the kappa image for the prior. This should work something like this:

```
recon_obj.get_objective_function().get_prior().set_kappa_sptr(OSEM_image)
recon_obj.get_objective_function().get_prior().get_kappa_sptr()
```

The question is, can this be added to `GeneralisedPrior.h` and therefore exposed, or is there a prior that does not support the kappa methodology? It seems to work with PLS, RDP, QP, (log-cosh)

robbietuk:
Apolgies, it is possible to create a new a prior with the kappa image using
```
import stir
RPD = stir.RelativeDifferencePrior3DFloat()
RPD.set_penalisation_factor(0.00001)
RPD.set_kappa_sptr(OSEM_image)
recon_obj.get_objective_function().set_prior_sptr(RPD)
```
Yet it remains impossible to change `gamma` and `kappa_sptr` from `recon_obj.get_objective_function().get_prior()`

KrisThielemans:
it would certainly be possible to have a voxel-dependent weight for all priors. The original "kappa" is a specific way to do that that only applies to cases where the prior is a sum of pair-wise priors i.e.
```
prior =  sum_ij kappa_i kappa_j R_ij(lambda_i, lambda_v)
```
with usually
```
R_ij(lambda_i, lambda_v) = w_ij R(lambda_i, lambda_v)
```
with usually `w_ij` chosen as a function of the distance between the 2 voxels (so not location dependent).

This wouldn't be applicable/appropriate to all priors I guess (especially non-local-means type of things). In that case, you'd have something like
```
prior =  sum_i kappa_i^2 R_i(lambda_i; lambda)
```
(taking all of the image). This is what we assumed in Tsai's paper.

Enabling this would be part of a long-overdue clean-up of the `GeneralisedPrior` hierarchy.

KrisThielemans:
> Yet it remains impossible to change `gamma` and `kappa_sptr` from `recon_obj.get_objective_function().get_prior()`

yes, this is because currently we cannot downcast in Python. We have a similar problem for the objective functions, but added a `ToPoisson...` function by hand https://github.com/UCL/STIR/blob/aae4c92e9cb63699997573b4e02b78fdb86d373d/src/swig/stir.i#L1601-L1607

[...]

PR #709, opened 2020-10-09
Generate SWIG prior factory

### LINKED ISSUE #284: SWIG: add downcasting
A difficulty is that some members return a base class ptr, which cannot be downcasted in python. Example
```python
input=recon.get_input_data()
```
Then `input` is of type `ExamData` and there's no way to get at members of a `ProjData` (or whatever it is).

This might be possible as mentioned for instance at https://github.com/tpaviot/pythonocc-generator/issues/28

robbietuk:
Adds downcasting via a swig factory for priors. 

This should address #662 and #707

### DIFF
--- a/src/swig/stir.i
+++ b/src/swig/stir.i
@@ -1576,14 +1576,14 @@ namespace stir {
 %shared_ptr(stir::FBP2DReconstruction);
 %shared_ptr(stir::FBP3DRPReconstruction);
 
-// tell SWIG to convert the return of get_prior_ptr etc to the appropriate object
 /* This currently fails, so is still commented out
-
 %newobject *::get_objective_function;
 %factory(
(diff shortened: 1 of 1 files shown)


robbietuk:
Sorry force pushed to stop travis.

I took a look at your changes and it doesnt seem to have resolved the issue. 

```
>> QP = self.recon.get_objective_function().get_prior()
Exception ignored in: '_wrap_delete_QuadraticPrior3DFloat'
TypeError: in method 'delete_QuadraticPrior3DFloat', argument 1 of type 'stir::QuadraticPrior< float > *'
```

KrisThielemans:
ah. it generates the code, but then something goes wrong later with the SWIG type mechanisms. Of course, it's quite weird it's calling a `delete` as well... sigh

KrisThielemans:
Fixed in #1712

PR #716, opened 2020-10-15
Log cosh prior

robbietuk:
Adds the logcosh prior to STIR.

This is from https://github.com/UCL/UCL-STIR/tree/New_priors . A pull request was too difficult so I have manually copied the files over.

### DIFF
--- a/src/include/stir/recon_buildblock/LogcoshPrior.h
+++ b/src/include/stir/recon_buildblock/LogcoshPrior.h
@@ -26,7 +26,7 @@
  \author Sanida Mustafovic
  \author Yu-Jung Tsai
  \author Robert Twyman
-
+ \author Zeljko Kereta
  */
 
 
@@ -49,11 +49,11 @@ START_NAMESPACE_STIR
  \brief
  A class in the GeneralisedPrior hierarchy. This implements a logcosh Gibbs prior.
 
- The gradient of the prior is computed as follows:
-
- \f[
- g_r = \sum_dr w_{dr} tanh(\lambda_r - \lambda_{r+dr}) * \kappa_r * \kappa_{r+dr}
+ The log-cosh function is given by:
+  \f[
+ f_r = \sum_{dr} w_{dr} log(cosh(\lambda_r - \lambda_{r+dr})) * \kappa_r * \kappa_{r+dr}
  \f]
+
  where \f$\lambda\f$ is the image and \f$r\f$ and \f$dr\f$ are indices and the sum
  is over the neighbourhood where the weights \f$w_{dr}\f$ are non-zero.
 
@@ -71,6 +71,8 @@ START_NAMESPACE_STIR
  Logcosh Prior Parameters:=
  ; next defaults to 0, set to 1 for 2D inverse Euclidean weights, 0 for 3D
  only 2D:= 0
+ ; scalar controls the transition between the quadratic (smooth) and linear (edge-preserving) nature of the function
+ ; scalar:=
  ; next can be used to set weights explicitly. Needs to be a 3D array (of floats).
  ' value of only_2D is ignored
  ; following example uses 2D 'nearest neighbour' penalty
@@ -135,9 +137,10 @@ public:
                          const BasicCoordinate<3,int>& coords,
                          const DiscretisedDensity<3,elemT> &current_image_estimate);
 
-    virtual Succeeded
-    add_multiplication_with_approximate_Hessian(DiscretisedDensity<3,elemT>& output,
-                                                const DiscretisedDensity<3,elemT>& input) const;
+    //! Compute the multiplication of the hessian of the prior multiplied by the input.
+    virtual Succeeded accumulate_Hessian_times_input(DiscretisedDensity<3,elemT>& output,
+                                                     const DiscretisedDensity<3,elemT>& current_estimate,
+                                                     const DiscretisedDensity<3,elemT>& input) const;
 
     //! get penalty weights for the neigbourhood
     Array<3,float> get_weights() const;
@@ -155,11 +158,19 @@ public:
     //! set kappa image
     void set_kappa_sptr(const shared_ptr<DiscretisedDensity<3,elemT> >&);
 
+    //! Get the scalar value
+    float get_scalar() const;
+
+    //! Set the scalar value
+    void set_scalar(float scalar_v);
+
 protected:
     //! can be set during parsing to restrict the weights to the 2D case
     bool only_2D;
-    //!
+
+    //! controls the transition between the quadratic (smooth) and linear (edge-preserving) nature of the prior
     float scalar;
+
     //! filename prefix for outputing the gradient whenever compute_gradient() is called.
     /*! An internal counter is used to keep track of the number of times the
      gradient is computed. The filename will be constructed by concatenating
@@ -173,27 +184,30 @@ protected:
      That initialisation should be moved to a new set_up() function.
      */
     mutable Array<3,float> weights;
+
     //! Filename for the \f$\kappa\f$ image that will be read by post_processing()
     std::string kappa_filename;
 
     virtual void set_defaults();
     virtual void initialise_keymap();
     virtual bool post_processing();
+
 private:
+    //! Spatially variant penalty penalty image ptr
     shared_ptr<DiscretisedDensity<3,elemT> > kappa_ptr;
 
-    static inline float surrogate(float x)
-    {
-      const float eps = 0.1;
+    //! The surrogate is defined as tanh(x)/x
+    /*!
+     x is the should be the difference between the ith and jth voxel.
+     However, use the taylor expansion if the x is too small to prevent division by 0.
+    */
+    static inline float surrogate(float x){
+      const float eps = 0.01;
+      //  use Taylor of tanh: tanh(x)/x ~= (x - x^3/3)/x = 1- x^2/3
       if (fabs(x)<eps)
-      {
-        //  use Taylor: tanh(x) = x - x^3/3
-        return 1- square(x)/3;
-      }
+      { return 1- square(x)/3; }
       else
-      {
-        return tanh(x)/x;
-      }
+      { return tanh(x)/x; }
     }
 };
 
--- a/src/include/stir/recon_buildblock/LogcoshPrior.h
+++ b/src/include/stir/recon_buildblock/LogcoshPrior.h
@@ -86,102 +86,102 @@ START_NAMESPACE_STIR
 
 
  */
-    template <typename elemT>
-    class LogcoshPrior:  public
-                         RegisteredParsingObject< LogcoshPrior<elemT>,
-                                 GeneralisedPrior<DiscretisedDensity<3,elemT> >,
-                                 PriorWithParabolicSurrogate<DiscretisedDensity<3,elemT> >
-                         >
-    {
-    private:
-        typedef
-        RegisteredParsingObject< LogcoshPrior<elemT>,
-                GeneralisedPrior<DiscretisedDensity<3,elemT> >,
-                PriorWithParabolicSurrogate<DiscretisedDensity<3,elemT> > >
-                base_type;
+template <typename elemT>
+class LogcoshPrior:  public
+                     RegisteredParsingObject< LogcoshPrior<elemT>,
+                             GeneralisedPrior<DiscretisedDensity<3,elemT> >,
+                             PriorWithParabolicSurrogate<DiscretisedDensity<3,elemT> >
+                     >
+{
+private:
+    typedef
+    RegisteredParsingObject< LogcoshPrior<elemT>,
+            GeneralisedPrior<DiscretisedDensity<3,elemT> >,
+            PriorWithParabolicSurrogate<DiscretisedDensity<3,elemT> > >
+            base_type;
 
-    public:
-        //! Name which will be used when parsing a GeneralisedPrior object
-        static const char * const registered_name;
+public:
+    //! Name which will be used when parsing a GeneralisedPrior object
+    static const char * const registered_name;
 
-        //! Default constructor
-        LogcoshPrior();
+    //! Default constructor
+    LogcoshPrior();
 
-        //! Constructs it explicitly
-        LogcoshPrior(const bool only_2D, float penalization_factor);
+    //! Constructs it explicitly
+    LogcoshPrior(const bool only_2D, float penalization_factor);
 
-        //! Constructs it explicitly with scalar
-        LogcoshPrior(const bool only_2D, float penalization_factor, const float scalar);
+    //! Constructs it explicitly with scalar
+    LogcoshPrior(const bool only_2D, float penalization_factor, const float scalar);
 
-        virtual bool
-        parabolic_surrogate_curvature_depends_on_argument() const
-        { return false; }
+    virtual bool
+    parabolic_surrogate_curvature_depends_on_argument() const
+    { return false; }
 
-        //! compute the value of the function
-        double
-        compute_value(const DiscretisedDensity<3,elemT> &current_image_estimate);
+    //! compute the value of the function
+    double
+    compute_value(const DiscretisedDensity<3,elemT> &current_image_estimate);
 
-        //! compute gradient
-        void compute_gradient(DiscretisedDensity<3,elemT>& prior_gradient,
-                              const DiscretisedDensity<3,elemT> &current_image_estimate);
+    //! compute gradient
+    void compute_gradient(DiscretisedDensity<3,elemT>& prior_gradient,
+                          const DiscretisedDensity<3,elemT> &current_image_estimate);
 
-        //! compute the parabolic surrogate for the prior
-        /*! in the case of logcosh priors this will just be the sum of weighting coefficients*/
-        void parabolic_surrogate_curvature(DiscretisedDensity<3,elemT>& parabolic_surrogate_curvature,
-                                           const DiscretisedDensity<3,elemT> &current_image_estimate);
+    //! compute the parabolic surrogate for the prior
+    /*! in the case of logcosh priors this will just be the sum of weighting coefficients*/
+    void parabolic_surrogate_curvature(DiscretisedDensity<3,elemT>& parabolic_surrogate_curvature,
+                                       const DiscretisedDensity<3,elemT> &current_image_estimate);
 
-        //! compute Hessian
-        void compute_Hessian(DiscretisedDensity<3,elemT>& prior_Hessian_for_single_densel,
-                             const BasicCoordinate<3,int>& coords,
-                             const DiscretisedDensity<3,elemT> &current_image_estimate);
+    //! compute Hessian
+    void compute_Hessian(DiscretisedDensity<3,elemT>& prior_Hessian_for_single_densel,
+                         const BasicCoordinate<3,int>& coords,
+                         const DiscretisedDensity<3,elemT> &current_image_estimate);
 
-        virtual Succeeded
-        add_multiplication_with_approximate_Hessian(DiscretisedDensity<3,elemT>& output,
-                                                    const DiscretisedDensity<3,elemT>& input) const;
+    virtual Succeeded
+    add_multiplication_with_approximate_Hessian(DiscretisedDensity<3,elemT>& output,
+                                                const DiscretisedDensity<3,elemT>& input) const;
 
-        //! get penalty weights for the neigbourhood
-        Array<3,float> get_weights() const;
+    //! get penalty weights for the neigbourhood
+    Array<3,float> get_weights() const;
 
-        //! set penalty weights for the neigbourhood
-        void set_weights(const Array<3,float>&);
+    //! set penalty weights for the neigbourhood
+    void set_weights(const Array<3,float>&);
 
-        //! get current kappa image
-        /*! \warning As this function returns a shared_ptr, this is dangerous. You should not
-         modify the image by manipulating the image refered to by this pointer.
-         Unpredictable results will occur.
-         */
-        shared_ptr<DiscretisedDensity<3,elemT> > get_kappa_sptr() const;
+    //! get current kappa image
+    /*! \warning As this function returns a shared_ptr, this is dangerous. You should not
+     modify the image by manipulating the image refered to by this pointer.
+     Unpredictable results will occur.
+     */
+    shared_ptr<DiscretisedDensity<3,elemT> > get_kappa_sptr() const;
 
-        //! set kappa image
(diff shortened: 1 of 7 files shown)


src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -0,0 +1,189 @@
+//
+//
+/*
+ Copyright (C) 2000- 2011, Hammersmith Imanet Ltd
+ Copyright (C) 2020, UCL
```
KrisThielemans, line 5: add 2016,2016, which is when @tsai-yujung did this work

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
+//
+//
+/*
+ Copyright (C) 2000- 2011, Hammersmith Imanet Ltd
+ Copyright (C) 2020, UCL
```
KrisThielemans, line 5: dates

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+            // dz==0,dy==0,dx==0 will have weight 0, so we can just include it in the loop
+            elemT current =
+                    weights[dz][dy][dx];
+
+            if (do_kappa)
+              current *=
+                      (*kappa_ptr)[z][y][x] * (*kappa_ptr)[z+dz][y+dy][x+dx];
+
+            diagonal += current;
+            prior_Hessian_for_single_densel_cast[z+dz][y+dy][x+dx] = -current*this->penalisation_factor;
```
KrisThielemans, line 494: this seems a copy of the QP code. If we don't need it, you could throw an error (and say so in the doxygen)

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+              for (int dy=min_dy;dy<=max_dy;++dy)
+                for (int dx=min_dx;dx<=max_dx;++dx)
+                {
+                  elemT current =
+                          weights[dz][dy][dx] * input[z+dz][y+dy][x+dx];
+
+                  if (do_kappa)
+                    current *=
+                            (*kappa_ptr)[z][y][x] * (*kappa_ptr)[z+dz][y+dy][x+dx];
+                  result += current;
```
KrisThielemans, line 647: QP code?

KrisThielemans (changes requested):

src/recon_buildblock/LogcoshPrior.cxx
zeljkozeljko: Would that be in the method add_multiplication_with_approximate_hessian?

In this case, if I’m not mistaken, the needed functionality should be to return
\sum_{i a neighbour of j} w_{ij}  gamma(x_i-x_j) * f_i
(to clarify, the sum goes over “i”, which are neighbours of the voxel “j”)

where x is the current image estimate and f the input (NB: in practice for SEM I will  be using f=x), and gamma(x) = rho’(x)/x, for rho being the potential function.
For the quadratic prior rho(x)=x^2/2 so gamma(x) = 1, and thus that should refer to the quadratic prior

So, in case of logcosh, ie rho(x) = delta^2 log cosh( abs(x) / delta), where delta is a predefined parameter (it is not that beta parameter that multiplies the prior), we have
gamma(x)  = delta * tanh(abs(x) / delta) / abs(x)
Defining a temp variable (as the shorthand here) I guess it should  read

temp = abs(current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx])/delta;
weights[dz][dy][dx] * tanh(temp)/temp * input[z+dz][y+dy][x+dx];

NB 1: there might be numerical issues if abs(x)=0, ie if the two neighbouring entries have the same value. The correct value in that case is 1, since lim_{x->0} tanh(x)/x =1, whereas if you instead use tanh(x)/(x+eps),  the value will be wrong since tanh(0) =0.
NB 2: in s(VR)EM I will also be computing 
f_j * \sum_{i a neighbour of j} w_{ij}  gamma(x_i-x_j)
where, again, f = x. As long as the method above works, this shouldn’t be a problem since I can just compute
\sum_{i a neighbour of j} w_{ij}  gamma(x_i-x_j) * 1 ——> ie by inputting ones
and then multiply the output (entry wise) with f. It also shouldn’t be that much slower I guess

zeljkozeljko: Apologies for the long comment. I can also write it in a more readable format with latex and then attach that here

KrisThielemans: thanks. There's 2 places (easier to see on the `Files` tab) where I believe there's a copy-paste of the Quadratic prior, in the surrogate (`rho'(x)/x`), and in the Hessian (`rho''(x)`).

Would be useful to write this in function of `static inline` functions for both. Will make the clean-up easier.

NB1: yes, correct. What about using the [Taylor series for `tang`](https://www.planetmath.org/taylorseriesofhyperbolicfunctions) and say
```c++
float  surrogate(float x)
{
  if (abs(x)<eps)
  { 
   //  use Taylor: tanh(x) = x - x^3/3
   return 1- square(x)/2;
  }
  else
  {
   return tanh(x)/x;
  }
}
```
(need to put the scale factor in)

zeljkozeljko: I just noticed (ie found! the log cosh files) so NB 1 is mostly addressed in the parabolic_surrogate_curvature (where you add 0.01 to both the numerator and the denominator, giving tanh(abs(x)+0.01)/(abs(x)+0.01). this of course introduces a small error at x=0 (ie adding an epsilon introduces an error of size roughly epsilon^2/3), but that should be ok

zeljkozeljko: Apologies, I replied before seeing your comment! Yes, that should work, though I guess with square(x)/3 instead of square(x)/2?

KrisThielemans: oops. correct

src/recon_buildblock/LogcoshPrior.cxx
robbietuk: Logcosh does have a second derivitive
`(x sech^2(|x|/scalar)) / (|x|)`, where x is `temp = x_i - x_j`. I think we add it. This is actually a nice and well behaved function.

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+             log(cosh(current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx])) *
+             (*kappa_ptr)[z][y][x] * (*kappa_ptr)[z+dz][y+dy][x+dx];
+             */
+            for (int dz=min_dz;dz<=max_dz;++dz)
+              for (int dy=min_dy;dy<=max_dy;++dy)
+                for (int dx=min_dx;dx<=max_dx;++dx)
+                {
+                  elemT temp = current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx];
+
+                  elemT current = weights[dz][dy][dx] *(1/(this->scalar*this->scalar))*log(cosh(this->scalar*temp));
```
robbietuk, line 284: Why is there (`1/ scalar^2`)? Should this not be (`scalar^2`).

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+             */
+#if 1
+            elemT gradient = 0;
+            for (int dz=min_dz;dz<=max_dz;++dz)
+              for (int dy=min_dy;dy<=max_dy;++dy)
+                for (int dx=min_dx;dx<=max_dx;++dx)
+                {
+                  elemT temp = current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx];
+
+                  elemT current = weights[dz][dy][dx] * (1/this->scalar) * tanh(this->scalar*temp);
```
robbietuk, line 363: `1/scalar` here too...

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+          {
+            const int min_dx = max(weights[0][0].get_min_index(), min_x-x);
+            const int max_dx = min(weights[0][0].get_max_index(), max_x-x);
+
+            /* formula:
+             sum_dx,dy,dz
+             weights[dz][dy][dx] *
+             tanh(current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx]) *
+             (*kappa_ptr)[z][y][x] * (*kappa_ptr)[z+dz][y+dy][x+dx];
+             */
```
robbietuk, line 354: There is no mention of `scalar` here

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+                            weights[dz][dy][dx] * current_image_estimate[z+dz][y+dy][x+dx];
+
+                            if (do_kappa)
+                                current *= (*kappa_ptr)[z+dz][y+dy][x+dx];
+
+                            gradient -= current;
+                        }
+                // multiply with central kappa
+                if (do_kappa)
+                    gradient *= (*kappa_ptr)[z][y][x];
```
robbietuk, line 415: Is this needed?

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+                for (int dx=min_dx;dx<=max_dx;++dx)
+                {
+                  // psi'(t)/t = tanh/t
+                  elemT current;
+                  elemT temp = fabs(current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx]);
+
+                  //if (temp <= 0.01)
+                  //    current = weights[dz][dy][dx];
+                  //else
+                  current = weights[dz][dy][dx] * (1/(this->scalar))*tanh(this->scalar*(temp+0.01))/(temp+0.01);
```
robbietuk, line 561: Remove `+0.01` and replace with taylor (https://github.com/UCL/STIR/pull/716#discussion_r505725258).

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+       sprintf(filename, "normalised_gradient%d.v",count);
+       write_basic_interfile(filename, parabolic_surrogate_curvature);
+       }*/
+    }
+
+    template <typename elemT>
+    Succeeded
+    LogcoshPrior<elemT>::
+    add_multiplication_with_approximate_Hessian(DiscretisedDensity<3,elemT>& output,
+                                                const DiscretisedDensity<3,elemT>& input) const
```
robbietuk, line 589: Possibly not enough arguments. In theory this would requires 3 arguments (`output, input, current_image_estimate`). In our planned usage of this method, `input = current_image_estimate`, but what if it doesn't? Does this ever exist? QP does not suffer from this problem because x=1 and therefore `current_image_estimate` is ignored and there are only two arguments (`output, input`).

src/recon_buildblock/LogcoshPrior.cxx
robbietuk: I believe this is resolved now, can someone please check

src/recon_buildblock/LogcoshPrior.cxx
robbietuk: My first glance math was not up to scratch, I understand now why these `(1/scalar)` multiplications are present.

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+  elemT diagonal = 0;
+  for (int dz=min_dz;dz<=max_dz;++dz)
+    for (int dy=min_dy;dy<=max_dy;++dy)
+      for (int dx=min_dx;dx<=max_dx;++dx)
+      {
+        // (temp sech^2(|temp|/scalar)) / (|temp|)
+        // sech(temp) = 1/cosh(temp)
+        elemT temp = current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx];
+        elemT current = weights[dz][dy][dx] * (temp / fabs(temp)) *
+                square((1/ cosh(fabs(temp) / this->scalar)));
```
KrisThielemans, line 488: this will run into trouble at `temp==0` as usual. another Taylor expansion?

src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -0,0 +1,203 @@
...
+/*!
+ \ingroup priors
+ \brief
+ A class in the GeneralisedPrior hierarchy. This implements a logcosh Gibbs prior.
+
+ The gradient of the prior is computed as follows:
+
+ \f[
+ g_r = \sum_dr w_{dr} tanh(\lambda_r - \lambda_{r+dr}) * \kappa_r * \kappa_{r+dr}
+ \f]
```
KrisThielemans, line 56: I'd replace this with the logcosh formula prior itself, as opposed to giving the gradient. It will need alpha

src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -0,0 +1,203 @@
...
+ penalty is computed. If \f$\kappa\f$ is not set, this class will effectively
+ use 1 for all \f$\kappa\f$'s.
+
+ By default, a 3x3 or 3x3x3 neigbourhood is used where the weights are set to
+ x-voxel_size divided by the Euclidean distance between the points.
+
+ \par Parsing
+ These are the keywords that can be used in addition to the ones in GeneralPrior.
+ \verbatim
+ Logcosh Prior Parameters:=
```
KrisThielemans, line 71: seems to miss alpha

src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -0,0 +1,203 @@
...
+     */
+    shared_ptr<DiscretisedDensity<3,elemT> > get_kappa_sptr() const;
+
+    //! set kappa image
+    void set_kappa_sptr(const shared_ptr<DiscretisedDensity<3,elemT> >&);
+
+protected:
+    //! can be set during parsing to restrict the weights to the 2D case
+    bool only_2D;
+    //!
```
KrisThielemans, line 161: add something here

src/recon_buildblock/LogcoshPrior.cxx
KrisThielemans: seems there's no C++ sign function weirdly enough. Could [write our own](https://stackoverflow.com/questions/1903954/is-there-a-standard-sign-function-signum-sgn-in-c-c), or use [std::copysign](http://www.cplusplus.com/reference/cmath/copysign/) although it might be cryptic, so would need doc
```
In any case, let's move this to its own function as well (I suggest Hessian as well, although maybe somewhat weird for a 1D function)
std::copysign(square((1/ cosh(fabs(temp) / this->scalar))), temp)

**Review by KrisThielemans — CHANGES_REQUESTED** [review]

File: src/include/stir/recon_buildblock/LogcoshPrior.h [hunk 15]
```diff
@@ -0,0 +1,203 @@
...
+
+    virtual void set_defaults();
+    virtual void initialise_keymap();
+    virtual bool post_processing();
+private:
+    shared_ptr<DiscretisedDensity<3,elemT> > kappa_ptr;
+
+    static inline float surrogate(float x)
+    {
+      const float eps = 0.1;
```
robbietuk, line 187: Change to `0.01`

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+
+  elemT diagonal = 0;
+  for (int dz=min_dz;dz<=max_dz;++dz)
+    for (int dy=min_dy;dy<=max_dy;++dy)
+      for (int dx=min_dx;dx<=max_dx;++dx)
+      {
+        // (temp sech^2(|temp|/scalar)) / (|temp|)
+        // sech(temp) = 1/cosh(temp)
+        elemT temp = current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx];
+        elemT current = weights[dz][dy][dx] * (temp / fabs(temp)) *
```
robbietuk, line 487: the usage of `(temp / fabs(temp))` is odd here and can be removed.

src/recon_buildblock/LogcoshPrior.cxx
robbietuk, line 488: No need for `fabs()`

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,666 @@
...
+        const int min_dx = max(weights[0][0].get_min_index(), min_x-x);
+        const int max_dx = min(weights[0][0].get_max_index(), max_x-x);
+
+        elemT gradient = 0;
+        for (int dz=min_dz;dz<=max_dz;++dz)
+          for (int dy=min_dy;dy<=max_dy;++dy)
+            for (int dx=min_dx;dx<=max_dx;++dx)
+            {
+              // psi'(t)/t = tanh/t
+              elemT temp = fabs(current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx]);
```
robbietuk, line 557: Remove `fabs()`

robbietuk (commented):
Comments following discussion with Zeljko

src/recon_buildblock/LogcoshPrior.cxx
robbietuk: The function was wrong due to the abs(). Instead, the Hessian is simply `sech^2(scalar * temp)`. This is a well behaved function

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,609 @@
...
+        const int min_dx = max(weights[0][0].get_min_index(), min_x-x);
+        const int max_dx = min(weights[0][0].get_max_index(), max_x-x);
+
+        elemT result = 0;
+        for (int dz=min_dz;dz<=max_dz;++dz)
+          for (int dy=min_dy;dy<=max_dy;++dy)
+            for (int dx=min_dx;dx<=max_dx;++dx)
+            {
+              elemT temp = input[z][y][x] - input[z+dz][y+dy][x+dx];
+              elemT current = weights[dz][dy][dx] * surrogate(temp) * input[z+dz][y+dy][x+dx];
```
robbietuk, line 585: In `accumulate_Hessian_times_input()`: This is not the hessian. This is the `parabolic_surrogate_curvature_times_input`

robbietuk:
@zeljkozeljko @KrisThielemans I think this is done. Can you both please check over. I have, offline, tested this prior with the test in #695, which passed. 

PS. We will do on the surrogate work in a later PR as it will affect other classes.

KrisThielemans:
@zeljkozeljko please confirm...

src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -0,0 +1,231 @@
...
+    virtual void initialise_keymap();
+    virtual bool post_processing();
+
+private:
+    //! Spatially variant penalty penalty image ptr
+    shared_ptr<DiscretisedDensity<3,elemT> > kappa_ptr;
+
+    //! The surrogate of the logcosh function is tanh(x)/x
+    /*!
+     * @param x is the should be the difference between the ith and jth voxel.
```
KrisThielemans, line 202:
```suggestion
     * @param x is should be the difference between the ith and jth voxel.
```

KrisThielemans (changes requested):
looks fine (I didn't check in detail, but seems straightforward enough).

Can you remove updates to `release_4.1.htm` and put them in 5.0?

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -0,0 +1,598 @@
...
+                      (1/(this->scalar*this->scalar))*log(cosh(this->scalar*temp));
+
+              if (do_kappa)
+                current *= (*kappa_ptr)[z][y][x] * (*kappa_ptr)[z+dz][y+dy][x+dx];
+              result += static_cast<double>(current);
+            }
+      }
+    }
+  }
+  return result * this->penalisation_factor/2.0;
```
robbietuk, line 304: I was a little surprised to find that this is needed. The test of #695 only passes with this halving.

robbietuk (commented):
I think we are finally ready. We will need some way of testing all these methods, including the surrogates, in future.

src/recon_buildblock/LogcoshPrior.cxx
KrisThielemans: this is the problem that Tsai found. It appears because it's a derivative of a sum of symmetric functions. Similar factors can appear in the Hessian and surrogate.

robbietuk: We have no way of testing for those right now. I think with the current implementations, the hessian values need to be multiplied by 2. 

The surrogate is not as obvious for me.

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -506,16 +498,14 @@ LogcoshPrior<elemT>::parabolic_surrogate_curvature(DiscretisedDensity<3,elemT>&
             for (int dx=min_dx;dx<=max_dx;++dx)
             {
               // psi'(t)/t = tanh/t
-              elemT temp = fabs(current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx]);
-              elemT current  = weights[dz][dy][dx] * (1/(this->scalar))* surrogate(this->scalar*temp);
+              elemT temp =current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx];
+              elemT current = weights[dz][dy][dx] * (1/(this->scalar))* surrogate(this->scalar * temp);
```
zeljkozeljko: i'm not sure if there should be an additional 1/scalar in front here.  surrogate is already tanh(x)/x, so what this outputs is  weight/scalar * tanh(x * scalar) / (x * scalar) =   weight * tanh(x * scalar) / x. is that correct?

robbietuk: Yep, this is wrong, there should be not 1/scalar. It is also wrong for the taylor case too.

PR #717, opened 2020-10-15
fix STIR matlab ProjData::fill compilation error

KrisThielemans:
#612 forgot to update the MATLAB code


KrisThielemans:
@ludovicabrusaferri this should fix your problem. Can you try this out? Hardly anyone else still uses the MATLAB wrapper as far as I know.

ludovicabrusaferri:
It works!

Issue #718, opened 2020-10-16
add_multiplication_with_approximate_Hessian - Issues


robbietuk:
`add_multiplication_with_approximate_Hessian` is a slightly confusing method.

Poisson log-likelihood
-----
The method computes `A'((1/y)A(input))`, where `A`and `A'` are the forward and backward projection operations, `input`  could be a vector of 1s and `y` the measured data
https://github.com/UCL/STIR/blob/997f672f69749466f296feb4f322fefcd2b834b8/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L809-L893

This is a method that saves on computation by not computing the true hessian, but an approximation, at `current_estimate` `A' ((y/ybar^2) A(input))`, where `ybar` is the forward projected  `current_estimate`. This is therefore the reason for `approximate` in `add_multiplication_with_approximate_Hessian`. Therefore the arguments for this function are `output, input` and it does not require `current_estimate`.

Priors
------
The prior functions are slightly different however as they are not computationally costly to compute. Additionally, up until recently the QP is the only prior with an implementation of `add_multiplication_with_approximate_Hessian` and the Hessian of the QP is 1, and independent of the `current_estimate`. Therefore the implementation works as expected. https://github.com/UCL/STIR/blob/997f672f69749466f296feb4f322fefcd2b834b8/src/recon_buildblock/QuadraticPrior.cxx#L586-L658

However, in general, the prior input arguments for this method in https://github.com/UCL/STIR/blob/997f672f69749466f296feb4f322fefcd2b834b8/src/recon_buildblock/GeneralisedPrior.cxx#L62-L71
are not sufficient as the Hessian of a prior may not be independent of the `current_estimate`, e.g. logcosh https://github.com/UCL/STIR/pull/716

______
Proposed Resolution
-----
- Addition of another method `accumulate_Hessian_times_input` that takes 3 arguments (`output, current_estimate, input`) and computes the second derivative at `current_estimate` and multiplies by `input`. This would need to be done for both Poisson log-likelihood and priors with surrogate methods, i.e. QP and logcosh.

----- 
EDIT:
This issue has been expanded to cover all aspects of `add_multiplication_with_approximate_Hessian` which include likelihood classes and priors.

- [x] Addition of Likelihood method `accumulate_Hessian_times_input` #733

- [x] Additon of Prior method `accumulate_Hessian_times_input` #719

[...]

PR #719, opened 2020-10-16
Accumulate hessian times input for Priors

robbietuk:
Resolves the prior aspect of problem with `add_multiplication_with_approximate_Hessian` as discussed in #718

### DIFF
--- a/src/recon_buildblock/GeneralisedObjectiveFunction.cxx
+++ b/src/recon_buildblock/GeneralisedObjectiveFunction.cxx
@@ -271,8 +271,9 @@ add_multiplication_with_approximate_sub_Hessian(TargetT& output,
     {
       // TODO used boost:scoped_ptr
       shared_ptr<TargetT>  prior_output_sptr(output.get_empty_copy());
-      if (this->prior_sptr->accumulate_Hessian_times_input(*prior_output_sptr, output, output) == Succeeded::no)
-	    return Succeeded::no;
+      if (this->prior_sptr->add_multiplication_with_approximate_Hessian(*prior_output_sptr, output) ==
+	  Succeeded::no)
+	return Succeeded::no;
 
 
        // (*prior_output_sptr)/= num_subsets;
--- a/src/recon_buildblock/GeneralisedObjectiveFunction.cxx
+++ b/src/recon_buildblock/GeneralisedObjectiveFunction.cxx
@@ -271,9 +271,8 @@ add_multiplication_with_approximate_sub_Hessian(TargetT& output,
     {
       // TODO used boost:scoped_ptr
       shared_ptr<TargetT>  prior_output_sptr(output.get_empty_copy());
-      if (this->prior_sptr->add_multiplication_with_approximate_Hessian(*prior_output_sptr, output) ==
-	  Succeeded::no)
-	return Succeeded::no;
+      if (this->prior_sptr->accumulate_Hessian_times_input(*prior_output_sptr, output, output) == Succeeded::no)
+	    return Succeeded::no;
 
 
        // (*prior_output_sptr)/= num_subsets;
--- a/src/recon_buildblock/QuadraticPrior.cxx
+++ b/src/recon_buildblock/QuadraticPrior.cxx
@@ -589,6 +589,16 @@ QuadraticPrior<elemT>::
 add_multiplication_with_approximate_Hessian(DiscretisedDensity<3,elemT>& output,
                                             const DiscretisedDensity<3,elemT>& input) const
 {
+  return accumulate_Hessian_times_input(output, input, input);
+}
+
+template <typename elemT>
+Succeeded
+QuadraticPrior<elemT>::
+accumulate_Hessian_times_input(DiscretisedDensity<3,elemT>& output,
+                               const DiscretisedDensity<3,elemT>& current_estimate,
+                               const DiscretisedDensity<3,elemT>& input) const
+{
   // TODO this function overlaps enormously with parabolic_surrogate_curvature
   // the only difference is that parabolic_surrogate_curvature uses input==1
 
--- a/src/recon_buildblock/QuadraticPrior.cxx
+++ b/src/recon_buildblock/QuadraticPrior.cxx
@@ -596,7 +596,7 @@ template <typename elemT>
 Succeeded
 QuadraticPrior<elemT>::
 accumulate_Hessian_times_input(DiscretisedDensity<3,elemT>& output,
-                               const DiscretisedDensity<3,elemT>& current_estimate,
+                               const DiscretisedDensity<3,elemT>& /*current_estimate*/,
                                const DiscretisedDensity<3,elemT>& input) const
 {
   // TODO this function overlaps enormously with parabolic_surrogate_curvature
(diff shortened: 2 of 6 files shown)


src/recon_buildblock/GeneralisedObjectiveFunction.cxx
```diff
@@ -271,9 +271,8 @@ add_multiplication_with_approximate_sub_Hessian(TargetT& output,
     {
       // TODO used boost:scoped_ptr
       shared_ptr<TargetT>  prior_output_sptr(output.get_empty_copy());
-      if (this->prior_sptr->add_multiplication_with_approximate_Hessian(*prior_output_sptr, output) ==
-	  Succeeded::no)
-	return Succeeded::no;
+      if (this->prior_sptr->accumulate_Hessian_times_input(*prior_output_sptr, output, output) == Succeeded::no)
+	    return Succeeded::no;
```
robbietuk, line 275: I believe this is the only implementation of the `add_multiplication_with_approximate_Hessian ` for the prior to be replaced with `accumulate_Hessian_times_input`. I assume the usage is correct for the prior.

robbietuk:
TODO:
- [ ] Add something to release notes
- [ ] Add some documentation for `accumulate_Hessian_times_input` (`add_multiplication_with_approximate_Hessian` is also undocumented)

src/recon_buildblock/GeneralisedObjectiveFunction.cxx
KrisThielemans: sadly not. The "approximate_Hessian" for the log-likelihood (or in general data-fitting) term is supposed to be roughly independent of the `current_estimate`. We have no idea what the `input` will be. For instance, for OSSPS, `input` is an image with ones
https://github.com/UCL/STIR/blob/b79bb323c1f1abab02f397f04ea8a13995b133c2/src/iterative/OSSPS/OSSPSReconstruction.cxx#L234-L237

Luckily, this function `add_multiplication_with_approximate_sub_Hessian` seems unused in STIR. OSSPS only uses `add_multiplication_with_approximate_sub_Hessian_without_penalty` and the surrogate for the prior
https://github.com/UCL/STIR/blob/29c04d7ac773a8c7ae8911f49aa3b6d6ce447bbc/src/iterative/OSSPS/OSSPSReconstruction.cxx#L377-L378

Maybe we should just remove this function (i.e. have it later with the "true" Hessian one)

src/recon_buildblock/QuadraticPrior.cxx
```diff
@@ -588,6 +588,16 @@ Succeeded
...
                                             const DiscretisedDensity<3,elemT>& input) const
+{
+  return accumulate_Hessian_times_input(output, input, input);
+}
+
+template <typename elemT>
+Succeeded
+QuadraticPrior<elemT>::
+accumulate_Hessian_times_input(DiscretisedDensity<3,elemT>& output,
+                               const DiscretisedDensity<3,elemT>& current_estimate,
```
KrisThielemans, line 599:
```suggestion
                               const DiscretisedDensity<3,elemT>& /* current_estimate */,
```
comment out to avoid compiler warnings

KrisThielemans (changes requested):
We might have to change the `accumulate` terminology again, although I like it. We have for instance `add_subset_sensitivity`. Maybe you can have a look at how many other `add_` functions we have?

robbietuk:
> Maybe you can have a look at how many other add_ functions we have?

As far as I can tell from search using Clion and grep, the only `add_*` methods are `add_multiplication_with_approximate_Hessian` type methods (including w/o priors and `actual_add_*`) and the mentioned `add_subset_sensitivity`.  The majority of other `add_*`methods relate to parsing keys or additive sinograms etc.


----- 
In
https://github.com/UCL/STIR/blob/0d69a428f88311a04564238f1eef53c7c73070de/src/include/stir/recon_buildblock/GeneralisedObjectiveFunction.h#L200-L205 documents why `add_*` is used. `accumulate_*` would also be suitable. I guess it is whether it is worth changing the names of all the methods. 

In replacement of  `accumulate_Hessian_times_input(output, current_image_estimate, input)` [the new proposed methods] we could use `add_Hessian_times_input(output, current_image_estimate, input)` to maintain consistant terminology. But I do find the `add_` quite confusing as it implies to me that the function contains some addition (which it technically does to `output`). 

I do think `accumulate_` sound better for "Hessian" methods. It also might be better for`add_subset_sensitivity()`, which would become `accumulate_subset_sensitivity()`

src/recon_buildblock/GeneralisedObjectiveFunction.cxx
robbietuk: I have resolved by reverting.

zeljkozeljko:
I quite agree regarding the discussion about the naming (ie I find "accumulate" to be more intuitive and I'm clearly not sure what does "add" refer to there apart from that a method is added maybe?). But as you alluded to, there might be some legacy/consistency arguments. From my end either would work.

robbietuk:
> I'm clearly not sure what does "add" refer to there apart from that a method

`add_` refers to the fact that the approximate hessian is added to the content `output`

KrisThielemans (changes requested):
looks pretty good to me. some documentation on what they are, release notes, done?

KrisThielemans:
let's stick to `accumulate` then by the way. We can also do a small PR change the `add_subset_sensivity` to `accumulate_subset_sensitivity` later on.

robbietuk:
> let's stick to accumulate then by the way. We can also do a small PR change the add_subset_sensivity to accumulate_subset_sensitivity later on.

Okay. Do you want the older `add_multiplication_with_approximate_Hessian()` converted into `accumulate_multiplication_with_approximate_Hessian()` in this later PR?

Documentation and release notes incoming.

KrisThielemans:
> Do you want the older add_multiplication_with_approximate_Hessian() converted into accumulate_multiplication_with_approximate_Hessian() in this later PR?

We can do that. I think we should do these things in 2 stages: put in new name and flag up   old for deprecation, next major release: remove old name. So, I've quickly created #732

robbietuk:
@KrisThielemans Question: for the prior and maybe in general, why are we accumulating/adding? As `add_multiplication_with_approximate_Hessian()` for the prior is only used once in STIR
https://github.com/UCL/STIR/blob/0d69a428f88311a04564238f1eef53c7c73070de/src/recon_buildblock/GeneralisedObjectiveFunction.cxx#L273-L274
and it acts on an empty target, would not be a good opotunity to do away with the accumulate/add terminologies for this method?

My suggestion would be to do something similar to `compute_gradient(prior_gradient, current_estimate)` and `parabolic_surrogate_curvature(parabolic_surrogate_curvature, current_image_estimate)` and fill the `output` argument with uniform 0's imediately. This would then rename THIS method to `Hessian_times_input()`, which I think is a lot clearer. 

Now in general, for the liklihood / datafitting aspect, this may not make sense to all the methods but some of them it may. We can deal with that later?

robbietuk:
Ready to merge I think.

KrisThielemans:
We could always have a method in the base class that doesn't accumulate, and let that call `fill(0);accumulate*`. Easy to do that later if it becomes useful. So I'll merge now.

PR #722, opened 2020-10-20
Multiple Energy Window Code

ludovicabrusaferri:
This is a Pull Request for merging the multiple energy window code into master


KrisThielemans:
Travis fails because of undefined virtual energy functions for `CListModeDataGEHDF5`. Probably easiest not to derive from `CListEnergy`?

I didn't check why Appveyor or codacy failed...

ludovicabrusaferri:
Ok I have removed the dependency. I guess it will be the same for  CListRecordECAT too? Let me know.

KrisThielemans:
Possibly we cannot just delete it as `LmToProjData`  uses `record.energy().get_energyA_in_keV()`.

The only way I see to resolve that is by doing thing like
```c++
// before the loop, do a `dynamic_cast` to test if we have energy
auto clist_energy_ptr =  dynamic_cast<CListEnergy const *> (record);

for (...)
{
  if (clist_energy_ptr)
  {
    // do checks on energy
   }
}
```
a bit ugly. Maybe you see another way?

KrisThielemans:
Another problem is a much bigger one: what do we do with SPECT where there's only 1 energy window. Maybe @danieldeidda has some great ideas...

danieldeidda:
> 
> 
> Another problem is a much bigger one: what do we do with SPECT where there's only 1 energy window. Maybe @danieldeidda has some great ideas...

sorry I need to read through this PR, but why 1 energy window would be a problem?

KrisThielemans:
>  why 1 energy window would be a problem?

a lot of the code talks about pairs. This is unavoidable (and correct) for scatter, but will be a pain for SPECT, as there it doesn't make sense. Somehow we'd have to "hide" it I guess. Or replace a "pair" with a single index.

We could shift the single "pair" index into `Bin` I think. That way, it seems we can let `LmToProjData` be agnostic about energy windows. I haven't thought about this properly though...

ludovicabrusaferri:
I can see indeed this could be an issue for spect. I need to think about it carefully too , I'll come back to this asap

KrisThielemans:
Travis: seems some things are failing with the `recon_test_pack`: ROOT support test scripts (e.g. `run_ROOT_gate.sh`) but also `run_scatter_tests.sh` (although that only with `error relative to sup-norm of first data-set = 0.435529 %` at the very end)

Appveyor: a lot of warnings make it hard to see where the errors are (some are due to main STIR, so I'll fix the main culprit in `ExamInfo`) but I found
```
C:\projects\stir\src\scatter_buildblock\scatter_detection_modelling.cxx(244): error C2065: 'M_PI': undeclared identifier [C:\projects\stir\build\src\scatter_buildblock\scatter_buildblock.vcxproj]
C:\projects\stir\src\scatter_buildblock\scatter_detection_modelling.cxx(244): error C2789: 'den1': an object of const-qualified type must be initialized [C:\projects\stir\build\src\scatter_buildblock\scatter_buildblock.vcxproj]
```
For the first error, use `_PI` (`M_PI` isn't portable, `_PI` is defined in `stir/common.h` which is always included). The 2nd is probably just a consequence of the first.

src/include/stir/listmode/CListRecordROOT.h
```diff
@@ -127,6 +127,45 @@ class CListTimeROOT : public ListTime
...
+/*! \ingroup listmode
+ */
+class CListEnergyROOT : public ListEnergy
+{
+public:
+
+    void init_energy_from_data(double energy1, double energy2)
+    {
+        energyA = energy1;
+        energyB = energy2;
```
robbietuk, line 140: As mentioned in #736, I cant figure out the benifit having `energyA` adn `energyB`.

PR #723, opened 2020-10-20
Tidy Patlak code

### LINKED ISSUE #700: Does reusing time-frame info from previous outputs make sense in indirect Patlak modelling?
When calling `apply_patlak_to_images`

the code does the following:

https://github.com/UCL/STIR/blob/7b4c27470ad175a80f29e9aabe945b6831a4f159/src/modelling_utilities/apply_patlak_to_images.cxx#L86-L105

i.e. it uses an existing file as a template, then will write over it. 

I can only see 2 possible cases where the code for the template is used

1. You are using the same data as before, or similar, and then use the output file as template. But this is unnecesary, as we can create it without the template
2. You are using different dataframes than before, and thus your output will have bad info, because you are using it as template, instead of writting the new one, which you can do. 

So, I only see this piece of code as a "trap". Either doesn't do anything bad but its useless, or does something bad/unexpected, but also by not really adding anything at all, because removing the code would have avoided the unexpected behaviour and result in a good output file. 

Any reason to keep it?

### LINKED ISSUE #702: Tidy up Patlak code
Just opening an issue to keep track of all the things that need tidy up in the Patlak code (will make a PR for this, testing codes now).

- [ ] ~~Indentation all over the place (pending, maybe not fix)~~ . Not here, #98 
- [x] Add more comments
- [x] `PlasmaData::get_if_decay_corrected()` -> `PlasmaData::get_is_decay_corrected()` and other typos
- [x] Unnecessary second if (unnecessary error maybe?)
https://github.com/UCL/STIR/blob/7b4c27470ad175a80f29e9aabe945b6831a4f159/src/modelling_buildblock/PatlakPlot.cxx#L178-L181
- [ ] ~~Remove unnecessary extra code just to make sure one can index 1-based (will this have impact in direct recon? it should not)~~ Don't fix. 
      https://github.com/UCL/STIR/blob/7b4c27470ad175a80f29e9aabe945b6831a4f159/src/modelling_buildblock/PatlakPlot.cxx#L145-L150
 - [x] `brain_patlak_model_array` not just brain

AnderBiguri:
Just some minor cleanup on the code, mostly fix typos and add comments. 

Fix #702

Also fix #700


src/modelling_buildblock/PatlakPlot.cxx
```diff
@@ -262,9 +279,13 @@ PatlakPlot::apply_linear_regression(ParametricVoxelsOnCartesianGrid & par_image,
             const int max_i_index = dyn_image[1][k][j].get_max_index();
             for ( int i = min_i_index; i<= max_i_index; ++i)
               { 
+                // Patlak Linear regression is applied to teh data in the format:
+                // C(t)/Cp(t)=Ki*\int{Cp(t)}/Cp(t)+Vb
```
KrisThielemans, line 283: I doubt it. I don't think we divide by Cp(t) as discussed

AnderBiguri: It happens 4 lines below the comment: 

https://github.com/UCL/STIR/blob/5bfc3a7b313228137706a16f9bbb18c0c8cd492f/src/modelling_buildblock/PatlakPlot.cxx#L287

Its not Cp(t) as in continuous form, you are right. Its the value of Cp at t=(frame_start_t+frame_end_t)/2 multiplied by the length. 

Happy to change the notation, or add an explanation of this (I propose the second).

AnderBiguri: However this code has been written in such a way that all that is hidden from this function, and in fact, one could change the plasma data to actually divide by Cp(t) instead, and this code would not change. This is why I did not add the explanation first time.

AnderBiguri: Hopefully the current text is clearer.

KrisThielemans: ok. I mis-remembered the indirect estimation then ... sorry!

some spelling mistakes in the comment ("taht", "its")

KrisThielemans (commented):
could you remove the duplication with `create_model_matrix` then, while you're cleaning up? Avoids having to duplicate clarifications!

By the way, when only changing comments, please put `[ci skip]` in the first line of the commit message. Saves time and resources.

src/test/modelling/test_modelling.cxx
```diff
@@ -178,18 +178,21 @@ void modellingTests::run_tests()
...
       const unsigned int starting_frame=23;
-      ModelMatrix<2> stir_model_matrix=(patlak_plot.get_model_matrix(sample_plasma_data_in_frames,time_frame_def,starting_frame));
+      patlak_plot._plasma_frame_data=sample_plasma_data_in_frames;
+      patlak_plot._frame_defs=time_frame_def;
+      patlak_plot._starting_frame=starting_frame;
+      patlak_plot.set_up();
+      ModelMatrix<2> stir_model_matrix=(patlak_plot.get_model_matrix());
       ModelMatrix<2> mathematica_model_matrix;
       mathematica_model_matrix.read_from_file(this->add_directory("math_model_matrix.in"));
-      stir_model_matrix.uncalibrate(10);
```
AnderBiguri, line 184: The current `create_model_matrix()` does the uncalibration inside. `ModelMatrix` only allows you to do this once, thus `stir_model_matrix` can not be uncalibrated. If we leave this line here, the tests in 194 failed with the division by 10. 

Hopefully this does not change function, as `ModelMatrix.uncalibrate` is being tested here anyway, just inside `patlak_plot.set_up()` in the call to `patlak_plot_create_model_matrix()`

KrisThielemans: I can see we don't want to call `uncalibrate` here, but I cannot see why we'd get some results. `create_model_matrix` uses `his->_cal_factor`, but that should default to 1, not 10. Wouldn't you have to set it above (before `set_up()`)?

KrisThielemans: ah sorry. I see you removed the `/10.F.` below. ok, test should work. But I suggest to set the `_cal_factor` anyway, and use the value of `cal_factor as opposed to a hard-wired `10.F` below.

AnderBiguri: @KrisThielemans done! Test pass in my machine, well see how CI goes.

KrisThielemans (approved):

KrisThielemans:
thanks. all's well. Please add a brief line in the release_5.htm (developers only, renaming of the 2 functions). add `ci skip` 😉

KrisThielemans:
ok to squash-merge?

AnderBiguri:
Sure

PR #731, opened 2020-10-28
Xapyb

gfardell:
- Renamed and changed signature: axpby to xapyb(x, a, y, b)
  - This passes unit test with Array<4, float>
  - Builds with swig python wrappers

 - Implemented vector version as overload of xapyb where a and b are now the same types as x and y.

This seems unable to distinguish the template versions when trying to instantiate templates
It starts in the vector version of:
```Array::xapyb(Array<N>, Array<N>,Array<N>,Array<N>)```
Before calling the scalar base class version:
```NumericVectorWithOffset::xapyb(NumericVectorWithOffset, scalar, NumericVectorWithOffset, scalar)```

 - By using full iterators in Array::xapby to avoid the recursive calls or calling the base class implementation gets me further. 

However it seems it still tries to instantiate the wrong template:

```include/stir/Array.inl: In instantiation of ‘void stir::Array<num_dimensions, elemT>::xapyb(const stir::Array<num_dimensions, elemT>&, elemT2, const stir::Array<num_dimensions, elemT>&, elemT2) [with elemT2 = stir::NumericVectorWithOffset<stir::Array<3, float>, float>; int num_dimensions = 4; elemT = float]’:```

instead of calling the vector version with each argument as Array<4,float> it casts a and b as `NumericVectorWithOffset<stir::Array<3, float>, float>` and now choses the template intended for scalars by setting this as elemT2.
This goes beyond my c++ knowledge...
    - It seems that the flexibility to cast `Array<N>` as `NumericVectorWithOffset<Array<N-1>>` causes some strange behaviour with the templates
    - Can I restrict elemT2 to scalar types?
    - I can’t see why it thinks `Array::xapyb(Array<N>, elemT2, Array<N>, elemT2)` is a better match than `Array::xapyb(Array<N>, Array<N>,Array<N>,Array<N>)`

- To get around these (temporarily) I decided to not overload xapyb and instead name the vector version xapyb_vec
  - Implemented xapyb_vec
  - This passes unit tests for both Array<4,float> and NumericVectorWithOffset<Array<4,float>,float> 

 - However when I try to build swig wrappers for python, I get a failure with:

```include/stir/NumericVectorWithOffset.inl:332:18: error: no match for ‘operator=’ (operand types are ‘stir::Array<2, float>’ and ‘stir::NumericVectorWithOffset<stir::Array<1, float>, float>’)

include/stir/Array.h:78:7: note: candidate: stir::Array<2, float>& stir::Array<2, float>::operator=(const stir::Array<2, float>&)
 class Array : public NumericVectorWithOffset<Array<num_dimensions-1, elemT>, elemT>

include/stir/Array.h:78:7: note:   no known conversion for argument 1 from ‘stir::NumericVectorWithOffset<stir::Array<1, float>, float>’ to ‘const stir::Array<2, float>&’
```

It seems that when passed a NumericVector of Arrays the operation a * x + b * y returns the base type (Numeric Vector) and not an Array type. 
This somehow seems swig specific, I can’t recreate it in the unit tests. My tests seem to pass using a NumericVectors of Arrays, so maybe they aren't doing what I think they are.


KrisThielemans:
If memory serves me right, template overloading with functions is a bit unpredictable, but with classes partial template instantiation rules are much clearer. A trick I've used is to forward the call to a class, i.e.
```c++
template <class T1, class T2>
class forward
{
  void f(T1&a, T2&b) { ...}
};
template <class T1, int num_dimensions, class elemT2>
class forward
{
  void f(T1&a, Array<num_dimensions, elemT2>&b) { ...}
};

template <class T1, class T2> void f(T1& a, T2& b)
{ forward::f(a,b); }
```

For the SWIG problem, I'm not sure but would want to see the full error and relevant bit of SWIG code.  However, probably best to fix the other thing first.

KrisThielemans:
@gfardell This should be fine now. As you'll see, I've restricted the template scope, and let SWIG ignore `NumericVectorWithOffset::paxby` (which we don't need in Python anyway), which resolved the problems you mentioned above.

Would you be able to wrap this up?
- addition of the `self` version (probably one could call call the other, but whatever works)
- add yourself as author and STFC in the copyright header
- add a line in `documentation/release_4.1.htm` (in the section for developers)

gfardell:
@KrisThielemans thank you, I'll take a look and finish it off.

I  also need to add back in the old axpby again and mark as depreciated.

gfardell:
swig doesn't like the `DEPRECATED` decorator. Do you have any swig tips @KrisThielemans ?

https://github.com/UCL/STIR/blob/5f5a8b0ca9d5333d1b2cf2867d4c19243f0695f2/src/include/stir/NumericVectorWithOffset.h#L133

src/include/stir/NumericVectorWithOffset.h
```diff
@@ -126,11 +128,23 @@ class NumericVectorWithOffset : public VectorWithOffset<T>
   //! dividing the elements of the current vector by an \c elemT 
   inline NumericVectorWithOffset & operator/= (const elemT &v);
 
-  //! a*x+b*y, where a and b are scalar, and x and y are vectors
+  //! DEPRECATED a*x+b*y
```
KrisThielemans, line 131: as above

src/include/stir/deprecated.h
```diff
@@ -0,0 +1,39 @@
+/*
+    Copyright (C) 2000 PARAPET partners
+    Copyright (C) 2000- 2007, Hammersmith Imanet Ltd
+    Copyright (C) 2018-2019, University College London
```
KrisThielemans, line 4: somewhat tricky as we got it from the internet. I'd just delete them, but add STFC of course

src/include/stir/deprecated.h
```diff
@@ -0,0 +1,39 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License for more details.
+    See STIR/LICENSE.txt for details
+*/
+#ifndef __stir_deprecated_H__
+#define __stir_deprecated_H__
+/*!
+  \file 
+  \ingroup buildblock
+  \brief This file declares a deprecation function.
```
KrisThielemans, line 21:
```suggestion
  \brief This file declares a deprecation macro.
```

src/include/stir/deprecated.h
```diff
@@ -0,0 +1,39 @@
...
+  \author PARAPET project
```
KrisThielemans, line 22: you

src/include/stir/deprecated.h
```diff
@@ -0,0 +1,39 @@
...
+*/
+START_NAMESPACE_STIR
+
+// Deprecation function. With C++14, could use [[deprecated("some message")]]
```
KrisThielemans, line 26:
```suggestion
//! Deprecation macro

 \par Usage:
 \code
 STIR_DEPRECATED void some_old_function();
 \endcode
 \par Note:
 
  With C++14, could use <code>[[deprecated("some message")]]</code>. However, STIR does not require C++-14 (yet).
```

src/swig/stir.i
```diff
@@ -984,6 +984,9 @@ T * operator-> () const;
   }
 #endif
 
+%include "stir/deprecated.h"
```
KrisThielemans, line 987:
```suggestion
```
don't let SWIG parse it.

KrisThielemans (changes requested):

KrisThielemans:
sorry, forgot to add in my review to change `DEPRECATED` to `STIR_DEPRECATED`.

KrisThielemans:
I suspect you also need to define the macro to empty in `scr/swig/CMakeLists.txt` (like we do for `STIR_NAMESPACE_STIR` etc)

KrisThielemans:
Appveyor fails https://ci.appveyor.com/project/KrisThielemans/stir/builds/36604180/job/4k4acxo0cfx0cat5
```
  C:\projects\stir\src\test\test_Array.cxx(771): error C2899: typename cannot be used outside a template declaration [C:\projects\stir\build\src\test\test_Array.vcxproj]
  C:\projects\stir\src\test\test_Array.cxx(804): error C2899: typename cannot be used outside a template declaration [C:\projects\stir\build\src\test\test_Array.vcxproj]
```

src/swig/stir.i
gfardell: If I don't parse it first I get a build error ` Error: Syntax error in input(3)`

from: https://github.com/UCL/STIR/blob/f9cbaf54995caa4210c9dd080a06a8bc25abda51/src/include/stir/NumericVectorWithOffset.h#L133

documentation/release_4.1.htm
```diff
@@ -62,7 +63,8 @@ <h3>New functionality</h3>
     <tt>10:01:01.00+0130</tt>.<br />
     <strong>Warning:</strong> If the time-zone is not specified, we assume it's in the current time-zone without DST. This is dangerous,
     but allows us to subtract time-fields.</li>
-  <li>Added a utility <tt>list_lm_info</tt>.
+  <li>Added a utility <tt>list_lm_info</tt>.</li>
+  <li>Implementaion of <tt>a*x+b*y</tt> methods <tt>xapyb</tt> and <tt>sapyb</tt> where a and b can be scalar or vector types.</li>
```
KrisThielemans, line 67:
```suggestion
  <li>Implementation of <tt>a*x+b*y</tt> methods <code>xapyb</code> and <code>sapyb</code> where <tt>a</tt> and <tt>b</tt> can be scalar or vector types.</li>
```

documentation/release_4.1.htm
```diff
@@ -26,6 +26,7 @@ <h2>Overall summary</h2>
         <li>Ander Biguri (UCL) for extending the Signa support to other GE scanners using RDF9</li>
         <li>Richard Brown (UCL) for wrapping of NiftyPET's GPU functionality for projecting and unlisting</li>
         <li>Robert Twyman (UCL) for Relative Difference Prior, computation of the log-likelihood modification, and GATE geometry check fixes. </li>  
+        <li>Gemma Fardell (UKRI) for implementing xapyb and sapyb array methods. </li>
```
KrisThielemans, line 29:
```suggestion
        <li>Gemma Fardell (UKRI) for implementing <code>xapyb</code> and <code>sapyb</code> array methods. </li>  
```

documentation/release_4.1.htm
```diff
@@ -88,8 +90,10 @@ <h3>Changed functionality</h3>
     <li><code>ScatterEstimation</code> has now many more <code>set*</code> functions such that most
       (all?) parameters can be set from C++/Python (i.e. without parsing).
     </li>
+    <li><tt>axpby</tt> has been deprecated and replaced with <tt>xapyb</tt>.</li>
```
KrisThielemans, line 93: can you check, but I'm fairly sure that axpby was introduced by Richard after STIR 4.0. Not 100% sure. (you can check the release notes, or `git blame`). If so, I would rephrase this as "xpaby was introduced on the release_4 branch after STIR 4.0 but is now deprecated...."

src/include/stir/deprecated.h
```diff
@@ -0,0 +1,38 @@
...
+    GNU Lesser General Public License for more details.
+    See STIR/LICENSE.txt for details
+*/
+#ifndef __stir_deprecated_H__
+#define __stir_deprecated_H__
+/*!
+  \file 
+  \ingroup buildblock
+  \brief This file declares a deprecation macro.
+  \author PARAPET project
```
KrisThielemans, line 21:
```suggestion
```
remove this. PARAPET was before 2000 :-)

src/swig/stir.i
KrisThielemans: ok. I'd somewhat prefer to modify
https://github.com/UCL/STIR/blob/29f0027429c0ba7f42a8e877eeff3f957768f5b8/src/swig/CMakeLists.txt#L36
to set `DSTIR_DEPRECATED=""` but i guess if it works it's fine.

KrisThielemans (changes requested):
textual changes

documentation/release_4.1.htm
gfardell: Yes, it looks like it was introduced in 4.1 so I've made the change.

src/swig/stir.i
gfardell: I've updated it to unset the macro, thanks for pointing out the line. I'm a rabbit in headlights when I try to read cmake files!

KrisThielemans (approved):
great. should be fine to merge, but waiting for CI.

One Travis job fails due to a time-out related to xcode. we'll ignore that

KrisThielemans:
@gfardell could you check if SIRF works fine with this?

gfardell:
> @gfardell could you check if SIRF works fine with this?

I will do

gfardell:
My SIRF PET unit tests pass, and I tested with a python script that calls axpby fine on sirf.STIR.AcquisitionData.

Is there a better way to test it? I haven't really used SIRF before beyond some demos at the symposium.

KrisThielemans:
great.

I understand @evgueni-ovtchinnikov  uses SIRF `apxby`, which can call STIR's `axpby`, so we didn't break anything. Obviously, SIRF will then have to transition to the new naming convention as well. (your next job?). I'll wait for one more Appveyor job to complete and then merge.

thanks!

Issue #664, opened 2020-08-21
Spatially variant penalty strength (kappa images) have limited documentation and no executable to compute


robbietuk:
Refering to `Spatially variant penalty strength` as `kappa images`.
I believe with the publishing of [Tsai, Y.-J., et.al, (2020). Benefits of Using a Spatially-Variant Penalty Strength With Anatomical Priors in PET Reconstruction. https://doi.org/10.1109/TMI.2019.2913889] we can now add some more details to the kappa images in STIR. Currently there is very little information about how to to use these images, let alone compute them. 

There are two methods, proposed in the document, one based solely on the measured data and the other includes an initial estimate (which requires some pre-processing). I do not believe there is a method in STIR to compute the kappa image, but it can be used in the reconstruction with various priors. 

There should be an executable to compute the kappa image. Arguments should be: `output_kappa_filename`, `measured_data_sinogram`, `normalisation_sinogram`, `additive_sinogram`, and `template_image`. `template_image` can either  a template for the forward/backward projection operations (Eq.24) or the `x_init` (Eq.25). Additionally there should be a toggle between these two methods of computing kappa images.

Additionally, this method should be accompanied by a summary regarding to the purpose of the kappa image in the documentation.

PR #733, opened 2020-10-29
Spatially variant penalty strength (kappa) computation utility and accumulate_Hessian_times_input methods for likelihood

robbietuk:
Kappa Computation Utility
------
This will address issue #664
The spatially variant penalty strength (kappa) is detailed in 
> Tsai, Y.-J., Schramm, G., Ahn, S., Bousse, A., Arridge, S., Nuyts, J., Hutton, B. F., Stearns, C. W., & Thielemans, K. (2020). Benefits of Using a Spatially-Variant Penalty Strength With Anatomical Priors in PET Reconstruction. IEEE Transactions on Medical Imaging, 39(1), 11–22. https://doi.org/10.1109/TMI.2019.2913889

The two methods to compute kappa are:
 - `A' [ A(1) / y ]` where `y` is the measured data, and `A`/`A'` is the forward and backwards operator. This method is computed using `add_multiplication_with_approximate_Hessian()`. Divisions are elementwise.
 - `A' [ (y / ybar^2) A(1) ]`, where `ybar` is expected data. This method is computed using `accumulate_Hessian_times_input()`.

________
accumulate_Hessian_times_input (log-likelihood)
--------
This expands upon the work done in PR #719 and adds `accumulate_Hessian_times_input` to the log-likelihood term. See umbrella issue #718 for more details.

This method is added to the following classes:
- GeneralisedObjectiveFunction
- PoissonLogLikelihoodWithLinearModelForMeanAndProjData
- PoissonLogLikelihoodWithLinearModelForMeanAndGatedProjDataWithMotion
- PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData

These are all classes that contain the `add_multiplication_with_approximate_Hessian` type methods. The MotionandGated and KineticModel classes may be removed?

`compute_spatially_variant_penalty_strength` provides usage of the `accumulate_Hessian_times_input()` method.

________
Todo
- [x] Documentation
- [x] Test the output against my own shell script to compute kappa.
- [x] Cleanup 
- [x] Check if kappa computation can support priors. This will require the prior to have `accumulate_Hessian_times_input()` and `add_multiplication_with_approximate_Hessian()` methods.
- [x] Add to swig
- [x] Multiprocessing? Distributable method? [ e: No need for distributable for now. ]
- [x] STIR users guide section

### DIFF
--- a/examples/samples/compute_sqrt_Hessian_row_sum.par
+++ b/examples/samples/compute_sqrt_Hessian_row_sum.par
@@ -1,4 +1,4 @@
-Hessian Row Sum Computation Parameters :=
+Square Root Hessian Row Sum Computation Parameters :=
   
   ; The script requires an input image to be used as:
   ; - image estimate at which the square root of the Hessian row sum should be computed, or
--- a/examples/samples/compute_sqrt_Hessian_row_sum.par
+++ b/examples/samples/compute_sqrt_Hessian_row_sum.par
@@ -1,4 +1,4 @@
-Square Root Hessian Row Sum Computation Parameters :=
+SqrtHessianRowSum Parameters :=
   
   ; The script requires an input image to be used as:
   ; - image estimate at which the square root of the Hessian row sum should be computed, or
@@ -10,7 +10,8 @@ Square Root Hessian Row Sum Computation Parameters :=
 
   ; bool to indicate if an approximate Hessian should be used or the true Hessian.
   ; This changes the usage of the `input image filename :=` key.
-  use approximate Hessian := 0
+  ; Default value is 1
+  use approximate Hessian := 1
 
   ; Define the objective function. 
   ; This should be the same objective function as used for reconstructions
@@ -39,6 +40,7 @@ Square Root Hessian Row Sum Computation Parameters :=
 
     ; The use of a prior in the computation of the sqrt Hessian row sum is optional
     ; but will not work with an approximate Hessian row sum computation as priors do not have an approximate Hessian
+    ; Default value is 0
     compute with penalty := 0
 
     prior type := Quadratic Prior
--- a/examples/samples/spatially_variant_penalty_strength.par
+++ b/examples/samples/spatially_variant_penalty_strength.par
@@ -1,11 +1,11 @@
 Hessian Row Sum Computation Parameters :=
   
   ; The script requires an input image to be used as:
-  ; - image estiamte at which the hessian should be computed, or
-  ; - an image template for a backprojection into
+  ; - image estimate at which the Hessian should be computed, or
+  ; - an image template to specify the geometry of the approximate Hessian
   input image filename := 
 
-  ; The output filename of the spatially variant penalty strength image
+  ; The output filename of the row-sum
   output filename := 
 
   ; bool to indicate if an approximate hessian should be used or the true hessian. 
@@ -32,16 +32,7 @@ Hessian Row Sum Computation Parameters :=
     projector pair using matrix parameters :=
       matrix type := Ray Tracing
       ray tracing matrix parameters :=
-        disable caching := 0
-        store only basic bins in cache := 1
-        restrict to cylindrical fov := 1
         number of rays in tangential direction to trace for each bin := 50
-        use actual detector boundaries := 1
-        do symmetry 90degrees min phi := 1
-        do symmetry 180degrees min phi := 1
-        do symmetry swap segment := 1
-        do symmetry swap s := 1
-        do symmetry shift z := 1
       end ray tracing matrix parameters :=
     end projector pair using matrix parameters :=
 
--- a/examples/samples/spatially_variant_penalty_strength.par
+++ b/examples/samples/spatially_variant_penalty_strength.par
@@ -1,16 +1,16 @@
 Hessian Row Sum Computation Parameters :=
   
   ; The script requires an input image to be used as:
-  ; - image estimate at which the Hessian should be computed, or
+  ; - image estimate at which the square root of the Hessian row sum should be computed, or
   ; - an image template to specify the geometry of the approximate Hessian
   input image filename := 
 
-  ; The output filename of the row-sum
+  ; The output filename of the sqrt Hessian row-sum computation
   output filename := 
 
-  ; bool to indicate if an approximate hessian should be used or the true hessian. 
+  ; bool to indicate if an approximate Hessian should be used or the true Hessian.
   ; This changes the usage of the `input image filename :=` key.
-  use approximate hessian := 0
+  use approximate Hessian := 0
 
   ; Define the objective function. 
   ; This should be the same objective function as used for reconstructions
@@ -37,7 +37,8 @@ Hessian Row Sum Computation Parameters :=
     end projector pair using matrix parameters :=
 
 
-    ; The use of a prior is optional
+    ; The use of a prior in the computation of the sqrt Hessian row sum is optional
+    ; but will not work with an approximate Hessian row sum computation as priors do not have an approximate Hessian
     compute with penalty := 0
 
     prior type := Quadratic Prior
--- a/examples/samples/spatially_variant_penalty_strength.par
+++ b/examples/samples/spatially_variant_penalty_strength.par
@@ -1,0 +1,60 @@
+Hessian Row Sum Computation Parameters :=
+  
+  ; The script requires an input image to be used as:
+  ; - image estiamte at which the hessian should be computed, or
+  ; - an image template for a backprojection into
+  input image filename := 
+
+  ; The output filename of the spatially variant penalty strength image
+  output filename := 
+
+  ; bool to indicate if an approximate hessian should be used or the true hessian. 
+  ; This changes the usage of the `input image filename :=` key.
+  use approximate hessian := 0
+
+  ; Define the objective function. 
+  ; This should be the same objective function as used for reconstructions
+  objective function type:= PoissonLogLikelihoodWithLinearModelForMeanAndProjData
+  PoissonLogLikelihoodWithLinearModelForMeanAndProjData Parameters:=
+
+    ; Measured data and additive sinograms
+    input file := 
+    additive sinogram := 
+
+    Bin Normalisation type := From ProjData
+    Bin Normalisation From ProjData :=
+    ; normalistion sinogram
+      normalisation projdata filename:= 
+    End Bin Normalisation From ProjData:=
+
+    ; Optional definition of a projector
+    projector pair type := Matrix
+    projector pair using matrix parameters :=
+      matrix type := Ray Tracing
+      ray tracing matrix parameters :=
+        disable caching := 0
+        store only basic bins in cache := 1
+        restrict to cylindrical fov := 1
+        number of rays in tangential direction to trace for each bin := 50
+        use actual detector boundaries := 1
+        do symmetry 90degrees min phi := 1
+        do symmetry 180degrees min phi := 1
+        do symmetry swap segment := 1
+        do symmetry swap s := 1
+        do symmetry shift z := 1
+      end ray tracing matrix parameters :=
+    end projector pair using matrix parameters :=
+
+
+    ; The use of a prior is optional
+    compute with penalty := 0
+
+    prior type := Quadratic Prior
+    Quadratic Prior Parameters:=
+      penalisation factor := 0.0
+      only 2D:= 0
+    END Quadratic Prior Parameters:=
+
+  end PoissonLogLikelihoodWithLinearModelForMeanAndProjData Parameters:=
+
+END :=
--- a/src/include/stir/recon_buildblock/Hessian_row_sum.h
+++ b/src/include/stir/recon_buildblock/Hessian_row_sum.h
@@ -18,14 +18,14 @@
 /*!
   \file
   \ingroup recon_buildblock
-  \brief Declaration of class stir::KappaComputation
+  \brief Declaration of class stir::Hessian_row_sum
 
   \author Robert Twyman
   \author Kris Thielemans
 */
 
-#ifndef STIR_KAPPACOMPUTATION_H
-#define STIR_KAPPACOMPUTATION_H
+#ifndef STIR_HESSIAN_ROW_SUM_H
+#define STIR_HESSIAN_ROW_SUM_H
 
 #include "stir/info.h"
 #include "stir/IO/OutputFileFormat.h"
@@ -62,44 +62,53 @@ class Succeeded;
   https://doi.org/10.1109/TMI.2019.2913889
 */
 template <typename TargetT>
-class KappaComputation:
+class Hessian_row_sum:
         public ParsingObject
 {
-    //All methods need documenting
 public:
-    KappaComputation();
+    Hessian_row_sum();
 
-    //!The principle function used to load the input image from file, setup, compute the spatially variant penalty
-    //! strength, and save it.
+    //! The main function to call to compute spatially variant penalty strength images
+    //! Loads the input image from file, sets up, compute Hessian row sum (with input = ones), sqrts the result,
+    //! and saves it.
     void process_data();
 
     //! get and set methods for the objective function
+    //@{
     GeneralisedObjectiveFunction<TargetT > const& get_objective_function();
     void set_objective_function_sptr(const shared_ptr<GeneralisedObjectiveFunction<TargetT > > &obj_fun);
+    //@}
 
     //! get and set methods for the input image
+    //@{
     shared_ptr<TargetT> get_input_image();
     void set_input_image(shared_ptr <TargetT > const& image);
+    //@}
 
     //! get method for returning the kappa_image
-    shared_ptr<TargetT> get_kappa_image_target_sptr();
-    //! This method resets
-    void reset_kappa_image_target_sptr(shared_ptr <TargetT > const& image);
+    shared_ptr<TargetT> get_output_target_sptr();
+
+    //! This method resets the output target to uniform 0s
+    void reset_output_target_sptr(shared_ptr <TargetT > const& image);
 
     //! get and set methods for use approximate hessian bool
+    //@{
     bool get_use_approximate_hessian();
     void set_use_approximate_hessian(bool use_approximate);
+    //@}
 
     //! get and set methods for the compute with penalty bool
+    //@{
     bool get_compute_with_penalty();
     void set_compute_with_penalty(bool with_penalty);
+    //@}
 
     //! Computes the objective function Hessian at the current image estimate. Can use the penalty's Hessian
     //! kappa_image_sptr is the
-    void compute_kappa_at_current_image_estimate();
+    void compute_Hessian_row_sum();
 
     //! Computes the approximate Hessian of the objective function. Cannot use penalty's Hessian
-    void compute_kappa_with_approximate();
+    void compute_approximate_Hessian_row_sum();
 
 protected:
 
@@ -108,7 +117,7 @@ private:
     shared_ptr<GeneralisedObjectiveFunction<TargetT> >  objective_function_sptr;
 
     //! The filename the for the output spatially variant penalty strength
-    std::string kappa_filename;
+    std::string output_filename;
 
     //! Used to load an image as a template or current image estimate to compute hessian at
     std::string input_image_filename;
@@ -117,7 +126,7 @@ private:
     shared_ptr<TargetT> input_image;
 
     //! The variable to which kappa computation methods will populate
-    shared_ptr<TargetT> kappa_image_target_sptr;
+    shared_ptr<TargetT> output_target_sptr;
 
     //! Used to toggle the approximate hessian or hessian at current image estimate should be computed
     //! This toggles the usage of input_image.
@@ -139,4 +148,4 @@ private:
 };
 
 END_NAMESPACE_STIR
-#endif //STIR_KAPPACOMPUTATION_H
+#endif //STIR_HESSIAN_ROW_SUM_H
--- a/src/include/stir/recon_buildblock/Hessian_row_sum.h
+++ b/src/include/stir/recon_buildblock/Hessian_row_sum.h
@@ -73,6 +73,9 @@ public:
     //! and saves it.
     void process_data();
 
+    //! Saves the output image as output_filename
+    void save_output();
+
     //! get and set methods for the objective function
     //@{
     GeneralisedObjectiveFunction<TargetT > const& get_objective_function();
--- a/src/recon_buildblock/Hessian_row_sum.cxx
+++ b/src/recon_buildblock/Hessian_row_sum.cxx
@@ -59,7 +59,7 @@ template <typename TargetT>
 void
 Hessian_row_sum<TargetT>::initialise_keymap()
 {
-  parser.add_start_key("Hessian_row_sum Computation Parameters");
+  parser.add_start_key("Hessian Row Sum Computation Parameters");
   parser.add_key("output filename", &output_filename);
   parser.add_key("input image filename", &input_image_filename);
   parser.add_key("use approximate hessian", &use_approximate_hessian);
@@ -183,7 +183,7 @@ process_data()
   }
   else
   {
-    //Compute the Hessian_row_sum image with the full hessian at the current image estimate.
+    // Compute the Hessian_row_sum image with the full hessian at the input image estimate.
     // The input image here is assumed to be the current_image_estimate at the point the hessian will be computed
     compute_Hessian_row_sum();
   }
@@ -205,7 +205,7 @@ compute_Hessian_row_sum()
   auto ones_image_sptr = input_image->get_empty_copy();
   std::fill(ones_image_sptr->begin_all(), ones_image_sptr->end_all(), 1.F);
 
-  info("Computing the spatially variant penalty strength at the current image estimate, this may take a while.");
+  info("Computing Hessian row sum, this may take a while...");
   if (get_compute_with_penalty())
     objective_function_sptr->accumulate_Hessian_times_input(*output_target_sptr, *input_image, *ones_image_sptr);
   else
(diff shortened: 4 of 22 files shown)


src/utilities/compute_spatially_variant_penalty_strength.cxx
```diff
@@ -0,0 +1,200 @@
...
+
+  shared_ptr<DiscretisedDensity<3, float>> output_image;
+  output_image = static_cast<const shared_ptr<DiscretisedDensity<3, float>>>(current_image_estimate->get_empty_copy());
+  output_image->fill(0.);
+
+  shared_ptr<DiscretisedDensity<3, float>> ones_image;
+  ones_image = static_cast<const shared_ptr<DiscretisedDensity<3, float>>>(current_image_estimate->get_empty_copy());
+  ones_image->fill(1.);
+
+  objective_function_sptr->set_up(current_image_estimate);
```
robbietuk, line 161: This currently needs setup because of a check on image characteristics in the sub functionality of `accumulate_Hessian_times_input` -> https://github.com/robbietuk/STIR/blob/5aa14155e728ced619e3ce1f1a7b8d4b9eb45422/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L818-L819

src/utilities/compute_spatially_variant_penalty_strength.cxx
```diff
@@ -0,0 +1,200 @@
...
+  shared_ptr<DiscretisedDensity<3,float>>
+          output_image(read_from_file<DiscretisedDensity<3,float>>(input_image_filename));
+  output_image->fill(0.);
+
+  shared_ptr<DiscretisedDensity<3, float>> ones_image;
+  ones_image = static_cast<const shared_ptr<DiscretisedDensity<3, float>>>(output_image->get_empty_copy());
+  ones_image->fill(1.);
+
+  objective_function_sptr->add_multiplication_with_approximate_Hessian(*output_image, *ones_image);
+  output_file_format_sptr->write_to_file(kappa_filename, *output_image);
```
robbietuk, line 181: This is currently where the method fails.

robbietuk:
The current output of `add_multiplication_with_approximate_Hessian_without_penalty()` and `accumulate_Hessian_times_input` is what we call kappa^2. 
@KrisThielemans, how do I square root `DiscretisedDensity` objects?

robbietuk (commented):
Move this to recon_buildblock and template.

src/utilities/compute_spatially_variant_penalty_strength.cxx
```diff
@@ -0,0 +1,223 @@
...
+{
+  // Square root the output
+  typename DiscretisedDensity<3,float>::const_full_iterator output_iter = output_image.begin_all_const();
+  const typename DiscretisedDensity<3,float>::const_full_iterator end_prior_output_iter = output_image.end_all_const();
+  typename DiscretisedDensity<3,float>::full_iterator tmp_iter = output_image.begin_all();
+  while (output_iter!=end_prior_output_iter)
+  {
+    *tmp_iter = sqrt(*output_iter);
+    ++tmp_iter; ++output_iter;
+  }
```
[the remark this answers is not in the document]
KrisThielemans: Looks like `auto` in a lambda is C++14. sorry. I think here you can safely use `float&`.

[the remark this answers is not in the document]
KrisThielemans: by the way, the `for_each` choice makes more sense as it's much clearer that the function modifies the object (and is therefore much easier to thread for a smart compiler)

src/recon_buildblock/KappaComputation.cxx
```diff
@@ -0,0 +1,264 @@
...
+  typename TargetT::full_iterator tmp_iter = image.begin_all();
+  while (output_iter!=end_prior_output_iter)
+  {
+    *tmp_iter = sqrt(*output_iter);
+    ++tmp_iter; ++output_iter;
+  }
+}
+
+template class KappaComputation<DiscretisedDensity<3,float> >;
+//template class KappaComputation<ParametricVoxelsOnCartesianGrid >;
```
robbietuk, line 262: Cannot enable due to `fill()` method usages on images.

src/recon_buildblock/KappaComputation.cxx
```diff
@@ -0,0 +1,264 @@
...
+  }
+  return false;
+}
+
+template <typename TargetT>
+PoissonLogLikelihoodWithLinearModelForMean<TargetT > const&
+KappaComputation<TargetT>::
+get_objective_function()
+{
+  return static_cast<PoissonLogLikelihoodWithLinearModelForMean<TargetT >&> (*objective_function_sptr);
```
robbietuk, line 95: Is this correct now? I am not sure the `static_cast` is needed

src/recon_buildblock/KappaComputation.cxx
```diff
@@ -0,0 +1,264 @@
...
+  return kappa_image_target_sptr;
+}
+
+template <typename TargetT>
+void
+KappaComputation<TargetT>::
+reset_kappa_image_target_sptr(shared_ptr <TargetT > const& image)
+{
+  kappa_image_target_sptr =  unique_ptr<TargetT>(image->get_empty_copy());
+  kappa_image_target_sptr->fill(0);
```
robbietuk, line 136: Does `get_empty_copy()` already `fill(0)`? If so, L136 is not needed

src/recon_buildblock/KappaComputation.cxx
```diff
@@ -0,0 +1,264 @@
...
+template <typename TargetT>
+void
+KappaComputation<TargetT>::
+compute_kappa_with_approximate()
+{
+  reset_kappa_image_target_sptr(input_image);
+  info("Computing the spatially variant penalty strength using approximate hessian, this may take a while.");
+  // Setup image
+  auto ones_image_sptr = input_image->get_empty_copy();
+  ones_image_sptr->fill(1.);
```
robbietuk, line 233: Second `fill()` method that is casuing problems with Templates.

src/utilities/compute_spatially_variant_penalty_strength.cxx
```diff
@@ -0,0 +1,72 @@
...
+using std::endl;
+
+
+
+
+START_NAMESPACE_STIR
+static void print_usage_and_exit()
+{
+  //todo:update usage
+  std::cerr<<"\nUsage: compute_spatially_variant_penalty_strength template_proj_data\n";
```
robbietuk, line 53: Usage need updating

src/utilities/compute_spatially_variant_penalty_strength.cxx
```diff
@@ -0,0 +1,72 @@
...
+
+  Computes a spatially variant penalty strength, either using:
+
+  todo: add methods and documentation
+  See Tsai, Y.-J., Schramm, G., Ahn, S., Bousse, A., Arridge, S., Nuyts, J., Hutton, B. F., Stearns, C. W.,
+    & Thielemans, K. (2020). Benefits of Using a Spatially-Variant Penalty Strength With Anatomical Priors
+    in PET Reconstruction. IEEE Transactions on Medical Imaging, 39(1), 11–22. https://doi.org/10.1109/TMI.2019.2913889
+   for more details
+
+  \author Robert Twyman
```
robbietuk, line 37: Need more information. Include an example parameter file.

robbietuk (commented):
As is obvious, I have gone quite heavy on the usage of the word `Kappa` in naming. This is probably not the best name for the methodology to compute spatially variant penalty strengths. Please feel free to suggest. It is named kappa in other area's of STIR, including in the priors.

src/recon_buildblock/KappaComputation.cxx
robbietuk: Instead we can use `this->get_objective_function().fill_nonidentifiable_target_parameters(* ones_image_sptr, 1.);`

src/recon_buildblock/KappaComputation.cxx
```diff
@@ -0,0 +1,264 @@
...
+{
+  this->set_defaults();
+}
+
+template <typename TargetT>
+void
+KappaComputation<TargetT>::
+set_defaults()
+{
+  objective_function_sptr.reset(new PoissonLogLikelihoodWithLinearModelForMeanAndProjData<TargetT>);
```
robbietuk, line 53: @KrisThielemans This is currently the problem I have when trying to allow other templates including 
```
template class KappaComputation<ParametricVoxelsOnCartesianGrid >;
```
Ideally `objective_function_sptr` should be set to different types of objective_function depending on what `TargetT` (template) is, but I dont know how to do this and because I can I think I am miss understanding the usage of abstract classes.

src/recon_buildblock/KappaComputation.cxx
KrisThielemans: I had forgotten we have that function. However, I think it does the "complement" (it sets the *non*-identifiable voxels...)

So, another STL one
```c++
std::fill(ones_image_sptr.begin_all(), ones_image_sptr->end_all(), 1.F);
```

src/recon_buildblock/KappaComputation.cxx
KrisThielemans: probably does, but not guaranteed, so best to fill anyway

src/recon_buildblock/KappaComputation.cxx
KrisThielemans: I think the solution is not to set a default for `objective_function_sptr`. What input data would the default have anyway?

src/recon_buildblock/KappaComputation.cxx
```diff
@@ -0,0 +1,264 @@
...
+  info("Spatially variant penalty strength (Kappa) has been computed and saved as " + kappa_filename + ".");
+}
+
+template <typename TargetT>
+void
+KappaComputation<TargetT>::
+compute_kappa_at_current_image_estimate()
+{
+  reset_kappa_image_target_sptr(input_image);
+  info("Computing the spatially variant penalty strength at the current image estimate, this may take a while.");
```
KrisThielemans, line 211: info isn't accurate as we could just as well use this for preconditioning

KrisThielemans (changes requested):
I see no reason to let `KappaComputation` have a `PoissonLogLikelihoodWithLinearModelForMean`. Why not `GeneralisedObjectiveFunction`?

Renaming `Kappa`: I think Fessler calls it "certainty". Not 100% sure why, and it might apply only to his kappa (which differs from ours). Could check. However, if we compute `H.1`, why not call that `Hessian_row_sum`? That then implies we have that the current kappa would be `sqrt_Hessian_row_sum` or something...

src/recon_buildblock/KappaComputation.cxx
```diff
@@ -0,0 +1,264 @@
...
+  // Save the output
+  output_file_format_sptr->write_to_file(kappa_filename, *kappa_image_target_sptr);
+
+  info("Spatially variant penalty strength (Kappa) has been computed and saved as " + kappa_filename + ".");
+}
+
+template <typename TargetT>
+void
+KappaComputation<TargetT>::
+compute_kappa_at_current_image_estimate()
```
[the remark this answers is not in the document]
robbietuk: I agree. That fits in the the preconditioning aspect too.

src/recon_buildblock/KappaComputation.cxx
```diff
@@ -0,0 +1,264 @@
...
+{
+  reset_kappa_image_target_sptr(input_image);
+  info("Computing the spatially variant penalty strength at the current image estimate, this may take a while.");
+  auto ones_image_sptr = input_image->get_empty_copy();
+  ones_image_sptr->fill(1);
+
+  if (get_compute_with_penalty())
+    objective_function_sptr->accumulate_Hessian_times_input(*kappa_image_target_sptr, *input_image, *ones_image_sptr);
+  else
+    objective_function_sptr->accumulate_Hessian_times_input_without_penalty(*kappa_image_target_sptr, *input_image, *ones_image_sptr);
```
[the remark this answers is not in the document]
robbietuk: Yes that is the general idea. I think we have briefly discussed before the idea of computing the hessian of the objective function instead of only the likelihood. This functionality didnt exist before this PR because `accumulate_Hessian_times_input` didnt exist and prior approximate hessians were weird. 

However, we need to be able to turn this off. There are priors without hessians in STIR (e.g. RDP) but we still want to be able to compute `accumulate_Hessian_times_input_without_penalty`

robbietuk:
>I see no reason to let KappaComputation have a PoissonLogLikelihoodWithLinearModelForMean. Why not GeneralisedObjectiveFunction?

I tried this to begin with but GeneralisedObjectiveFunction does not have the methods: `set_recompute_sensitivity`, `set_use_subset_sensitivities`, and `set_sensitivity_filename`, which allow me to speed up the objective function setup. Could ignore these however and take the computational penalty associated with computing the sensitivities (or set thes from parameter file). 

> Renaming Kappa: I think Fessler calls it "certainty". Not 100% sure why, and it might apply only to his kappa (which differs from ours). Could check. However, if we compute H.1, why not call that Hessian_row_sum? That then implies we have that the current kappa would be sqrt_Hessian_row_sum or something...

`Hessian_row_sum` is a good name.

src/recon_buildblock/KappaComputation.cxx
robbietuk: Removing this default solved a lot of problems.

KrisThielemans:
> > I see no reason to let KappaComputation have a PoissonLogLikelihoodWithLinearModelForMean. Why not GeneralisedObjectiveFunction?
> 
> I tried this to begin with but GeneralisedObjectiveFunction does not have the methods: `set_recompute_sensitivity`, `set_use_subset_sensitivities`, and `set_sensitivity_filename`, which allow me to speed up the objective function setup. Could ignore these however and take the computational penalty associated with computing the sensitivities (or set thes from parameter file).

ok, but bad idea to restrict a general bit of code to enable an optimisation. You can put the bit where you need `PoissonLogLikelihoodWithLinearModelForMean` in an `if` with a `dynamic_cast`

src/recon_buildblock/KappaComputation.cxx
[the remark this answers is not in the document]
robbietuk: I have changed my mind. The computation of  `H.1` computation is done in `GeneralisedObjectiveFunction` and `KappaComputation` (renamed `Hessian_row_sum` in future) simply selects which Hessian method to use. Other classes, such as the reconstruction classes, will have access to the `accumulate_Hessian_times_input()` type methods in `GeneralisedObjectiveFunction` where it is very simple to setup the "Kappa" computation method.

Perhaps `Hessian_row_sum` doesnt need its own class but we wanted it to be a `Utilities` and present in `recon_buildblock` for python usage.

robbietuk:
Not quite done but I think we can start to finalise things. The image is a comparison between my shell script I use to compute `kappa` images and the STIR executable from an initial OSEM image guess. Minor scaling differences but those are due to the scanner sensitivity at the edges of the FOV. There is a slight difference between the two images but this is most likely due to the edge of the FOV.
image
I would say this is a sucess.

KrisThielemans:
agreed. difference in background are certainly due to "divide_and_truncate" etc. I'm happy with this

robbietuk:
I dont think OpenMP is working correctly. The computation time of the sensitivity for D690 is `CPU 1143.57s, wall-clock 88.3149s`. However, computation of the Hessian row sum is very slow.

robbietuk:
As I understand STIR now, all multithreading goes through `distributable` and as it turns out, the current OpenMP code in STIR masters `actual_add_multiplication_with_approximate_sub_Hessian_without_penalty` and `actual_accumulate_sub_Hessian_times_input_without_penalty` does not work on multiple threads and is therefore very slow.  I do not know the reason for this, but I am guessing it has something to do with OpenMP and setting the number of threads outside of `distributable`. 

During `actual_add_multiplication_with_approximate_sub_Hessian_without_penalty`, if I add the sequence

```
std::cout << "a: " << omp_get_num_threads() << "\n";
omp_set_num_threads(4);
std::cout << "b: " << omp_get_num_threads() << "\n";
```
I get a return of 
```
a: 1
b: 1
```

robbietuk:
Implementation into Swig remains but that is a minor fix coming. OpenMP now works for each of the `for` loops

examples/samples/spatially_variant_penalty_strength.par
```diff
@@ -0,0 +1,60 @@
+Hessian Row Sum Computation Parameters :=
+  
+  ; The script requires an input image to be used as:
+  ; - image estiamte at which the hessian should be computed, or
+  ; - an image template for a backprojection into
```
KrisThielemans, line 5:
```suggestion
  ; - image estimate at which the Hessian should be computed, or
  ; - an image template to specify the geometry of the approximate Hessian
```

examples/samples/spatially_variant_penalty_strength.par
```diff
@@ -0,0 +1,60 @@
...
+  input image filename := 
+
+  ; The output filename of the spatially variant penalty strength image
```
KrisThielemans, line 8:
```suggestion
  ; The output filename of the row-sum
```

examples/samples/spatially_variant_penalty_strength.par
```diff
@@ -0,0 +1,60 @@
...
+        disable caching := 0
+        store only basic bins in cache := 1
+        restrict to cylindrical fov := 1
+        number of rays in tangential direction to trace for each bin := 50
+        use actual detector boundaries := 1
+        do symmetry 90degrees min phi := 1
+        do symmetry 180degrees min phi := 1
+        do symmetry swap segment := 1
+        do symmetry swap s := 1
+        do symmetry shift z := 1
```
KrisThielemans, line 44:
```suggestion
        number of rays in tangential direction to trace for each bin := 50
```

src/include/stir/recon_buildblock/Hessian_row_sum.h
```diff
@@ -0,0 +1,154 @@
...
+#include "stir/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.h"
+#include "stir/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndGatedProjDataWithMotion.h"
+#include "stir/recon_buildblock/PoissonLogLikelihoodWithLinearKineticModelAndDynamicProjectionData.h"
+
+START_NAMESPACE_STIR
+
+class Succeeded;
+
+/*! \ingroup recon_buildblock
+  \brief Implementations of the two spatially variant penalty strength (kappa) computation methods propose by Tsai et.al (2020).
```
KrisThielemans, line 45: change all references to kappa I guess. row-sum or sqrt row-sum by the way?

src/include/stir/recon_buildblock/Hessian_row_sum.h
```diff
@@ -0,0 +1,154 @@
...
+    //@{
+    bool get_compute_with_penalty();
+    void set_compute_with_penalty(bool with_penalty);
+    //@}
+
+    //! Computes the objective function Hessian at the current image estimate. Can use the penalty's Hessian
+    //! kappa_image_sptr is the
+    void compute_Hessian_row_sum();
+
+    //! Computes the approximate Hessian of the objective function. Cannot use penalty's Hessian
```
KrisThielemans, line 113: > Cannot use penalty's Hessian

Probably best to explain way somewhere. Code should check that the penalty is zero?

src/recon_buildblock/Hessian_row_sum.cxx
```diff
@@ -0,0 +1,244 @@
...
+Hessian_row_sum<TargetT>::
+set_use_approximate_hessian(bool use_approximate)
+{
+  use_approximate_hessian = use_approximate;
+}
+
+template <typename TargetT>
+bool
+Hessian_row_sum<TargetT>::
+get_compute_with_penalty()
```
KrisThielemans, line 157:
```suggestion
get_compute_with_penalty() const
```

src/recon_buildblock/Hessian_row_sum.cxx
```diff
@@ -0,0 +1,244 @@
...
+{
+  // We could simply use image->get_empty_copy() but to be sure all voxels are 0, we set it
+  output_target_sptr =  unique_ptr<TargetT>(image->get_empty_copy());
+  std::fill(output_target_sptr->begin_all(), output_target_sptr->end_all(), 0.F);
+}
+
+template <typename TargetT>
+bool
+Hessian_row_sum<TargetT>::
+get_use_approximate_hessian()
```
KrisThielemans, line 141:
```suggestion
get_use_approximate_hessian() const
```

KrisThielemans (changes requested):
- Renaming to `Hessian-row-sum` is good, except that
   - I guess we compute the sqrt of it?
   - utility and par file still has the old name

- `reset_output_target_sptr` is very non-STIR. I think this function has to be `set_up()`. I should also call `set_up` for the obj-function.

- process_data() should not call set_up(), nor read the input data. The latter can be done in `post_processing()`. Ideally we have a `set_input` as well (callable from Python)

- save_output() is also non-standard. Simplest is to merge into `process_data`?

src/include/stir/recon_buildblock/Hessian_row_sum.h
robbietuk: `accumulate_sub_Hessian_times_input` does the check within it incase penalty is zero.

robbietuk:
- >Renaming to Hessian-row-sum is good, except that
I guess we compute the sqrt of it?
utility and par file still has the old name

You are correct, everything was referencing the Hessian-row-sum but we do not have any need for that as a utility outside of reconstruction classes. I have therefore renamed `Hessian_row_sum` to `SqrtHessianRowSum`. This formatting puts it better inline with other class and filenames names while being a better discription.

- > reset_output_target_sptr is very non-STIR. I think this function has to be set_up(). I should also call set_up for the obj-function.
 process_data() should not call set_up(), nor read the input data. The latter can be done in post_processing().
 save_output() is also non-standard. Simplest is to merge into process_data?

Done 

- >Ideally we have a set_input as well (callable from Python)

I have `set_input_image()`, which I think we should keep as the name because the objective function has the `set_input()` for sinogram data.

 Questions:
 - Should I alter some of the example parameter files and reconstruction class **documentation** referencing this utility for `kappa_image_filename` in the prior keys? 
 - This then leads to the discussion point, what should the utility (`compute_spatially_variant_penalty_strength`) and example parameter file (`compute_spatially_variant_penalty_strength.par`) be named? We have three conventions going right now, `SqrtHessianRowSum` w.r.t the STIR object used to compute them. The ultility `compute_spatially_variant_penalty_strength`, and the keys, variables, and methods used in the objective function (`kappa`). The utility and reconstruction keys (user facing) should be the same.

robbietuk:
After discussion, the Utility should be renamed to `compute_sqrt_Hessian_row_sum`. "Kappa" references may be found throughout the code and the documentation. In future this should be corrected but is out of scope of this PR. For now, the users guide entry to `compute_sqrt_Hessian_row_sum` and `kappa image :=` should reference one another.

src/recon_buildblock/SqrtHessianRowSum.cxx
```diff
@@ -0,0 +1,237 @@
...
+    compute_Hessian_row_sum();
+  }
+
+  // Square Root the output of the Hessian_row_sum methods
+  std::for_each(output_target_sptr->begin_all(), output_target_sptr->end_all(),
+                [](float& a) { return a=sqrt(a); } );
+
+  // Save the output
+  output_file_format_sptr->write_to_file(output_filename, *output_target_sptr);
+  info("Output image of sqrt Hessian row sum has been computed and saved as " + output_filename + ".");
```
KrisThielemans, line 196:
```suggestion
  if (!output_filename.empty())
    {
      output_file_format_sptr->write_to_file(output_filename, *output_target_sptr);
      info("Output image of sqrt Hessian row sum has been computed and saved as " + output_filename + ".");
    }
```

src/recon_buildblock/SqrtHessianRowSum.cxx
```diff
@@ -0,0 +1,237 @@
...
+  this->set_defaults();
+}
+
+template <typename TargetT>
+void
+SqrtHessianRowSum<TargetT>::
+set_defaults()
+{
+  output_file_format_sptr = OutputFileFormat<TargetT>::default_sptr();
+  output_filename = "SqrtHessianRowSum.hv";
```
KrisThielemans, line 52:
```suggestion
  output_filename = "";
```

src/recon_buildblock/SqrtHessianRowSum.cxx
```diff
@@ -0,0 +1,237 @@
...
+  set_use_approximate_hessian(true);
+  set_compute_with_penalty(false);
+}
+
+template <typename TargetT>
+void
+SqrtHessianRowSum<TargetT>::initialise_keymap()
+{
+  parser.add_start_key("Square Root Hessian Row Sum Computation Parameters");
```
KrisThielemans, line 61:
```suggestion
  parser.add_start_key("SqrtHessianRowSum Parameters");
```

examples/samples/compute_sqrt_Hessian_row_sum.par
```diff
@@ -0,0 +1,52 @@
+Square Root Hessian Row Sum Computation Parameters :=
```
KrisThielemans, line 1:
```suggestion
SqrtHessianRowSum Parameters :=
```

KrisThielemans (changes requested):

robbietuk:
Awaiting Travis

KrisThielemans:
@robbietuk Travis is happy. All good to merge?

robbietuk:
Very 😄

PR #740, opened 2020-11-09
provide more defaults for ScatterEstimation

KrisThielemans:
Trying to make it easier to run scatter estimation by removing need for files...

would be nice to get rid of need for `run_reconstruction.par` but difficulty is number of subsets.


KrisThielemans:
Appveyor failed because of a timeout for one job, which is unrelated to this PR.

PR #746, opened 2020-11-13
Update master from rel4 13 nov 2020

KrisThielemans:


KrisThielemans:
merging as previous run failed just one job where the scatter threshold was just exceeded (5.05% as opposed to 5%) in one job. I've therefore increased it to 6%.

Issue #694, opened 2020-09-22
Generalisation of Priors class


robbietuk:
There is a lot of code duplication for each of the priors. The priors implemented right now are:
- Quadratic
- Relative Difference
- Parallel Level Sets
- FilterRoot

There are additional priors that could easily be added: 
- Logcosh
- Huber
- Bowsher

Each of these priors has its own iterator of the computation of value and gradient and generally act over a local neighbourhood. 
All priors (except FilterRoot) an implementation of the usage of the spatially variable penalty strength (kappa). 
There are a selection of priors (i.e. QP, logcosh, hubber) can utilise `parabolic_surrogate_curvature`. 

It seems that a lot of this functionality should belong in `Generalised Prior` (https://github.com/UCL/STIR/blob/master/src/recon_buildblock/GeneralisedPrior.cxx) or more likely an intermediary class, e.g. `Gibbs prior` that contains all the common methods, such as an iterator for the function and the gradient.

This issue is just to note some ideas regarding the generalisation of the priors.

robbietuk:
#662 _(Unable to set/get prior kappa image from STIR-python (matlab))_ and #707 _(STIR-python OSMAPOSL priors are GeneralisedPrior objects rather than child class priors)_ are related to this.

robbietuk:
In light of #901, I am more so inclinded to add an intermediate class, i.e. `Gibbs prior`, that the Quadratic, Relative Difference Prior, Parallel Level Sets, LogCosh, and (Huber + Bowsher) are child classes of. This class should contain not only iterators but virtual functions for gradients and Hessian second derivitives. Each of the classes can then override those functions.

In #901 I have begun to add `off_diagonal_second_derivative` and `diagonal_second_derivative` methods, which will make `accumulate_Hessian_times_input` for QP, logcosh and RDP 100% code duplication.

KrisThielemans:
This was all done in #1629

PR #758, opened 2020-11-26
Logcosh amendments

robbietuk:
Minor changes to Logcosh I forgot. 

- Logcosh was not setup properly in `recon_buildblock_registries` and therefore methods such as `OSMAPOSL` could not use it. The error was:
```
FactoryRegistry: key LogCosh not found in current registry
5 possible values are:
FilterRootPrior
None
PLS
Quadratic
Relative Difference Prior
```
 - Renamed `Logcosh` to `Logcosh Prior`

 - Updated the `UsersGuide` to have a Logcosh subsection and example usage.

### DIFF
--- a/documentation/STIR-UsersGuide.tex
+++ b/documentation/STIR-UsersGuide.tex
@@ -4200,7 +4200,7 @@ This implements a quadratic Gibbs prior. The gradient of the prior is computed a
   \end{verbatim}
 
 
-{ \subsubsubsection{Relative Difference}
+{ \subsubsubsection{Relative Difference} \label{sec:priors:Relative_Difference}
 }
 This implements a Relative Difference Prior RDP, proposed by J. Nuyts, et.al., 2002. The gradient of the prior is computed as follows:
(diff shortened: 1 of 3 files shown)


KrisThielemans:
looks like we failed to standardise naming for `registered_name` and the parsing keywords. Will have to do this when we revise the whole prior hierarchy (#694). somewhat related is also #759.

In any case, I don't think we want the `Prior` in the `registered_name` as we have
```
prior type := Logcosh Prior
```
makes more sense to have
```
prior type := Logcosh
```

Best to create yet another issue for standardising the `registered_name` and add it to the developer guide.

robbietuk:
I agree, `prior type := ` should be Logcosh/Quadratic/Relative Difference. The start keys should include the word prior e.g. `Logcosh Prior Parameters:=`, `Quadratic Prior Parameters:=`, or `Relative Difference Prior Parameters:=`

I will adapt for Logcosh and create a `registered_name` issue.

Issue #765, opened 2020-12-04
apxby backwards compatibility


KrisThielemans:
@gfardell SIRF is currently breaking on Travis ([example](https://travis-ci.org/github/SyneRBI/SIRF/jobs/747224489)
```
/home/travis/build/SyneRBI/SIRF/src/xSTIR/cSTIR/stir_data_containers.cpp:109:13: error: ‘class stir::ProjData’ has no member named ‘axpby’
     data()->axpby(a, *x->data(), b, *y->data());
```
admittedly, this is on https://github.com/SyneRBI/SIRF/pull/543 but the failure has nothing to do with that PR.

@evgueni-ovtchinnikov , heads-up on this.

Obviously, we will need to adapt SIRF according to new convention. @gfardell can you create a SIRF issue for that (if it doesn't exist yet)

gfardell:
https://github.com/SyneRBI/SIRF/issues/834

gfardell:
Ah, I see the problem.
https://github.com/UCL/STIR/blob/f2e894de6d861114cc9525659e1948e11724a202/src/include/stir/ProjData.h#L310

I updated the class to use `xapby` instead of `axpby`. Should I change that back for now, or include both versions? Or should we just focus on updating SIRF?

gfardell:
https://github.com/UCL/STIR/blob/f2e894de6d861114cc9525659e1948e11724a202/src/include/stir/ProjDataInMemory.h#L103 also now uses `xapyb`

evgueni-ovtchinnikov:
I vote for having both versions  for now

KrisThielemans:
that was the original plan, yes

KrisThielemans:
fixed by #767

evgueni-ovtchinnikov:
SIRF Travis builds  2768.3 and 2768.4 failed with the same `no member named ‘axpby’` error.

KrisThielemans:
This seems to be on the `DEVEL` builds only. I've now merged `release_4` (and hence #767) on to `master` so it should be fixed now.

I've restarted the 2 failed travis jobs in https://travis-ci.org/github/SyneRBI/SIRF/builds/748388755. Please close this if they succeed (or fail because of a different reason!)

evgueni-ovtchinnikov:
all well now, thanks!

PR #766, opened 2020-12-04
Axpby fix

gfardell:
@KrisThielemans removing this line might be needed for #765. It was needed before I unset the macro, but I guess now should return full (old) functionality.


KrisThielemans:
no, this still breaks. In any case, #765 is a C++ error. (SIRF doesn't use STIR's Python interface)

PR #769, opened 2020-12-09
Logcosh function and approximation when value >30

### LINKED ISSUE #768: Logcosh value evaluation returns inf with larger values
Interesting one, this time with exponentials.
The logcosh value evaluation 
https://github.com/UCL/STIR/blob/10e5fd409d62ac21fa9453ecf6238ac3748544b6/src/recon_buildblock/LogcoshPrior.cxx#L292-L295
will return `inf` when `voxel_diff > 90.3167` (approx.) because of the evalution of `cosh(voxel_diff) ~=8.3e+38`. This is above the `float` max value of 2.7e+34.

Obviously this computation could be performed in `double` but will only allow evaluation up to `voxel_diff ~=  700`, no great improvement.

I think the function evaluation need to be split. The trigonometric identifier:`cosh(x) = ( e^x + e^(-x) )/ 2` gives one method. At `|x|>30` the impact of the e^(-|x|) is so minimal. Apply the logarithm `log[(1/2) * e^|x|] = log(1/2) + |x|` and the impact of missing term is mininal. 

Shall write a new function in the header for this, something like:
```
elemT voxel_diff = fabs(current_image_estimate[z][y][x] - current_image_estimate[z+dz][y+dy][x+dx]);
elemT loghalf = log(1/2);  // Define this much earlier in the method outside of for loops
if ( voxel_diff < 30){
   elemT current = weights[dz][dy][dx] * 
         (1/(this->scalar*this->scalar))*log(cosh(this->scalar*voxel_diff));
} else {
   elemT current = weights[dz][dy][dx] * 
         (1/(this->scalar*this->scalar)) * ( this->scalar*voxel_diff + loghalf ) ;
}
```
and add a little documentation explaining this rational.

robbietuk:
Fixes #768

At the threshold of 30 the `log(cosh(x)) = x - log(2)` within numerical precision.

### DIFF
--- a/src/include/stir/recon_buildblock/LogcoshPrior.h
+++ b/src/include/stir/recon_buildblock/LogcoshPrior.h
@@ -204,16 +204,16 @@ private:
     static inline float logcosh(const float d)
     {
       const float x = fabs(d);
-      if ( x < 30 ){
+      if ( x < 30.f ){
         return log(cosh(x));
       } else {
-        return x + log(0.5);
+        return x + log(0.5f);
       }
     }
 
     //! The surrogate of the logcosh function is tanh(x)/x
     /*!
-     * @param d is should be the difference between the ith and jth voxel.
+     * @param d should be the difference between the ith and jth voxel.
      However, it will use the taylor expansion if the x is too small (to prevent division by 0).
      * @param scalar is the logcosh scalar value controlling the priors transition between the quadratic and linear behaviour
      * @return the surrogate of the log-cosh function
--- a/src/include/stir/recon_buildblock/LogcoshPrior.h
+++ b/src/include/stir/recon_buildblock/LogcoshPrior.h
@@ -196,9 +196,24 @@ private:
     //! Spatially variant penalty penalty image ptr
     shared_ptr<DiscretisedDensity<3,elemT> > kappa_ptr;
 
+    //! The Log(cosh()) function and its approximation
+    /*!
+     Cosh(x) = 0.5(e^x + e^-x) is an exponential function.
+     Makes the approximation that when |x| > 30, e^-|x| << e^|x| and therefore log(cosh(x)) ~= |x| + log(0.5)
+    */
+    static inline float logcosh(const float d)
+    {
+      const float x = fabs(d);
+      if ( x < 30 ){
+        return log(cosh(x));
+      } else {
+        return x + log(0.5);
+      }
+    }
+
     //! The surrogate of the logcosh function is tanh(x)/x
     /*!
-     * @param d is should be the difference between the ith and jth voxel. The scalar factor may be included.
+     * @param d is should be the difference between the ith and jth voxel.
      However, it will use the taylor expansion if the x is too small (to prevent division by 0).
      * @param scalar is the logcosh scalar value controlling the priors transition between the quadratic and linear behaviour
      * @return the surrogate of the log-cosh function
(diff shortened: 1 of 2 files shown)


src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -196,9 +196,24 @@ class LogcoshPrior:  public
...
 
+    //! The Log(cosh()) function and its approximation
+    /*!
+     Cosh(x) = 0.5(e^x + e^-x) is an exponential function.
+     Makes the approximation that when |x| > 30, e^-|x| << e^|x| and therefore log(cosh(x)) ~= |x| + log(0.5)
+    */
+    static inline float logcosh(const float d)
+    {
+      const float x = fabs(d);
+      if ( x < 30 ){
```
KrisThielemans, line 207:
```suggestion
      if ( x < 30 .F){
```

src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -196,9 +196,24 @@ class LogcoshPrior:  public
...
+        return log(cosh(x));
+      } else {
+        return x + log(0.5);
```
KrisThielemans, line 210:
```suggestion
        return x + log(0.5.F);
```

src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -196,9 +196,24 @@ class LogcoshPrior:  public
...
+      }
+    }
+
     //! The surrogate of the logcosh function is tanh(x)/x
     /*!
-     * @param d is should be the difference between the ith and jth voxel. The scalar factor may be included.
+     * @param d is should be the difference between the ith and jth voxel.
```
KrisThielemans, line 216:
```suggestion
     * @param d should be the difference between the ith and jth voxel.
```

KrisThielemans (changes requested):
interesting! I like your solution

src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -196,9 +196,24 @@ class LogcoshPrior:  public
     //! Spatially variant penalty penalty image ptr
     shared_ptr<DiscretisedDensity<3,elemT> > kappa_ptr;
 
+    //! The Log(cosh()) function and its approximation
+    /*!
+     Cosh(x) = 0.5(e^x + e^-x) is an exponential function.
+     Makes the approximation that when |x| > 30, e^-|x| << e^|x| and therefore log(cosh(x)) ~= |x| + log(0.5)
```
KrisThielemans, line 202: @zeljkozeljko can you add your line on the bound here? https://github.com/UCL/STIR/issues/768#issuecomment-742181217

please add `[ci skip]` in the first line of the commit message

zeljkozeljko: I'm not sure I completely understood your message, but I can at least copy that comment here
Screenshot 2020-12-10 at 01 49 40
ie the  log(Cosh(x)) = -log(2) + |x| + log(1 + e^(-2|x|)) = -log(2) + |x|+ O(10^(-27)), for |x|>30

KrisThielemans: best to put it in the actual code. you should be able to edit the file on github. admittedly it'll mean creating a fork etc, so possibly more hassle than it's worth. Alternatively @robbietuk or me do it.

robbietuk: Done, it was easier for me to do it I think.

PR #770, opened 2020-12-15
BinNormalisation takes timing information from the ExamInfo member

danieldeidda:
BinNormalisation takes timing information from ExamInfo member and there is no nee to pass start and end time to each functions

### DIFF
--- a/documentation/release_5.0.htm
+++ b/documentation/release_5.0.htm
@@ -190,7 +190,7 @@ from the above):</H2>
 
 <h3>Backward incompatibities</h3>
 <ul>
-    <li>Classes that use<code>InterfilePDFSHeader</code> now contain a <code>shared_ptr&lt;ProjDataInfo&gt;</code> instead of a raw pointer, removing a memory leak.</li>
+    <li>Classes that use <code>InterfilePDFSHeader</code> now contain a <code>shared_ptr&lt;ProjDataInfo&gt;</code> instead of a raw pointer, removing a memory leak.</li>
     <li>Changes improving safety of use of <code>shared_ptr</code>:<br />
       In the previous version of STIR, the use of <code>shared_ptr</code> allowed unsafe access to
       the objects (although this never happened in distributed STIR code). We now prevent this with changes to the class interface (although there is still work to do):
--- a/documentation/release_5.0.htm
+++ b/documentation/release_5.0.htm
@@ -206,7 +206,7 @@ from the above):</H2>
     <li><code>PlasmaData::set_if_decay_corrected</code> and <code>PlasmaData::get_if_decay_corrected</code> renamed to <code>PlasmaData::set_is_decay_corrected</code> and <code>PlasmaData::get_is_decay_corrected</code></li>
     <li>Remove overloaded <code>PatlakPlot::get_model_matrix</code>, now it can only be called using class members</li>
     <li><code>BinNormalisation::set_up()</code> now need exam_info_sptr as input.</li>
-    <li><code>BinNormalisation::apply()</code> and <code>undo</code> now no longer accept start_time and end_time as they are taken from exam_info. To allow this to happen, class <code>Bin</code> has now an extra member <code>time_frame()</code>. Note that it has to be initialised explicitly (not via a constructor).</li>
+    <li><code>BinNormalisation::apply()</code> and <code>undo</code> now no longer accept start_time and end_time as they are taken from exam_info. To allow this to happen, class <code>Bin</code> has now an extra member <code>time_frame()</code>. Note that it defaults to 1. If this is incorrect, it has to be initialised explicitly (not via a constructor).</li>
 </ul>
 
 <h3>New functionality</h3>
--- a/src/include/stir/Bin.h
+++ b/src/include/stir/Bin.h
@@ -76,6 +76,8 @@ public:
   inline float get_bin_value()const; 
   //! set the value to be back projected 
   inline void set_bin_value( float v );
+  //! set bin timeframe 
+  inline void set_time_frame( int t );
   
   //! accumulate voxel's contribution during forward projection 
   inline Bin&  operator+=(const float dx);
@@ -98,6 +100,7 @@ private :
   int  axial_pos; 
   int  tangential_pos; 
   float bin_value;
+  int time_frame;
   
   
 };
--- a/src/include/stir/Bin.h
+++ b/src/include/stir/Bin.h
@@ -82,8 +82,6 @@ public:
   inline float get_bin_value()const; 
   //! set the value to be back projected 
   inline void set_bin_value( float v );
-  //! set time-frame number (1-based) 
-  inline void time_frame_num( int t );
   
   //! accumulate voxel's contribution during forward projection 
   inline Bin&  operator+=(const float dx);
--- a/src/include/stir/Bin.h
+++ b/src/include/stir/Bin.h
@@ -65,11 +65,15 @@ public:
   inline int tangential_pos_num()  const; 
   //! get view number
   inline int view_num() const; 
+  //! get time-frame number (1-based)
+  inline int time_frame_num() const;
   
   inline int& axial_pos_num(); 
   inline int& segment_num(); 
   inline int& tangential_pos_num(); 
   inline int& view_num(); 
+  inline int& time_frame_num();
+  
   
   //! get an empty copy
   inline Bin get_empty_copy() const;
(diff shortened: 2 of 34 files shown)


src/listmode_buildblock/LmToProjData.cxx
```diff
@@ -399,16 +399,16 @@ get_bin_from_event(Bin& bin, const ListEvent& event) const
...
-     const double start_time = frame_defs.get_start_time(current_frame_num);
-     const double end_time =frame_defs.get_end_time(current_frame_num);
-#endif
+//#ifndef FRAME_BASED_DT_CORR
+//     const double start_time = current_time;
+//     const double end_time = current_time;
+//#else
+//     const double start_time = frame_defs.get_start_time(current_frame_num);
+//     const double end_time =frame_defs.get_end_time(current_frame_num);
+//#endif
```
KrisThielemans, line 408: this is the bit that needs some thought here.

danieldeidda: yes I was not sure about this indeed.

KrisThielemans: this will be resolved once we fix #794

KrisThielemans:
This PR will make us more vulnerable to https://github.com/UCL/STIR/issues/793. Need to confirm if the `set_up` of the objective function uses the time frame info (i.e. exam_info) of the proj-data to set up the normalisation.

KrisThielemans (changes requested):
Also needs a change https://github.com/UCL/STIR/blob/5706f9fad5e219c3e82d7acd7fba8bafea6f791f/src/recon_buildblock/distributableMPICacheEnabled.cxx#L135, maybe other places there or in `experimental` as well.

I still need to think about `LmToProjData` and the `set_up()` of the `BinNormalisation` objects there. It currently happens with the `exam_info` of the listmode data. I think it needs to be done with one with the appropriate time frame definitions. The sad thing of course is that the normalisation there is a bad thing to do, but it is there because of how people did it when using LOR-repositioning.

KrisThielemans:
could you also sort out the conflict?

danieldeidda:
> 
> 
> Also needs a change
> 
> https://github.com/UCL/STIR/blob/5706f9fad5e219c3e82d7acd7fba8bafea6f791f/src/recon_buildblock/distributableMPICacheEnabled.cxx#L135
> , maybe other places there or in `experimental` as well.
> 
> I still need to think about `LmToProjData` and the `set_up()` of the `BinNormalisation` objects there. It currently happens with the `exam_info` of the listmode data. I think it needs to be done with one with the appropriate time frame definitions. The sad thing of course is that the normalisation there is a bad thing to do, but it is there because of how people did it when using LOR-repositioning.

Should I put override in all the experimental as well?

KrisThielemans:
> Should I put override in all the experimental as well?

can just as well do it. thanks!

KrisThielemans:
Need to change
https://github.com/UCL/STIR/blob/90ffafbd0cd73c4a8f9a2d8db6e03201d75317c2/src/listmode_buildblock/LmToProjData.cxx#L324
and below to something like
```c++
 auto all_frames_exam_info_sptr std::make_shared<ExamInfo>(lm_data_ptr->get_exam_info());
 all_frames_exam_info_sptr .set_time_frame_definitions(frame_defs);
 ... normalisation_ptr->set_up(all_frames_exam_info_sptr , proj_data_info_cyl_uncompressed_ptr) 
```

KrisThielemans (approved):
should be fine now

KrisThielemans (changes requested):
Please change `Bin::set_time_frame()` to `Bin::time_frame()` to fit with the others like `view_num()` etc (I know the `bin_value` is an outlier sadly).

Weirdly, we don't have an accessor for it. That means that `BinNormalisation` objects that use time cannot be currently correct. This is probably only the `BinNormalisationFromECAT7`.

danieldeidda:
> inline int&  time_frame_num()

Sorry I am confused. all these segment_num() etc seem getter functions to me they all return a member

src/include/stir/Bin.h
```diff
@@ -76,6 +82,8 @@ class Bin
   inline float get_bin_value()const; 
   //! set the value to be back projected 
   inline void set_bin_value( float v );
+  //! set time-frame number (1-based) 
+  inline void time_frame_num( int t );
```
KrisThielemans, line 86: remove this as not needed

src/listmode_buildblock/LmToProjData.cxx
```diff
@@ -653,6 +655,7 @@ process_data()
 		     // set value in case the event decoder doesn't touch it
 		     // otherwise it would be 0 and all events will be ignored
 		     bin.set_bin_value(1);
+             bin.time_frame_num(current_frame_num);
```
KrisThielemans, line 658:
```suggestion
             bin.time_frame_num() = current_frame_num;
```

src/include/stir/recon_buildblock/BinNormalisation.h
```diff
@@ -129,10 +130,18 @@ class BinNormalisation : public RegisteredObject<BinNormalisation>
 
     The default value for the symmetries means that TrivialDataSymmetriesForBins will be used.
   */
-  void undo(ProjData&,const double start_time, const double end_time, 
+  void undo(ProjData&, 
             shared_ptr<DataSymmetriesForViewSegmentNumbers> = shared_ptr<DataSymmetriesForViewSegmentNumbers>()) const; 
-  
+
+  //! old interface. do not use
+  STIR_DEPRECATED void undo(ProjData& p,const double /*start_time*/, const double /*end_time*/,
```
KrisThielemans, line 137: let's do the same for `apply`

documentation/release_5.0.htm
```diff
@@ -206,7 +206,7 @@ <h3>Backward incompatibities</h3>
     <li><code>PlasmaData::set_if_decay_corrected</code> and <code>PlasmaData::get_if_decay_corrected</code> renamed to <code>PlasmaData::set_is_decay_corrected</code> and <code>PlasmaData::get_is_decay_corrected</code></li>
     <li>Remove overloaded <code>PatlakPlot::get_model_matrix</code>, now it can only be called using class members</li>
     <li><code>BinNormalisation::set_up()</code> now need exam_info_sptr as input.</li>
-    <li><code>BinNormalisation::apply()</code> now does not need start_time and end_time as they are taken from exam_info. The same will be done with <code>undo()</code></li>
+    <li><code>BinNormalisation::apply()</code> and <code>undo</code> now no longer accept start_time and end_time as they are taken from exam_info. To allow this to happen, class <code>Bin</code> has now an extra member <code>time_frame()</code>. Note that it has to be initialised explicitly (not via a constructor).</li>
```
KrisThielemans, line 209:
```suggestion
    <li><code>BinNormalisation::apply()</code> and <code>undo</code> now no longer accept start_time and end_time as they are taken from exam_info. To allow this to happen, class <code>Bin</code> has now an extra member <code>time_frame()</code>. Note that it defaults to 1. If this is incorrect, it has to be initialised explicitly (not via a constructor).</li>
```

KrisThielemans (changes requested):

documentation/release_5.0.htm
```diff
@@ -190,7 +190,7 @@ <h3>Major bugs fixed</h3>
 
 <h3>Backward incompatibities</h3>
 <ul>
-    <li>Classes that used<code>InterfilePDFSHeader</code> now contains a <code>shared_ptr&lt;ProjDataInfo&gt;</code> instead of a raw pointer, removing a memory leak.</li>
+    <li>Classes that use<code>InterfilePDFSHeader</code> now contain a <code>shared_ptr&lt;ProjDataInfo&gt;</code> instead of a raw pointer, removing a memory leak.</li>
```
KrisThielemans, line 193:
```suggestion
    <li>Classes that use <code>InterfilePDFSHeader</code> now contain a <code>shared_ptr&lt;ProjDataInfo&gt;</code> instead of a raw pointer, removing a memory leak.</li>
```

KrisThielemans (changes requested):
tiny typo

KrisThielemans (approved):
all fine now. Let's just see if SIRF compiles with this.

danieldeidda:
> 
> 
> all fine now. Let's just see if SIRF compiles with this.

I have tried SIRFSuperbuild with this and it builds

Issue #750, opened 2020-11-15
modify stir:Scanner to allow changing the number of virtual crystals


KrisThielemans:
values are currently hard-wired. This is even more confusing since #617

PR #776, opened 2020-12-22
Virtual xtal interface.

NikEfth:
Interface for the virtual xtals. 

Update CListModeDataROOT.cxx, Scanner.h, and 4 more files...

### DIFF
--- a/src/IO/InterfileHeader.cxx
+++ b/src/IO/InterfileHeader.cxx
@@ -600,6 +600,12 @@ InterfilePDFSHeader::InterfilePDFSHeader()
   num_transaxial_crystals_per_block = 0;
   add_key("number of crystals_per_block in transaxial direction",
 	  &num_transaxial_crystals_per_block);
+  num_virtual_axial_crystals_per_block = 0;
+  add_key("number of virtual crystals_per_block in axial direction",
+      &num_virtual_axial_crystals_per_block);
+  num_virtual_transaxial_crystals_per_block = 0;
+  add_key("number of virtual crystals_per_block in transaxial direction",
+      &num_virtual_transaxial_crystals_per_block);
   num_axial_crystals_per_singles_unit = -1;
   add_key("number of crystals_per_singles_unit in axial direction",
 	  &num_axial_crystals_per_singles_unit);
@@ -1095,6 +1101,10 @@ bool InterfilePDFSHeader::post_processing()
       num_axial_crystals_per_block = guessed_scanner_ptr->get_num_axial_crystals_per_block();
     if (num_transaxial_crystals_per_block<=0)
       num_transaxial_crystals_per_block = guessed_scanner_ptr->get_num_transaxial_crystals_per_block();
+    if (num_virtual_axial_crystals_per_block<=0)
+      num_virtual_axial_crystals_per_block = guessed_scanner_ptr->get_num_virtual_axial_crystals_per_block();
+    if (num_virtual_transaxial_crystals_per_block<=0)
+      num_virtual_transaxial_crystals_per_block = guessed_scanner_ptr->get_num_virtual_transaxial_crystals_per_block();
     if (num_axial_crystals_per_singles_unit < 0) 
       num_axial_crystals_per_singles_unit = 
         guessed_scanner_ptr->get_num_axial_crystals_per_singles_unit();
@@ -1191,6 +1201,22 @@ bool InterfilePDFSHeader::post_processing()
 	warning("Interfile warning: num_transaxial_crystals_per_block (%d) is expected to be %d.\n",
 		num_transaxial_crystals_per_block, guessed_scanner_ptr->get_num_transaxial_crystals_per_block());
 	mismatch_between_header_and_guess = true;
+      }
+    if (
+    guessed_scanner_ptr->get_num_virtual_axial_crystals_per_block()>0 &&
+    num_virtual_axial_crystals_per_block!= guessed_scanner_ptr->get_num_virtual_axial_crystals_per_block())
+      {
+    warning("Interfile warning: num_virtual_axial_crystals_per_block (%d) is expected to be %d.\n",
+        num_virtual_axial_crystals_per_block, guessed_scanner_ptr->get_num_virtual_axial_crystals_per_block());
+        mismatch_between_header_and_guess = true;
+      }
+    if (
+    guessed_scanner_ptr->get_num_virtual_transaxial_crystals_per_block()>0 &&
+    num_virtual_transaxial_crystals_per_block!= guessed_scanner_ptr->get_num_virtual_transaxial_crystals_per_block())
+      {
+    warning("Interfile warning: num_transaxial_crystals_per_block (%d) is expected to be %d.\n",
+        num_transaxial_crystals_per_block, guessed_scanner_ptr->get_num_transaxial_crystals_per_block());
+    mismatch_between_header_and_guess = true;
       }
     if ( guessed_scanner_ptr->get_num_axial_crystals_per_singles_unit() > 0 &&
          num_axial_crystals_per_singles_unit != 
@@ -1302,6 +1328,8 @@ bool InterfilePDFSHeader::post_processing()
 		num_transaxial_blocks_per_bucket,
 		num_axial_crystals_per_block,
 		num_transaxial_crystals_per_block,
+                num_virtual_axial_crystals_per_block,
+                num_virtual_transaxial_crystals_per_block,
 		num_axial_crystals_per_singles_unit,
                 num_transaxial_crystals_per_singles_unit,
                 num_detector_layers,
--- a/src/buildblock/Scanner.cxx
+++ b/src/buildblock/Scanner.cxx
@@ -124,7 +124,7 @@ Scanner::Scanner(Type scanner_type)
     set_params(E931, string_list("ECAT 931"),  
                8, 192, 2 * 256, 
                510.0F, 7.0F, 13.5F, 3.129F, 0.0F, 
-               2, 4, 4, 8, 4, 8 * 4, 1,
+               2, 4, 4, 8, 0, 0, 4, 8 * 4, 1,
                0.37f, 511.f);
     // 16 BUCKETS per ring in TWO rings - i.e. 32 buckets in total
 
@@ -135,7 +135,7 @@ Scanner::Scanner(Type scanner_type)
     set_params(E951, string_list("ECAT 951"), 
                16, 192, 2 * 256, 
                510.0F, 7.0F, 6.75F, 3.12932F, 0.0F, 
-               1, 4, 8, 8, 8, 8 * 4, 1);
+               1, 4, 8, 8, 0, 0, 8, 8 * 4, 1);
     break;
 
   case E953:
@@ -143,7 +143,7 @@ Scanner::Scanner(Type scanner_type)
     set_params(E953, string_list("ECAT 953"), 
                16, 160, 2 * 192, 
                382.5F, 7.0F, 6.75F, 3.12932F, static_cast<float>(15.*_PI/180), 
-               1, 4, 8, 8, 8, 8 * 4, 1);
+               1, 4, 8, 8, 0, 0, 8, 8 * 4, 1);
     break;
 
   case E921:
@@ -151,7 +151,7 @@ Scanner::Scanner(Type scanner_type)
     set_params(E921, string_list("ECAT 921", "ECAT EXACT", "EXACT"), 
                24, 192, 2* 192, 
                412.5F, 7.0F, 6.75F, 3.375F, static_cast<float>(15.*_PI/180),
-               1, 4, 8, 8, 8, 8 * 4, 1);
+               1, 4, 8, 8, 0, 0, 8, 8 * 4, 1);
     break;
 
   case E925:
@@ -159,7 +159,7 @@ Scanner::Scanner(Type scanner_type)
     set_params(E925, string_list("ECAT 925", "ECAT ART"), 
                24, 192, 2* 192, 
                412.5F, 7.0F, 6.75F, 3.375F, static_cast<float>(15.*_PI/180),
-               3, 4, 8, 8, 8, 8 * 4, 1);
+               3, 4, 8, 8, 0, 0, 8, 8 * 4, 1);
     break;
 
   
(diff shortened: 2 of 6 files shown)


NikEfth:
This PR addresses the issues raised in #750

src/buildblock/Scanner.cxx
```diff
@@ -782,7 +826,7 @@ check_consistency() const
       {
 	const int dets_axial =
 	  get_num_axial_blocks() *
-	  get_num_axial_crystals_per_block();
+      (get_num_axial_crystals_per_block() + get_num_virtual_axial_crystals_per_block());
```
KrisThielemans, line 829: shouldn't be changed I believe

src/buildblock/Scanner.cxx
```diff
@@ -817,12 +861,12 @@ check_consistency() const
 	      this->get_name().c_str());
     else
       {
-	if ( get_num_detectors_per_ring() % get_num_transaxial_crystals_per_singles_unit() != 0)
+    if ( get_num_detectors_per_ring() % (get_num_transaxial_crystals_per_singles_unit() + get_num_virtual_transaxial_crystals_per_block()) != 0)
```
KrisThielemans, line 864: shouldn't be changed

src/buildblock/Scanner.cxx
```diff
@@ -817,12 +861,12 @@ check_consistency() const
...
 	  { 
 	    warning("Scanner %s: inconsistent transaxial singles unit info:\n"
-		    "\tnum_detectors_per_ring %d should be a multiple of num_transaxial_crystals_per_singles_unit %d",
+            "\tnum_detectors_per_ring %d should be a multiple of num_transaxial_crystals_per_singles_unit + virtual %d",
```
KrisThielemans, line 867: shouldn't be changed

src/buildblock/Scanner.cxx
```diff
@@ -817,12 +861,12 @@ check_consistency() const
...
 		    this->get_name().c_str(),
-		    get_num_detectors_per_ring(), get_num_transaxial_crystals_per_singles_unit()); 
+            get_num_detectors_per_ring(), (get_num_transaxial_crystals_per_singles_unit() + get_num_virtual_transaxial_crystals_per_block()));
```
KrisThielemans, line 869: shouldn't be changed

KrisThielemans (changes requested):
Most of this is fine. However, you're changing the interpretation of the "standard" function like `get_num_axials_crystal_per_block` to mean the "non-virtual" ones, while it was "all of them". Although I agree that the naming is confusing, I'd rather avoid changing this as it probably has much wider consequences. For instance, what does `get_num_rings` mean? It was "number of all rings, non-virtual+virtual". If we'd change that, lots of things will have to be adapted.

NikEfth:
The definition of the scanner overall is very circular and confusing and probably we should re-write from scratch. 

However, I took the lead from the mCT example were the number of the detectors per ring includes the virtual crystals (13+1)*48. Respectively the num_rings should include the virtual rings. 

The changes, I made allow us to design a scanner if a 2-fold way, 1 gap per block (in mCT every 13 crystals) or 1 gap every several blocks. The later is in better agreement with most scanners I would know, where the block separation between blocks is small of e.g. half crystal, but there is 1 or  2 crystal between the rotated blocks (sectors in GATE). 

For example, my scanner has 18 sectors with 4 transaxial blocks with 8 transaxial crystals.
Now both number of detectors per ring 18*(4*8+1) and 18*4*(8+1) can pass the consistency test. 
And soon the CListModeDataROOT will be able to know were to push the gap.  

I think that bein able to reposition the gaps is quite important.

NikEfth:
I would like to add that the modification does not interfere with the singles units, which can still be independent.

src/listmode_buildblock/CListModeDataROOT.cxx
```diff
@@ -263,37 +261,45 @@ check_scanner_match_geometry(std::string& ret, const shared_ptr<Scanner>& scanne
...
     if (scanner_sptr->get_num_rings() != root_file_sptr->get_num_rings())
     {
-        stream << "the number of rings, ";
+        stream << "the number of rings ("<< scanner_sptr->get_num_rings() << ", " <<
+                  root_file_sptr->get_num_rings() << "), ";
         ok = false;
     }
 
-    if (scanner_sptr->get_num_detectors_per_ring() != root_file_sptr->get_num_dets_per_ring())
+    if ((scanner_sptr->get_num_detectors_per_ring() - scanner_sptr->get_num_virtual_transaxial_crystals()) != root_file_sptr->get_num_dets_per_ring())
```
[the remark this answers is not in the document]
KrisThielemans: I guess we could change it after all, if we say that the `root_file_sptr->get_num_dets_per_ring()` returns  the number of physical detectors per ring, which is probably logical.

Needs serious documentation though

src/include/stir/Scanner.inl
```diff
@@ -160,7 +160,7 @@ Scanner::get_num_axial_blocks() const
 int
 Scanner::get_num_transaxial_blocks() const
 {
-  return num_detectors_per_ring/num_transaxial_crystals_per_block;
+  return num_detectors_per_ring/(num_transaxial_crystals_per_block+num_virtual_transaxial_crystals_per_block);
```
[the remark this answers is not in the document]
KrisThielemans: I still disagree on this one. Currently `num_transaxial_crystals_per_block` includes both. We shouldn't be adding.

KrisThielemans:
> The definition of the scanner overall is very circular and confusing and probably we should re-write from scratch

yeah. at some point... You cleaned at least the `set_params` stuff up in #304. Let's not attempt a major rewrite until #304 and #577 are merged.

> The changes, I made allow us to design a scanner if a 2-fold way, 1 gap per block (in mCT every 13 crystals) or 1 gap every several blocks. The later is in better agreement with most scanners I would know, where the block separation between blocks is small of e.g. half crystal, but there is 1 or 2 crystal between the rotated (bigger) blocks (r-sectors in GATE).

I certainly agree that many scanners would have bigger gaps between "buckets" than "blocks". However, I hadn't realised that you wanted to fix that with this PR. The name of the PR and previous conversation led me to think you only wanted to be able to set the number of virtual crystals (which is what most of this PR is about).

In fact, I cannot see how you could fix this without having additional members `num_axial_virtual_crystals_per_bucket` etc (or proably better `num_additional_axial_virtual_crystals_per_bucket` as it'd add up to the one between blocks).

To me adding that facility seems to need a fair amount of work, but maybe I'm wrong.

NikEfth:
Well, I am not really fixing it, for that we would need a ```num_virtual_crystal_per_bucket()``` function, but just letting the consistency test pass and the make the CListModeROOT do the right thing. 
The key logic is that based on the number of detector per ring you set, we can figure out how many virtual crystals there are. Is the divisor of block number or block*crystal. 

I understand that some tests succeed but there might be further implications down the line.

src/buildblock/Scanner.cxx
NikEfth: But this is what you have in mCT.
This is an important comment: If you don't include the virtual crystals in the total number of detectors per ring the unlisting fails. The sinogram is broken. I could send an image to show you.

NikEfth: It has something to do with the inversion in the tangestial positions when the tang_pos is larger than the number of views. Which currently is calculated from the total number of detectors per ring.

src/include/stir/Scanner.inl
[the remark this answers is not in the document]
NikEfth: Then in come down to what we include in the total number of detectors in the template.
If the total number of detectors include the virtual or not. 
My initial notion is that they should not. BUT if you don't you have problems with the get_bin() function (wrong inversions). 

So, until we restructure the class, the total number of rings and detectors per ring should include the virtual.

src/buildblock/Scanner.cxx
```diff
@@ -698,34 +712,64 @@ int
...
+{
+//  switch(get_type())
+//    {
+//    case E1080:
+//    case Siemens_mCT:
+//      return 1;
+//    default:
+//      return 0;
+//    }
+    return num_virtual_axial_crystals_per_block * get_num_axial_blocks();
```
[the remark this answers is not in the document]
NikEfth: Yes I know. 
I don't know what to propose here. Say that you add one that the end. I don't think that it actually changes much. Where could this create a problem? Normalisation? Multiple bed positions? I guess the axial middle of the scanner will include that ring and that might be a problem.

What we could do is check with the num_rings + gaps - 1,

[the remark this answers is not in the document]
NikEfth: @tniknejad Is this something you could change?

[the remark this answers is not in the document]
KrisThielemans: I'm very confused. you introduced this function. We have `get_num_virtual_axial_crystals_per_block` which is well-defined. why do we need this?

NikEfth:
hmm, after all, I introduced the gaps in buckets. 
My main question is whether you prefer the num_of_detectors_per_ring in the template to be the physical detectors only, or both gaps and detectors.

KrisThielemans:
> I think that in the template we should set the number of physical detectors and the then get functions to do the calculations. However, this means that we cannot have set functions. There have to become explicitly set_physical, or very very very well documented. What do you think?

While reading/setting things in terms of physical detectors would make a lot of sense, changing the meaning of `get_num_rings()` to suddenly be `get_num_physical_rings()` would break quite a lot of stuff. Given my pace of accepting PRs, I'd strongly recommend not attempting that... Of course, making it clearer in the documentation what is what would help.

Of course, having members  `get_num_physical_rings()`,  `get_num_non_physical_rings()` (with the sum `get_num_rings()` ) is an excellent idea. Providing an alias for `get_num_rings()` that makes it clear what it means is of course fine as well. We could then consider deprecating the old names.

src/buildblock/Scanner.cxx
KrisThielemans: >  If you don't include the virtual crystals in the total number of detectors per ring the unlisting fails. The sinogram is broken.

of course. Siemens does all calculations/sinograms in terms of "total number". Nowhere in their files there are any "virtual" ones (even not in the norm, presumably because they do gap-filling)

src/include/stir/Scanner.h
```diff
@@ -188,9 +192,10 @@ class Scanner
   //! \name Functions returning geometrical info
   //@{
 
-  //! get number of rings
+  //! get number of rings. This function returns the total number of rings + gaps
   inline int get_num_rings() const;
-  //! get the number of detectors per ring
+  //! get the number of detectors per ring. This function returns the total
+  //! number of detectors + gasp in the ring.
```
KrisThielemans, line 198:
```suggestion
  //! number of (physical and virtual) detectors in the ring.
```

src/include/stir/Scanner.h
```diff
@@ -188,9 +192,10 @@ class Scanner
   //! \name Functions returning geometrical info
   //@{
 
-  //! get number of rings
+  //! get number of rings. This function returns the total number of rings + gaps
```
KrisThielemans, line 195:
```suggestion
  //! get number of rings. This function returns the total number of  (physical and virtual) rings
```

src/include/stir/Scanner.h
```diff
@@ -302,9 +311,11 @@ class Scanner
   // zlong, 08-04-2004, add set_methods
   //! set scanner type
   inline void set_type(const Type & new_type);
-  //! set number of rings
+  //! set number of rings. This function will set the total number of rings
+  //! reseting the virtual crystals to the axial direction to 0.
```
KrisThielemans, line 315:
```suggestion
  //! resetting the virtual crystals in the axial direction to 0.
```

src/include/stir/Scanner.h
```diff
@@ -302,9 +311,11 @@ class Scanner
...
   inline void set_num_rings(const int & new_num);
-  //! set the namber of detectors per ring
+  //! set the namber of detectors per ring. This function will set the total number
+  //! of detectors per ring, reseting the number of virtual crystals to 0.
```
KrisThielemans, line 318:
```suggestion
  //! set the number of detectors per ring. This function will set the total number
  //! of detectors per ring, resetting the number of virtual crystals (in transaxial direction) to 0.
```

src/include/stir/Scanner.inl
```diff
@@ -154,25 +164,28 @@ Scanner::get_num_axial_blocks() const
...
   // need to take this into account.
-  return (num_rings+get_num_virtual_axial_crystals_per_block())/num_axial_crystals_per_block;
+  return (num_rings)/num_axial_crystals_per_block;
 }
 
 int
 Scanner::get_num_transaxial_blocks() const
 {
-  return num_detectors_per_ring/num_transaxial_crystals_per_block;
+  return num_transaxial_crystals_per_block == 0 ? num_detectors_per_ring :
```
KrisThielemans, line 173: `num_transaxial_crystals_per_block` should never be zero. It doesn't make sense in STIR. You might have only 1 block, but no crystals in a block means no crystals

src/include/stir/Scanner.inl
```diff
@@ -154,25 +164,28 @@ Scanner::get_num_axial_blocks() const
...
+                                                  num_detectors_per_ring/(num_transaxial_crystals_per_block);
 }
 
 int
 Scanner::get_num_axial_buckets() const
 {
-  return get_num_axial_blocks()/num_axial_blocks_per_bucket;
+  return num_axial_blocks_per_bucket == 0 ? get_num_axial_blocks() :
```
KrisThielemans, line 180: same as above

src/listmode_buildblock/CListModeDataROOT.cxx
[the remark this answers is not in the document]
KrisThielemans: But I would still find it very confusing if `get_num_dets_per_ring()` means different things for `Scanner` and the `ROOT` stuff

src/IO/InterfileHeader.cxx
```diff
@@ -600,6 +600,12 @@ InterfilePDFSHeader::InterfilePDFSHeader()
   num_transaxial_crystals_per_block = 0;
   add_key("number of crystals_per_block in transaxial direction",
 	  &num_transaxial_crystals_per_block);
+  num_virtual_axial_crystals_per_block = 0;
```
KrisThielemans, line 603: seems we'd need some bucket stuff here as well

KrisThielemans (changes requested):
I've given this another look and added some comments. I didn't check consistency of everything. I also didn't think about the "singles units" stuff.

I guess I'm fine with changing the internal meaning of some of the private variables like `Scanner::num_rings` if you prefer, but I don't think we should change the meaning of `get_num_rings()` etc (which you don't of course, so that's good). It could be confusing for a developer of `Scanner`, but it can be addressed by documentation.

I really think though that we cannot change the meaning of `num_rings` etc in the Interfile header. It's too late for that.

KrisThielemans:
@NikEfth I cannot recall where we were with this, but aside from the comments above, it probably needs a revisit now that #833 has been merged.

Issue #785, opened 2021-01-04
isotope parsing should be lifted to InterfileHeader


KrisThielemans:
It's currently in `InterfileImageHeader`

PR #787, opened 2021-01-05
fixing #785: moving isotope name to interfile header

danieldeidda:


src/IO/InterfileHeader.cxx
```diff
@@ -265,7 +266,12 @@ bool InterfileHeader::post_processing()
       catch(...)
         {}
     }
-
+  
+  if (!isotope_name.empty()){
+      this->exam_info_sptr->set_radionuclide(isotope_name);
+      info("isotope name is " + isotope_name);
```
KrisThielemans, line 272:
```suggestion
```
none of the other parsing keywords provide this feedback. People can get it via `list*_info` etc.

KrisThielemans (changes requested):
thanks. tiny change please (use `[ci skip]`)

PR #788, opened 2021-01-06
Add script to change License

KrisThielemans:
First step in #97.

This script seems to be able to change all files, including C++, Python etc, independent of comment characters.


KrisThielemans:
sorry. on wrong branch

PR #789, opened 2021-01-06
add script to change license to Apache 2.0

KrisThielemans:
First step in #97.

This script seems to be able to change all files, including C++, Python etc, independent of comment characters.

### DIFF
--- a/scripts/maintenance/change-license.sed
+++ b/scripts/maintenance/change-license.sed
@@ -1,0 +1,23 @@
+# sed script to replace the GNU license statement with an Apache 2.0 one.
+# Copyright (C) 2021 University College London
+# SPDX-License-Identifier: Apache-2.0
+# Author: Kris Thielemans
+#
+# Thanks to https://www.grymoire.com/Unix/Sed.html
+#
+/.*This file is free software; you can redistribute it and\/or modify.*/ {
+# first put 10 lines into the pattern space (as the whole GNU license statement takes ~8 lines)
+N;
+N;
+N;
+N;
+N;
+N;
+N;
+N;
+N;
+N;
+# now replace the GNU license statement with an Apache one
+# preserving any characters used at the start of the line
+  s#(.*)This file is free software; you can redistribute it and/or modify.*\n\1it under the terms.*GNU .*General Public License for more details.#\1SPDX-License-Identifier: Apache-2.0#;
+}
--- a/scripts/maintenance/change-license.sh
+++ b/scripts/maintenance/change-license.sh
@@ -1,0 +1,19 @@
+#! /bin/sh
+# This script runs sed to change the GNU license to Apache 2.0
+# Usage:
+#    change-license.sh <some-path>
+# This will change all files in that path and sub-folders.
+# <some-path> can be replaced by arguments to "find" to limit the scope.
+#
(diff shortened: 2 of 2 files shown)


KrisThielemans:
Would need to exclude FORE code unfortunately, as no signed agreement to change that to Apache 2.0

KrisThielemans:
@casperdcl @dvolgyes if you happen to know this kind of stuff, what do you think? It seems to work fine for me.

dvolgyes:
Hi,

Obviously, there are two parts: software development and software distribution.
For instance, let's assume you have Mozilla License and a GPL license.
You can combine these codes on your computer, but you cannot publish them together.
Most licenses don't care about limitations unless you actually distribute code (binary or source).

The other part is when you cannot even use them together, regardless of distribution,
because you need to publish the code even if you don't want to distribute it (AGPLv3).
This case is kind of known as "closing the cloud loophole". E.g. Google/Amazon/Microsoft use GPL linux
internally, but they do not publish modifications because the code is not distributed in any form,
they just sell cloud services. 

So, if there is no AGPL code in the code repository, your users can combine it with anything,
as long as they don't sell a device or software with the code. 

If you want to enable code share, e.g. inside a project, you can always go for permissive licenses,
but you might end up with the MacOS-situation: they use BSD Unix, but they did not publish anything.

Many people feel that Apache is the most liberal licensing. This isn't true, MIT and BSD-new are more
permissive, and you might end up with issues in case  GPL v1 or v2. (unless it is GPLv1/2 or _newer_)

Here is a compatibility graph:
https://en.wikipedia.org/wiki/License_compatibility#/media/File:Floss-license-slide-image.svg

Long story short: I would recommend BSD-new license for permissive license, unless you have
a strong argument for Apache. It can be combined with BSD, but there are combinations
where Apache doesn't work with other open source licenses. 
It is not too serious issue, Apache v2 is still quite liberal, but BSD is easier, so I would go for BSD.

(Basically, I have two kind of projects: BSD and AGPLv3, depending if I want very permissive or very strict license.
In case of STIR, I would vote for BSD wherever it is possible.)

Quick explanation: "BSD", or "BSD-new" refer to the so called BSD 3-clause license. Old, v1, etc., refers to the 4-clause license. "Simplified" or "FreeBSD" license refers to the 2-clause BSD license.

KrisThielemans:
Thanks @dvolgyes . Helpful, but it wasn't what I tried to ask.

I'm fairly well aware of the different licenses (although your explanation has taught me some more). However, the decision to use Apache 2.0 was made a while ago. I cannot change it to a more permissive license, as I don't hold the copyright (nor does UCL). for most of the code. I did need to get about 15 letters from various institutions agreeing to use Apache 2.0. Don't want to do that again!

Instead, what I was after is a check of the `sh` and `sed` script to see if it does what it's supposed to (and nothing else!), i.e. replace the text in all our files that says it's GPL or LPGL with simply ` SPDX-License-Identifier: Apache-2.0`.

sorry to be unclear!

dvolgyes:
Hi,

I see. I misunderstood it, i will look into the scripts tomorrow. :)

KrisThielemans:
Merging this PR. We'll do a separate PR for the actual update of the headers.

PR #799, opened 2021-01-19
Bowsher prior

KrisThielemans:
This PR currently contains my 2017 quick attempt to create a Bowsher prior. This is largely a copy-paste of the `QuadraticPrior`, which is very bad practice...

This PR currently only exists to help someone who wants to implement the Bowsher. It probably doesn't even compile anymore!

@tsai-yujung tried it at the time and as far as I recall found problems with it.

### DIFF
--- a/src/recon_buildblock/QuadraticBowsherPrior.cxx
+++ b/src/recon_buildblock/QuadraticBowsherPrior.cxx
@@ -1,0 +1,764 @@
+/*
+    Copyright (C) 2000- 2011, Hammersmith Imanet Ltd
+	Copyright (C) 2017, University College London
+    This file is part of STIR.
+
+    This file is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


KrisThielemans:
@ashgillman possibly this is helpful, but probably not...

src/recon_buildblock/QuadraticBowsherPrior.cxx
```diff
@@ -0,0 +1,764 @@
...
+  this->only_2D = false;
+  this->kappa_ptr.reset();  
+  this->weights.recycle();
+  //this->anatomical_image_ptr.reset();
+}
+
+template <>
+const char * const 
+QuadraticBowsherPrior<float>::registered_name =
+  "NewAnatomical";
```
danieldeidda, line 133: Just having a look. But why "NewAnatomical" and not just "Bowsher"?

Issue #802, opened 2021-01-20
compute..Hessian.par file error


KrisThielemans:
currently has
```
    prior type := Quadratic Prior
    Quadratic Prior Parameters:=
      penalisation factor := 0.0
      only 2D:= 0
    END Quadratic Prior Parameters:=
```
but our keyword is still `Quadratic`

robbietuk:
https://github.com/UCL/STIR/blob/84a000431372059707ca65cabadb8fc9eba1759a/examples/samples/compute_sqrt_Hessian_row_sum.par#L46-L50
Oops, missed it. PR coming.

PR #803, opened 2021-01-20
compute..Hessian.par file error fix

robbietuk:
Addresses #802

### DIFF
--- a/examples/samples/compute_sqrt_Hessian_row_sum.par
+++ b/examples/samples/compute_sqrt_Hessian_row_sum.par
@@ -43,7 +43,7 @@ SqrtHessianRowSum Parameters :=
     ; Default value is 0
     compute with penalty := 0
 
-    prior type := Quadratic Prior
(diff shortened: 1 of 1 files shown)


KrisThielemans:
rest of the parameters ok?

robbietuk:
I can't see any further problems 😄

robbietuk:
Hold on, I need to check one or two more things.

robbietuk:
Okay, all okay. I retested the script and the output images with a restricted FOV have a large value ring around the edge of the FOV. I wanted to quickly investigate this but it should be okay. Turning the restricted FOV off removes this ring. This is therefore a consequence of projector sensitivity?
image

KrisThielemans:
sure. it's `P' (1/y) P 1`, so near the edge of the FOV, `y` gets small, so the row-sum gets large (only well-defined with background term of course). And the final backprojection will only update voxels that are with the circular FOV.

PR #810, opened 2021-02-01
HKEM allows iterative kernel to be fixed to a chosen FULL ITERATION  …

danieldeidda:
…number;

- the member stop_iterative_kernel_at_iter_num has been added;
- default value is zero;


src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -534,6 +541,15 @@ set_hybrid(const bool arg)
   this->hybrid = arg;
 }
 
+template <typename TargetT>
+void
+KOSMAPOSLReconstruction<TargetT>::
+set_stop_iterative_kernel_at_iter_num(const int arg)
+{
+  this->_already_set_up = false;
```
KrisThielemans, line 549: is this necessary? i.e. do we really need to call `set_up()` again after using this?

KrisThielemans (commented):
I'm not familiar enough with the code to see if the logic is alright unfortunately. I'm guessing that this allows you to freeze the "PET" kernel after a few updates. That makes sense as a strategy. I find the naming then not ideal and might prefer `freeze_kernel_at...`., but I could be wrong.

I wonder if it wouldn't be better to work in terms of sub_iteration_num as opposed to iter_num. The rest of STIR works that way, and it's therefore probably less confusing.

Needs an entry in release notes and potentially usersguide

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -764,6 +780,8 @@ void KOSMAPOSLReconstruction<TargetT>::compute_kernelised_image(
       // Going to need the full emission regional normalised differences
       int dimf_row = this->num_voxels;
       int dimf_col = this->num_non_zero_feat-1;
+      
+      if(!this->iterative_kernel_image_fixed_sptr)
```
KrisThielemans, line 784: this might fail if you use the same object twice for reconstruction. It could still be non-zero from the previous run?

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -968,15 +986,31 @@ update_estimate(TargetT &current_alpha_coefficent_image)
...
   unique_ptr< TargetT > multiplicative_update_image_ptr
     (current_alpha_coefficent_image.get_empty_copy());
 
   const int subset_num=this->get_subset_num();  
   info(boost::format("Now processing subset #: %1%") % subset_num);
 
+  
+  if (this->stop_iterative_kernel_at_iter_num>0 &&
+      this->subiteration_num%this->get_num_subsets()==0 &&
+      this->subiteration_num/this->get_num_subsets()==this->stop_iterative_kernel_at_iter_num)
```
ashgillman, line 1001: I'm a bit confused here with the two `if`s, can you add some comments to explain the logic? i.e., why do we need this `if` and `iterative_kernel_image_sptr`? Can't we replace the two sets of `if`s here with

```
  if (this->stop_iterative_kernel_at_iter_num<=0 ||
       this->subiteration_num/this->get_num_subsets()<this->stop_iterative_kernel_at_iter_num) {
    iterative_kernel_image_sptr=shared_ptr<TargetT>(current_alpha_coefficent_image.clone());
}
```
I believe this would stop updating iterative kernel_image_sptr when `this->subiteration_num%this->get_num_subsets()==0` by default.

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -208,12 +211,14 @@ class KOSMAPOSLReconstruction:
...
  //kernel parameters
   int num_neighbours,num_non_zero_feat,num_elem_neighbourhood,num_voxels,dimz,dimy,dimx;
+  int stop_iterative_kernel_at_iter_num;
   std::vector<double> sigma_m;
   bool only_2D;
   bool hybrid;
   double sigma_p;
   double sigma_dp, sigma_dm;
   BasicCoordinate<3,int> min_ind, max_ind;
+  shared_ptr<TargetT> iterative_kernel_image_fixed_sptr;
```
ashgillman, line 221: Possibly could rename/simplify to `iterative_kernel_image_sptr`? It may or may not be fixed, just represents a sptr to the image for the iterative kernel.

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
```diff
@@ -167,6 +167,7 @@ set_defaults()
   this->only_2D = 0;
   this->kernelised_output_filename_prefix="";
   this->hybrid=0;
+  this->stop_iterative_kernel_at_iter_num=0;
```
ashgillman, line 170: I think it might be better to set the default to -1 and change the conditions to `<0` instead of `<=0`? I guess a user would never want to fix the kernel at iteration 0, but its technically valid, right? Plus, its more conventional to use a negative "special value" (by special I mean this is either the iteration number or -1 is a special case meaning never).

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -208,12 +211,14 @@ class KOSMAPOSLReconstruction:
   shared_ptr<TargetT> kpnorm_sptr;
  //kernel parameters
   int num_neighbours,num_non_zero_feat,num_elem_neighbourhood,num_voxels,dimz,dimy,dimx;
+  int stop_iterative_kernel_at_iter_num;
```
ashgillman, line 214: I agree with @KrisThielemans  that "stop" is confusing because it doesn't actually stop using the kernel. `freeze_iterative_kernel_at_iter_num` might be clearer

ashgillman (changes requested):

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: no probably not. I saw all the setters had this and I completely forgot why this was, so I put it just in case but can remove

danieldeidda:
> 
> 
> I'm not familiar enough with the code to see if the logic is alright unfortunately. I'm guessing that this allows you to freeze the "PET" kernel after a few updates. That makes sense as a strategy. I find the naming then not ideal and might prefer `freeze_kernel_at...`., but I could be wrong.
> 
> I wonder if it wouldn't be better to work in terms of sub_iteration_num as opposed to iter_num. The rest of STIR works that way, and it's therefore probably less confusing.
> 
> Needs an entry in release notes and potentially usersguide

Yes you are correct, that's the strategy. I don't have a strong opinion about the name, your suggestion sounds good to me. I will had the docs as well

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: mh I am confused about this. iterative_kernel_image_fixed_sptr points to a copy of the update only if we set "stop_kernel..." and we are in the exact iteration we chose. After that the pointer should not change as is meant to be "fixed". But I am probably missing something.

@ashgillman you mean to create a function for this condition? if (self->still_updating_iterative_kernel())

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: Sure definitely the comment needs to be added. 

the first condition is used to set the member  "iterative_kernel_image_fixed_sptr"  to the iteration we want while the second defines "iterative_kernel_image_sptr" which is used when we call "compute_kernel...()".  This can be = to the current update if we did not set the keyword "stop..." or if the current iteration is < stop_kernel... otherwise  iterative_kernel_image_sptr= iterative_kernel_image_fixed_sptr

this->subiteration_num%this->get_num_subsets()==0  this condition is not enough because this would stop the kernel at every full iteration

src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
danieldeidda: Mh actually this is meant to be fixed to whatever we chose in the parfile

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: Yes you right. I was unsure about this but I agree. I think the iteration number starts from 1 though so we probably need to trhough an error if the user set it to 0

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
KrisThielemans: > I am probably missing something.

My concern is what happens if you do (in Python or whatever).
```
recon = KOSMAPOSLReconstruction(...)
recon.set_input(...)
recon.reconstruct()
..
recon.set_input(...)
recon.reconstruct()
```
The 2nd recon would be affected with what you do before.

(It's quite different to think about interactive work vs command line)

robbietuk:
If I may make a small suggestion, I think the phrasing `fix`/`fixed` etc can be confusing and I would suggest using `freeze`/`frozen` terminology. This is only an opinion and feel free to ignore me.

danieldeidda:
> 
> 
> If I may make a small suggestion, I think the phrasing `fix`/`fixed` etc can be confusing and I would suggest using `freeze`/`frozen` terminology. This is only an opinion and feel free to ignore me.

Sure you're right it will be more consistent withe the keyword "freeze_kernel..."

danieldeidda:
I forgot to add: the reconstruction is tested and running OK ATM

src/iterative/KOSMAPOSL/KOSMAPOSLReconstruction.cxx
danieldeidda: Mh I see. good point!

danieldeidda:
I think it's all good now have a look. And perhaps I can create  a last commit for the tests

danieldeidda:
> Needs an entry in release notes and potentially usersguide

which release? 5.0?

KrisThielemans:
> which release? 5.0?

yes please

Issue #809, opened 2021-01-31
BinNormalisation::set_up fails on ExamInfo equality again for ECAT8 norm


KrisThielemans:
as before, and `ExamInfo` printed are the same. This was when running the SIRF recon on NEMA data

danieldeidda:
Does this happens only with this particular dataset?

KrisThielemans:
It's the only one I tried. sorry.

KrisThielemans:
This has now been resolved

PR #812, opened 2021-02-02
calibration factor is read from the interfile header. BinNormalisatio…

danieldeidda:
…n derived classes will need to check whether this has been read and if not read somewhere else

### DIFF
--- a/src/IO/InterfileHeader.cxx
+++ b/src/IO/InterfileHeader.cxx
@@ -175,6 +175,8 @@ InterfileHeader::InterfileHeader()
   add_key("originating system", &exam_info_sptr->originating_system);
   ignore_key("GENERAL DATA");
   ignore_key("GENERAL IMAGE DATA");
+  
+  add_key("calibration factor", &calibration_factor); 
   add_key("isotope name", &isotope_name); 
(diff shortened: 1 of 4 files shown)


danieldeidda:
travis error seems to say that calibartion factor is different in that check. Maybe connected to #809

KrisThielemans:
I guess best to use "approx equal" checks for all floats in `ExamInfo::operator==`

danieldeidda:
> 
> 
> I guess best to use "approx equal" checks for all floats in `ExamInfo::operator==`

But one is -1 and the other is something to the power of -38. Which I think means that calib_factor needs o be initialised somewhere

danieldeidda:
This needs to set defaults for calibration factor

KrisThielemans:
change https://github.com/UCL/STIR/blob/e3322df118ec0fd04a988839e3cef15b063f60e8/src/include/stir/ExamInfo.h#L133 to
use `(both calib factors <=0) || (abs(ratio -1) <= 1E-4)` or something like this.

I would move the implementation of `ExamInfo::operator==` to the .cxx

KrisThielemans (approved):

PR #813, opened 2021-02-05
Speed up ROOT file OI by turning off all unneeded branches

robbietuk:
A 1 line addition that turns off all branches for reading. When calling `stream_ptr->SetBranchAddress("time1", &time1);` etc., this turns the specific branch back on.

To test this, I unlisted a root file containing True 10.53M coincidence events (I have an SSD). There are no differences between the output sinograms.

 - STIR master: CPU time = 49.4s
 - FasterRootIO: CPU time = 21.7s

@NikEfth Do you have any comment on this? I don't probably dont understand ROOT to the same degree.


NikEfth:
No, I am hapy with the modification

KrisThielemans:
nice! 1 line mod, half the speed. feel free to find a few more of these!

Let's just wait for Travis etc, but i cannot see that there will be a problem

PR #814, opened 2021-02-10
Faster root IO

robbietuk:
Uses GetEntry() on individual branches as the checks are performed. These changes are suggestions provided by a [discussion on the ROOT forums](https://root-forum.cern.ch/t/tchain-get-entry-acceleration/43468/5).

This is an extention on #813, which allowed unlisting of 10.53M events in `21.78s CPU time`. This PR improves the unlisting to `11.97s CPU time` (when unlisting with no-exclusions). The sinograms, created by unlisting a ROOT file with STIR master and this PR are identical. 

@KrisThielemans https://github.com/UCL/STIR/pull/813#issuecomment-774195119 I found another one.

### DIFF
--- a/src/IO/InputStreamFromROOTFile.cxx
+++ b/src/IO/InputStreamFromROOTFile.cxx
@@ -127,15 +127,16 @@ InputStreamFromROOTFile::set_up(const std::string & header_path)
     stream_ptr->Add(fullfilename.c_str());
     // Turn off all branches
     stream_ptr->SetBranchStatus("*",0);
+
     // Branches are turned back on by SetBranchAddress()
-    stream_ptr->SetBranchAddress("time1", &time1);
-    stream_ptr->SetBranchAddress("time2", &time2);
-    stream_ptr->SetBranchAddress("eventID1",&eventID1);
-    stream_ptr->SetBranchAddress("eventID2",&eventID2);
-    stream_ptr->SetBranchAddress("energy1", &energy1);
-    stream_ptr->SetBranchAddress("energy2", &energy2);
-    stream_ptr->SetBranchAddress("comptonPhantom1", &comptonphantom1);
-    stream_ptr->SetBranchAddress("comptonPhantom2", &comptonphantom2);
+    stream_ptr->SetBranchAddress("time1", &time1, &br_time1);
+    stream_ptr->SetBranchAddress("time2", &time2, &br_time2);
+    stream_ptr->SetBranchAddress("eventID1",&eventID1, &br_eventID1);
+    stream_ptr->SetBranchAddress("eventID2",&eventID2, &br_eventID2);
+    stream_ptr->SetBranchAddress("energy1", &energy1, &br_energy1);
+    stream_ptr->SetBranchAddress("energy2", &energy2, &br_energy2);
+    stream_ptr->SetBranchAddress("comptonPhantom1", &comptonphantom1, &br_comptonPhantom1);
+    stream_ptr->SetBranchAddress("comptonPhantom2", &comptonphantom2, &br_comptonPhantom2);
 
     if (read_optional_root_fields)
     {
--- a/src/IO/InputStreamFromROOTFile.cxx
+++ b/src/IO/InputStreamFromROOTFile.cxx
@@ -69,6 +69,7 @@ InputStreamFromROOTFile::set_defaults()
     singles_readout_depth = -1;
     exclude_scattered = false;
     exclude_randoms = false;
+    check_energy_window_information = true;
     low_energy_window = 0.f;
     up_energy_window = 1000.f;
     read_optional_root_fields=false;
@@ -87,6 +88,7 @@ InputStreamFromROOTFile::initialise_keymap()
     this->parser.add_key("name of input TChain", &this->chain_name);
     this->parser.add_key("exclude scattered events", &this->exclude_scattered);
     this->parser.add_key("exclude random events", &this->exclude_randoms);
+    this->parser.add_key("check energy window information", &this->check_energy_window_information);
     this->parser.add_key("offset (num of detectors)", &this->offset_dets);
     this->parser.add_key("low energy window (keV)", &this->low_energy_window);
     this->parser.add_key("upper energy window (keV)", &this->up_energy_window);
@@ -194,17 +196,19 @@ InputStreamFromROOTFile::check_brentry_randoms_scatter_energy_conditions(Long64_
   }
 
   // Energy condition.
-  if (br_energy1->GetEntry(brentry) == 0)
-    return false;
-  if (br_energy2->GetEntry(brentry) == 0)
-    return false;
-  // Check both energy values are within window
-  if (this->get_energy1_in_keV() < this->low_energy_window ||
-      this->get_energy1_in_keV() > this->up_energy_window ||
-      this->get_energy2_in_keV() < this->low_energy_window ||
-      this->get_energy2_in_keV() > this->up_energy_window)
-  {
-    return false;
+  if (this->check_energy_window_information){
+    if (br_energy1->GetEntry(brentry) == 0)
+      return false;
+    if (br_energy2->GetEntry(brentry) == 0)
+      return false;
+    // Check both energy values are within window
+    if (this->get_energy1_in_keV() < this->low_energy_window ||
+        this->get_energy1_in_keV() > this->up_energy_window ||
+        this->get_energy2_in_keV() < this->low_energy_window ||
+        this->get_energy2_in_keV() > this->up_energy_window)
+    {
+      return false;
+    }
   }
   return true;
 }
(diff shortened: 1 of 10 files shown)


robbietuk:
There is a lot of overlap with the two derived classes but I cannot figure out how to parse `brentry` to a function in `InputStreamFromROOTFile`. Do I need to make it a class variable of `InputStreamFromROOTFile` to do so?

src/IO/InputStreamFromROOTFileForCylindricalPET.cxx
```diff
@@ -68,30 +68,76 @@ Succeeded
 InputStreamFromROOTFileForCylindricalPET::
 get_next_record(CListRecordROOT& record)
 {
-
+    auto brentry = stream_ptr->LoadTree(static_cast<Long64_t>(current_position));
```
KrisThielemans, line 71: why do you need to declare this out of the loop? If you do, it's fine, but then I wouldn't initialise it here (as you do it below)

KrisThielemans:
A factor 2 speed-up but ~40 line change. Not as good as the previous one 😄 !

Great stuff.

I think that indeed we can have in `InputStreamFromROOTFile`
```c++
virtual Succeeded get_next_record(CListRecordROOT&) = 0;
```
with a lot of the common code. By making it pure virtual, you force the derived classes to implement it. There is no way to force them to call the one from `InputStreamFromROOTFile` (as `InputStreamFromROOTFile::get_next_record()`), but obviously you would. 

> I cannot figure out how to parse `brentry` to a function in `InputStreamFromROOTFile`

this confuses me, as it's a local variable. Obviously, `br_eventID1` etc should be in  `InputStreamFromROOTFile`

src/IO/InputStreamFromROOTFileForCylindricalPET.cxx
robbietuk: Yep, this was a mistake.

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -163,6 +164,51 @@ InputStreamFromROOTFile::set_up(const std::string & header_path)
...
+bool
+InputStreamFromROOTFile::check_brentry_randoms_scatter_energy_conditions(Long64_t brentry)
+{
+  if (brentry < 0)
+    return false;
+
+  // Scatter event condition.
+  if (this->exclude_scattered) {
+    if (br_comptonPhantom1->GetEntry(brentry) == 0)
+      return false;
```
robbietuk, line 176: Previously these `br_comptonPhantom1->GetEntry(brentry) == 0` checks would `return Succeeded::no;`. With this change, this will now result in a `continue;` in the daughter class, which causes the next entry in the file to be processed. 

I really need the ability to return 3 things, what about?
 - `return 0;` - all good
 - `return 1;` - Event failed the conditions
 - `return 2;` - `GetEntry()== 0`

robbietuk (commented):
One concern I have is that here I am suggesting we use `TTree` methods. This change uses 
```
Long64_t brentry = stream_ptr->LoadTree(static_cast<Long64_t>(current_position));
```
As I understand it, `TChain` contains `TTree`s which contain `events` with `Branches`.
I am concerned that if multiple root files are setup in a `TChain`, we might encounter a problem... Not sure, I will test.

edit: I thought we could provide multiple root files to be unlisted but I cant find details in the users guide... Any ideas?

KrisThielemans:
>  I thought we could setup multiple root files to be unlisted but I cant find details in the users guide... Any ideas?

no, I don't think we can at the moment. Some of the code might look like that, but it was potentially copied from the ECAT966 code (which can do that), but this was never implemented for ROOT

src/IO/InputStreamFromROOTFileForECATPET.cxx
```diff
@@ -66,31 +66,68 @@ Succeeded
 InputStreamFromROOTFileForECATPET::
 get_next_record(CListRecordROOT& record)
 {
-    while(true)
-    {
-        if (current_position == nentries)
-            return Succeeded::no;
-
+  auto brentry = stream_ptr->LoadTree(static_cast<Long64_t>(current_position));
```
KrisThielemans: sam e problem here.

Please no white-space changes

src/IO/InputStreamFromROOTFile.cxx
KrisThielemans: `GetEntry()== 0` is really an `error`, is that correct? The new STIR policy is then to use `error` (which will throw an exception) as opposed to returning `Succeeded::no` (and hoping that the user will check, which half the time they don't).  Possibly that solves your problem. If not, feel free to create an `enum` (I guess `protected` inside here) such that the names tell people what's going on.

src/IO/InputStreamFromROOTFileForECATPET.cxx
KrisThielemans: sorry. I was checking on the outdated version

src/IO/InputStreamFromROOTFile.cxx
robbietuk: I think yes, but also no. https://root.cern.ch/doc/master/classTBranch.html#a4f9c8c55ab275930f2fc2f9262e31f91
The key line here is 
> The function returns the number of bytes read from the input buffer. If entry does not exist, the function returns 0. If an I/O error occurs, the function returns -1.

STIR does not check for an I/O error and therefore we really need:
```
if (br_*->GetEntry(brentry) <= 0); 
  error("I have errored!");  // Report an error message.
```

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -127,42 +128,88 @@ InputStreamFromROOTFile::set_up(const std::string & header_path)
...
+  if (br_energy2->GetEntry(brentry) == 0)
+    return false;
+  // Check both energy values are within window
+  if (this->get_energy1_in_keV() < this->low_energy_window ||
+      this->get_energy1_in_keV() > this->up_energy_window ||
+      this->get_energy2_in_keV() < this->low_energy_window ||
+      this->get_energy2_in_keV() > this->up_energy_window)
+  {
+    return false;
+  }
```
robbietuk, line 208: Can we have an if around if energy information is default we can ignore these tests.

KrisThielemans: What does a "default" mean? It might not be the same for GATE macro as STIR. I guess you could introduce yet another `bool check_energy_window_information` defaulting to `true`, but I feel it's creating work for not much benefit. Maybe I'm wrong.

robbietuk: Yes, we can (and probably should) do this. By default I meant the class defaults:
https://github.com/UCL/STIR/blob/a08cb7504aac957742e2b7a81240a38a62f2e489/src/IO/InputStreamFromROOTFile.cxx#L71-L72

robbietuk: >but I feel it's creating work for not much benefit. Maybe I'm wrong.

It was easy.

robbietuk:
Awaiting travis and I will conduct an offline test on ECATPET root data (travis only tests CylindericalPET). I think it is then good to merge.

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -68,6 +69,7 @@ InputStreamFromROOTFile::set_defaults()
     singles_readout_depth = -1;
     exclude_scattered = false;
     exclude_randoms = false;
+    check_energy_window_information = true;
```
KrisThielemans, line 72: do we document default values of all these `bool`s somewhere? Should be in the doxygen, I guess best for `set_defaults()` (can put it in the .cxx).

Also, the class doc in the .h files needs to have this new variable (and refer to `set_defaults()` for more info. Finally, the `examples/samples/*.hroot` files need to have it.

(you can see why I thought this was some work... 😄 )

src/include/stir/IO/InputStreamFromROOTFileForECATPET.h
```diff
@@ -133,8 +134,19 @@ class InputStreamFromROOTFileForECATPET : public
...
 
+    //! \name TBranches for ECAT PET
+    //@{
+    TBranch *br_crystalID1 = nullptr;
+    TBranch *br_crystalID2 = nullptr;
+    TBranch *br_blockID1 = nullptr;
+    TBranch *br_blockID2 = nullptr;
+    //@}
+
+    //! \name ROOT Variables, e.g. to hold data from each entry.
```
KrisThielemans, line 145: e.g. -> i.e.

KrisThielemans (changes requested):
some comments on documentation only.

Do add this to release_5.0.htm

KrisThielemans (approved):

PR #823, opened 2021-02-18
Fix822 ffp early stop with open mp

NikEfth:
This is a small patch to fix early stop when forward projecting with OpenMP. 
The issue is not 100% reproducible but I have experienced many times.

### DIFF
--- a/src/recon_buildblock/ForwardProjectorByBin.cxx
+++ b/src/recon_buildblock/ForwardProjectorByBin.cxx
@@ -47,6 +47,8 @@
 #include <boost/format.hpp>
 #include <iostream>
 
+#include <omp.h>
+
 START_NAMESPACE_STIR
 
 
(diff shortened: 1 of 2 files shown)


NikEfth:
@KrisThielemans Please wait before you review this. It will fail without OpenMP, I forgot to fix that.

KrisThielemans:
I don't think we've ever seen this problem, @AnderBiguri and @danieldeidda can probably comment as they both have "monster" machines. But obviously that doesn't mean that it doesn't exist (wtih multi-threading, you never know).

However, I cannot see why this modification would help. In the original code, the `viewgrams` are a local variable, so they are thread-specific already.

NikEfth:
I agree, that `viewgrams` is  local and there should not be a difference. 
As you know for the normalisation code I had to forward project a model of the source used in the simulation. 
Many times, the projection was finishing without reported problems, but a few megabytes short.
With the suggested modification I have not faced the problem again.

NikEfth:
P.S. 
Maybe is an OPENMP version, specific problem. I am using Arch Linux, which means I have the latest version.

KrisThielemans:
I really can't see how this would help. The affected variables are local or read-only. I'm going to close this therefore. Feel free to re-open (with some evidence...).

PR #833, opened 2021-02-24
ml_norm for virtual crystals

gefeichen:
normfile
I got it by dividing normalized projdata by raw projdata (mCT).
And no zero value found in https://github.com/UCL/STIR/blob/882c261bc23af642865e61f00b19d902dcf3a5f3/src/utilities/find_ML_normfactors3D.cxx#L162 after make_fan_data_remove_gaps


gefeichen:
sinogram after norm
And another image:  upper is a cylinder sinogram, below is a sinogram after normalisation

src/buildblock/ML_norm.cxx
```diff
@@ -1035,45 +1035,54 @@ void make_fan_data_remove_gaps(FanProjData& fan_data,
...
+    const int num_transaxial_crystals_per_block =
+            proj_data_info_ptr->get_scanner_sptr()->
+                    get_num_transaxial_crystals_per_block();
+    const int num_axial_crystals_per_block =
+            proj_data_info_ptr->get_scanner_sptr()->
+                    get_num_axial_crystals_per_block();
+
+    const int num_physical_transaxial_crystals_per_block = num_transaxial_crystals_per_block - virtual_transaxial_crystals;
+
+    const int num_physical_axial_crystals_per_block = num_axial_crystals_per_block - virtual_axial_crystals;
```
KrisThielemans, line 1062: I guess we could just as well provide the member functions for these 2 as well.

src/buildblock/ML_norm.cxx
```diff
@@ -1199,19 +1224,29 @@ void set_fan_data_add_gaps(ProjData& proj_data,
...
-                    if ((ra+1)% num_axial_crystals_per_block == 0) continue;
-                    if ((rb+1)% num_axial_crystals_per_block == 0) continue;
-                    if ((a+1)% num_transaxial_crystals_per_block == 0) continue;
-                    if ((b+1)% num_transaxial_crystals_per_block == 0) continue;
-                    
-                    int new_a = a - a/num_transaxial_crystals_per_block;
-                    int new_b = b - b/num_transaxial_crystals_per_block;
-                    int new_ra = ra - ra/ num_axial_crystals_per_block;
-                    int new_rb = rb - rb/num_axial_crystals_per_block;
+                    (*segment_ptr)[bin.axial_pos_num()][bin.view_num()][bin.tangential_pos_num()] = 0;
```
KrisThielemans, line 1228: I suggest we use an extra input argument to the function `gap_value`, defaulting to 0. It will also make it clearer what's happening here.

KrisThielemans (changes requested):
This looks very nice to me.  Thanks a lot.

I have 2 suggestions in the review below. In addition, I think we should get rid of the old `make_fan_data` and its inverse. Otherwise we keep too much code duplication.

Add yourself as author and whoever owns your copyright in the headers. Also, add a note in release_5.0.htm in section `changed functionality`.

KrisThielemans:
I believe this supersedes your other PR. If so, please close it by referring to this one.

I don't understand the first picture. Can you provide a bit more detail?

gefeichen:
> I believe this supersedes your other PR. If so, please close it by referring to this one.
> 
> I don't understand the first picture. Can you provide a bit more detail?

I used a cylinder source in gPET to cover the whole FOV and generated corresponding model projection data.

It's hard to visualize the norm parameter. So I just use manip_segments to divide model_data by measure_data.  And the result shows in the first picture. Maybe it's useless.

KrisThielemans:
easier to look at the inverse of it:
```
stir_math -s --including-first --power -1 eff_factors.hs norm_factors.hs
```
could do the trick

KrisThielemans:
I'd also compare measured data with model*eff_factors. I think you get the latter with `apply_normfactors3D` with `apply_or_undo=1`.

gefeichen:
The absolute difference of measure and model*eff_factors is 19%
`sum(abs(model*eff-measure))/sum(measure)`
I don't know whether it is correct.

Below is eff and geo factors. left:eff   right:geo
normeffgeo

KrisThielemans:
> The absolute difference of measure and model*eff_factors is 19%

they will be different because of noise of course. Hard to quantify that. But 19% doesn't look too bad. efficiencies look a bit strange but could be alright.

I note that my text was confusing. with "eff_factors" I meant "eff*geo" (i.e. the total efficiency, also known as "1/norm_factors"). I'll call them "total_eff_factors" below.

Things to compare measured with model*total_eff_factors:
- visual comparison of 
- profiles. you will probably have to do some averaging to reduce noise.
- reconstruction of the gate data used for normalisation (`lm_to_projdata` with excluding scatters and randoms, no background in recon).
- reconstruction of other gate data  (`lm_to_projdata` with excluding scatters and randoms, no background in recon).
- reconstruction of other gate data full pipeline (although that doesn't have anything to do with the norm)

Are you using https://github.com/UCL/STIR-GATE-Connection? If so, it tells you how to handle the norm (currently `BinNormalisationFromProjData` needs the norm-factors, not the total_eff_factors)

src/buildblock/ML_norm.cxx
```diff
@@ -1102,52 +1085,10 @@ void make_fan_data_remove_gaps(FanProjData& fan_data,
...
-          }
-    proj_data.set_segment(*segment_ptr);
-  }
-}
-
 /// **** This function make proj_data from fan_data while adding the intermodule gaps **** ////
 /// *** fan_data doesn't have gaps, proj_data has gaps *** ///
 void set_fan_data_add_gaps(ProjData& proj_data,
-                  const FanProjData& fan_data)
+                  const FanProjData& fan_data, const int gap_value)
```
KrisThielemans, line 1091: ` float`  please

src/include/stir/ML_norm.h
```diff
@@ -228,17 +228,12 @@ get_fan_info(int& num_rings, int& num_detectors_per_ring,
...
 void make_fan_data_remove_gaps(FanProjData& fan_data,
                    const ProjData& proj_data);
 
-void set_fan_data(ProjData& proj_data,
-                  const FanProjData& fan_data);
-
 void set_fan_data_add_gaps(ProjData& proj_data,
-                  const FanProjData& fan_data);
+                  const FanProjData& fan_data,
+                  const int gap_value=0);
```
KrisThielemans, line 236:
```suggestion
                  const float gap_value=0.F);
```

KrisThielemans (changes requested):
can you merge `master` in here to resolve the conflict? sorry.

KrisThielemans (approved):

KrisThielemans:
@gefeichen, will your mCT tests be affected by #838 and/or #837?

NikEfth:
Hi , I am sorry for the delay, I will test with my data tonight.

src/buildblock/ML_norm.cxx
```diff
@@ -1035,45 +996,54 @@ void make_fan_data_remove_gaps(FanProjData& fan_data,
...
     const int new_half_fan_size = new_fan_size/2;
-    const int num_axial_blocks_in_max_delta = max_delta/num_axial_crystals_per_block;
-    const int new_max_delta = max_delta - num_axial_blocks_in_max_delta - 1;
-    const int new_num_detectors_per_ring = num_detectors_per_ring - num_transaxial_blocks;
-    const int new_num_rings = num_rings - num_axial_blocks;
-    
+    const int num_axial_blocks_in_max_delta = max_delta/(num_axial_crystals_per_block);
+    const int new_max_delta = max_delta - (num_axial_blocks_in_max_delta)*virtual_axial_crystals;
+    const int num_physical_detectors_per_ring = num_detectors_per_ring - num_transaxial_blocks*virtual_transaxial_crystals;
+    const int num_physical_rings = num_rings - (num_axial_blocks-1)*virtual_axial_crystals;
```
NikEfth, line 1032: I agree with this.

NikEfth (commented):
Hi @gefeichen 
Thank you for your good work, 
I don't have mCT data but from my data (which as we had discussed are slightly different ) (and your code) I saw that the last segment was coming out empty. 
I don't think that this actually has to do with your modifications but, could you please check you last, and second to last segments?  I believe that those would be +54 and +53 (?), might be wrong, though.

gefeichen:
Hi @NikEfth 
   Thanks for your checking.
   Could you provide more information about your data? Like ring numbers, virtual crystal number, and position

KrisThielemans:
This seems ready to me now, and we have the signed letter.

@NikEfth if you have any final comments, let us know. Otherwise I'll merge

PR #839, opened 2021-03-04
small fix to LmToProjData store_delayeds

robbietuk:


KrisThielemans (approved):
funny that one! It might speed it up a tiny tiny bit, but probably not at all.

robbietuk:
Probably wont but the line above says it sets it to false, so it probably should set it to false 😄

PR #842, opened 2021-03-16
scripts for GE PETCT RDF9 processing 

KrisThielemans:
I'm putting it here as a start, but it is very preliminary right now. 
- Assumes you want to process sinograms (but the GE-Signa scripts handle listmode, so can be used here)
- Nac_recon.sh also does norm etc, so this needs to be split
- various assumptions along the way
- a copy of a lot of files from the Signa directory


KrisThielemans:
abandoned over #875

PR #845, opened 2021-03-19
use_actual_detector_boundaries to use view_offset

### LINKED ISSUE #844: ProjMatrixByBinUsingRayTracing with use_actual_detector_boundaries does not use view_offset information
This results in `view_offset (degrees)` being ignored (see L583)

https://github.com/UCL/STIR/blob/63d8d2373113a3f90eb67dfae2b429b80de9e9bd/src/recon_buildblock/ProjMatrixByBinUsingRayTracing.cxx#L561-L583

robbietuk:
Fixes #844

Previously, when `use_actual_detector_boundaries` is enabled, the projector did not take into account the `azimuthal_angle_offset` (view_offset angle)


robbietuk:
I believe that `ProjMatrixByBinUsingRayTracing` is the only class to use this key.

KrisThielemans:
> I believe that `ProjMatrixByBinUsingRayTracing` is the only class to use this key/variable.

correct

PR #849, opened 2021-04-08
Fix issue with find_ML_singles_from_delayed when delayed sinogram is out of cwd

### LINKED ISSUE #848: find_ML_singles_from_delayed fails if data is not in cwd
Running:
```
DelayedData=${ABS_PATH_TO_DELAYED}/Sino_SumSim.Delayed_S1R1.hs
factors=singles_from_delayed
num_iterations=10
find_ML_singles_from_delayed ${factors} ${DelayedData} ${num_iterations}
```
gets the following warning, which is really an error:
```

Display iterations which are a multiple of [0,10 D:0]: 
Compute KL distance between fan-sums at iterations which are a multiple of [0,10 D:0]: 
Write output at iterations which are a multiple of [0,10 D:10]: 
WARNING: Interfile warning: 'energy resolution' (0.124) is expected to be -1.000. Currently, the energy resolution and the reference energy, are used only in scatter correction.

WARNING: Error opening output file fansums_for_${ABS_PATH_TO_DELAYED}/Sino_SumSim.Delayed_S1R1.dat
```

This is because of:
https://github.com/UCL/STIR/blob/9c37b88d609b1a0c32617fed0feef77b16791ba0/src/utilities/find_ML_singles_from_delayed.cxx#L105-L110

A little more additional string manipulation is needed. What about?
```
        std::string fan_sum_name = "fansums_for_";
        {
          // extract filename
          char sep = '/';
#ifdef _WIN32
          sep = '\\';
#endif
          std::string tmp = argv[2];
          size_t i = tmp.rfind(, tmp.length());
          fan_sum_name += tmp.substr(i+1, tmp.length() - i);
         }
	fan_sum_name.erase(fan_sum_name.begin() + fan_sum_name.rfind('.'),
			   fan_sum_name.end());
	fan_sum_name += ".dat"; 
        }
```

I believe the 
```
#ifdef _WIN32
          sep = '\\';
#endif
```
is needed by perhaps not. Anyway, I will make a PR. With this change it appears to work if an abs path is given.

robbietuk:
Fixes #848
If the delayed data sinogram was a path, saving/loading the fan_sum data would error because the path would be `fansums_for_${FullPath}.dat` 

This fix extracts the filename from a path and so it will save the fansums as `fansums_for_${filename}.dat`

### DIFF
--- a/src/utilities/find_ML_singles_from_delayed.cxx
+++ b/src/utilities/find_ML_singles_from_delayed.cxx
@@ -103,16 +103,7 @@ int main(int argc, char **argv)
       // write fan sums to file
       {
         std::string fan_sum_name = "fansums_for_";
-        {
-          // extract filename
-          char sep = '/';
-#ifdef _WIN32
-          sep = '\\';
-#endif
-          std::string fname = argv[2];
-          size_t i = fname.rfind(sep, fname.length());
-          fan_sum_name += fname.substr(i+1, fname.length() - i);
-        }
+        fan_sum_name += get_filename(argv[2]);
 	fan_sum_name.erase(fan_sum_name.begin() + fan_sum_name.rfind('.'), 
 			   fan_sum_name.end());
 	fan_sum_name += ".dat"; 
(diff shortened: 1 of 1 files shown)


KrisThielemans:
we have some path manipulation functions in `utilities.h`. Best to use that...

robbietuk:
> we have some path manipulation functions in `utilities.h`. Best to use that...

I now get the same result as before :) Letting travis run just in case.

PR #850, opened 2021-04-09
Adds the ability to exclude_trues when unlisting ROOT files

robbietuk:
There are two debugging variables used in unlisting ROOT files, i.e. `exclude_randoms` and `exclude_scattered`. Therefore True (`trues`) events are always unlisted making the unlisting of only scattered and random events difficult, requiring 2 unlistings and `stir_math`. 

The current unlisting combination, always include `trues` results in the following combination
 x | **Exclude Randoms=1** | **Exclude Randoms=0**
--- | --- | ---
**Exclude Scattered=1** | `T` | `T` + `R`
**Exclude Scattered=0** | `T` + `S` | `T` + `R` + `S` + `RS`

where `T` = True and unscattered event, `R` = Random and unscattered event, `S` = True but Scattered event, and `RS` is a Random and Scattered event.

____
**`exclude_trues=1`**

This PR addes the ability to `exclude_trues=1`, resulting in the following possible combinations
 x | **Exclude Randoms=1** | **Exclude Random=0**
--- | --- | ---
**Exclude Scattered=1** | None | `R`
**Exclude Scattered=0** | None | `R` + `RS`

which allows the unlisting of Random and Random scattered event.

____
**`exclude_unscattered=1`**

It also adds the ability to `exclude_unscattered=1`, resulting in the following possible combinations
 x | **Exclude Randoms=1** | **Exclude Randoms=0**
--- | --- | ---
**Exclude Scattered=1** | None | None
**Exclude Scattered=0** | `S` | `S` + `RS`

--------
 Additionally, in the `set_up`, if all events will be excluded then an error will be thrown, i.e. 
```exclude_unscattered=1 && exclude_scattered=1```
 AND 
```exclude_trues=1 && exclude_randoms=1```

TODO (pending general approval):

- [ ] Add documentation regarding usage
- [ ] Update the example `*.hroot` files
- [ ] Add to usersguide regarding the possible combinations detailed here.
- [ ] Add to release notes

### DIFF
--- a/documentation/STIR-UsersGuide.tex
+++ b/documentation/STIR-UsersGuide.tex
@@ -799,6 +799,51 @@ in conjunction with the \textit{LMF} library. However, these might be obsolete
 as the OpenGATE project might be distributing files
 to enable STIR to read \textit{LMF} format files.
 
+\subsubsubsection{ROOT exclusion keys configurations}
+The \textit{exclude xxx events := 0} keys can be used to process specific types of events.
+Below is are sets event types that will be processed with an exclusion or combination of exclusions.
+Not not mentioned, the exclusion of event type is false.
+
+\begin{table}[h]
+    \centering
+    \begin{tabular}{ | c | c | }
+        \hline
+        \textbf{Event(s) type} & \textbf{Exclusion Configuration}
+        \\ \hline \hline
+        All events & \textit{None}
+
+        \\ \hline \hline
+        \vtop{\hbox{\strut Scattered from same eventID }
+        \hbox{\strut Unscattered from same eventID}}
+        & \textit{exclude random events := 1}
+        \\ \hline
+        \vtop{\hbox{\strut Unscattered from same eventID }
+        \hbox{\strut Unscattered from different eventIDs}}
+        & \textit{exclude scattered events := 1}
+        \\ \hline
+        \vtop{\hbox{\strut Scattered from different eventIDs }
+        \hbox{\strut Unscattered from different eventIDs}}
+        & \textit{exclude non-random events := 1}
+        \\ \hline
+        \vtop{\hbox{\strut Scattered from same eventID}
+        \hbox{\strut Scattered from different eventIDs}}
+        & \textit{exclude unscattered events := 1}
+
+        \\ \hline \hline
+        Unscattered from same eventID & \vtop{\hbox{\strut \textit{exclude random events := 1} }\hbox{\strut \textit{exclude scattered events := 1}}}
+        \\ \hline
+        Unscattered from different eventIDs &  \vtop{\hbox{\strut \textit{exclude non-random events := 1} }\hbox{\strut \textit{exclude scattered events := 1}}}
+        \\ \hline
+        Scattered from same eventID  & \vtop{\hbox{\strut \textit{exclude random events := 1} }\hbox{\strut \textit{exclude unscattered events := 1}}}
+        \\ \hline
+        Scattered from different eventIDs & \vtop{\hbox{\strut \textit{exclude non-random events := 1} }\hbox{\strut \textit{exclude unscattered events := 1}}}
+        \\ \hline
+    \end{tabular}
+    \caption{A guide on how to process specific types of events from a ROOT file using various exclusion keys and combinations of keys.}
+    \label{tab:ROOT_file_exclusions}
+\end{table}
+
+
 \subsection{Multi header format}
 \label{multi-file-format}
 It is often useful to be able to constract a single dataset from multiply components (e.g. images
--- a/documentation/STIR-UsersGuide.tex
+++ b/documentation/STIR-UsersGuide.tex
@@ -770,8 +770,13 @@ GATE_Cylindrical_PET Parameters :=
   number of crystals_Z := 4
 
   Singles readout depth := 1
+
+  ; exclude different event types
+  exclude non-random events := 0
+  exclude unscattered events := 0
   exclude scattered events := 0
   exclude random events := 0
+
   offset (num of detectors) := 0
 
   low energy window (keV) := 0      ; Default 0
--- a/documentation/STIR-UsersGuide.tex
+++ b/documentation/STIR-UsersGuide.tex
@@ -800,9 +800,9 @@ as the OpenGATE project might be distributing files
 to enable STIR to read \textit{LMF} format files.
 
 \subsubsubsection{ROOT exclusion keys configurations}
-The \textit{exclude xxx events := 0} keys can be used to process specific types of events.
+The \textit{exclude xxx events := 1} keys can be used to process specific types of events.
 Below is are sets event types that will be processed with an exclusion or combination of exclusions.
-Not not mentioned, the exclusion of event type is false.
+The exclusions of an event type is false by default.
 
 \begin{table}[h]
     \centering
--- a/documentation/STIR-UsersGuide.tex
+++ b/documentation/STIR-UsersGuide.tex
@@ -801,7 +801,7 @@ to enable STIR to read \textit{LMF} format files.
 
 \subsubsubsection{ROOT exclusion keys configurations}
 The \textit{exclude xxx events := 1} keys can be used to process specific types of events.
-Below is are sets event types that will be processed with an exclusion or combination of exclusions.
+Below are sets event types that will be processed with an exclusion or combination of exclusions.
 The exclusions of an event type is false by default.
 
 \begin{table}[h]
--- a/src/IO/InputStreamFromROOTFile.cxx
+++ b/src/IO/InputStreamFromROOTFile.cxx
@@ -30,6 +30,7 @@
 #include "stir/IO/FileSignature.h"
 #include "stir/error.h"
 #include "stir/FilePath.h"
+#include "stir/info.h"
 
 #include <TROOT.h>
 #include <TSystem.h>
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by robbietuk** [comment]
This still is not perfect. `T`, `S`, and `R` events can be isolated and unlisted, but `RS` cannot be. 

An alternative would be to replace the current `exclude_*` keywords with `include_trues`, `include_randoms`, `include_scattered`, and `include_random_scattered`. These could then stack ontop of one another allowing the user clear and more obvious control over what type of events are unlisted.

**Comment by KrisThielemans** [comment]
thanks. It's true that `include` keywords might be easier to understand. The question is if we need this.. If we do, then there's going to be a bit of a problem with backwards compatibility. I guess we'd have to check if an `exclude` one is set (meaning we'd have to change the type to `int` as it cannot be done with `bool`), and if so, do some check. A reasonable amount of work...

**Comment by robbietuk** [comment]
> The question is if we need this.. 

I'd like to have the ability to do this in the STIR-GATE-Connection debugging tools. 

I think the `exclude` keys can be kept, I would remove those added by this PR though... I would add a set of variables that are `include_trues`, `include_randoms`, `include_scattered`, and `include_scattered_randoms`, which are all true by default. 

Consider the following two examples: 
 - Setting `include scatteded events :=0` in a `hroot` file will then set the varibale `include_scattered=false` and all scattered (or `S` as I have denoted it above) would be disabled. However, this would leave `RS` enabled. Additionally `include true events :=0` and `include random events :=0` woudl allow `RS` only events to be unlisted.

 - In another example, set `exclude scattered events := 0` (current code). This will disable both the `S` and `RS` data types and would override any `include` arguments. This overriding would be done by evaluting the status of the `exclude` variabels after the `includes` during the `set_up`.

The `get_next_entry` ROOT code would only depend on the `include` variables. This should also be printed during `set_up`.

Obviously this might be too much for STIR. I could implement this on my own branch and leave it out of STIR master but I do think that this `include` is much simpler and is only complicated by maintaining the backwards compatibility.

**Comment by KrisThielemans** [comment]
The details seem tricky. First of all, what is the difference between `exclude_xxx=1` and `include_xxx=0`? If there is one, is that going to be expected by people?

Having `include_scattered_randoms` seems non-intuitive to me. Are we saying that `include_randoms=1` but `include_scattered_randoms=0` only gives unscattered randoms? Maybe that makes some sense, but what does `include_randoms=1 && include_scatter=0` do then? We easily get into semantics here...

I'm not sure why we'd want to isolate scattered randoms anyway. The scatter correction is NOT aiming to get these, and they WILL be included in any randoms correction method. Making an option that confuses people for this corner case seems like a bad idea to me.

Maybe I'm saying that the current PR does what is needed?

**Comment by robbietuk** [comment]
> Are we saying that `include_randoms=1` but `include_scattered_randoms=0` only gives unscattered randoms?

Almost but by default `include_xxx` would be true so all event types are excepted unless specified. So to get ONLY unscattered (randoms) you would need `include_scattered=0` AND  `include_trues=0` AND `include_scattered_randoms=0` AND `include_randoms=1` (but this last one would be default.

> Maybe that makes some sense, but what does include_randoms=1 && include_scatter=0 do then?

On its own, this would process `T` + `R` + `RS` events. 

I think an easier way to look at this is to take the default as `T` + `R` + `S` + `RS` events are process.  The `include_xxx` bools would control a different type events:
 - `include_trues` controls `T`
 - `include_randoms` controls `R`
 - `include_scattered` controls `S`
 - `include_scattered_randoms` controls `RS`

Perhaps renaming `include_trues` to `include_unscattered_trues` etc would be more clear?

> Maybe I'm saying that the current PR does what is needed?

I agree more and more as I wrote this. This PR would be enough to isolate every event type, but it isnt obvious...

- `T` -> `exclude_scattered=1 ` + `exclude_randoms=1`
- `R` -> `exclude_scattered=1 ` + `exclude_trues=1`
- `S` -> `exclude_unscattered=1 ` + `exclude_randoms=1`
- `RS` -> `exclude_unscattered=1 ` + `exclude_trues=1`

Combinations are generally simpler but I dont think there is a way to get `T` + `RS` (not that I have any idea why you would want too). BUT I dont like the `exclude_xxx` keywords as double negitives are needed.

**Comment by robbietuk** [comment]
If we continue with the `exclude_xxx` and this PR, could I add the following to the end of `set_up` to make it clear exactly what is being doing. It will help with user debugging. I ask because I think it is very unSTIR like...
```
    {
      // Show which event types will be unlisted based upon the exclusion criteria
      bool trues = true;
      bool randoms = true;
      bool scattered = true;
      bool scattered_randoms = true;
      if ( exclude_trues || this->exclude_unscattered)
        trues = false;
      if ( exclude_randoms || exclude_unscattered )
        randoms= false;
      if ( exclude_scattered || exclude_trues )
        scattered = false;
      if ( exclude_scattered || exclude_randoms )
        scattered_randoms = false;
      std::string status = "InputStreamFromROOTFile: Processing data with the following event type inclusions:"
                           "\n    True events:             " + std::to_string(trues) +
                           "\n    Random events:           " + std::to_string(randoms) +
                           "\n    Scattered events:        " + std::to_string(scattered) +
                           "\n    Scattered Random events: " + std::to_string(scattered_randoms);
      warning(status);
    }
```

**Comment by KrisThielemans** [comment]
> Perhaps renaming include_trues to include_unscattered_trues etc might be clearer?

oh no, now we get into the discussion on "what does `trues` mean". I have heard it used for "non-randoms", or for "non-random&non-scattered" (what you use here). I don't think that there's an agreed terminology on that one. sorry...

> I dont like the `exclude_xxx` keywords as double negatives are needed.

ah well. Looks like Greeks (i.e. @NikEfth) and Dutch/Flemish speakers (i.e. me) don't care about those! What's wrong with "don't exclude these guys"?

> It will help with user debugging. I ask because I think it is very unSTIR like...

I don't mind. You'd have to have good terminology though (I don't think anyone will agree with your terminology of "randoms" as being "unscattered randoms").

Also, you'd have to use `info(status, 2)` (as it's not a `warning`, and the verbosity 2 means that you'd get it only if you really want to).

**Comment by robbietuk** [comment]
> What's wrong with "don't exclude these guys"?

It overcomplicates everything and makes me think...

File: src/IO/InputStreamFromROOTFile.cxx [hunk 1]
```diff
@@ -164,6 +169,38 @@ InputStreamFromROOTFile::set_up(const std::string & header_path)
         stream_ptr->SetBranchAddress("sourcePosZ2",&sourcePosZ2, &br_sourcePosZ2);
     }
 
+
+    {
+      // Ensure that two conflicting exclusions are not applied
+      if (this->exclude_trues && this->exclude_randoms)
+        error("InputStreamFromROOTFile: Both the exclusion of true and random events has been set. Therefore, "
+              "no data will be processes.");
```
KrisThielemans, line 177:
```suggestion
              "no data will be processed.");
```

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -164,6 +169,38 @@ InputStreamFromROOTFile::set_up(const std::string & header_path)
...
+      if (this->exclude_scattered && this->exclude_unscattered)
+        error("InputStreamFromROOTFile: Both the exclusion of scattered and unscattered events has been set. Therefore, "
+              "no data will be processes.");
```
KrisThielemans, line 180:
```suggestion
              "no data will be processed.");
```

KrisThielemans (changes requested):
ok. looks good.

now some user-doc and release notes.

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -86,7 +89,9 @@ InputStreamFromROOTFile::initialise_keymap()
     this->parser.add_key("name of data file", &this->filename);
     this->parser.add_key("Singles readout depth", &this->singles_readout_depth);
     this->parser.add_key("name of input TChain", &this->chain_name);
+    this->parser.add_key("exclude true events", &this->exclude_trues);
     this->parser.add_key("exclude scattered events", &this->exclude_scattered);
+    this->parser.add_key("exclude unscattered events", &this->exclude_unscattered);
     this->parser.add_key("exclude random events", &this->exclude_randoms);
```
robbietuk, line 95: Are we happy with these keys, `exclude true events` in particular? This refers to any event from the SAME eventID

robbietuk:
> now some user-doc and release notes.

Release_5_0?

KrisThielemans:
> > now some user-doc and release notes.
> 
> Release_5_0?

sure

src/IO/InputStreamFromROOTFile.cxx
KrisThielemans: I'm sure you're not going to like `exclude non-random events`, although then we can have a triple negative!

what about `exclude events with same ID`

robbietuk: `exclude non-random events` is cleaner and it is obvious it the the oposite of `exclude random events`... 😞 I'll do it

documentation/STIR-UsersGuide.tex
```diff
@@ -794,6 +799,51 @@ \subsubsection{ROOT files as output by OpenGATE}
...
+        Unscattered from different eventIDs &  \vtop{\hbox{\strut \textit{exclude non-random events := 1} }\hbox{\strut \textit{exclude scattered events := 1}}}
+        \\ \hline
+        Scattered from same eventID  & \vtop{\hbox{\strut \textit{exclude random events := 1} }\hbox{\strut \textit{exclude unscattered events := 1}}}
+        \\ \hline
+        Scattered from different eventIDs & \vtop{\hbox{\strut \textit{exclude non-random events := 1} }\hbox{\strut \textit{exclude unscattered events := 1}}}
+        \\ \hline
+    \end{tabular}
+    \caption{A guide on how to process specific types of events from a ROOT file using various exclusion keys and combinations of keys.}
+    \label{tab:ROOT_file_exclusions}
+\end{table}
```
robbietuk, line 844: From overleaf, a generation of this insert
image

documentation/STIR-UsersGuide.tex
```diff
@@ -794,6 +799,51 @@ \subsubsection{ROOT files as output by OpenGATE}
 as the OpenGATE project might be distributing files
 to enable STIR to read \textit{LMF} format files.
 
+\subsubsubsection{ROOT exclusion keys configurations}
+The \textit{exclude xxx events := 1} keys can be used to process specific types of events.
+Below is are sets event types that will be processed with an exclusion or combination of exclusions.
```
KrisThielemans, line 804:
```suggestion
Below are sets event types that will be processed with an exclusion or combination of exclusions.
```

KrisThielemans (changes requested):

KrisThielemans (approved):

PR #851, opened 2021-04-15
Examinfoequal

danieldeidda:
Modify  ExamInfo == operator. Checks all floats are equal up to a limit

### DIFF
--- a/src/buildblock/ExamInfo.cxx
+++ b/src/buildblock/ExamInfo.cxx
@@ -73,8 +73,8 @@ ExamInfo::parameter_info() const
 
 bool 
 ExamInfo::operator == (const ExamInfo &p1) const {      
-    return  (this->up_energy_thres/p1.up_energy_thres -1.)<=1E-3 &&
-            (this->low_energy_thres/p1.low_energy_thres-1.) <=1E-3 &&
+    return  abs(this->up_energy_thres/p1.up_energy_thres -1.)<=1E-3 &&
+            abs(this->low_energy_thres/p1.low_energy_thres-1.) <=1E-3 &&
             this->radionuclide==p1.radionuclide &&
             this->time_frame_definitions==p1.time_frame_definitions &&
 //              this->branching_ratio==p1.branching_ratio &&
@@ -83,6 +83,6 @@ ExamInfo::operator == (const ExamInfo &p1) const {
             this->imaging_modality==p1.imaging_modality &&
             this->originating_system==p1.originating_system &&
             this->patient_position==p1.patient_position &&
-            (this->start_time_in_secs_since_1970/p1.start_time_in_secs_since_1970-1.)<=1E-3; }
+            abs(this->start_time_in_secs_since_1970/p1.start_time_in_secs_since_1970-1.)<=1E-3; }
 
 END_NAMESPACE_STIR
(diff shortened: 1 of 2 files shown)


AnderBiguri:
Hi Daniel!

I think you are implementing this worngly, you sould `abs(var1-var2)<=eps`.

Example of why:

seconds_since_1970 now: 1618485350
seconds_since_1970 tomorrow: 1618571851

seconds_since_1970_now/seconds_since_1970_tomorrow-1.)= -0.00005344279

danieldeidda:
> 
> 
> Hi Daniel!
> 
> I think this way of comparing can have some bugs, you should `abs(var1-var2)<=eps`.
> 
> Example of why:
> 
> seconds_since_1970 now: 1618485350
> seconds_since_1970 tomorrow: 1618571851
> 
> `(1618485350/1618571851-1.)= -0.00005344279 `
> 
>     * is negative, meaning if the difference was higher, you may get a large negative number. In any case, this is enough to pass the check, regardless of the number.
> 
>     * is below tolerance in absolute terms, with an entire day of difference, so it should not be.
> 
>     * If they are `int` (I think they are?) the division will be rounded, and can pass the check with a `-1`

Well spotted thanks!

KrisThielemans (changes requested):
sadly, these will fail horribly when they the denominator is zero,
as we know what the units are for these things, I'd favour an absolute comparison
```
abs(this->up_energy_thres-p1.up_energy_thres ) > 1 /* keV*/
...
abs(this->start_time_in_secs_since_1970-p1.start_time_in_secs_since_1970) > .5 /* sec */
```

KrisThielemans:
I would also modify `parameter_info()` like in
```c++
 s << "Scan start time: " << std::fixed << std::setprecision(12) <<this->start_time_in_secs_since_1970
   << std::setprecision(5) // reset for further floats
```

KrisThielemans (approved):
thanks. did you try and print if it looks ok (e.g. with `list_projdata_info --all`)?

danieldeidda:
> 
> 
> thanks. did you try and print if it looks ok (e.g. with `list_projdata_info --all`)?

Yes:

Modality: nucmed
Calibration Factor: 1.000000000000
Radionuclide: ^90^Yttrium
Patient position: unknown
Scan start time: 0.000000000000
Time frame start - end (duration), all in secs: 0.00000 - 7200.00000 (7200.00000)
number of energy windows:=1
energy window lower level[1] := -1.00000
energy window upper level[1] := -1.00000
ProjDataInfoCylindricalArcCorr :=

KrisThielemans:
ok, seems like `setprecision` now gives "digits after the dot" (I guess because of the `fixed`), so arguably we could use `setprecision(1)` or so.

if you do that, take the opportunity to update the copyright header, but commit with `ci skip`

KrisThielemans (changes requested):
need to include `<iomanip>`

danieldeidda:
> 
> 
> ok, seems like `setprecision` now gives "digits after the dot" (I guess because of the `fixed`), so arguably we could use `setprecision(1)` or so.

1 instead of 12? or instead of 5?

KrisThielemans:
> 1 instead of 12? or instead of 5?

I guess can just be 1 for all of them (after the calib factor). no?

KrisThielemans (approved):

KrisThielemans:
Thanks. By the way, it has to be `[ci skip]`, not `[ci-skip]`...

Waiting for Travis and Appveyor

danieldeidda:
> 
> 
> Thanks. By the way, it has to be `[ci skip]`, not `[ci-skip]`...
> 
> Waiting for Travis and Appveyor

ops sorry :D

KrisThielemans:
All well. I've cancelled the Appveyor after 2 succesful builds. Merging

PR #855, opened 2021-04-17
BinNormalisation time-frame fixes

### LINKED ISSUE #853: BinNormalisation crash when no time-frame info present
Tracking https://github.com/SyneRBI/SIRF/issues/892 down, shows that we have a crash at
https://github.com/UCL/STIR/blob/56f742977c837ccd51c2f3e5fc1fb8f78c2353cb/src/recon_buildblock/BinNormalisation.cxx#L193-L194
This is because there is no time frame info in the test data. We should handle this gracefully.

it's not so obvious how to though. Some `BinNormalisation` objects would need the time-frame info, others not. I guess their `set_up` function should check if it needs it or not, so that's fine. However, the `undo` function wouldn't know it.

But... it turns out that pretty much all these functions do not use the `start_time` and `end_time`, so possibly this problem disappears once we cut them out!

In any case, `exam_info_sptr->get_time_frame_definitions().get_start_time()` needs to do a check and write a more sensible error message.

KrisThielemans:
Fixes #853 

also included an unrelated minor fix in this PR (sorry)

### DIFF
--- a/src/buildblock/ProjDataInfoCylindrical.cxx
+++ b/src/buildblock/ProjDataInfoCylindrical.cxx
@@ -43,6 +43,7 @@
 #include "stir/round.h"
 #include "stir/numerics/norm.h"
 #include "stir/warning.h"
+#include "stir/info.h"
 #include <boost/format.hpp>
 #include <math.h>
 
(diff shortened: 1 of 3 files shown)


KrisThielemans:
@danieldeidda just removing the relevant lines has fixed the problem. phew!

PR #858, opened 2021-04-21
set_num_views() is now called in the post processing in BinNormalisat…

danieldeidda:
…ionSPECT:

if it's not set it could create issues in the reconstruction as it is used to estimate the interval between each angular position which is used for the decay correction.


KrisThielemans (commented):
I think this should be in `set_up()`, not `post_processing()`. The latter is only called after parsing. The former will be called always.

Also, please remove commented out lines. Cleaner...

PR #867, opened 2021-04-25
Work on modernizing, small clean-ups across codebase 

mahnen:
This is a pull request, to potentially tidy the code a bit. Checking out if you would be interested in this. The big part was done using clang-tidy with the following methods:

        "bugprone-argument-comment",
        "bugprone-assert-side-effect",
        "bugprone-bad-signal-to-kill-thread",
        "bugprone-bool-pointer-implicit-conversion",
        "bugprone-branch-clone",
        "bugprone-copy-constructor-init",
        "bugprone-dangling-handle",
        "bugprone-dynamic-static-initializers",
        "bugprone-exception-escape",
        "bugprone-fold-init-type",
        "bugprone-forward-declaration-namespace",
        "bugprone-forwarding-reference-overload",
        "bugprone-inaccurate-erase",
        "bugprone-incorrect-roundings",
        "bugprone-infinite-loop",
        "bugprone-integer-division",
        "bugprone-lambda-function-name",
        "bugprone-macro-parentheses",
        "bugprone-macro-repeated-side-effects",
        "bugprone-misplaced-operator-in-strlen-in-alloc",
        "bugprone-misplaced-widening-cast",
        "bugprone-move-forwarding-reference",
        "bugprone-multiple-statement-macro",
        "bugprone-narrowing-conversions",
        "bugprone-not-null-terminated-result",
        "bugprone-parent-virtual-call",
        "bugprone-posix-return",
        "bugprone-signed-char-misuse",
        "bugprone-sizeof-container",
        "bugprone-sizeof-expression",
        "bugprone-string-constructor",
        "bugprone-string-integer-assignment",
        "bugprone-string-literal-with-embedded-nul",
        "bugprone-suspicious-enum-usage",
        "bugprone-suspicious-memset-usage",
        "bugprone-suspicious-missing-comma",
        "bugprone-suspicious-semicolon",
        "bugprone-suspicious-string-compare",
        "bugprone-swapped-arguments",
        "bugprone-terminating-continue",
        "bugprone-throw-keyword-missing",
        "bugprone-too-small-loop-variable",
        "bugprone-undefined-memory-manipulation",
        "bugprone-undelegated-constructor",
        "bugprone-unhandled-self-assignment",
        "bugprone-unused-raii",
        "bugprone-unused-return-value",
        "bugprone-use-after-move",
        "bugprone-virtual-near-miss",
        "modernize-avoid-bind",
        "modernize-avoid-c-arrays",
        "modernize-concat-nested-namespaces",
        "modernize-deprecated-headers",
        "modernize-deprecated-ios-base-aliases",
        "modernize-loop-convert",
        "modernize-make-shared",
        "modernize-make-unique",
        "modernize-raw-string-literal",
        "modernize-redundant-void-arg",
        "modernize-replace-auto-ptr",
        "modernize-replace-random-shuffle",
        "modernize-return-braced-init-list",
        "modernize-shrink-to-fit",
        "modernize-unary-static-assert",
        "modernize-use-bool-literals",
        "modernize-use-default-member-init",
        "modernize-use-emplace",
        "modernize-use-equals-default",
        "modernize-use-equals-delete",
        "modernize-use-nodiscard",
        "modernize-use-noexcept",
        "modernize-use-nullptr",
        "modernize-use-override",
        "modernize-use-auto",
        "modernize-use-transparent-functors",
        "modernize-use-uncaught-exceptions",
        "performance-faster-string-find",
        "performance-for-range-copy",
        "performance-implicit-conversion-in-loop",
        "performance-inefficient-algorithm",
        "performance-inefficient-string-concatenation",
        "performance-inefficient-vector-operation",
        "performance-move-const-arg",
        "performance-move-constructor-init",
        "performance-no-automatic-move",
        "performance-noexcept-move-constructor",
        "performance-trivially-destructible",
        "performance-type-promotion-in-math-fn",
        "performance-unnecessary-copy-initialization",
        "performance-unnecessary-value-param"


mahnen:
Seems like this is incompatible with some of the (older) compilers in the CD. Anyways, would be interested in your opinion on this

Issue #827, opened 2021-02-21
use override attribute for virtual functions


KrisThielemans:
C++-11 allows using the `override` attribute to prevent errors where inadvertently the wrong signature is used, see e.g. https://en.wikipedia.org/wiki/C%2B%2B11#Explicit_overrides_and_final

#826 introduces this for the `BinNormalisation` hierarchy, but there are far more of course.  It's unfortunately not obvious how to automate this.

mahnen:
use clang-tidy like this:

`run-clang-tidy.py -header-filter='.*' -checks='-*,modernize-use-override' -fix`

see:

https://www.kdab.com/clang-tidy-part-1-modernize-source-code-using-c11c14/

mahnen:
I tried this in a custom branch, see pull request number 23 (closed), it kind of worked well.

KrisThielemans:
thanks @mahnen I think this is a great idea. I suppose the problem with #867 was:
- possible "too modern". I suppose `clang-tidy` should have settings for C++ version (STIR is still on C++-11).
-  automated tools can struggle with our work-arounds. For instance, the appveyor jobs failed because the automatic tool couldn't cope with a [manual work-around for (very old versions of Visual Studio)](vhttps://github.com/mahnen/STIR/blob/3d3520415f189fb416a1296fb0020a6753eb3bd8/src/buildblock/Scanner.cxx#L985-L988). Of course, those work-arounds are actually out-of-date already, so probably should be removed first.
- changing 460 files means a lot of code-review. (If it is only a matter of running an automatic tool, that'd be easier than if corrections than needed to be made by hand, but still some checking would be good.)

I think doing this in stages would solve these problems though. It should be paired with creating some scripts for doing the cleaning, such that people can run it on a PR. We could even ultimately do that via git hooks once #724 is merged (although run-time might become a problem).

Do you have a feel for a good order of doing this?

mahnen:
Ill try with the minimum modernize-use-override and create a pull req, then well see from there maybe?

KrisThielemans:
sure. If possible, do the PR in commits: first commit to add a script (in `scripts/maintenance`), another that has the result. 

Presumably running the script twice doesn't change anything anymore, so we can always run this after future merges.

mahnen:
pull request created, for your review #868

KrisThielemans:
I'm picking this up now, as we've removed all (?) the work-arounds and macros for old compilers.

[...]

PR #868, opened 2021-04-26
Modernize use override

mahnen:
pull request to adress #827

### DIFF
--- a/scripts/maintenance/modernize-use-override.sh
+++ b/scripts/maintenance/modernize-use-override.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+# This is an example script for use with clang-tidy. 
+# Use after coniguring the stir build in a build
+# folder, with -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+# so that 
+#
+#    compile_commands.json
+#
+# exists. Then run this with with clang-tidy like this:
+#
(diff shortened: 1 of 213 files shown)


mahnen:
@KrisThielemans : Don't know if I can do something about this, one of the runners timed out while it was going well...

KrisThielemans:
yes, that particular job is on the limit for some reason. I've committed something elsewhere that reduces the number of warnings that the compiler outputs, so possibly it'll bring it in line. So I suggest that we wait till those are merged.

Overall, this looked pretty good though. Thanks!

I think we need to follow @casperdcl 's [advice](https://github.com/UCL/STIR/pull/724#issuecomment-714437514), and commit the final patch with a specific author, to avoid you getting all the credit 😉 .

One thing that I'm not so sure about is the use of `compile_commands.json` exported by CMake. This works fine if all options are enabled and all dependencies are found. However, could it trouble otherwise ? e.g. if your system doesn't have ecat7, then CMake won't export the STIR ecat7 files, and `clang-tidy` will never touch them, and they could therefore become out-of-date. Of course, ideally we'd find this in CI, but unfortunately we don't test everything. (case in point: the ecat7 matrix library isn't freely available).

KrisThielemans:
I've re-run this on current STIR in #1367, so closing this. thanks for telling me how to do this!

Issue #873, opened 2021-04-30
Bug with Poisson log-likelihood compute_sub_gradient when not using subset sensitivities


robbietuk:
Issue
------
The method:
https://github.com/UCL/STIR/blob/dc6f3566f050b8d50f775ca531b06b9cffea3a59/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMean.cxx#L362-L390
is split into two halves. Firstly `compute_sub_gradient_without_penalty_plus_sensitivity()` is called, which computes (in distrubutable):
```
grad_backproj_m( y_m / [ forwardproj_m( x ) + b_m ] ),
```
where `m` is the subset index, `x` the image, and `b` is the background.

Secondly the "subset sensitivity" is subtracted:
```
backproj_m( y_m / [ forwardproj_m( x ) + b_m ] )  - S_m
``` 
where `S_m` is the subset sensitivity.

When `use_subset_sensitivies == false`, `set_up` computes the `total_sensitivity` (sum of all `S_m`). It then populates the `S_m` with `total_sensitivity / num_subsets`. Therefore, when `S_m` is subtracted in the gradient computation, it is not subtracting the correct `S_m`. When `use_subset_sensitivies == false` this is not an issue.

Short Term Solution
------
If `use_subset_sensitivies == false`, `compute_sub_gradient_without_penalty` should throw an error.

Long Term Solution
------
Functional changes, `divide_and_truncate` https://github.com/UCL/STIR/blob/dc6f3566f050b8d50f775ca531b06b9cffea3a59/src/buildblock/recon_array_functions.cxx#L183-L277
or a wrapper, prior to backprojection, should subtract 1's over the subset of data. A more in-depth impact assessment is needed.

robbietuk:
Maybe this should be a seperate issue. Currently `setup()` of the `PoissonLogLikelihood` objective functions will compute the subset sensitivities. This is because of the gradient computation. If the subset sensitivty subtraction when computing the gradients is removed (long term solution), these subset sensitivies would not be required when setting up the objective functions, saving significant preprocessing computation.

PR #874, opened 2021-04-30
compute_sub_gradient_without_penalty - throw error when use_subset_sensitivities is false

robbietuk:
Short term solution to #873


robbietuk:
Tests prevent any use of `compute_sub_gradient_without_penalty` when `use_subset_sensitivities = false`. I expect some tests to fail, if they dont, test should be added 😄

robbietuk:
This PR breaks many tests so will be closed as alternative solutions will be investigated.

PR #888, opened 2021-05-14
Move Demo4 to correct location

robbietuk:
I believe b3e6d38e7744a46c096483bc69822f97e44fbd37 resulted in demos 1, 2, and 3 being moved but demo4 was missed.


robbietuk:
It appears that there isnt a travis configuration that tests the demos, even building them.

KrisThielemans:
> It appears that there isnt a travis configuration that tests the demos, even building them.

that's true. please create an issue for it.

PR #760, opened 2020-11-30
Optional KL Divergence Computation

robbietuk:
Addresses #697 

Combines the objective function value computation to run through the same code as the gradient computation. For numerical computational reasons, when computing a single function/gradient evaluation (i.e. a single bin) an upper and lower threshold are appiled to prevent singlarities (divide by ~zero) and negative values. Both the functional and gradient values should be subject to the same bounds and therefore this PR combines these evaluations into `poisson_divide_and_truncate`. 

If a `log_likelihood_ptr` is parsed, the function will compute the function. Likewise, if the `do_division == true` the function will compute the gradient. 

Adds bool `use_KL_divergence` that is parsed from `PoissonLogLikelihoodWithLinearModelForMean` methods through to a set of functions in `recon_array_functions.cxx`. `use_KL_divergence` can be set at objective function level and by default this is off, so backward compatable. 

- [x] Test with evaluation of objective functions (Compare STIR master, and branch LL and KL computations). 

STIR master and branch LL are identical. KL produces value ~1000x smaller than LL.

- [ ] Documentation: Release notes, Users guide.
- [ ] Check mismatch between STIR master LL and this PR LL values (https://github.com/UCL/STIR/pull/760#issuecomment-780743088)

- [ ] Test MPI Code

PR #893, opened 2021-05-21
Removing requirements for compute_sub_gradient to use sensitivity image

robbietuk:
This is a fix for #873

Introduces the method
``` 
actual_compute_sub_gradient_without_penalty(TargetT& gradient, const TargetT &current_estimate, const int subset_num, const bool do_subtraction)
```

where the bool `do_subtraction` (name needs to change) indicates to distributable and the gradient computation whether to do the gradient computation as:
`do_subtraction = true` (e.g. `compute_sub_gradient_without_penalty`)
```
backproj_m [ y/ybar -1 ] 
```

or `do_subtraction = false` (e.g. `compute_sub_gradient_without_penalty_plus_sensitivity`)
```
backproj_m [ y/ybar ]
```


robbietuk:
Sorry, I thought a Draft PR didnt run Travis. It appears that using `override` results in ALOT of warnings along the lines of:
```
STIR/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndGatedProjDataWithMotion.h:80:5: warning: 'construct_target_ptr' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
```

KrisThielemans:
ok.  we'll fix all the `override` issues in one go as in #860. Let's remove them here then.

KrisThielemans:
> or `do_subtraction = false` (e.g. `compute_sub_gradient_without_penalty_plus_sensitivity`)
> 
> ```
> backproj_m [ y/ybar ] - backproj [ 1 ]
> ```

that'll have to be
```
 backproj_m [ y/ybar ]
```

robbietuk:
Corrected the above. Opinion on the `do_subtraction` name and/or the `actual_compute_sub_gradient_without_penalty()` method name?

robbietuk:
Need to fix the documentation regarding sensitivities: https://github.com/UCL/STIR/blob/42240decc608f18a2ad89f462309b5baa47c86cf/src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMean.h#L69-L72

robbietuk:
Interestingly the `test_PoissonLogLikelihoodWithLinearModelForMeanAndProjData` fails because the numerical gradient does not match the gradient. 

I think this is due to a normalisation issue with the new implementation that is caused by STIR's relucatnce to apply normalisation during the gradient computation and previous handling using the sensitivity subtraction.

Let me try and explain. The gradient computation does not want/or need the normalisation (`N`):
https://github.com/UCL/STIR/blob/dc6f3566f050b8d50f775ca531b06b9cffea3a59/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L1207-L1217

This is because it utilised the mathmatical shortcut:
Standard loglikelihood `grad = A^T [ Ny/ (NA(x) + b) - 1 ]`
STIR master `grad = A^T [ y/ (A(x) + b/N) ] - A^T [ 1/N ]`

`A` is the projector operator and `b` is an additive term (with the state of normalisation unknown, it isnt that important for this). This allows for the precomputation of the constant terms: additive (`b/N`) and sensitivity (`A^T [ 1/N ]`). Then,  the gradient computation was simple and the `RPC_process_related_viewgrams_gradient` does not apply the normalisation to either the expected or the measured data.

If this is the case, this PR modification might need some tweaking. Rather than subtracting ones, it should subtract `1/N`. Correct?

robbietuk:
Test passes on my end. I get the same result (ish)

gradient:
image

difference between `compute_sub_gradient_without_penalty_plus_sensitivity() - subset_sensitivity` and `gradient.hv`
image

robbietuk:
Test failing on OSSPS in recon_test_pack
https://github.com/UCL/STIR/blob/42240decc608f18a2ad89f462309b5baa47c86cf/recon_test_pack/run_tests.sh#L231-L246

This PR
--------

The error output
```
-------- Running OSSPS with a quadratic prior --------
Running OSSPS
---- Comparing output of OSSPS subiter 8 (should be identical up to tolerance)
Running compare_image

Maximum absolute error = 0.00227004
Maximum in (1st - 2nd) = 0.00227004
Minimum in (1st - 2nd) = -0.000148283
Error relative to sup-norm of first image = 1.01647 %

Image arrays deemed different
(tolerance used: 0.2 %)


Maximum absolute error = 0.00131911
Maximum in (1st - 2nd) = 0.00131911
Minimum in (1st - 2nd) = -1.35558e-05
Error relative to sup-norm of first image = 0.590663 %

Image arrays deemed different
(tolerance used: 0.05 %)

There were problems here!
```

The OSSPS reconstruction is fine, just the reconstructed image and reference are ~1% different.

STIR Master
----------
The difference between the reference  in STIR master are of magnetude e-04 % / e-05 %.

Discussion
-------

I think this is because the OSSPS reconstuction parameter file using `use subset sensitivities:=0`. 
In STIR master, :
`this->objective_function_sptr->compute_sub_gradient(*numerator_ptr, current_image_estimate, subset_num);` would be subtracting the subset sensitvity from `backproj[ y/ybar]` but the subset sensitivity is the `total_sensitivity/ num_subset`.
In this PR, the gradient will subtract the actual subset sensitivity (or the equivalent. 
Therefore, I think the issue is the test and the reference image (`test_image_OSSPS_PM_QP_8.hv`) is no longer correct for this objective function gradient in the OSSPS reconstruction.

Difference between master image and reference is virtually 0
Difference between PR image and reference:
image
Most of the error is in the low sensitivity region, which definitely makes sense if the total sensitivity was subtracted.

Proposed Solution
---------
Replace the OSSPS reference image.

robbietuk:
Now that I have confirmed that there is a difference between OSSPS when `use subset sensitivities:= 0` or `1`, the difference between the image reconstructed using STIR master with `use subset sensitivities:=1` and this PR is:
image

This image is coronal view (top/bottom are the end rings). Clearly there is an end plane issue but the majority of the interior is good.

I tested to see if this was a subset sensitivity issue in STIR master but comparing the resulting images output by the test when `use subset sensitivities:= 0` and `1`
image There are minor issues but the lack of end plane difference suggests that it is a result of this PR.


This is affected by changing from `zero end planes of segment 0:= 1` to `zero end planes of segment 0:= 0` but it does not resolve the issue, as shown here.
image

Maybe this is a normalisation subtraction issue in the PR?

The difference is subtle and does occur over 8 OSSPS sub-iterations with 4 subsets
image

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -613,26 +613,36 @@ PoissonLogLikelihoodWithLinearModelForMeanAndProjData<TargetT>::
...
+                                            const TargetT &current_estimate,
+                                            const int subset_num,
+                                            const bool do_subtraction)
 {
   assert(subset_num>=0);
   assert(subset_num<this->num_subsets);
-  distributable_compute_gradient(this->projector_pair_ptr->get_forward_projector_sptr(), 
-                                 this->projector_pair_ptr->get_back_projector_sptr(), 
+  distributable_compute_gradient(this->projector_pair_ptr->get_forward_projector_sptr(),
+                                 this->projector_pair_ptr->get_back_projector_sptr(),
```
KrisThielemans, line 631: bunch of white-space diffs? Let's get rid of them

KrisThielemans (changes requested):

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx
```diff
@@ -532,6 +532,28 @@ compute_sub_gradient_without_penalty_plus_sensitivity(TargetT& gradient,
         }
     }
     info(boost::format("Number of used events: %1%") % num_used_events);
+
+  if (!add_sensitivity)
+    {
+      if (!this->get_use_subset_sensitivities())
+        error("PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin::"
+              "actual_compute_sub_gradient_without_penalty(): cannot subtract subset sensitivity because "
+              "use_subset_sensitivities is false.");
```
robbietuk, line 541: This is not clean but I feel it is nessisery to prevent improper use of sensitivity subtraction.

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -1111,10 +1114,28 @@ void distributable_compute_gradient(const shared_ptr<ForwardProjectorByBin>& for
                               zero_seg0_end_planes,
                               log_likelihood_ptr,
                               additive_binwise_correction,
-                              /* normalisation info to be ignored */ shared_ptr<BinNormalisation>(), 0., 0.,
-                              &RPC_process_related_viewgrams_gradient,
+                              /* normalisation info to be ignored */ shared_ptr<BinNormalisation>(),
```
robbietuk, line 1117: This is actually incorrect in the case when `log_likelihood_ptr != NULL`. This will need to be corrected in #760 as to compute the log likelihood, you need the normalisation

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
robbietuk: I believe resolved but need to check back before merge

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -642,7 +652,7 @@ PoissonLogLikelihoodWithLinearModelForMeanAndProjData<TargetT>::
 actual_compute_objective_function_without_penalty(const TargetT& current_estimate,
                                                   const int subset_num)
 {
-  double accum=0.;  
+  double accum=0.;
```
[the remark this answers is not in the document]
robbietuk: I believe resolved but need to check back before merge

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx
```diff
@@ -427,8 +427,20 @@ PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin<Tar
...
+template <typename TargetT>
+void
+PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin<TargetT>::
+actual_compute_sub_gradient_without_penalty(TargetT& gradient,
+                                                      const TargetT &current_estimate,
+                                                      const int subset_num,
+                                                      const bool do_subtraction)
+{
+  if (do_subtraction)
+    error("List mode cannot currently do subtraction on the gradient computation.");
```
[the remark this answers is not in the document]
robbietuk: Can do `!add_sensitivity` (`= do_subtraction `) now by subtracting subset sensitivity at the end of the actual. Outdated

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx
robbietuk: Not outdated

src/include/stir/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMean.h
```diff
@@ -141,12 +132,8 @@ public  GeneralisedObjectiveFunction<TargetT>
   /*! 
     This function is used for instance by OSMAPOSL.
 
-    This computes
-    \f[ {\partial L \over \partial \lambda_v} + P_v =
-      \sum_b P_{bv} {y_b \over Y_b}
-      \f]
-    (see the class general documentation).
-    The sum will however be restricted to a subset.
```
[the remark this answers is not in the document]
robbietuk: Done

robbietuk:
Major changes since last comments:
- Resolved the [endplane issue](https://github.com/UCL/STIR/pull/893#issuecomment-848987351). It was a normalisation issue.
- Updated the OSSPS recon_test_pack reference image to this PR (previously failed because the reference was computed with `use_subset_sensitivities =0 ` which resulted in the wrong objective function.
- Added OSSPS `recon_test_pack` test to compare the output using subset senstitivites and not. These should be identical (expect for list mode right now [not tested in `recon_test_pack`], which I am not sure how to fix).
- Added `compute_sub_gradient_without_penalty` documentation
 
Marking as ready for review (pending travis). There are still a few issues:
- List mode `compute_sub_gradient_without_penalty()` has been made to error when `use_subset_sensitivities=true` because this is the only call now that does not use the "standard" objective function. Keep the error or let it run normally? Maybe add lots of warning? up to you @KrisThielemans 
- In `distributable_compute_gradient()` the `normalisation_sptr` should be parsed in both cases. This is probably safer to do anyway.

robbietuk:
Ready for review

PR #898, opened 2021-05-28
Fix bug with iteration limit

AnderBiguri:
Loops for 1 more argc than exits, errors with ambiguous message:

    	basic_string::_M_construct null not valid

I suspect this is not tested, but I will submit without the skipping command just in case.


AnderBiguri:
Github action fails, but I changed 1 character. I am confused.

KrisThielemans:
The github action tests white-space and will currently always fail. see email. We ignore it for now.

Issue #892, opened 2021-05-19
BinNormalisation can error-out because of originating_system


KrisThielemans:
`ExamInfo::operator==` checks for `originating_system` but this is not uniquely defined .For instance, @evgueni-ovtchinnikov [reported a case](https://github.com/SyneRBI/SIRF/issues/926#issuecomment-844185647_) where `this->originating_system` was `2008`, and `p1.originating_system` was `Siemens mMR`.

@danieldeidda, shall we remove it from the comparison? `BinNormalisation` will check for `proj_data_info` anyway. If we do, we should document that we're not checking it.

danieldeidda:
> 
> 
> `ExamInfo::operator==` checks for `originating_system` but this is not uniquely defined .For instance, @evgueni-ovtchinnikov [reported a case](https://github.com/SyneRBI/SIRF/issues/926#issuecomment-844185647_) where `this->originating_system` was `2008`, and `p1.originating_system` was `Siemens mMR`.
> 
> @danieldeidda, shall we remove it from the comparison? `BinNormalisation` will check for `proj_data_info` anyway. If we do, we should document that we're not checking it.

OK yes better to remove it

danieldeidda:
> 
> If we do, we should document that we're not checking it.

You mean to write something in documentation? or to throw a warning message?

KrisThielemans:
let's just add a bit of doxygen. Writing a warning is maybe ok, although we might see a lot of them. I guess I wouldn't, but feel free 😄

PR #900, opened 2021-06-02
 ExamInfo: removed check on originating system for operator ==: #892

danieldeidda:
fixing #892


src/include/stir/ExamInfo.h
```diff
@@ -45,6 +45,8 @@ START_NAMESPACE_STIR
 
   \todo This should be an abtract registered object, in order to serve as a complete
   base function for every input data type.
+  
+  Warning: the operator == does not check that originating system is consistent!
```
KrisThielemans, line 49: why not put it this in the member doc? Seems to make more sense. (do a `[ci skip]' commit)

Issue #705, opened 2020-10-06
PLS gradient has edge artefacts


KrisThielemans:
Tests in #695 fail for the PLS prior because of edge artefacts. Here is a difference image. PLS uses a uniform 1's anatomical image and the same randomised density (between 0 and 1 voxel values) as QP and RDP
![image](https://user-images.githubusercontent.com/13292409/94136976-9e54e980-fe5d-11ea-9f9e-d4eb55f2679b.png) 

_Originally posted by @robbietuk in https://github.com/UCL/STIR/pull/695#issuecomment-698273995_

PR #901, opened 2021-06-07
Priors/Hessian method corrections

robbietuk:
This PR was going to simply add the `accumulate_Hessian_times_input` method to the RDP but while implementing it it was discovered that the implementations in the `Quadratic` and `LogCosh` priors were also incorrect. Additionally improves priors testing (#153).

This PR will attempt to do the following for each of the aforementioned priors:

- [x] Correct implementation issues with `accumulate_Hessian_times_input` to compute the correct Hessian vector product
- [x] Clearly document what is being done
- [x] Add a variable `is_convex = false` to generalised prior. Set `is_convex = true` for all convex priors.
- [x] Add test for convexity, much like #902 by testing `x^T Hx >0`
- [x] Add `compute_Hessian()` to all convex methods
- [x] Add test for convex priors by checking `grad(x) - grad(x+eps) ~= H(x)`, where eps is a small perturbation.
- [x] Add `diagonal_second_derivative` and `off_diagonal_second_derivative` methods for use in all Hessian methods

Relative Difference Prior `accumulate_Hessian_times_input`
-------

- [x] RDP accumulate_Hessian_times_input
- [x] Added RDP Hessian documentation
- [x] Add div by 0 safety
- [x] Add epsilon to Hessian function


robbietuk:
These latest commits do not fix a possible mistake with the RDP's `accumulate_Hessian_times_input` but it does indicate convexity.

src/include/stir/recon_buildblock/RelativeDifferencePrior.h
```diff
@@ -193,6 +198,27 @@ class RelativeDifferencePrior:  public
...
+   * @param x is the target voxel.
+   * @param y is the neighbourhood voxel.
+   * @param gamma is the edge preservation value controlling the transition between the quadratic and linear behaviour
+   * @param eps is a small non-negative value included to prevent division by zero, see epsilon.
+   * @return the second derivative of the Relative Difference Prior
+
+   */
+  static inline float Hessian(const float x, const float y, const float gamma, const float eps)
+  {
+    if (x > 0.0 || y > 0.0 || eps > 0.0)
```
AnderBiguri, line 216: Just browsing, and I may be missing some context, so apologies if that is the case. 

Why should you check for `eps` here and return zero if negative? 

Would it not make more sense to have the `set_epsilon()` method validate that its positive (it has to be) and then assume here that `eps` is correct? This is a private static method, so no one else than this class can call it.

Even if that is not the case, if `eps` is an invalid value, why return `0.0` instead of error?

Also, perhaps a question for @KrisThielemans: why have **static private** methods here? optimization?

robbietuk: `eps` is an term that was added to prevent divide by zeros and should be non-negative. However, it can be zero and therefore ignored. The check here is to return 0.0 if `x`, `y`, AND `eps` are zero. Otherwise, the function would compute `0^2/0^3`.

AnderBiguri: makes sense!

robbietuk:
Implemented `compute_Hessian` for QP, logcosh, and RDP all the same style. This is complete code duplication for each of these methods in preperation for #694 changes.

I have rewritten `compute_Hessian` for the QP and Logcosh. Their implementations were not incorrect but required the assumption that `d^2/dx^2 = - d^2/dxdy`. The RDP does not abide by this constraint. I have also added the `diagonal_second_derivative` and `off-diagonal_second_derivative` methods (name may be changed) to each of the classes. 

Added `accumulate_Hessian_times_input` to RDP and modernised the QP and logcosh priors to use the same methodology (code duplication again in prep for changes #694).

Added `_is_convex` and `get _is_convex()` to Generalised Prior. Default is false and each derived prior sets to true (if is the case)

* Potential issue is the PLS prior (which is still broken #705 and disabled in testing)

Regarding testing:
 - Added `Verbosity::set(0)` calls to suppress `compute_gradient` info calls
 - Added `test_gradient `, `test_Hessian_convexity`,`test_Hessian_convexity_configuration ` and  `test_Hessian_against_numerical` methods. Hessian tests are only conducted if `_is_convex=true`

robbietuk:
Currently `test_Hessian_against_numerical` will fail for both RDP checks. 

This is because of an inconsistancy between off-diagonal elements in the numerical-Hessian and the Hessian. The RDP is very sensitivie if `eps` (the perturbation) is too large relative to the comparison voxel values. Comparing the numerical-Hessian and the Hessian, the error is usually small (~10%), but can be large (>~60%), and occurs when both voxels values (x_j,x_k) are small w.r.t. `eps`, ergo effectively a large change in the numerical gradient. 

I am fairly happy that the implementation of the RDP's Hessian is correct, but the test is the issue.

src/recon_test/test_priors.cxx
```diff
@@ -96,46 +133,253 @@ run_tests_for_objective_function(const std::string& test_name,
...
+          input[perturbation_coords[1]][perturbation_coords[2]][perturbation_coords[3]] = perturbed_voxels_original_value;
+
+          // Now compute the numerical-Hessian = (g(x+eps) - g(x))/eps
+          *pert_grad_and_numerical_Hessian_sptr -= *gradient_sptr;
+          *pert_grad_and_numerical_Hessian_sptr /= eps;
+
+          Verbosity::set(verbosity_default);
+          // Test if pert_grad_and_numerical_Hessian_sptr is all zeros.
+          // This can happen if the eps is too small. This is a quick test that allows for easier debugging.
+          if (pert_grad_and_numerical_Hessian_sptr->sum_positive() == 0.0)
```
robbietuk, line 358: if (pert_grad_and_numerical_Hessian_sptr->sum_positive() == 0.0 && Hessian_sptr->sum_positive() > 0.0)

robbietuk:
Create bools to toggle the use of various tests for different penalties. RDP with RDP_eps = 0 will likely allways fail the numerical Hessian test.

robbietuk:
Ready for review

src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -125,9 +125,10 @@ class LogcoshPrior:  public
...
     //! compute Hessian
-    void compute_Hessian(DiscretisedDensity<3,elemT>& prior_Hessian_for_single_densel,
-                         const BasicCoordinate<3,int>& coords,
-                         const DiscretisedDensity<3,elemT> &current_image_estimate);
+    virtual Succeeded
+    compute_Hessian(DiscretisedDensity<3,elemT>& prior_Hessian_for_single_densel,
+                    const BasicCoordinate<3,int>& coords,
+                    const DiscretisedDensity<3,elemT> &current_image_estimate) const;
 
     //! Compute the multiplication of the hessian of the prior multiplied by the input.
```
KrisThielemans, line 133: let's correct this one as well
```suggestion
    //! Compute the multiplication of the hessian of the prior (at \c current_estimate) and the \c input.
```

src/recon_buildblock/LogcoshPrior.cxx
```diff
@@ -376,18 +377,21 @@ compute_gradient(DiscretisedDensity<3,elemT>& prior_gradient,
...
                 const BasicCoordinate<3,int>& coords,
-                const DiscretisedDensity<3,elemT> &current_image_estimate)
+                const DiscretisedDensity<3,elemT> &current_image_estimate) const
 {
   assert(  prior_Hessian_for_single_densel.has_same_characteristics(current_image_estimate));
   prior_Hessian_for_single_densel.fill(0);
   if (this->penalisation_factor==0)
   {
-    return;
+    return Succeeded::yes;
```
KrisThielemans, line 390: so I'd revert these

src/recon_buildblock/RelativeDifferencePrior.cxx
```diff
@@ -430,6 +512,121 @@ add_multiplication_with_approximate_Hessian(DiscretisedDensity<3,elemT>& output,
...
+template <typename elemT>
+float
+RelativeDifferencePrior<elemT>::
+diagonal_second_derivative(const float x_j, const float x_k) const
+{
+  if (x_j > 0.0 || x_k > 0.0 || this->epsilon > 0.0)
+    return 2 * pow(2 * x_k + this->epsilon, 2) /
+           pow(x_j + x_k + this->gamma * abs(x_j - x_k) + this->epsilon, 3);
+  else
+    return 0.0;
```
KrisThielemans, line 615: I guess not defined? Should document what we do here. Is 0 a reasonable value?

src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -224,18 +225,19 @@ class LogcoshPrior:  public
...
-     This function returns the hessian of the logcosh function
-     * @param d the difference between the ith and jth voxel.
-     * @param scalar is the logcosh scalar value controlling the priors transition between the quadratic and linear behaviour
-     * @return the second derivative of the log-cosh function
+     Diagonal refers to the second derivative w.r.t. x_j only (i.e. diagonal of the Hessian)
+     Off-diagonal refers to the second derivative w.r.t. x_j and x_k (i.e. off-diagonal of the Hessian)
+     For LogCosh, the off diagonal is the negative of the diagonal.
+     * @param x_j is the target voxel.
+     * @param x_k is the voxel in the neighbourhood.
+     * @return the second order partial derivatives of the LogCosh Prior
```
KrisThielemans, line 235: I don't understand the definition of these. I certainly can't get it from the documentation.

prior is a symmetric function of p(x_j,x_k). (many of them are a function of (x_j-x_k), but not all). Is "diagonal" the second derivative w.r.t. to the first arg, and "off_diagonal" the derivative wr.t. each arg? That would make sense to have. I'd then call them differently though. somethng like `derivative_20() `and `derivative_11`().

src/include/stir/recon_buildblock/LogcoshPrior.h
```diff
@@ -224,18 +225,19 @@ class LogcoshPrior:  public
...
      */
-    static inline float Hessian(const float d, const float scalar)
-    {
-      const float x = d * scalar;
-      return square((1/ cosh(x)));
-    }
+    //@{
+    float diagonal_second_derivative(const float x_j, const float x_k) const;
+    float off_diagonal_second_derivative(const float x_j, const float x_k) const;
```
KrisThielemans, line 239: I believe `float` has to be `elemT`

KrisThielemans (changes requested):
some comments. I didn't check the tests yet

src/include/stir/recon_buildblock/GeneralisedPrior.h
```diff
@@ -102,6 +116,9 @@ class GeneralisedPrior:
   virtual void check(DataT const& current_estimate) const;
 
   bool _already_set_up;
+
+  //! Variable to indicate that the prior is a convex function, should be set in defaults by the derived class
+  bool _is_convex;
```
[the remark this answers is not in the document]
robbietuk: A pure virtual is safer I think.

src/include/stir/recon_buildblock/GeneralisedPrior.h
```diff
@@ -60,6 +60,17 @@ class GeneralisedPrior:
   virtual void compute_gradient(DataT& prior_gradient, 
 		   const DataT &current_estimate) =0; 
 
+  //! This should computes a single row of the Hessian
+  /*! Default implementation just call error(). This function needs to be overridden by the
+      derived class.
+      This method computes the Hessian of a particular voxel, indicated by \c coords, of the current image estimate.
+      The default method should overwrite the values in \c prior_Hessian_for_single_densel.
+   */
+  virtual Succeeded
```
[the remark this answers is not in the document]
robbietuk: Would you like this adjusted for all methods? `compute_Hessian`, `accumulate_Hessian_times_input` and `add_multiplication_with_approximate_Hessian`??

src/recon_buildblock/RelativeDifferencePrior.cxx
robbietuk: An interesting point that was not considered at the time. If x_j, x_k and epsilon = 0, it would compute `(0^2) / (0^3) = undefined`. However, I think that 0 is the only value we return and retain numerical stability but I do not think there is a gerneral correct answer here. Hopefully by setting the value to 0 we can minimise any impact of this instance.

KrisThielemans: probably right limit will be +inf, and left limit -inf then, which means the limit doesn't exist. Setting it to 0 is "reasonable" but should be documented.

src/include/stir/recon_buildblock/GeneralisedPrior.h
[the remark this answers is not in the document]
KrisThielemans: ok. naming of these is a bit of a mess of course. I guess we're breaking backwards compatibility already fairly hard, so what about renaming to `accumulate_Hessian_times_vector` and `accumulate_approximate_Hessian_times_vector`?

src/include/stir/recon_buildblock/LogcoshPrior.h
robbietuk: Yes. diagonal refers to the partial derivatve `d^2/d x_j^2` and off-diagonal is `d^2/(d x_j d x_k)`. I can rename the functions `derivative_20` and `derivative_1` and detail why these names are used. We could also create a `derivative_1` at a later date in gradient computations.

src/include/stir/recon_buildblock/LogcoshPrior.h
robbietuk: I believe we can do this in all instances and for all priors? 
```
    elemT diagonal_second_derivative(const elemT x_j, const elemT x_k) const;
    elemT off_diagonal_second_derivative(const elemT x_j, const elemT x_k) const;
```

src/include/stir/recon_buildblock/GeneralisedPrior.h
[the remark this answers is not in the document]
robbietuk: If we will rename, I would suggest `Hessian_vector_product`. I have seen that terminology utilised around the literature a fair amount.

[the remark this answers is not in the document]
robbietuk: I thought about this further and I dont think that these renames belong in this PR. It will impact a lot more than priors and is a little out of scope. I will create an issue that can be addressed after #901 and #902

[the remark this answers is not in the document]
KrisThielemans: makes sense

src/recon_buildblock/LogcoshPrior.cxx
robbietuk: Done

robbietuk:
> some comments. I didn't check the tests yet

I have addressed all pertinent suggestions

robbietuk:
Future progress on this PR should include an intermediate class for the impacted priors. This class should handle the for loops for each of the Hessian methods and create the virtual functions for the `derivative_20` and `derivative_11`.

robbietuk (commented):
I think this is ready to go, assuming the tests pass.

src/recon_buildblock/GeneralisedPrior.cxx
```diff
@@ -52,26 +52,37 @@ set_up(shared_ptr<const TargetT> const&)
...
+void
+GeneralisedPrior<TargetT>::
+compute_Hessian(TargetT& output,
+                const BasicCoordinate<3,int>& coords,
+                const TargetT& current_image_estimate) const
+{
+  if (this->is_convex())
+    error("GeneralisedPrior:\n  compute_Hessian implementation is not overloaded by your convex prior.");
+  else
+    error("GeneralisedPrior:\n  compute_Hessian is not implemented because the prior is not convex.");
```
KrisThielemans, line 64: I'd rephrase this. a non-convex prior could be 2nd order differentiable of course.
```suggestion
    error("GeneralisedPrior:\n  compute_Hessian is not implemented for this (non-convex) prior.");
```

KrisThielemans (changes requested):
looks good. One tiny comment. Also, can you add something to `release_5.1.htm`? You probably will have to merge `master` into here first to get that.

KrisThielemans (approved):

Issue #914, opened 2021-08-03
list_ROI_values - 75% of computation is constructing shapes


robbietuk:
I ran a profile of `list_ROI_values` for a (251,251,47 voxel) image and found that the majority of computation is in the construction of the shapes (`construct_volume`).

![image](https://user-images.githubusercontent.com/13292409/128080309-3cfe158b-cf86-4783-9555-12d8c4b5c235.png)

When running an ROI analysis on multiple images (custom code, see #913 ), repeating `construct_volume` result in needless additional computation.

This can be avoided by adding an intermediate function, parameterised by a VoxelsOnCartesianGrid object instead of `Shape3D`, that computes the ROI values. The main function (parameterised by the `Shape3D`)  can `construct_volume` and call the intermediate function.

robbietuk:
Resolved by #915

PR #913, opened 2021-08-03
ROIValues and compute_total_ROI_values exposed to python

robbietuk:
I have tested on my machine and I have access to `ROIValues` class and `compute_total_ROI_values`.

PR #915, opened 2021-08-03
Compute ROI values from voxelised shape

robbietuk:
Addresses #914 

Overloads `compute_total_ROI_values` and `compute_ROI_values_per_plane`The new functions take `discretised_shape` (a `VoxelsOnCartesianGrid<float>` object) as argument. This was implemented as 75% of ROI value computation is creating the shapes. This change allows that command to be run once.

### DIFF
--- a/documentation/release_5.0.htm
+++ b/documentation/release_5.0.htm
@@ -251,6 +251,10 @@ as used by <code>copy_to</code> and <code>fill_from</code>
   Before this, the <code>add_multiplication_with_approximate_Hessian(output, input)</code> method was used that assumed a quadratic function and therefore <code>hessian = 1</code>.
 </li>
 <li><code>Exam_info</code>, <code>TimeFrameDefinition</code> and <code>PatientPosition </code> have now a <code>==</code> operator</li>
+<li>
+    Overloaded <code>compute_total_ROI_values</code> and <code>compute_ROI_values_per_plane</code> functions to allow the parsing of <code>VoxelsOnCartesianGrid</code> instead of <code>Shape3D</code> as the ROI shape.
+    This allows the ROI volume to be constructed during preprocessing for multiple image analysis.
+</li>
 </ul>
 
 
--- a/documentation/release_5.0.htm
+++ b/documentation/release_5.0.htm
@@ -252,7 +252,7 @@ as used by <code>copy_to</code> and <code>fill_from</code>
 </li>
 <li><code>Exam_info</code>, <code>TimeFrameDefinition</code> and <code>PatientPosition </code> have now a <code>==</code> operator</li>
 <li>
-    Overloaded <code>compute_total_ROI_values</code> and <code>compute_ROI_values_per_plane</code> functions to allow the parsing of <code>VoxelsOnCartesianGrid</code> instead of <code>Shape3D</code> as the ROI shape.
+    Overloaded <code>compute_total_ROI_values</code> and <code>compute_ROI_values_per_plane</code> functions to allow the passing of  a<code>VoxelsOnCartesianGrid</code> object instead of a <code>Shape3D</code> as the ROI shape.
     This allows the ROI volume to be constructed during preprocessing for multiple image analysis.
 </li>
 </ul>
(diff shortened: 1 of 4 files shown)


robbietuk:
This is a seperate PR to #913 but we should wait until that PR is merged and then expand the `ROI_demo.py` to include a demonstration of this functionality.

src/include/stir/evaluation/compute_ROI_values.h
```diff
@@ -55,6 +64,11 @@ compute_total_ROI_values(const DiscretisedDensity<3,float>& image,
                          const CartesianCoordinate3D<int>& num_samples
 			 );
 
+ROIValues
+compute_total_ROI_values(const DiscretisedDensity<3,float>& image,
+                         const VoxelsOnCartesianGrid<float>& discretised_shape_ptr,
```
KrisThielemans, line 69:
```suggestion
                         const VoxelsOnCartesianGrid<float>& discretised_shape,
```

src/eval_buildblock/compute_ROI_values.cxx
```diff
@@ -127,6 +138,16 @@ compute_total_ROI_values(const DiscretisedDensity<3,float>& image,
     compute_total_ROI_values(values);
 }
 
+ROIValues
+compute_total_ROI_values(const DiscretisedDensity<3,float>& image,
+                         const VoxelsOnCartesianGrid<float>& discretised_shape_ptr,
```
KrisThielemans, line 143:
```suggestion
                         const VoxelsOnCartesianGrid<float>& discretised_shape,
```

src/eval_buildblock/compute_ROI_values.cxx
```diff
@@ -127,6 +138,16 @@ compute_total_ROI_values(const DiscretisedDensity<3,float>& image,
...
+                         const CartesianCoordinate3D<int>& num_samples)
+{
+  VectorWithOffset<ROIValues> values;
+    compute_ROI_values_per_plane(values, image, discretised_shape_ptr, num_samples);
```
KrisThielemans, line 147:
```suggestion
    compute_ROI_values_per_plane(values, image, discretised_shape, num_samples);
```

documentation/release_5.0.htm
```diff
@@ -251,6 +251,10 @@ <h3>New functionality</h3>
   Before this, the <code>add_multiplication_with_approximate_Hessian(output, input)</code> method was used that assumed a quadratic function and therefore <code>hessian = 1</code>.
 </li>
 <li><code>Exam_info</code>, <code>TimeFrameDefinition</code> and <code>PatientPosition </code> have now a <code>==</code> operator</li>
+<li>
+    Overloaded <code>compute_total_ROI_values</code> and <code>compute_ROI_values_per_plane</code> functions to allow the parsing of <code>VoxelsOnCartesianGrid</code> instead of <code>Shape3D</code> as the ROI shape.
```
KrisThielemans, line 255:
```suggestion
    Overloaded <code>compute_total_ROI_values</code> and <code>compute_ROI_values_per_plane</code> functions to allow the passing of  a<code>VoxelsOnCartesianGrid</code> object instead of a <code>Shape3D</code> as the ROI shape.
```

KrisThielemans (changes requested):
some comments, but I think the main one is that the new versions actually don't use `num_samples`. In fact, they require that the geometric info between image and `discretised_shape` are identical (something which is taken care of in the original version). So the new versions need to do a check on that. (There's a method for that I believe).

Also, to avoid a lot of casting in client code, I think it is better to have the `discretised_shape` arguments as `DiscretisedDensity<3,float>&` (and do the cast inside if it's needed).

robbietuk:
@KrisThielemans I thought drafts dont run travis... could you please kill these jobs as I forgot to use [ci skip]

robbietuk:
> (and do the cast inside if it's needed).

`test_ROIs` works locally and my python code works without the dynamic cast in the 
```
compute_ROI_values_per_plane(VectorWithOffset<ROIValues>& values,
                             const DiscretisedDensity<3,float>& density,
                             const DiscretisedDensity<3,float>& discretised_shape)
```
function. The dynamic cast to a `VoxelsOnCartesianGrid` is needed for the generation of the shape (due to voxel sizes?) and then all comparisons in the function are performed between elements of the arrays. Please correct me if I am wrong, my understanding of volume objects in STIR is limited.

Awaiting travis.

Issue #910, opened 2021-07-20
Travis osx, OpenMP using llvm clang 12.0.1 gives linking errors


robbietuk:
In two different PR's #901 and #902, the osx, OpenMP, using llvm clang builds are failing, see https://app.travis-ci.com/github/UCL/STIR/jobs/525748509. But it is only this build configuration, all others pass.

robbietuk:
This is a seperate travis macOS issue to that mentioned in #897

KrisThielemans:
error is https://app.travis-ci.com/github/UCL/STIR/jobs/532976171#L8622
```
Undefined symbols for architecture x86_64:
  "stir::BinNormalisation::set_up(std::__1::shared_ptr<stir::ExamInfo const> const&, std::__1::shared_ptr<stir::ProjDataInfo const> const&)", referenced from:
      vtable for stir::TrivialBinNormalisation in liblistmode_buildblock.a(LmToProjData.cxx.o)
  "stir::BinNormalisation::BinNormalisation()", referenced from:
      stir::LmToProjData::set_defaults() in liblistmode_buildblock.a(LmToProjData.cxx.o)
  "stir::BinNormalisation::~BinNormalisation()", referenced from:
      stir::LmToProjData::set_defaults() in liblistmode_buildblock.a(LmToProjData.cxx.o)
      stir::TrivialBinNormalisation::~TrivialBinNormalisation() in liblistmode_buildblock.a(LmToProjData.cxx.o)
      stir::TrivialBinNormalisation::~TrivialBinNormalisation() in liblistmode_buildblock.a(LmToProjData.cxx.o)
  "stir::TrivialBinNormalisation::registered_name", referenced from:
      stir::RegisteredParsingObject<stir::TrivialBinNormalisation, stir::BinNormalisation, stir::BinNormalisation>::get_registered_name() const in liblistmode_buildblock.a(LmToProjData.cxx.o)
  "stir::BinNormalisation::check(stir::ProjDataInfo const&) const", referenced from:
      vtable for stir::TrivialBinNormalisation in liblistmode_buildblock.a(LmToProjData.cxx.o)
  "stir::BinNormalisation::check(stir::ExamInfo const&) const", referenced from:
      vtable for stir::TrivialBinNormalisation in liblistmode_buildblock.a(LmToProjData.cxx.o)
  "typeinfo for stir::BinNormalisation", referenced from:
[...]

PR #921, opened 2021-08-21
KeyParser::parameter_info() works for vectorised keys

KrisThielemans:
ugly code, just like the reading... but at least `parameter_info()` will now list all keywords/values that we can read (I think).

This should now get a proper test ideally..


KrisThielemans:
travis failure due to unrelated #910

PR #922, opened 2021-08-22
Better error message when reading Interfile projdata 

KrisThielemans:
Reading of Interfile projection-data will now write a more explicit warning when the binary file is non-existent or empty improving over the cryptic `WARNING: read_data: error after reading from stream` that we wrote before).


KrisThielemans:
This change breaks too many things, in particular wherever we use an empty "template" projdata

PR #924, opened 2021-09-08
Add weights to Patlak linear regression

AnderBiguri:
Weights can be turned on or off in par file, assume Poisson distribution

- This PR assumes data is NOT decay corrected.

### DIFF
--- a/src/include/stir/modelling/PatlakPlot.h
+++ b/src/include/stir/modelling/PatlakPlot.h
@@ -139,6 +139,7 @@ class PatlakPlot : public RegisteredParsingObject<PatlakPlot, KineticModel>
   float _time_shift;   //!< Shifts the time to fit the timing of Plasma Data with the Projection Data.
   bool _in_correct_scale; //!< Switch to scale or not the model_matrix to the correct scale, according to the appropriate scale factor.
   bool _in_total_cnt;   //!< Switch to choose the values of the model to be in total counts or in mean counts.
+  bool _assume_poisson_distribution; //!< Assume that image is Poisson distributed, and weight linear regression accordingly [1/(counts/int{Cp(t)}^2)]
   std::string _blood_data_filename;   //!< Name of file in which the input function is stored
   PlasmaData _plasma_frame_data;    //!< Stores the plasma data into frames for brain studies
   std::string _time_frame_definition_filename;   //!< name of file to get frame definitions
--- a/src/modelling_buildblock/PatlakPlot.cxx
+++ b/src/modelling_buildblock/PatlakPlot.cxx
@@ -34,6 +34,7 @@ set_defaults()
   this->_time_shift=0.;
   this->_in_correct_scale=false;
   this->_in_total_cnt=false;
+  this->_assume_poisson_distribution=false;
 }
 
 const char * const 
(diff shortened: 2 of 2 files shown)


KrisThielemans (changes requested):
looks great. Can you add something to `release_5.0.htm` as new feature?

AnderBiguri:
@KrisThielemans Sure. Want to add some sort of check for decay correction? I don't remember well, but isn't one of the input parameters if the data is decay corrected?

KrisThielemans:
interesting. Checking through `PatlakPlot.cxx` there is no check for decay corrected images. All decay stuff is for the input-function.

Digging a bit deeper:
`PatlakPlot.h` doxygen says that images are "in decaying counts", but it isn't checked. In fact it can't be checked as `DynamicDiscretisedDensity` doesn't have a `get_if_decay_corrected` (or simply `decay_corrected()`).

Maybe you could also add that in? Should be a 5 min job? (hmmm...)

By the way, there's some outdated comments in the utility [here](https://github.com/UCL/STIR/blob/master/src/modelling_utilities/apply_patlak_to_images.cxx#L32). Once we remove that, we should also remove [this stuff](https://github.com/UCL/STIR/blob/master/src/modelling_utilities/apply_patlak_to_images.cxx#L24-L28).

AnderBiguri:
@KrisThielemans if I add to `DynamicDiscretisedDensity` `decay_corrected()`, this doesn't really solve much of a problem right? because we don't fill this nor use it anywhere. So I propose we just keep assuming everything is not decay corrected for now, adding the method to STIR that would check this and act accordingly seems like something that its its own PR, not just applicable to the Patlak stuff.

KrisThielemans:
@AnderBiguri I've updated the doc. Have a look if it makes sense please.

Note that we have the capability with `_in_total_cnt` to allow input in "activity" images (as opposed to usual STIR "counts"). I guess the Poisson weights would need modification then? (Maybe not because of the division with the cst of the matrix). could you check? If it gets too confusing, we could just error out for now...

AnderBiguri:
@KrisThielemans when the input is in "activity", is it decay corrected too? My gut says that it would make sense for it to be called activity only if its decay corrected, but if it can be non-decay corrected, then its just a multiplication by a constant, right? If that is the case, even if it the weight was not taking the image values into account, it would be the same.

 But indeed, the weights are reading the image values, so they are   `pixel_val/(\int Cp)^2`. I believe this equation only changes if there is decay correction, otherwise its just as is, irrelevant of `pixel_val` units, right?

KrisThielemans:
you're right that "activity" should be decay correct. We can have 3 cases:
- counts
- counts  / frame_duration (not sure if there's an accepted name for this, but we could call it "non-decay-corrected activity" or similar)
- activity = counts  / frame_duration * decay_factor

We currently say that we don't support decay corrected data, so I guess the last one is out. Only "counts" follow Poisson stats.

Regarding factors, global factors don't matter, as long as they're constant for every frame. However, that wouldn't be the case for any of these.

I think I wrote the eqs down somewhere but didn't get into the division by model-matrix. As this in itself has frame_duration and decay_factor contributions, this could be somewhat tricky, but then again for the weight calculation, it just seems a constant.

I have no time to check this better now. sorry. again, if it gets hairy, just throw an error and support only what you know is correct

PR #929, opened 2021-09-15
recon_demo.m was fixed

luciferlys:
Same error caused by using 'isequal' like in 'test_IO.m' and 'test_buildblock.m'. It's fixed now.


src/swig/test/matlab/test_buildblock.m
```diff
@@ -175,6 +175,7 @@
 assert(seg.find_min()==0);
 seg.fill(4);
 assert(seg.find_min()==4);
-assert(isequal(proj_data.set_segment(seg), success));
+assert(eq(proj_data.set_segment(seg), success)); % change isequal to eq
+seg_temp = proj_data.set_segment(seg);
```
robbietuk, line 179: I don't understand the purpose of this line addition change.

KrisThielemans (changes requested):
I'd rather have all comments like `% changed` removed before merging this. Possibly I do it myself.

PR #940, opened 2021-10-08
Super-block versions of find_ML_normfactors3D.cxx and apply_normfactors3D.cxx

emikhay:


KrisThielemans:
code doesn't compile. see https://github.com/UCL/STIR/runs/4224474062?check_suite_focus=true#step:7:655

emikhay:
@KrisThielemans @danieldeidda @VietAnhDao

KrisThielemans (changes requested):
looks good. can you please add the command line option and modify `release_5.0.htm`?

KrisThielemans:
@robbietuk this PR could affect the STIR-GATE-Connection. Most importantly, our cludge there of increasing the block-size for the 690 should not longer be necessary. https://github.com/UCL/STIR-GATE-Connection/issues/59. (still need to look into the offset)

PR #953, opened 2021-11-05
DiscretisedDensity write_to_file and python example

### LINKED ISSUE #952: Write to file for VoxelsOnCartesianGrid
The `ProjData` structure has a method `read_from_file` and a method `write_to_file`. However, the `VoxelsOnCartesianGrid` structure has a method `read_from_file` but no method `write_to_file`. 

Comments:
- it's possible to use `stir.write_image_to_file(filename, VoxelsOnCartesianGrid)` in Python
- there is no example of this in the python recon examples

robbietuk:
Fixes #952 and adds example usage to `recon_demo.py`

### DIFF
--- a/documentation/release_5.1.htm
+++ b/documentation/release_5.1.htm
@@ -48,6 +48,8 @@ none
   </li>    
   <li> Improvements to penalty functions including Hessian methods, see <a href="https://github.com/UCL/STIR/pull/901/">PR #901</a>.
   </li>
+  <li>Added <code>write_to_file(fname)</code> method to python <code>VoxelsOnCartesianGrid</code> objects and updated <code>recon_demo.py</code> as a demonstration.
+  </li>
 </ul>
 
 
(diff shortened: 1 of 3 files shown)


KrisThielemans (changes requested):
Thanks but I do not want to modify the C++ code. This creates (even more) circular dependencies between io and buildblock, which we should remove. Instead, we should add it in `stir.i` via `%extend`, like we do for `read_from_file` https://github.com/UCL/STIR/blob/84e4ddc6eb69621bc14af7ed3662febe52b180df/src/swig/stir.i#L963

robbietuk:
This took an embarrassingly long time to figure out extend and swig but it works for me now 😄

KrisThielemans (approved):

KrisThielemans:
looks good. let's wait for the tests.

Can you add a line in the latest release notes? Sadly probably means merging `master` into here as otherwise you won't have them (and we'd have conflicts).

robbietuk:
All should be good. These three comits were made since a rebase onto master today. There should not be a conflict.

PR #961, opened 2021-11-16
Randoms from singles fix and half life improvements

### LINKED ISSUE #960: Error in decay correction factor computation in `randoms_from_singles`
https://github.com/UCL/STIR/blob/07aad4371043136f6d2f0ee63893cfabb6e5104b/src/data_buildblock/randoms_from_singles.cxx#L79

This line is suppose to compute `2*lambda` by the equation. The convertion between `isotope_halflife` and `lambda` is handled in the `decay_correction_factor` function and so this line should read:
```
const double double_decay_corr_factor = decay_correction_factor(0.5*isotope_halflife, 0., duration); 
```

robbietuk:
- Fixes #960 by correcting the `decay_correction_factor` input.

- Goes further to unhardcode the `isotope_halflife` by getting from `projdata` (if exists) or from the `input_file`. To do this `GEHDF5Wrapper::get_halflife()` was added. This also required the `randoms_from_singles` function to have a `isotope_halflife` as an argument because that information was not necessarily present in the `projdata`.

- Improved the info output in `randoms_from_singles.cxx`, e.g:
 ```
INFO: Isotope half-life: 6586.2
RFS: decay correction factor: 1.20137,
time frame duration: 3600.
total correction factor from (singles_totals)^2 to randoms_totals: 3557.55.
```

### DIFF
--- a/src/data_buildblock/randoms_from_singles.cxx
+++ b/src/data_buildblock/randoms_from_singles.cxx
@@ -30,7 +30,7 @@
 START_NAMESPACE_STIR
 
 void randoms_from_singles(ProjData& proj_data, const SinglesRates& singles,
-                          const float coincidence_time_window, float isotope_halflife=-1.F)
+                          const float coincidence_time_window, float isotope_halflife)
 {
   if (isotope_halflife == -1.F)
     isotope_halflife = proj_data.get_exam_info().get_radionuclide().get_half_life();
--- a/src/data_buildblock/randoms_from_singles.cxx
+++ b/src/data_buildblock/randoms_from_singles.cxx
@@ -30,7 +30,7 @@
 START_NAMESPACE_STIR
 
 void randoms_from_singles(ProjData& proj_data, const SinglesRates& singles,
-                          const float coincidence_time_window)
+                          const float coincidence_time_window, const float isotope_halflife)
 {
   const int num_rings =
     proj_data.get_proj_data_info_sptr()->get_scanner_ptr()->get_num_rings();
@@ -73,13 +73,15 @@ void randoms_from_singles(ProjData& proj_data, const SinglesRates& singles,
        That leads to the formula below (as it turns out that the above ratio only depends t2-t1)
     */
     const double duration = frame_defs.get_duration(1);
-    warning("Assuming F-18 tracer!!!");
-    const double isotope_halflife = 6586.2;
     const double decay_corr_factor = decay_correction_factor(isotope_halflife, 0., duration);
     const double double_decay_corr_factor = decay_correction_factor(0.5*isotope_halflife, 0., duration);
     const double corr_factor = square(decay_corr_factor) / double_decay_corr_factor / duration;
-    info(boost::format("RFS: decay correction factor: %1%, time frame duration: %2%. total correction factor from (singles_totals)^2 to randoms_totals: %3%")
-         % decay_corr_factor % duration % (1/corr_factor),
+
+    info(boost::format("Isotope half-life: %1%\n"
+                       "RFS: decay correction factor: %2%,\n"
+                       "time frame duration: %3%.\n"
+                       "total correction factor from (singles_totals)^2 to randoms_totals: %4%.\n")
+         % isotope_halflife % corr_factor % duration % (1/corr_factor),
          2);
 
     multiply_crystal_factors(proj_data, total_singles,
--- a/src/data_buildblock/randoms_from_singles.cxx
+++ b/src/data_buildblock/randoms_from_singles.cxx
@@ -81,7 +81,7 @@ void randoms_from_singles(ProjData& proj_data, const SinglesRates& singles,
                        "RFS: decay correction factor: %2%,\n"
                        "time frame duration: %3%.\n"
                        "total correction factor from (singles_totals)^2 to randoms_totals: %4%.\n")
-         % isotope_halflife % corr_factor % duration % (1/corr_factor),
+         % isotope_halflife % decay_corr_factor % duration % (1/corr_factor),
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by KrisThielemans — CHANGES_REQUESTED** [review]
This is overall fine, but see specific comments. I think that the logic here regarding `exam_info` is a bit wrong: it should set nearly everything in `ExamInfo` from the input data. At present, the template is used. That seems to give scope for mistakes. But this isn't part of this fix I suppose :-).

I note #962, which explains why you had to set the half-life. Again, that should/can be fixed separately.

I mention all of this though because I don't really like that we now have to pass the half-life to `randoms_from_singles`. What about making it an optional parameter (i.e. default it to -1.F or so), and let if use the one from `proj_data.get_exam_info()` if it isn't set? Then at some point, this parameter will become obsolete again.

File: src/utilities/construct_randoms_from_GEsingles.cxx [hunk 1]
```diff
@@ -87,6 +87,11 @@ int main(int argc, char **argv)
   GE::RDF_HDF5::SinglesRatesFromGEHDF5  singles(input_filename);
   const float coincidence_time_window = input_file.get_coincidence_time_window();
 
-  randoms_from_singles(proj_data, singles, coincidence_time_window);
+  // If half-life exists in proj_data, use that value, otherwise get it from the input_file
+  const float isotope_halflife = proj_data.get_exam_info().get_radionuclide().get_half_life(false) != -1
+                                 ? proj_data.get_exam_info().get_radionuclide().get_half_life()
+                                 : input_file.get_halflife();
```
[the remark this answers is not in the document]
robbietuk: Are we therefore dependant on a fix of #962?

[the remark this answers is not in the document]
KrisThielemans: > I think this illistrates an issue with current code because of how projdata is constructed when a template is provided.
> ...
>  I also realise that this is becoming a construct_randoms_from_singles redesign

yes sure. Let's create an issue for this but not fix that here

> Is there any reason to keep the template sinogram option for the construct_randoms_from_GEsingles utility,

yes, to be able to specify different time frames and/or span/mashing.

> Are we therefore dependant on a fix of #962?

I wouldn't fix that here either...

My suggestion is just to optionally pass half-life, and if it isn't passed, get it from the `projdata`. But in the utility, we always pass it.

src/data_buildblock/randoms_from_singles.cxx
```diff
@@ -30,8 +30,11 @@
 START_NAMESPACE_STIR
 
 void randoms_from_singles(ProjData& proj_data, const SinglesRates& singles,
-                          const float coincidence_time_window)
+                          const float coincidence_time_window, float isotope_halflife=-1.F)
 {
+  if (isotope_halflife == -1.F)
+    isotope_halflife = proj_data.get_exam_info().get_radionuclide().get_half_life();
```
robbietuk, line 36: This relies on the `projdata` halflife being set but will error if not.

src/include/stir/data/randoms_from_singles.h
```diff
@@ -52,12 +52,11 @@ class SinglesRates;
...
   Random coincidence estimation from single event rates on the Discovery ST PET/CT scanner.
-  2003 IEEE Nuclear Science Symposium. Conference Record (IEEE Cat. No.03CH37515), 5, 3067–3069.
+  2003 IEEE Nuclear Science Symposium. Conference Record (IEEE Cat. No.03CH37515), 5, 3067-3069.
   https://doi.org/10.1109/NSSMIC.2003.1352545
 
   \todo Dead-time is currently completely ignored.
-  \todo The function currently assumes F-18 half-life.
 */
-void randoms_from_singles(ProjData& proj_data, const SinglesRates& singles, const float coincidence_time_window, const float isotope_halflife);
+void randoms_from_singles(ProjData& proj_data, const SinglesRates& singles, const float coincidence_time_window, float isotope_halflife=-1.F);
```
KrisThielemans, line 60: please document what -1 means

src/data_buildblock/randoms_from_singles.cxx
```diff
@@ -30,8 +30,11 @@
 START_NAMESPACE_STIR
 
 void randoms_from_singles(ProjData& proj_data, const SinglesRates& singles,
-                          const float coincidence_time_window, const float isotope_halflife)
+                          const float coincidence_time_window, float isotope_halflife=-1.F)
```
KrisThielemans, line 33:
```suggestion
                          const float coincidence_time_window, float isotope_halflife)
```
no default in the implementation, only in the declaration

src/data_buildblock/randoms_from_singles.cxx
KrisThielemans: what do you mean with "will error". Throw an error? If so, ok. But if you mean "silently return the wrong results", then that'd be a problem. I then suggest to add a check if it's still <0, then use F-18 (and write the warning).

KrisThielemans (changes requested):

src/data_buildblock/randoms_from_singles.cxx
robbietuk: It is okay as is, will call the `error()` function if `check==true`, which it does by default. https://github.com/UCL/STIR/blob/07aad4371043136f6d2f0ee63893cfabb6e5104b/src/buildblock/Radionuclide.cxx#L75-L80

KrisThielemans (approved):

KrisThielemans:
ok to squash-merge this one?

robbietuk:
Sure!

PR #969, opened 2021-11-24
Subset Hackathon: ProjData::get_subset

ashgillman:
Original plan
- [x] Add a std::vector of "views present" in a deriving ProjDataInfoSubsetByView . Keep a track of original ProjDataInfo. Use that for get_phi and all those functions first translate view to original_view, and vice versa. (complication: how to do this with view-mashing?)
- [ ] Add a keyword for it in the interfile stuff.
- [x] Write code to construct a new ProjData with fewer views and test.
- [x] See if projectors work with all view-symmetries switched off (i.e. write test)
- [ ] See if SIRF works ok (i.e. write test)
- [x] Work on DataSymmetriesForPET… and see if projectors with symmetries (probably not this hackathon)
- [ ]  Classes for generating which views below to which subset (a la `StandardSubsetter::get_views_for_subsets(int subset)` in `test_proj_data_info_subsets.cxx`.

Edit
- [x]  Test w/ unbalanced subsets?
- [x]  test backprojection
- [x]  test z shift symmetry
- [ ]  add Interfile keywords
- [ ]  test I/O
- [x]  test reduce_segment_range


ashgillman:
I've started on `ProjDataInfoSubsetByView`, it will need to be a concrete class, but will offload the majority of its work to a `full_proj_data_info_sptr`.

The testing still needs to be fleshed out a lot. `compare_views` will look like `compare_sinos` but using Viewgrams.

src/buildblock/ProjData.cxx
```diff
@@ -225,6 +229,27 @@ read_from_file(const string& filename,
...
+      for (int subset_view_num=0; subset_view_num<views.size(); ++subset_view_num)
+        {
+          const Viewgram<float> viewgram = this->get_viewgram(views[subset_view_num], segment_num);
+          Viewgram<float> subset_viewgram = subset_proj_data_info_sptr-> get_empty_viewgram(subset_view_num, segment_num);
+          subset_viewgram.fill(viewgram);
+          if (subset_proj_data.set_viewgram(subset_viewgram) != Succeeded::yes)
+            error("ProjData::get_subset failed to set a viewgram");
+        }
+    }
+}
```
evgueni-ovtchinnikov, line 251: `return subset_proj_data` missing?

src/buildblock/ProjData.cxx
```diff
@@ -225,6 +229,27 @@ read_from_file(const string& filename,
 //  this->exam_info_sptr.reset(new ExamInfo(new_exam_info));
 //}
 
+
+ProjDataInMemory
```
evgueni-ovtchinnikov, line 233: Frankly, I do not like ABC returning an instance of derived class - I firmly believe that ABC should not be aware about classes derived from it.
@ashgillman How about this trick that we use in SIRF: add this method to `ProjDataInMemory`:
```
virtual ProjData* same_proj_data
	(stir::shared_ptr<const stir::ExamInfo> sptr_exam_info,
	stir::shared_ptr<stir::ProjDataInfo> sptr_proj_data_info) const
{
	ProjData* ptr_pd =
		new ProjDataInMemory(sptr_exam_info, sptr_proj_data_info);
	return ptr_pd;
}
```
and same-prototype one to `ProjData`, returning zero pointer for now (until all other derived classes get proper ones eventually). It can then be used instead of what you have on line 238 to create `unique_ptr<ProjData>` to be returned by your `get_subset`.

ashgillman:
I need some help with the current errors during linking, as I'm not sure what's wrong for the linker

src/buildblock/ProjData.cxx
KrisThielemans: Seems that STIR indeed misses functionality like `same_proj_data`. We have `get_empty_copy` in the image front, but not for `ProjData`. However, this would be a considerable amount of work. In fact, it is virtually impossibly to create a GE/Siemens/whatever proj-data from scratch as we don't have all of the info on the `ExamInfo` and `ProjDataInfo` that would be needed for that. The usual strategy for that is to copy existing files, and then modify the things we need. Implementing even that is a large amount of work. We can currently only create new `ProjDataInterfile` and `ProjDataInMemory` objects.

It might even be entirely undesirable to have it return a `ProjData` of the same type.

Because of all this we took the past of least resistance at the moment. we could easily decide to return a `unique_ptr<ProjData>` and optional arguments such as a a filename, and `ios:openmode` flags in which case it will create a `ProjDataInterfile` accordingly. More than that I think is very much out-of-scope

src/buildblock/ProjData.cxx
KrisThielemans: yes indeed. feel free to commit (with `[ci skip]`)

evgueni-ovtchinnikov: @KrisThielemans committed but cannot push to `ashgillman/STIR`

KrisThielemans: ah, @ashgillman maybe you fix this? For now, @evgueni-ovtchinnikov  just do a PR on @ashgillman's branch.

KrisThielemans: I've committed this now. thanks for spotting.

KrisThielemans:
above linking problem was because the cxx file wasn't in the `CMakeLists.txt`.

Everything compiles/links now, but the test seg-faults. Various problems remain (some marked with TODO). Cause for seg-fault is that `ProjDataInfo` expects some private members to be set., which was a very bad design It really needs an intermediate class but I'll try to work around that for now.

KrisThielemans:
test works now!

KrisThielemans:
I've did a minimal modification to the symmetry code (i.e. disable view symmetries if it's a subset case). I believe this should now work ok.

So... we need to test projectors. Maybe like this
1. create image/projdata and forward project
2. get subset projdata and forward project image into that
3. compare
4. set other views in original projdata to zero and backproject
5. backproject subset projdata
6. compare

This should in principle be done for all projectors though.

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -91,30 +91,42 @@ run_tests()
...
+    ProjDataInMemory subset_proj_data = proj_data.get_subset(subset_views);
+
+    // loop over views in the subset data and compare them against the original "full" data
+    for(std::size_t i = 0; i < subset_views.size(); ++i){
+      // i runs from 0, 1, ... views_in_subset - 1 and indicates the view number in the subset
+      // the corresponding view in the original data is at subset_views[i]
+
+      // TODO: Check how to get number of segments from proj_data
+      // loop over all segments to check viewgram for all segments
+      for (int segment = 0; segment < proj_data.get_num_segments(); ++segment){
```
KrisThielemans, line 121:
```suggestion
      for (int segment = proj_data.get_min_segment_num(); segment < proj_data.get_max_segment_num(); ++segment){
```

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -91,30 +91,42 @@ run_tests()
...
+        // TODO: implement compare_views
+        compare_views(
+            proj_data.get_viewgram(subset_views[i], segment),
+            subset_proj_data.get_viewgram(i, segment), "Are viewgrams equal?");
+        // TODO also compare viewgram metadata
```
KrisThielemans, line 126: somewhat tricky as it will be different. However, we should at least compare `viewgram.get_proj_data_info_sptr()->get_phi(Bin(....))`. Possibly this doesn't need to be from the `viewgram` but just use the original `proj_data_info`, as those are pretty much guaranteed to be the same (could add a check if you want)

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -91,30 +91,42 @@ run_tests()
...
-    std::vector<int> odd_views;
-
-    for (int view = 0; view < proj_data.get_num_views(); ++ view)
-      {
-        if (view % 2 == 0)
-            even_views.push_back(view);
-        else
-            odd_views.push_back(view);
-      }
+  int view;
```
KrisThielemans, line 94:
```suggestion
```
do not predeclare variables in C++

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -91,30 +91,42 @@ run_tests()
...
-    ProjDataInMemory even_subset = proj_data.get_subset(even_views);
-    ProjDataInMemory odd_subset = proj_data.get_subset(odd_views);
+  for (int subset = 0; subset < num_subsets; ++subset){
+    // create a vector containg every num_subsest-th view starting at subset
+    // for num_subsets = 4 and subset = 0 this is [0, 4, 8, 12, ...]
+    // for num_subsets = 4 and subset = 1 this is [1, 5, 9, 13, ...]
 
-    int segment = 0; // for segment
+    subset_views.clear();
+    view = subset;
```
KrisThielemans, line 105:
```suggestion
    int view = subset;
```

KrisThielemans: standard STIR naming is `view_num` though

KrisThielemans (changes requested):

src/buildblock/ProjData.cxx
evgueni-ovtchinnikov: @KrisThielemans the most important thing here is to return `unique_ptr<ProjData>` whichever way the subset object is created, otherwise it will not be possible to wrap it into `PETAcquisitionData` without cloning (one copy already being created at present)

evgueni-ovtchinnikov: @KrisThielemans sorry to be such a bore but I tried this in VS 2019:
```
#include <iostream>
#include <memory>

class A {
public:
	int i;
	A(int v) : i(v) {}
	A(const A& a)
	{
		std::cout << "copy constructor called\n";
		i = a.i;
	}
};

A f()
{
	A a(8);
	return a;
}

int main()
{
	A a(42);
	std::cout << "calling f()...\n";
	a = f();
	std::cout << "calling f() with make_shared...\n";
	auto b = std::make_shared<A>(f());
	return 0;
}
```
and I got
```
calling f()...
copy constructor called
calling f() with make_shared...
copy constructor called
copy constructor called
```
On VM with Ubuntu20.04 and g++ version 9.3.0, I get
```
calling f()...
calling f() with make_shared...
copy constructor called
```
So 
* apparently VS 2019 is not clever enough to avoid copying return value, and 
* it looks like
```
make_shared<A>(a)
```
actually simply does this (as I always thought):
```
shared_ptr<A>(new A(a))
```
So, with `ProjData::get_subset`, why not take this IMHO simplest possible way to avoid unnecessary copies:
```
unique_ptr<ProjData>
ProjData::get_subset(const std::vector<int>& views) const {
...
auto uptr_subset_proj_data = unique_ptr<ProjData>(new ProjDataInMemory(exam_info_sptr, subset_proj_data_info_sptr));
subset_proj_data *uptr_subset_proj_data;
...
return uptr_subset_proj_data;
}
```

KrisThielemans: interesting. makes me wonder what Microsoft plans are for their C++ compiler if they even didn't do that. thanks for trying!  please do change the code then...

evgueni-ovtchinnikov: @KrisThielemans I have just forked Ashley's STIR branch (could not push to it) but it does not have your latest edits - where can I get them from? UCL/STIR does not seem to have `subset_object` branch.

evgueni-ovtchinnikov: @KrisThielemans forget about my previous comment - I opened a file before forking and forgot that leafpad does not automatically update edited file.

evgueni-ovtchinnikov: I get this build error:
```
/home/sirfuser/devel/build/sources/STIR/src/recon_buildblock/DataSymmetriesForBins_PET_CartesianGrid.cxx:141:88: error: ‘using element_type = const class stir::ProjDataInfo’ {aka ‘const class stir::ProjDataInfo’} has no member named ‘get_org_proj_data_info_sptr’; did you mean ‘construct_proj_data_info’?

if(!is_null_ptr(dynamic_cast<const ProjDataInfoCylindrical*>(proj_data_info_ptr->get_org_proj_data_info_sptr().get())))
```
Indeed, `proj_data_info_ptr` is declared as `ProjDataInfo`, which does not have that new method. It is cast to `ProjDataInfoSubsetByView` in the quoted line
```
  if(!is_null_ptr(dynamic_cast<const ProjDataInfoSubsetByView *>(proj_data_info_ptr.get())))
```
but the result of the cast is not assigned to any variable.
I edited as follows
```
  const ProjDataInfoSubsetByView *ptr=dynamic_cast<const ProjDataInfoSubsetByView *>(proj_data_info_ptr.get());
  if(!is_null_ptr(ptr))
```
and the error message disappeared.
@KrisThielemans But I thought you already managed to build ok, and `git fetch` does not fetch - were you working on some other fork/branch?

KrisThielemans:
we're green now!

I think it would be best to move the current content of `test_proj_data_info_subsets.cxx` to `test_proj_data_info` in a separate method there, and call its `test_generic_proj_data_info` at the end, that way we have extra tests for free. The tests with projectors should be in a class in `recon_test` anyway. Any volunteer to move the code?

src/buildblock/ProjData.cxx
```diff
@@ -225,6 +229,28 @@ read_from_file(const string& filename,
...
 //}
 
+
+unique_ptr<ProjData>
+ProjData::get_subset(const std::vector<int>& views) const
+{
+  auto subset_proj_data_info_sptr =
+    std::make_shared<ProjDataInfoSubsetByView>(proj_data_info_sptr, views);
+  auto subset_proj_data_ptr = new ProjDataInMemory(exam_info_sptr, subset_proj_data_info_sptr);
+  auto subset_proj_data = *subset_proj_data_ptr;
```
KrisThielemans, line 239: risk of copy, so
```suggestion
  auto& subset_proj_data = *subset_proj_data_ptr;
```

src/include/stir/ProjData.h
```diff
@@ -127,6 +128,10 @@ class ProjData : public ExamData
   virtual Succeeded 
     set_sinogram(const Sinogram<float>&) = 0;
 
+  //! construct projection data that stores a subset of the views
+  unique_ptr<ProjData>
```
KrisThielemans, line 132:
```suggestion
  unique_ptr<ProjDataInMemory>
```
this avoids casting below. This should automatically convert to a `unique_ptr<ProjData>` anyway

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -0,0 +1,220 @@
...
+
+    subset_views.clear();
+    view = subset;
+
+    while(view < proj_data.get_num_views()){
+      subset_views.push_back(view);
+      view += num_subsets;
+    }
+
+    ProjDataInMemory& subset_proj_data = (ProjDataInMemory&)*proj_data.get_subset(subset_views);
```
KrisThielemans, line 130: Dangerous and will fail for reasons:
- C-style cast (equivalent to `static_cast`, and the latter is more verbose but much preferred) will not do any run-time checks, so as soon a `get_subset()` would return something else, this would crash with a seg-fault.
- you have a reference to a temp-object, which could/disappear immediately.

As we don't need to know what type it is, we can resolve this either like this
```suggestion
    auto subset_proj_data_uptr = proj_data.get_subset(subset_views);
    auto& subset_proj_data = *auto subset_proj_data_uptr;
```
or just use the first line and replace `proj_data.get*` with `proj_data_uptr->get*` (which I think is cleaner)

KrisThielemans (changes requested):

KrisThielemans:
build errors are due to SWIG not supporting `unique_ptr` (quite sadly). Let me commit a work-around for that

ashgillman:
Running into the error here:

https://github.com/UCL/STIR/pull/969/files#diff-bf2211a394ea137258920c27b0445c9d25b67ac04c48e0e1f015aa229ca8ca80R143-R144

"can only handle projection data corresponding to a cylinder"

I haven't had a chance to probe into it yet. But I'm a bit confused, non-arc-corrected data you generate for the ProjData, @KrisThielemans, should be cylindrical?

KrisThielemans:
confusing. what's the whole error (it should print the typename, which will likely be mangled, in which case running `nm` on it might help. And are you sure it's that line (as the one below has an identical error, which isn't very good of course)

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -0,0 +1,332 @@
...
+  return std::make_shared<ProjDataInMemory>(exam_info_sptr, proj_data_info_sptr);
+}
+
+shared_ptr<VoxelsOnCartesianGrid<float> > TestProjDataInfoSubsets::construct_test_image_data(const ProjDataInfo &template_projdatainfo)
+{
+  cerr << "\tGenerating default image of Ellipsoid" << endl;
+  shared_ptr<VoxelsOnCartesianGrid<float> > image = std::make_shared<VoxelsOnCartesianGrid<float> >(template_projdatainfo);
+
+  // make radius 0.8 FOV
+  auto radius = image->get_physical_coordinates_for_indices(image->get_min_indices()) * 0.8;
```
KrisThielemans, line 98: this will set `radius.z()=0`

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -0,0 +1,332 @@
...
+void TestProjDataInfoSubsets::
+test_forward_projection_is_consistent(
+    const shared_ptr<const VoxelsOnCartesianGrid<float> >& input_image_sptr,
+    const shared_ptr<const ProjData>& template_sino_sptr,
+    const shared_ptr<ForwardProjectorByBin>& fwd_projector_sptr,
+    bool use_symmetries, int num_subsets)
+{
+  cerr << "\tTesting Subset forward projection is consistent" << endl;
+  
+  auto full_forward_projection = ProjDataInMemory(*template_sino_sptr);
```
KrisThielemans, line 246: while this compiles, it means that `full_forward_projection` will be a reference to an object that is immediately deleted.

KrisThielemans (changes requested):
I'm fixing these myself, but there's still something else going on

KrisThielemans:
I've fixed some issues but still have a test failure.
```
	Subset 0, forward projecting subset

WARNING: forward_project. Imaging modality unknown for either the image or the projection data or both.
Going ahead anyway.

ERROR: ForwardProjectorByBin set-up with different geometry for projection data.
Set_up was with
ProjDataInfoCylindricalNoArcCorr := 
Scanner parameters:= 
Scanner type := ECAT 953
Number of rings                          := 16
Number of detectors per ring             := 384
Inner ring diameter (cm)                 := 76.5
Average depth of interaction (cm)        := 0.7
Distance between rings (cm)              := 0.675
Default bin size (cm)                    := 0.312932
View offset (degrees)                    := 15
Maximum number of non-arc-corrected bins := 160
Default number of arc-corrected bins     := 160
Number of blocks per bucket in transaxial direction         := 4
Number of blocks per bucket in axial direction              := 1
Number of crystals per block in axial direction             := 8
Number of crystals per block in transaxial direction        := 8
Number of detector layers                                   := 1
Number of crystals per singles unit in axial direction      := 8
Number of crystals per singles unit in transaxial direction := 32
end scanner parameters:=

start vertical bed position (mm) := 0
start horizontal bed position (mm) := 0

Segment_num range:           (-3, 3)
Number of Views:                24
Number of axial positions per seg: {5 15 25 31 25 15 5 }
Number of tangential positions: 64
Azimuthal angle increment (deg):   7.5
Azimuthal angle extent (deg):      180
ring differences per segment: 
(-15,-13)(-12,-8)(-7,-3)(-2,2)(3,7)(8,12)(13,15)
End :=

Called with
ProjDataInfoCylindricalNoArcCorr := 
Scanner parameters:= 
Scanner type := ECAT 953
Number of rings                          := 16
Number of detectors per ring             := 384
Inner ring diameter (cm)                 := 76.5
Average depth of interaction (cm)        := 0.7
Distance between rings (cm)              := 0.675
Default bin size (cm)                    := 0.312932
View offset (degrees)                    := 15
Maximum number of non-arc-corrected bins := 160
Default number of arc-corrected bins     := 160
Number of blocks per bucket in transaxial direction         := 4
Number of blocks per bucket in axial direction              := 1
Number of crystals per block in axial direction             := 8
Number of crystals per block in transaxial direction        := 8
Number of detector layers                                   := 1
Number of crystals per singles unit in axial direction      := 8
Number of crystals per singles unit in transaxial direction := 32
end scanner parameters:=

start vertical bed position (mm) := 0
start horizontal bed position (mm) := 0

Segment_num range:           (-3, 3)
Number of Views:                24
Number of axial positions per seg: {5 15 25 31 25 15 5 }
Number of tangential positions: 64
Azimuthal angle increment (deg):   7.5
Azimuthal angle extent (deg):      180
ring differences per segment: 
(-15,-13)(-12,-8)(-7,-3)(-2,2)(3,7)(8,12)(13,15)
End :=

End of tests. Please correct errors !
```
This is cryptic as I believe they list the same, but that's likely because we haven't adjusted `parameter_info` yet so it doesn't say it's a subset projdatainfo. I can't immediately see the coding error though.

Worse, there is then a segfault in the destructors (probably a double destruction). Probably too much casting going on.

I have to go on and do other stuff though.

ashgillman:
Thanks for decrypting my last commit which must have been mysterious. I intended to post an update here but was pulled away before I had time.

ashgillman:
My C++ debugging foo isn't so strong, perhaps @evgueni-ovtchinnikov might have a suggestion? (more around my logic below than anything specific to the STIR codebase)

It seems to be crashing while de-allocating a `VectorWithOffset` that perhaps belongs to a `ProjDataInfoCylindricalNoArcCorr` that is being de-allocated, while in turn belong to a `ProjDataInfoSubsetByView` being de-allocated? (all during the de-allocation of the projectors).

So perhaps the `ProjDataInfoSubsetByView::org_proj_data_info_sptr` has an issue with reference counting such that it's already been de-allocated before `ProjDataInfoSubsetByView` gets to it? Would that be consistent with this error?

The backtrace is:

```
#0  0x00000000006e5d0d in __gnu_cxx::__exchange_and_add (__val=-1, __mem=0x1f)                                        
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/ext/atomicity.h:82                     
#1  __gnu_cxx::__exchange_and_add_dispatch (__val=-1, __mem=0x1f)                                                     
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/ext/atomicity.h:82                     
#2  __gnu_cxx::__exchange_and_add_dispatch (__val=-1, __mem=0x1f)                                                     
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/ext/atomicity.h:78                     
#3  std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x17)                                         
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:152             
#4  std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0xcd5a00, __in_chrg=<optimised out>)       
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:730             
#5  std::__shared_ptr<std::vector<std::pair<int, int>, std::allocator<std::pair<int, int> > >, (__gnu_cxx::_Lock_polic
y)2>::~__shared_ptr (this=0xcd59f8, __in_chrg=<optimised out>)                                                        
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:1169            
#6  std::shared_ptr<std::vector<std::pair<int, int>, std::allocator<std::pair<int, int> > > >::~shared_ptr (          
    this=0xcd59f8, __in_chrg=<optimised out>)                                                                         
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr.h:103                  
#7  stir::VectorWithOffset<std::shared_ptr<std::vector<std::pair<int, int>, std::allocator<std::pair<int, int> > > > >
::_destruct_and_deallocate (this=0x9ef098) at /home/gil2a4/src/STIR/src/include/stir/VectorWithOffset.inl:83          
#8  stir::VectorWithOffset<std::shared_ptr<std::vector<std::pair<int, int>, std::allocator<std::pair<int, int> > > > >
::_destruct_and_deallocate (this=0x9ef098) at /home/gil2a4/src/STIR/src/include/stir/VectorWithOffset.inl:70          #9  stir::VectorWithOffset<std::shared_ptr<std::vector<std::pair<int, int>, std::allocator<std::pair<int, int> > > > >
::~VectorWithOffset (this=0x9ef098, __in_chrg=<optimised out>)                                                        
    at /home/gil2a4/src/STIR/src/include/stir/VectorWithOffset.inl:304                                                
#10 stir::VectorWithOffset<stir::VectorWithOffset<std::shared_ptr<std::vector<std::pair<int, int>, std::allocator<std:
:pair<int, int> > > > > >::_destruct_and_deallocate (this=0x9eea18)                                                   
    at /home/gil2a4/src/STIR/src/include/stir/VectorWithOffset.inl:83                                                 
#11 stir::VectorWithOffset<stir::VectorWithOffset<std::shared_ptr<std::vector<std::pair<int, int>, std::allocator<std:
:pair<int, int> > > > > >::_destruct_and_deallocate (this=0x9eea18)                                                   
    at /home/gil2a4/src/STIR/src/include/stir/VectorWithOffset.inl:70                                                 
#12 stir::VectorWithOffset<stir::VectorWithOffset<std::shared_ptr<std::vector<std::pair<int, int>, std::allocator<std:
:pair<int, int> > > > > >::~VectorWithOffset (this=0x9eea18, __in_chrg=<optimised out>)                               
    at /home/gil2a4/src/STIR/src/include/stir/VectorWithOffset.inl:304                                                
#13 stir::ProjDataInfoCylindrical::~ProjDataInfoCylindrical (this=this@entry=0x9ee810, __in_chrg=<optimised out>)     
    at /home/gil2a4/src/STIR/src/include/stir/ProjDataInfoCylindrical.h:46                                            
#14 0x00000000006e6631 in stir::ProjDataInfoCylindricalNoArcCorr::~ProjDataInfoCylindricalNoArcCorr (this=0x9ee810,   
    __in_chrg=<optimised out>) at /home/gil2a4/src/STIR/src/include/stir/VectorWithOffset.inl:70                      
#15 stir::ProjDataInfoCylindricalNoArcCorr::~ProjDataInfoCylindricalNoArcCorr (this=0x9ee810,                         
    __in_chrg=<optimised out>) at /home/gil2a4/src/STIR/src/include/stir/ProjDataInfoCylindricalNoArcCorr.h:81        
#16 0x0000000000726f88 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x8fb970)               
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:148             
#17 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x8fb970)                                     
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:148             
#18 std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0x9f4e98, __in_chrg=<optimised out>)       
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:730             
#19 std::__shared_ptr<stir::ProjDataInfo, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr (this=0x9f4e90,                  
    __in_chrg=<optimised out>)                                                                                        
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:1169 
#20 std::shared_ptr<stir::ProjDataInfo>::~shared_ptr (this=0x9f4e90, __in_chrg=<optimised out>)             [135/8000]
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr.h:103                  
#21 stir::ProjDataInfoSubsetByView::~ProjDataInfoSubsetByView (this=0x9f4e00, __in_chrg=<optimised out>)              
    at /home/gil2a4/src/STIR/src/include/stir/ProjDataInfoSubsetByView.h:35                                           
#22 stir::ProjDataInfoSubsetByView::~ProjDataInfoSubsetByView (this=0x9f4e00, __in_chrg=<optimised out>)                  at /home/gil2a4/src/STIR/src/include/stir/ProjDataInfoSubsetByView.h:35                                           
--Type <RET> for more, q to quit, c to continue without paging--l                                                     
#23 0x000000000053b928 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x9d0250)               
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:148             
#24 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x9d0250)                                     
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:148             
#25 0x00000000005d9e35 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0x9da8c0,            
    __in_chrg=<optimised out>)                                                                                        
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:1169            
#26 std::__shared_ptr<stir::ProjDataInfo const, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr (this=0x9da8b8,            
    __in_chrg=<optimised out>)                                                                                        
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:1169            
#27 std::shared_ptr<stir::ProjDataInfo const>::~shared_ptr (this=0x9da8b8, __in_chrg=<optimised out>)                 
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr.h:103                  
#28 stir::BackProjectorByBin::~BackProjectorByBin (this=0x9da7f0, __in_chrg=<optimised out>)                          
    at /home/gil2a4/src/STIR/src/recon_buildblock/BackProjectorByBin.cxx:50                                           
#29 0x0000000000633305 in stir::RegisteredParsingObject<stir::BackProjectorByBinUsingProjMatrixByBin, stir::BackProjec
torByBin, stir::BackProjectorByBin>::~RegisteredParsingObject (this=0x9da7f0, __in_chrg=<optimised out>)              
    at /home/gil2a4/src/STIR/src/include/stir/RegisteredParsingObject.h:78                                            
#30 stir::BackProjectorByBinUsingProjMatrixByBin::~BackProjectorByBinUsingProjMatrixByBin (this=0x9da7f0,             
    __in_chrg=<optimised out>)                                                                                        
    at /home/gil2a4/src/STIR/src/include/stir/recon_buildblock/BackProjectorByBinUsingProjMatrixByBin.h:49            #31 stir::BackProjectorByBinUsingProjMatrixByBin::~BackProjectorByBinUsingProjMatrixByBin (this=0x9da7f0,             
    __in_chrg=<optimised out>)                                                                                        
    at /home/gil2a4/src/STIR/src/include/stir/recon_buildblock/BackProjectorByBinUsingProjMatrixByBin.h:49            
#32 0x0000000000636b00 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x9d8340)               
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:148             
#33 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x9d8340)                                     
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:148             
#34 std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0x9d2fd8, __in_chrg=<optimised out>)       
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:730             
#35 std::__shared_ptr<stir::BackProjectorByBin, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr (this=0x9d2fd0,            
    __in_chrg=<optimised out>)                                                                                        
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:1169            
#36 std::shared_ptr<stir::BackProjectorByBin>::~shared_ptr (this=0x9d2fd0, __in_chrg=<optimised out>)                 
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr.h:103                  
#37 stir::ProjectorByBinPair::~ProjectorByBinPair (this=0x9d2f40, __in_chrg=<optimised out>)                          
    at /home/gil2a4/src/STIR/src/include/stir/recon_buildblock/ProjectorByBinPair.h:52                                
#38 stir::RegisteredParsingObject<stir::ProjectorByBinPairUsingProjMatrixByBin, stir::ProjectorByBinPair, stir::Projec
torByBinPair>::~RegisteredParsingObject (this=0x9d2f40, __in_chrg=<optimised out>)                                    
    at /home/gil2a4/src/STIR/src/include/stir/RegisteredParsingObject.h:78                                            
#39 stir::ProjectorByBinPairUsingProjMatrixByBin::~ProjectorByBinPairUsingProjMatrixByBin (this=0x9d2f40,             
    __in_chrg=<optimised out>)                                                                                        
    at /home/gil2a4/src/STIR/src/include/stir/recon_buildblock/ProjectorByBinPairUsingProjMatrixByBin.h:34            
#40 0x000000000053b928 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x9d2f30)               
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:148             
#41 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x9d2f30)                                     
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:148             
#42 0x000000000051f18e in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0x7fffffffa088,
    __in_chrg=<optimised out>)                                                                                        
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:1169            
#43 std::__shared_ptr<stir::ProjectorByBinPairUsingProjMatrixByBin, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr (      
    this=0x7fffffffa080, __in_chrg=<optimised out>)                                                                   
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:1169            
#44 std::shared_ptr<stir::ProjectorByBinPairUsingProjMatrixByBin>::~shared_ptr (this=0x7fffffffa080,                  
--Type <RET> for more, q to quit, c to continue without paging--                                                      
    __in_chrg=<optimised out>)                                                                                        
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/bits/shared_ptr.h:103                  
#45 stir::TestProjDataInfoSubsets::test_forward_projection_is_consistent (this=0x7fffffffa320, input_image_sptr=...,  
    template_sino_sptr=..., fwd_projector_sptr=..., use_symmetries=<optimised out>, num_subsets=10)                   
    at /home/gil2a4/src/STIR/src/test/test_proj_data_info_subsets.cxx:262                                             
#46 0x000000000053aa0d in stir::TestProjDataInfoSubsets::run_tests (this=0x7fffffffa320)                              
    at /nix/store/fvf3qjqa5qpcjjkq37pb6ypnk1mzhf5h-gcc-9.3.0/include/c++/9.3.0/ext/atomicity.h:96                     
#47 0x0000000000530b4e in main (argc=<optimised out>, argv=<optimised out>)                                           
    at /home/gil2a4/src/STIR/src/test/test_proj_data_info_subsets.cxx:319 
```

ashgillman:
Sorry, pressed the wrong button

src/buildblock/ProjDataInfoSubsetByView.cxx
```diff
@@ -0,0 +1,202 @@
...
+
+ProjDataInfoSubsetByView::ProjDataInfoSubsetByView(const shared_ptr<const ProjDataInfo> full_proj_data_info_sptr,
+                                                   const std::vector<int>& views)
+  :
+  ProjDataInfo(full_proj_data_info_sptr->get_scanner_sptr(),
+               VectorWithOffset<int>(full_proj_data_info_sptr->get_min_segment_num(),
+                                     full_proj_data_info_sptr->get_max_segment_num()), // filled in below
+               views.size(),
+               full_proj_data_info_sptr->get_num_tangential_poss()),
+  org_proj_data_info_sptr(full_proj_data_info_sptr->clone()),
```
ashgillman, line 33: Is it possible this is the cause of our segfault? ProjDataInfoCylindricalNoArcCorr doesn't have a copy constructor (called by clone) implemented, nor does ProjDataInfo. But ProjDataInfo has shared_ptr members. Are these being shallow copied?

ashgillman: In which case we've uncovered a bug with `ProjDataInfo`? But also is it surprising to the user that if you creat a `ProjDataInfoSubsetByView` and then modify it, the `org_proj_data_info_sptr` changes, but the original `proj_data_info_sptr` doesn't.

src/buildblock/ProjData.cxx
```diff
@@ -225,6 +229,40 @@ read_from_file(const string& filename,
...
+  auto subset_proj_data_ptr = new ProjDataInMemory(exam_info_sptr, subset_proj_data_info_sptr);
+  auto& subset_proj_data = *subset_proj_data_ptr;
+
+  for (int segment_num=get_min_segment_num(); segment_num<=get_max_segment_num(); ++segment_num)
+    {
+      for (int subset_view_num=0; subset_view_num<views.size(); ++subset_view_num)
+        {
+          const Viewgram<float> viewgram = this->get_viewgram(views[subset_view_num], segment_num);
+          Viewgram<float> subset_viewgram = Viewgram<float>(viewgram, subset_proj_data_info_sptr, subset_view_num, segment_num);
+          if (subset_proj_data.set_viewgram(subset_viewgram) != Succeeded::yes)
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: @ashgillman just a matter of style: `a.b` looks neater than `a_ptr->b`, does it not? (BTW the former should have been used on line 257 too - `subset_proj_data_ptr` is only needed on line 263.)

src/buildblock/ProjDataInfoSubsetByView.cxx
KrisThielemans: @rijobro has introduced a lot of `shared_ptr<const A>` in arguments to prevent this trouble with these things. That means introducing more `clone()`, which is less efficient, but does prevent surprising things like modifications of one object affecting another one. I'm not 100% sure of course if we've caught all of these.

In any case, I didn't think that in this test code we are modifying such settings. And even if we did, that wouldn't cause a segfault as far as I can see, just surprising behaviour.

src/test/test_proj_data_info_subsets.cxx
evgueni-ovtchinnikov: This is what I did trying to localize the source of segfault.

Firstly, I replaced 247 with
```
  ProjDataInMemory full_forward_projection(*template_sino_sptr);
```
(I actually believe clever enough compiler does that anyway, but just in case).
Next I put `return` after line 269 - still segfalts, so segfault does not seem to be caused by 270 (or at least not only by 270, which causes that weird alleged 'different geometry' error we observe).
I then put
```
if (use_symmetries)
```
before 127 so that it does not affect full projection - no segfault!

So, it looks like the trouble is caused by `proj_pair_sptr->set_up`.

KrisThielemans:
Valgrind reports the following
```
	Subset 0, creating subset forward projector
	Setting up default projector pair, ProjectorByBinPairUsingProjMatrixByBin
==5440== Invalid write of size 4
==5440==    at 0x8C8DA8: stir::ProjDataInfoSubsetByView::ProjDataInfoSubsetByView(std::shared_ptr<stir::ProjDataInfo const>, std::vector<int, std::allocator<int> > const&) (ProjDataInfoSubsetByView.cxx:40)
==5440==    by 0x8C8FD2: stir::ProjDataInfoSubsetByView::clone() const (ProjDataInfoSubsetByView.cxx:58)
==5440==    by 0x63226F: stir::ProjDataInfo::create_shared_clone() const (ProjDataInfo.inl:31)
==5440==    by 0x7272AB: stir::ProjectorByBinPair::set_up(std::shared_ptr<stir::ProjDataInfo const> const&, std::shared_ptr<stir::DiscretisedDensity<3, float> const> const&) (ProjectorByBinPair.cxx:45)
==5440==    by 0x727C57: stir::ProjectorByBinPairUsingProjMatrixByBin::set_up(std::shared_ptr<stir::ProjDataInfo const> const&, std::shared_ptr<stir::DiscretisedDensity<3, float> const> const&) (ProjectorByBinPairUsingProjMatrixByBin.cxx:88)
==5440==    by 0x5EBC6D: stir::TestProjDataInfoSubsets::construct_projector_pair(std::shared_ptr<stir::ProjDataInfo const> const&, std::shared_ptr<stir::VoxelsOnCartesianGrid<float> const> const&, bool) (test_proj_data_info_subsets.cxx:127)
==5440==    by 0x5EC66E: stir::TestProjDataInfoSubsets::test_forward_projection_is_consistent(std::shared_ptr<stir::VoxelsOnCartesianGrid<float> const> const&, std::shared_ptr<stir::ProjData const> const&, std::shared_ptr<stir::ForwardProjectorByBin> const&, bool, int) (test_proj_data_info_subsets.cxx:266)
==5440==    by 0x5EBF54: stir::TestProjDataInfoSubsets::run_tests() (test_proj_data_info_subsets.cxx:154)
==5440==    by 0x5ECAF9: main (test_proj_data_info_subsets.cxx:323)
==5440==  Address 0x90130b0 is 16 bytes before a block of size 26 alloc'd
==5440==    at 0x4C3017F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==5440==    by 0x7347737: gdcm::PrivateDict::LoadDefault() (in /data/devel/install/lib/libitkgdcmDICT-4.13.so.1)
==5440==    by 0x73450A1: gdcm::Dicts::LoadDefaults() (in /data/devel/install/lib/libitkgdcmDICT-4.13.so.1)
==5440==    by 0x73440AF: _GLOBAL__sub_I_gdcmGlobal.cxx (in /data/devel/install/lib/libitkgdcmDICT-4.13.so.1)
==5440==    by 0x4010732: call_init (dl-init.c:72)
==5440==    by 0x4010732: _dl_init (dl-init.c:119)
==5440==    by 0x40010C9: ??? (in /lib/x86_64-linux-gnu/ld-2.27.so)
```
indicating a crash in
```c++
  for (int subset_view_num=0; subset_view_num<static_cast<int>(views.size()); ++subset_view_num)
    {
      org_view_to_view_num[view_to_org_view_num[subset_view_num]] = subset_view_num;
    }
```

src/buildblock/ProjDataInfoSubsetByView.cxx
```diff
@@ -0,0 +1,202 @@
...
+      this->set_max_axial_pos_num(full_proj_data_info_sptr->get_max_axial_pos_num(segment_num), segment_num);
+    }
+  this->set_bed_position_horizontal(full_proj_data_info_sptr->get_bed_position_horizontal());
+  this->set_bed_position_vertical(full_proj_data_info_sptr->get_bed_position_vertical());
+}
+
+
+ProjDataInfoSubsetByView* ProjDataInfoSubsetByView::clone() const
+{
+  return new ProjDataInfoSubsetByView(this->org_proj_data_info_sptr, this->org_view_to_view_num);
```
KrisThielemans, line 58: (Probably my) mistake. This had to be `view_to_org_view_num`. Caused segfaults

KrisThielemans:
segfault resolved. geometry run-time error is still there

evgueni-ovtchinnikov:
`ProjDataInfo::operator>=` that is used for comparing `ProjDataInfo` objects starts with comparing `typeid`'s:
```
  if (typeid(*this) != typeid(proj_data_info))
    return false;
```
In our case `typeid`'s are different:
```
N4stir32ProjDataInfoCylindricalNoArcCorrE
N4stir24ProjDataInfoSubsetByViewE
```
which is why we get that error message.

KrisThielemans:
ok, but then the error makes sense, the projector has to be set-up with the correct projdatainfo. It wasn't but I thought I had fixed that (possibly introducing a bug...)

KrisThielemans:
the error was due to a bug in `operator>=`. I've fixed this now (although could be improved). Interestingly, I think with this fix, it might be a step along the way of being able to project a subset with the matrix for the full data, although that probably still needs other changes.

Nevertheless, we now have another crash. Valgrind says
```
	Subset 0, forward projecting subset

WARNING: forward_project. Imaging modality unknown for either the image or the projection data or both.
Going ahead anyway.

INFO: Processing view 0 of segment -3
==24726== Thread 4:
==24726== Invalid read of size 8
==24726==    at 0x635A42: stir::VectorWithOffset<int>::check_state() const (VectorWithOffset.inl:63)
==24726==    by 0x6335C0: stir::VectorWithOffset<int>::operator[](int) const (VectorWithOffset.inl:124)
==24726==    by 0x632B3B: stir::ProjDataInfoCylindrical::get_average_ring_difference(int) const (ProjDataInfoCylindrical.inl:141)
==24726==    by 0x6F05F4: stir::DataSymmetriesForBins_PET_CartesianGrid::find_transform_z(int, int) const (DataSymmetriesForBins_PET_CartesianGrid.inl:73)
==24726==    by 0x6F0AF1: stir::DataSymmetriesForBins_PET_CartesianGrid::find_sym_op_general_bin(int, int, int, int) const (DataSymmetriesForBins_PET_CartesianGrid.inl:169)
==24726==    by 0x6F1684: stir::DataSymmetriesForBins_PET_CartesianGrid::find_symmetry_operation_from_basic_bin(stir::Bin&) const (DataSymmetriesForBins_PET_CartesianGrid.inl:368)
==24726==    by 0x70F130: stir::ProjMatrixByBin::get_proj_matrix_elems_for_one_bin(stir::ProjMatrixElemsForOneBin&, stir::Bin const&) const (ProjMatrixByBin.inl:57)
==24726==    by 0x721DA2: stir::ForwardProjectorByBinUsingProjMatrixByBin::actual_forward_project(stir::RelatedViewgrams<float>&, stir::DiscretisedDensity<3, float> const&, int, int, int, int) (ForwardProjectorByBinUsingProjMatrixByBin.cxx:140)
==24726==    by 0x6AB095: stir::ForwardProjectorByBin::actual_forward_project(stir::RelatedViewgrams<float>&, int, int, int, int) (ForwardProjectorByBin.cxx:296)
==24726==    by 0x6AAFA3: stir::ForwardProjectorByBin::forward_project(stir::RelatedViewgrams<float>&, int, int, int, int) (ForwardProjectorByBin.cxx:272)
==24726==    by 0x6AAD0A: stir::ForwardProjectorByBin::forward_project(stir::RelatedViewgrams<float>&) (ForwardProjectorByBin.cxx:224)
==24726==    by 0x6AB469: stir::ForwardProjectorByBin::forward_project(stir::ProjData&, int, int, bool) [clone ._omp_fn.0] (ForwardProjectorByBin.cxx:209)
==24726==  Address 0x90af8d0 is 32 bytes before a block of size 32 in arena "client"
==24726== 
```

evgueni-ovtchinnikov:
I know very little about STIR implementation of forward projection, but the output
```
INFO: Processing view 0 of segment -3

INFO: Processing view 1 of segment -3
Segmentation fault (core dumped)
```
looks suspicious to me - so far as I can see (and remember), full forward projection always starts with segment 0, then proceeds to segments 1, 2 etc.

Though maybe this is simply because symmetries are not used for subset projection? How can one switch off the use of symmetries for full projection for comparison?

KrisThielemans:
the segfault is caused by what valgrind says. I haven't figured out yet which array it's flagging up, but given the stack trace it seems one of the ring-diff arrays: https://github.com/UCL/STIR/blob/e8d250feb8dfbb00fcb25ab63d08282496ac6751/src/include/stir/ProjDataInfoCylindrical.inl#L141

The relevant loop is
https://github.com/UCL/STIR/blob/e8d250feb8dfbb00fcb25ab63d08282496ac6751/src/recon_buildblock/ForwardProjectorByBin.cxx#L193-L203

Therefore forward projection will effectively start with the most negative segment if the seg-symmetry is switched off, which @ashgillman  did (I guess to first check it without that complication)
```
    auto subset_proj_pair_sptr = construct_projector_pair(
      subset_forward_projection_sptr->get_proj_data_info_sptr(), input_image_sptr,
      /*use_symmetries =*/false);
```
so this looks fine to me.

Having said that, we switch symmetries off, but it still is in `DataSymmetries*` functions, so that's a bit weird as well. Seems to need some more step-wise debugging.

KrisThielemans:
it's a `static_cast` in `DataSymmetriesForBins_PET_CartesianGrid::find_transform_z`, which gets called even if all symmetries are off

KrisThielemans:
I think I've fixed it. 2 problems. test works in my system. valgrind is ok

ashgillman:
Some more TODO from Devel meeting:
- Test w/ unbalanced subsets?
- test backprojection
- test z shift symmetry
- add Interfile keywords
- test I/O
- test reduce_segment_range

ashgillman:
I've found a failing edge-case. Running a subset with no views (get_subset with empty list) initially works, but leads to a segfault at `ProjDataInfoSubsetByView::get_org_bin` called when setting up Projectors.

ashgillman:
I'm not sure what the desired behaviour would be for an empty subset. Fail?

It might be nice to be able to have empty subsets, but its not clear how functions like `get_tantheta` work when there are no bins.

ashgillman:
Could we conceive of an algorithm where every subset has its views randomly bootstrapped such that there could be "double-ups" of some views? Should we test that?

src/buildblock/ProjDataInfoSubsetByView.cxx
```diff
@@ -67,6 +67,11 @@ Bin ProjDataInfoSubsetByView::get_org_bin(const Bin& bin) const
   return org_bin;
 }
 
+std::vector<int> ProjDataInfoSubsetByView::get_org_views() const
```
KrisThielemans, line 70:
```suggestion
std::vector<int> ProjDataInfoSubsetByView::get_org_view_nums() const
```
and some doxygen

KrisThielemans:
> 

I suggest to fail. it creates a lot of headaches afterwards, and I cannot see the use of it

KrisThielemans:
> Could we conceive of an algorithm where every subset has its views randomly bootstrapped such that there could be "double-ups" of some views? Should we test that?

At present I think there is one place where we assume no replication (I believe when checking the number of views in the equality operator or so), but probably nowhere else.

Not so sure if this is worth coping with. I guess most stochastic algorithms would randomly sample so replication could occur. 
Random sampling ensuring that all of the views are used and no duplication probably makes most sense. @epapoutsellis, @ClaireDelplancke, @gschramm?

evgueni-ovtchinnikov:
To quick test `get_subset()`, I tried it on the subset of all views, i.e. with `range(nv)` as the argument, where `nv` is the number of views (I used `PET\acquisition_data.py` in SIRF branch `subset_object` modified accordingly).

This is what I got for `my_forward_projection.hs` (the number of sinograms shown reduced to 15 for convenience):
Figure_1
Figure_2
Clearly, something went wrong.

ClaireDelplancke:
There are two different situations, @ashgillman @KrisThielemans I'm not sure which one we're referring to. 

- The list of views for one given subset is redundant, say [1, 5, 1]. I think this is what @ashgillman is alluding to. This situation doesn't arise in SPDHG nor in mini-batch gradient-based algorithms. 

- Subsets are overlapping, for example the subset chosen at the first iteration is [1,2] and the subset chosen at the second iteration is [1,5]. This situation isn't ruled out by the theory, but would it be an issue with respect to the current implementation? In that case, one can certainly use only samplings inducing non-overlapping subsets (this is known as "serial sampling" in SPDHG). I don't think anyone has observed an advantage of either type of sampling on the other at the moment, as @gschramm said, but it could occur in the future that we find situations where overlapping subsets work better than non-overlapping ones.

robbietuk:
>  I don't think anyone has observed an advantage of either type of sampling on the other at the moment, as @gschramm said, but it could occur in the future that we find situations where overlapping subsets work better than non-overlapping ones.

We did something like this for Fully3D 2021 ("Investigation of Subset Methodologies Applied to Penalised Iterative PET Reconstruction") with the Randomised Batches, `m` projection angles sampled at random (without replacement) at each update. This allows for overlapping subsets but the subsets are not constant throughout the reconstruction.

ashgillman:
@ClaireDelplancke  rightly highlights two possible definitions of "overlapping". I think @robbietuk , you are referring to the second, as you say "without replacement".

The current solution works fine for number 2: each "get_subset" is completely independent and so can easily have multiple subsets referring to the same view.

The current solution doesn't work for number 1, sampling with replacement. It seems like there is little appetite for this so lets not support it for now?

ashgillman:
@evgueni-ovtchinnikov Are you able to push that test to https://github.com/SyneRBI/SIRF/pull/1028? It looks like they may be from different segments

KrisThielemans:
> The current solution doesn't work for number 1, sampling views with replacement within the same subset. It seems like there is little appetite for this so lets not support it for now? 

agreed. I think I documented it, but could add a test at construction time I guess. (not sure which, maybe create `std::set<int>(views.begin(), views.end())` and check size remains the same)

KrisThielemans:
Appveyor fails as (old) Visual C++ doesn't like using `or`

ashgillman:
Whoops, Python habits. I was hoping to knock off a few more checklist items last night, should have the backproj done by Thu

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -373,28 +452,52 @@ test_forward_projection_for_one_subset(
...
+
+    auto subset_back_projector_sptr = construct_projector_pair(
+      input_sino_sptr->get_proj_data_info_sptr(), template_image_sptr, /*use_symmetries=*/false)
+      ->get_back_projector_sptr() ;
+
+    VoxelsOnCartesianGrid<float> subset_back_projection = *template_image_sptr->clone();
+    subset_back_projector_sptr->back_project(subset_back_projection, *input_sino_sptr);
+    back_projection_sum += subset_back_projection;
+  }
+  back_projection_sum /= num_subsets;  // Why do I have to do this...?!
```
ashgillman, line 494: @KrisThielemans  could you please check this line? I expected when I wrote the test that I would sum the subsets, but I instead averaged them to run the test. Hopefully I just misunderstood the backproj units (activity vs counts)

KrisThielemans: I can't check the code at the moment, but backprojection sums. a problem with using `template_image_sptr->clone()`? Better to guarantee it starts with zero (i.e. use `get_uniform_copy` I guess)

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -0,0 +1,531 @@
...
+      _test_image_sptr, _input_sino_sptr);
+    cerr << "repeat with an 'unusual' number of subsets, 13" << endl;
+    test_forward_projection_is_consistent(
+      _test_image_sptr, _input_sino_sptr, /*use_z_symmetries=*/false, /*use_z_symmetries=*/false, /*num_subsets=*/13);
+    cerr << "repeat with z shift symmetries" << endl;
+    test_forward_projection_is_consistent(
+      _test_image_sptr, _input_sino_sptr, /*use_z_symmetries=*/true, /*use_z_symmetries=*/false);
+    cerr << "repeat with all symmetries" << endl;
+    test_forward_projection_is_consistent(
+      _test_image_sptr, _input_sino_sptr, /*use_z_symmetries=*/true, /*use_z_symmetries=*/true);
```
ashgillman, line 222: Not sure if we expected this one to run? It passes locally. Of course, some of the symmetries are automatically disabled later.

src/test/test_proj_data_info_subsets.cxx
ashgillman: Disregard, this is my mistake. I backproject the original image every time, not the subset!!

ashgillman: (but I need to go to bed now, I've pushed but its producing a segfault. I'll look tomorrow afternoon)

ashgillman:
Hmm, I'm a bit stumped with this segfault @KrisThielemans 

I think this is the relevant valgrind output:

```
pure virtual method called
terminate called without an active exception
==898810== 
==898810== Process terminating with default action of signal 6 (SIGABRT)
==898810==    at 0x4C11FFA: raise (in /nix/store/0c7c96gikmzv87i7lv3vq5s1cmfjd6zf-glibc-2.31-74/lib/libc-2.31.so)
==898810==    by 0x4BFC527: abort (in /nix/store/0c7c96gikmzv87i7lv3vq5s1cmfjd6zf-glibc-2.31-74/lib/libc-2.31.so)
==898810==    by 0x493E692: ??? (in /nix/store/c10296m7xgm3ksibcklb2xf48jr635x3-gcc-9.3.0-lib/lib/libstdc++.so.6.0.28)
==898810==    by 0x4949F05: ??? (in /nix/store/c10296m7xgm3ksibcklb2xf48jr635x3-gcc-9.3.0-lib/lib/libstdc++.so.6.0.28)
==898810==    by 0x4949F70: std::terminate() (in /nix/store/c10296m7xgm3ksibcklb2xf48jr635x3-gcc-9.3.0-lib/lib/libstdc++.so.6.0.28)
==898810==    by 0x494AC0E: __cxa_pure_virtual (in /nix/store/c10296m7xgm3ksibcklb2xf48jr635x3-gcc-9.3.0-lib/lib/libstdc++.so.6.0.28)
==898810==    by 0x634705: create_shared_clone (ProjDataInfo.inl:31)
==898810==    by 0x634705: stir::ProjectorByBinPair::set_up(std::shared_ptr<stir::ProjDataInfo const> const&, std::shared_ptr<stir::DiscretisedDensity<3, float> const> const&) (ProjectorByBinPair.cxx:45)
==898810==    by 0x634D88: stir::ProjectorByBinPairUsingProjMatrixByBin::set_up(std::shared_ptr<stir::ProjDataInfo const> const&, std::shared_ptr<stir::DiscretisedDensity<3, float> const> const&) (ProjectorByBinPairUsingProjMatrixByBin.cx
x:88)
==898810==    by 0x5385CA: stir::TestProjDataInfoSubsets::construct_projector_pair(std::shared_ptr<stir::ProjDataInfo const> const&, std::shared_ptr<stir::VoxelsOnCartesianGrid<float> const> const&, bool, bool) (test_proj_data_info_subset
s.cxx:169)
==898810==    by 0x53A60A: stir::TestProjDataInfoSubsets::test_back_projection_is_consistent(std::shared_ptr<stir::ProjData const> const&, std::shared_ptr<stir::VoxelsOnCartesianGrid<float> const> const&, int) (test_proj_data_info_subsets
.cxx:493)
==898810==    by 0x53D0FF: stir::TestProjDataInfoSubsets::run_tests() (test_proj_data_info_subsets.cxx:228)
==898810==    by 0x531068: main (test_proj_data_info_subsets.cxx:526)

```

KrisThielemans:
indicates a problem with `create_shared_clone`. Possibly `ProjDataInfo::create_shared_clone` should have been pure virtual but isn't, such that the compiler would have told us that we need it for the subset class.

ashgillman:
I'll try as you suggest. I'm still a bit confused as to why this would be the case.

`ProjDataInfo::create_shared_clone()` calls `this->clone()`. Since `ProjDataInfo::clone()` is virtual, why doesn't `ProjDataInfoSubsetByView::clone()` get called?

KrisThielemans:
ah no, forget my comments. `create_shared_clone` isn't `virtual` at all. Nevertheless, it seems to be calling a pure virtual from the stack trace, which might mean it's a problem with the call to `clone`. We do override that one, but maybe there's something wrong there? Try to add an `override` directive to `ProjDataInfoSubsetByView::clone`  to let the compiler know and warn if we make a mistake there.

KrisThielemans:
sorry, my comments above were entirely incorrect. I guess that happens when trying to read `valgrind` output in a multi-threaded case. I just picked on the wrong one.

The problem was that the backprojector was called without `start_accumulating_in_new_image` (and there was no check on this). Also, the paradigm `Type& var = *something->clone()` leads to memory leaks.

I've cleaned up the C++ a little bit as well. It's still somewhat inconsistent, but fine.

Tests are working for me now.

KrisThielemans:
merge with `master` creates some conflicts. This will need a bit of care.

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -370,84 +374,57 @@ generate_full_forward_projection(
...
 
 
-VoxelsOnCartesianGrid<float> TestProjDataInfoSubsets::
+shared_ptr<VoxelsOnCartesianGrid<float>> TestProjDataInfoSubsets::
 generate_full_back_projection(
     const shared_ptr<const ProjData>& input_sino_sptr,
     const shared_ptr<const VoxelsOnCartesianGrid<float> >& template_image_sptr,
     bool use_z_symmetries, bool use_other_symmetries)
 {
-  bool input_is_nonzero = false;
```
ashgillman: Are you sure we don't want this? I included it originally to double check I'd generated test data correctly. But if the tests are run on empty input data they will always trivially pass.

src/buildblock/ProjData.cxx
```diff
@@ -235,32 +235,22 @@ ProjData::get_subset(const std::vector<int>& views) const
 {
   auto subset_proj_data_info_sptr =
     std::make_shared<ProjDataInfoSubsetByView>(proj_data_info_sptr, views);
-  auto subset_proj_data_ptr = new ProjDataInMemory(exam_info_sptr, subset_proj_data_info_sptr);
-  auto& subset_proj_data = *subset_proj_data_ptr;
+  unique_ptr<ProjDataInMemory> subset_proj_data_uptr(new ProjDataInMemory(exam_info_sptr, subset_proj_data_info_sptr));
 
+  //TODOTOF loop here
```
ashgillman, line 240: Is this TODO needed before merge?

ashgillman:
Thanks very much for looking over that issue. I had been banging my head against the wall and doubt I'd have solved that.

ashgillman:
Okay, I think that was all the tests we wanted implemented for this PR. All remaining items are for subset I/O, for which I will create a new PR, and to SIRF tests, which are covered in https://github.com/SyneRBI/SIRF/pull/1028

src/buildblock/ProjData.cxx
KrisThielemans: no. I put it in such that we don't forget when merging to the TOF branch

src/test/test_proj_data_info_subsets.cxx
KrisThielemans: I know, but I guess we know now that you don't have a bug in the test... I want to reduce the number of loops over data as much as possible as that will change with TOF

KrisThielemans:
thanks! can you add something to `documentation/release_5.0.htm` if not done yet? Then we should merge

src/buildblock/ProjData.cxx
KrisThielemans: this was all done.

ashgillman:
Don't merge yet, I just pushed a failing test!

src/buildblock/ProjDataInfoSubsetByView.cxx
```diff
@@ -0,0 +1,265 @@
...
+  // check all of smaller_proj_data_info org_views are in this subset
+  for (int view_num_idx = 0; view_num_idx < smaller_proj_data_info.get_num_views(); ++view_num_idx)
+  {
+    int org_view_num = smaller_proj_data_info.view_to_org_view_num[view_num_idx];
+    if (std::find(this->view_to_org_view_num.begin(), this->view_to_org_view_num.end(), org_view_num)
+          == this->view_to_org_view_num.end())
+    { return false; }
+  }
+
+  return base_type::operator>=(proj);
```
ashgillman, line 241: So the issue is that the `base_type::operator>=(proj)` returns False when we want True because it works by reducing the size of `this` until it should equal that of `proj`, and then compares for equality.
We need to do something like this:
```c++
  // base_type::operator>= works by reducing the larger ProjDataInfo (this) until they're equal.
  // We'll need to, therefore, compare with a subset of this that only includes the views of
  // smaller_proj_data_info
  std::vector<int> smaller_proj_data_view_nums;
  for (auto org_view_num : smaller_proj_data_info.view_to_org_view_num)
    smaller_proj_data_view_nums.push_back(this->org_view_to_view_num[org_view_num]);
  ProjDataInfoSubsetByView this_projdata_info_with_other_views(
    this->create_shared_clone(), smaller_proj_data_view_nums);

  // I need to call something like:
  base_type::operator>=(this_projdata_info_with_other_views, smaller_proj_data_info);
```
How can I call `ProjDataInfo::operator>=` here? In Python I would do like `ProjDataInfo.operator>=(this_projdata_info_with_other_views, smaller_proj_data_info)` because `self`/`this` is passed around as an argument. Can I do this?

ashgillman: An alternative is to have a non-overridden function in `ProjDataInfo`, like `ProjDataInfo::is_superset`, than can be accessed, and have `ProjDataInfo::operator>=` call it.

KrisThielemans: You can call `base_type::operator>=` by the way, no problem. But I can't see how that'd help.  In the end, we have to compare if the views are there, and if the `org` is also larger.. Can you be more explicit on which test fails and why? to me the `if (typeid(*this) != typeid(proj)) ... else`  statement seems wrong.

ashgillman:
Updated release info. Except for remaining very minor and ignoreable comments (blindly equals, Ellipsoid include) this is ready to go from my perspective.

src/buildblock/ProjDataInfoSubsetByView.cxx
KrisThielemans: > You wrote this, so I tried to guess intent here.

😄 . Ok, I understand my own logic again!

> I think `ProjDataInfoSubsetByView::operator>=` should check that anything subset-specific is equal, and then use `ProjDataInfo::operator>=` to check the rest.

well, we have above all this `if (!((*this->org_proj_data_info_sptr) >= (*smaller_proj_data_info.org_proj_data_info_sptr)))...`, so it's doing the check in the other order, but that seems fine to me.

So why do we not simply return `true` after the check on `views` ? I'm missing something I guess.

In any case, I don't like the introduction of `is_superset` too much. We now have a virtual function `operator >=` calling another virtual `is_superset`. We override `operator>=`, but not `is_superset`, which from a normal developer perspective is very weird. Hopefully we don't need `is_superset` therefore...

KrisThielemans:
When all done, could you run `clang-format` on all the new files (not the existing ones)?

src/buildblock/ProjDataInfoSubsetByView.cxx
```diff
@@ -0,0 +1,265 @@
...
+}
+
+bool ProjDataInfoSubsetByView::blindly_equals(const root_type * const p) const
+{
+  // can do static_cast as operator== already checked type
+  auto that = static_cast<const self_type *>(p);
+  if (this->org_view_to_view_num != that->org_view_to_view_num)
+    return false;
+
+  return ((*this->org_proj_data_info_sptr) == (*that->org_proj_data_info_sptr));
```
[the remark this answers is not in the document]
ashgillman: No: blindly_equals is a private helper for `==` so that won't have any effect.

ashgillman:
Done and visually ran my eyes over it. I noted that clang-format w/ `Standard: Cpp11` will remove formatting like:
`NestedTemplate<A<B> >`
to:
`NestedTemplate<A<B>>`.

I think we no longer support older C++ standards anyway? Otherwise we'd need to change that to `Standard: Cpp03`.

https://github.com/Microsoft/vscode-cpptools/issues/1102

src/buildblock/ProjDataInfoSubsetByView.cxx
ashgillman: Yeah I agree it is quite ugly.

But I see your argument now as to whether we need to do any check. I wanted to confirm segments, tangential_pos, axial_pos, etc. were `<=` for the subset. But since we already check these for the `org_proj_data_info` and every time we update for the subset we update for the original, that should be covered.

There is one technicality to bypass this check: if  the `org_proj_data` was modified without the subset info. But I don't think this can happen as outside the class you can only access a readonly pointer to `org_proj_data`. We just need to be careful within the class: I just noticed that `set_num_tangential_poss` doesn't update both subset and original.

ashgillman: Removed `is_subset` and fixed `set_num_tangential_poss`

KrisThielemans:
> I think we no longer support older C++ standards anyway? 

Indeed, STIR `master` is now on C++-11.

src/test/test_proj_data_info_subsets.cxx
```diff
@@ -0,0 +1,603 @@
...
+// #include "stir/recon_buildblock/BackProjectorByBinUsingProjMatrixByBin.h"
+#include "stir/recon_buildblock/ProjMatrixByBinUsingRayTracing.h"
+#include "stir/ProjDataInMemory.h"
+#include "stir/Viewgram.h"
+#include "stir/Scanner.h"
+#include "stir/ProjDataInfoCylindricalNoArcCorr.h"
+#include "stir/ProjDataInfoSubsetByView.h"
+#include "stir/Shape/Ellipsoid.h"
+#include "stir/Shape/DiscretisedShape3D.h"
+// ^ should we have to include this? Should be included from Ellipsoid? Bug?
```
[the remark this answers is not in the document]
KrisThielemans: I just deleted the repetition of the Shape files and kept only `Ellipsoid` and that is working fine.

KrisThielemans (approved):

KrisThielemans:
All done! Thanks a lot @ashgillman !

Could you create a new issue with the remaining items in the TODO list?

PR #971, opened 2021-11-29
[CMake] use ROOT::Tree target if it exists

### LINKED ISSUE #954: ROOT now requires C++17
not sure from which version, but it now uses `std::string_view`. We'll have to check this in our `CMakeLists.txt`

KrisThielemans:
Recent ROOT versions come with a ROOTConfig.cmake which defines targets. Using these targets means appropriate compiler options will be used by STIR, including C++ version.

This prevents problems with std::string_view when the compiler used for STIR does not match the compiler used for ROOT.

Fixes #954


robbietuk:
We currently have no root tests and so I am testing on my machine. How is the `CERN_ROOT_DEBUG` set?

KrisThielemans:
> How is the `CERN_ROOT_DEBUG` set?

Command-line CMake:
```
cd build-dir
cmake -DCERN_ROOT_DEBUG=ON .
```
CMake GUI:
1. create new BOOL variable `CERN_ROOT_DEBUG`
2. set it to `ON`
3. configure and generate

robbietuk:
Got it. Apparently the `ccmake` command tool does not have the option to add variables so I needed to use `cmake -DCERN_ROOT_DEBUG=ON .` and when configuring, got the message:
```
-- Found ROOT::Tree CMake target, so will use that as opposed to CERN_ROOT_LIBRARIES
```

But I am getting a build error, which is possibly [an unrelated issue due to ROOT on MacOS 12.0.1](https://root-forum.cern.ch/t/root-6-24-06-can-not-work-on-macos-monterey/47409).

KrisThielemans:
ok. thanks for checking! I'll merge therefore.

Issue #972, opened 2021-11-29
split parallelproj GPU projections into chunks


gschramm:
When dealing with big projection data sets (e.g. full TOF sinograms), having the ability to split the 
GPU projections from parallelproj into chunks to avoid out of memory errors on the GPU.

PR is coming

PR #973, opened 2021-11-29
implement chunk-wise GPU projections using the parallelproj projectors

gschramm:
Chunk-wise GPU projections are implemented by adding the possibility of splitting the memory transfer and kernel execution into more than 1 chunks.

PR for this [issue](https://github.com/UCL/STIR/issues/972#issue-1066238358)


KrisThielemans:
I guess the SWIG mods need to be within `ifdef` to avoid problems when Parallelproj isn't found. I'd have to dig which variable but am short of time at the moment. sorry.

gschramm:
I will have a look myself (nut not a swig expert yet ;)

gschramm:
which identifier for ifdef gets set when STIR was built with parallelproj? I see the "parallelproj_built_with_cuda" identifier in the code, but I guess we also want the SWIG stuff when parllelproj was built without CUDA

KrisThielemans:
Probably best to copy the mechanism via CMake
https://github.com/UCL/STIR/blob/b768f7afd154a89ce50d7ab16f35ef1dda7ddc78/src/swig/CMakeLists.txt#L33-L35, like this
```CMake
if (parallelproj_FOUND)
   SET(CMAKE_SWIG_FLAGS ${CMAKE_SWIG_FLAGS} -DHAVE_parallelproj)
endif()
```
what do you think?

gschramm:
> Probably best to copy the mechanism via CMake
> 
> https://github.com/UCL/STIR/blob/b768f7afd154a89ce50d7ab16f35ef1dda7ddc78/src/swig/CMakeLists.txt#L33-L35
> , like this
> 
> ```cmake
> if (parallelproj_FOUND)
>    SET(CMAKE_SWIG_FLAGS ${CMAKE_SWIG_FLAGS} -DHAVE_parallelproj)
> endif()
> ```
> 
> what do you think?

Modified swig CMakeLists.txt and stir.i accordingly, see
https://github.com/gschramm/STIR/blob/940619f56b40dc28f773bf4d2b2822cf11dca00c/src/swig/CMakeLists.txt#L36
https://github.com/gschramm/STIR/blob/940619f56b40dc28f773bf4d2b2822cf11dca00c/src/swig/stir.i#L100
https://github.com/gschramm/STIR/blob/940619f56b40dc28f773bf4d2b2822cf11dca00c/src/swig/stir.i#L1777

But now I get weird compiler errors a la:
```
src/swig/CMakeFiles/_stir.dir/stirPYTHON_wrap.cxx:144672:200: error: expected primary-expression before \u2018)\u2019 token
144672 |     return (void *) new stir::shared_ptr< stir::RegisteredParsingObject< stir::BackProjectorByBinParallelproj,stir::BackProjectorByBin > >(*(stir::shared_ptr< stir::BackProjectorByBinParallelproj > *)x);
```

Any idea why that happens? (does not if I comment out all the ifdef HAVE_parallelprojs)

KrisThielemans:
Turns out that in the preamble of stir.i (i.e. L100), we are using the C++ macros, while in the rest we are using SWIG-defined macros (set with `CMAKE_SWIG_FLAGS`). So, for the former, we need to use whatever we have in `config,h`, i.e.
https://github.com/UCL/STIR/blob/b768f7afd154a89ce50d7ab16f35ef1dda7ddc78/src/cmake/STIRConfig.h.in#L68

Obviously it is weird that some of these are called `HAVE_*` and others `STIR_with_*`. I'm sure @rijobro had a good brainwave when doing that for NiftyPET, and I just copied it.

KrisThielemans:
@gschramm would you be able to merge `master` in here again to check? Otherwise good to go I guess.

Ideally we'd also include a build on GHA where it installs parallelproj and finds it, but that might take too much of your time.

gschramm:
@KrisThielemans  master is merged back in. What does GHA mean?

KrisThielemans:
sorry. GitHub Actions. yours is currently still running at https://github.com/UCL/STIR/actions/runs/1927809483 (see the "Actions" tab).

this is governed by https://github.com/UCL/STIR/blob/master/.github/workflows/build-test.yml

KrisThielemans:
I've tried to add it to GHA. let's see.

However, it looks like the `find_package(parallelproj 0.8)` still finds 0.7. Not sure why. Any clues?

PR #974, opened 2021-12-03
Add verbosity to SqrtHessianRowSum class

robbietuk:
Similar to the [Reconstruction class](https://github.com/UCL/STIR/blob/master/src/recon_buildblock/Reconstruction.cxx). No need for python accessable methods because verbosity is global?


KrisThielemans:
> No need for python accessable methods because verbosity is global?

correct

robbietuk:
Perfect! I should have also mentioned. I tested this locally 😄

PR #989, opened 2022-01-28
Relax the condition on z position in FORE

melkatib1:
A condition (i.e. z_in_mm/half_distance_between_rings - z > 0.0001) is set on z position to prevent rebinning to non existing z-positions (sinograms), however, this value of 0.0001 may not always be satisfied (due to numerical precision). For example, I got an error message when trying FORE with a PET having more than 1000 axial positions, although SSRB goes well; For verification I done manual calculation with the returned z value, and that gives a difference of 0.0035 (z_in_mm/half_distance_between_rings - z = 0.0035) wich is indeed > 0.0001 (so the condition is not satisfied), although the z position in mm is correct and corresponds to one of the PET's actual axial positions.
After re-building and re-installing STIR on virtual machine with this value of 0.0001 relaxed to 0.005F the FORE is running well for all z positions.

Also I deleted the // TODO comment.


KrisThielemans (approved):
thanks for the testing!

Issue #938, opened 2021-10-06
ROOT view/detector offset


KrisThielemans:
The rotation for images corresponding to ROOT files is currently confusing and largely undocumented. There are 2 keywords in the `.hroot` file:
https://github.com/UCL/STIR/blob/4221fc32585b71ae1d68febba0fef82f1f819a1b/examples/samples/root_header.hroot#L13
https://github.com/UCL/STIR/blob/4221fc32585b71ae1d68febba0fef82f1f819a1b/examples/samples/root_header.hroot#L45

# offset (num of detectors)
The `offset (num of detectors)` is used to change the detector numbering
https://github.com/UCL/STIR/blob/4221fc32585b71ae1d68febba0fef82f1f819a1b/src/IO/InputStreamFromROOTFileForCylindricalPET.cxx#L119-L120

The `offset (num detectors)` keyword is defined [here](https://github.com/UCL/STIR/blob/4221fc32585b71ae1d68febba0fef82f1f819a1b/src/IO/InputStreamFromROOTFile.cxx#L89), but there is no default set for it in [InputStreamFromROOTFile::set_defaults](https://github.com/UCL/STIR/blob/4221fc32585b71ae1d68febba0fef82f1f819a1b/src/IO/InputStreamFromROOTFile.cxx#L59) which is arguably a bug.

# additional rotations in the code
Unfortunately, going through the code, there are additional offsets being applied
https://github.com/UCL/STIR/blob/4221fc32585b71ae1d68febba0fef82f1f819a1b/src/IO/InputStreamFromROOTFileForCylindricalPET.cxx#L115-L116
where https://github.com/UCL/STIR/blob/4221fc32585b71ae1d68febba0fef82f1f819a1b/src/IO/InputStreamFromROOTFileForCylindricalPET.cxx#L53-L55
(note that this uses integer division, which will be surprising) and

https://github.com/UCL/STIR/blob/4221fc32585b71ae1d68febba0fef82f1f819a1b/src/listmode_buildblock/CListRecordROOT.cxx#L69-L70
where
https://github.com/UCL/STIR/blob/4221fc32585b71ae1d68febba0fef82f1f819a1b/src/listmode_buildblock/CListRecordROOT.cxx#L29
This is a 90 degrees rotation, but it also uses integer division, so this would create some **small unexpected rotation if the total number of crystals is not a multiple of 4.**

# view offset
[...]

PR #991, opened 2022-02-17
GATE/ROOT STIR LOR consistency test

robbietuk:
Adds a consistency test between GATE and STIR data that was initially described in https://github.com/UCL/STIR/issues/938#issuecomment-1039817967. This is based upon work conducted previously by @eliseemond for a TOF test. 

Simulate emission of (12) point sources in GATE with known `original` position. Read the output ROOT files using STIR finding the closest voxel to the `original` position and measure distance. If distance is greater than a given tolerance then mark event as `failed_event`. Ensure that the majority of events are within the tolerance.

- `tolerance` is currently set to `1.5 * norm(grid_spacing).
- 50% of events must not "fail" for test to pass.

A possible alternative, test the `mean(abs(distance)) < tolerance`? 

This PR has a lot of overlap with a similar TOF test in development https://github.com/NikEfth/STIR/pull/23. I suggest this PR is merged into UCL/STIR and then UCL/STIR into the TOF branch.

REMANING TODO
-----
- [x] Fix READMEs
- [ ] Cleanup directory. Add sub directories to contain data etc.
- [ ] Remove unnecessary GATE files 
- [ ] Upload pre-generated ROOT files to Zenodo and add download methodology. Do this during the test? Check if ROOT data exists before test?
- [ ] Investigate reasoning for large `distance` results. Some values are up to `10* tolerance`
- [ ] No need to generate the STIR_images (12 * 16MB = 168MB). They are not used in the Gate simulations and they could be generated in the test. Can modify `generate_image.cxx` and create a class methods to generate and save the image. The utilility can call both but this test can simply generate and keep in memory... thoughts?


recon_test_pack/ROOT_STIR_consistency/run_pretest_script.sh
```diff
@@ -0,0 +1,17 @@
+#! /bin/bash
+
+for I in {1..12}
+do
+	generate_image generate_image${I}.par
+	cd Gate_macros
+	# Create main GATE macro files from template
+	sed -e s/SOURCENAME/test${I}/ main_D690_template.mac > main_D690_test${I}.mac
```
robbietuk, line 8: `sed` issues may occur on MacOS. I personally have to use `gsed` for this command to work. This is documented [elsewhere](https://unix.stackexchange.com/a/131940) but may lead to issues for us in this case.

robbietuk:
Current status. All tests pass with the Gate generated data and take about 1.5 seconds each. The tests are stacked as follows.
```
      Start  1: test_consistency_root_1
 1/73 Test  #1: test_consistency_root_1 ..............................................   Passed    1.61 sec
      Start  2: test_consistency_root_2
 2/73 Test  #2: test_consistency_root_2 ..............................................   Passed    1.36 sec
      Start  3: test_consistency_root_3
 3/73 Test  #3: test_consistency_root_3 ..............................................   Passed    1.36 sec
      Start  4: test_consistency_root_4
 4/73 Test  #4: test_consistency_root_4 ..............................................   Passed    1.36 sec
      Start  5: test_consistency_root_5
 5/73 Test  #5: test_consistency_root_5 ..............................................   Passed    1.29 sec
      Start  6: test_consistency_root_6
 6/73 Test  #6: test_consistency_root_6 ..............................................   Passed    1.30 sec
      Start  7: test_consistency_root_7
 7/73 Test  #7: test_consistency_root_7 ..............................................   Passed    1.34 sec
      Start  8: test_consistency_root_8
 8/73 Test  #8: test_consistency_root_8 ..............................................   Passed    1.35 sec
      Start  9: test_consistency_root_9
 9/73 Test  #9: test_consistency_root_9 ..............................................   Passed    1.33 sec
      Start 10: test_consistency_root_10
10/73 Test #10: test_consistency_root_10 .............................................   Passed    1.34 sec
      Start 11: test_consistency_root_11
11/73 Test #11: test_consistency_root_11 .............................................   Passed    1.28 sec
      Start 12: test_consistency_root_12
12/73 Test #12: test_consistency_root_12 .............................................   Passed    1.29 sec
```

recon_test_pack/ROOT_STIR_consistency/run_pretest_script.sh
KrisThielemans: it's true, but the pretest script is only run to create the files, so presumably only needs running once in a while. Let's live with it (maybe just add a TODO comment in the file)

robbietuk: Agreed. The hroot and root files can be downloaded. `run_pretest_script.sh` does not need to be run for each test.

KrisThielemans (commented):
looks good. I'm a bit confused on the relation between test_view_offset_root.cxx and test_consistency_root.cxx.

Also, a question is where we put the GATE macros. etc. It feels strange to me to have them in `recon_test_pack`, as that is currently standalone (scripts and data, used on "installed STIR"). I note that files for `recon_test_pack/run_root_GATE.sh` sit in `examples/GATE_test_macro` (even that's incomplete). That is of course a bit confusing as well.  Depends a bit how independent the test is from these files. (We could have them in `examples` but include them in the Zenodo upload for instance)

KrisThielemans:
>  I'm a bit confused on the relation between test_view_offset_root.cxx and test_consistency_root.cxx.

oh, you want to merge this first, and then sort out the TOF stuff later. agreed.

markus-jehl:
No, I haven't looked at generate_image before.

robbietuk:
Two step process:

1.  Split `GenerateImage.compute()` into two to add `save_image()` method.
2. Move `GenerateImage` to its own header and source and tie to Shape. I may well have made some mistake or not finished the documentation. 

New test pass with this modification 😄

src/include/stir/Shape/GenerateImage.h
```diff
@@ -0,0 +1,98 @@
...
+  \file
+  \ingroup Shape
+
+  \brief Declaration of class stir::GenerateImage
+
+  \author Kris Thielemans
+  \author Sanida Mustafovic
+  \author Robert Twyman
+
+  *See generate_image utility documentation for parameter file example and usage documentation.
```
robbietuk, line 21: Documentation of the parameter files here? We don't particulaly want duplicate documentation?

src/Shape_buildblock/GenerateImage.cxx
```diff
@@ -0,0 +1,307 @@
...
+                                                     output_voxel_size_y,
+                                                     output_voxel_size_x));
+  this->out_density_ptr.reset(current_image.clone());
+#endif
+  std::vector<float >::const_iterator value_iter = values.begin();
+  for (std::vector<shared_ptr<Shape3D> >::const_iterator iter = shape_ptrs.begin();
+          iter != shape_ptrs.end();
+  ++iter, ++value_iter)
+  {
+    std::cerr << "Next shape\n"; //(**iter).parameter_info() << '\n';
```
KrisThielemans, line 283: probably comment this out. it isn't very useful. we'd have to call `info` instead (at verbosity 2) if you feel you want to keep it.

src/Shape_buildblock/GenerateImage.cxx
```diff
@@ -0,0 +1,307 @@
...
+GenerateImage::
+save_image()
+{
+  output_file_format_sptr->write_to_file(output_filename, *out_density_ptr);
+  return Succeeded::yes;
+}
+
+shared_ptr<DiscretisedDensity<3, float>>
+GenerateImage::
+get_out_density_ptr()
```
KrisThielemans, line 302: let's call this `get_output_sptr`

src/include/stir/Shape/GenerateImage.h
```diff
@@ -0,0 +1,98 @@
+//
+//
+/*
+    Copyright (C) 2022, University College London
```
KrisThielemans, line 4: license and copyright

src/include/stir/Shape/GenerateImage.h
KrisThielemans: As this class can be exposed to Python, it indeed makes most sense to move doc here, and refer to it in the utility

src/include/stir/Shape/GenerateImage.h
```diff
@@ -0,0 +1,98 @@
...
+#define __stir_Shape_GenerateImage_h__
+
+
+
+START_NAMESPACE_STIR
+
+class GenerateImage : public KeyParser
+{
+public:
+    //! Constructor requires a parameter filename passed. See \generate_image utility for details.
```
KrisThielemans, line 47: adjust doxygen

src/include/stir/Shape/GenerateImage.h
```diff
@@ -0,0 +1,98 @@
...
+    GenerateImage(const char * const par_filename);
```
KrisThielemans, line 48:
```suggestion
    explicit GenerateImage(const char * const par_filename);
```

src/recon_test/test_view_offset_root.cxx
```diff
@@ -0,0 +1,242 @@
...
+    cerr << argv[0]
+         << ": Error opening root header file: " << argv[1] << "\nExiting.\n";
+
+    return EXIT_FAILURE;
+  }
+  ifstream in2(argv[2]);
+  if (!in2)
+  {
+    cerr << argv[0]
+         << ": Error generate image parameter filename: " << argv[2] << "\nExiting.\n";
```
KrisThielemans, line 233:
```suggestion
         << ": Error opening generate_image parameter filename: " << argv[2] << "\nExiting.\n";
```

src/recon_test/test_view_offset_root.cxx
```diff
@@ -0,0 +1,242 @@
...
+  USING_NAMESPACE_STIR
+
+  if (argc != 3)
+  {
+    cerr << "Usage : hroot_filename generate_image_par_filename\n\n";
+    return EXIT_FAILURE;
+  }
+
+
+  ifstream in(argv[1]);
```
KrisThielemans, line 221: put all these checks in its own block `{ ... }` such that the files get closed before passing it on. Doesn't matter, but cleaner

src/utilities/generate_image.cxx
```diff
@@ -1,6 +1,6 @@
 /*
   Copyright (C) 2003-2011, Hammersmith Imanet Ltd
-  Copyright (C) 2018-2020, University College London
+  Copyright (C) 2018-2022, University College London
```
KrisThielemans, line 3: this copyright can just be UCL 2022, as there will be essentially nothing left in the file.

KrisThielemans (changes requested):
some comments on generate_image. Add a line about it in `release_5.0.htm` (in the developer's section I guess). Can then also add a line about the test there somewhere

robbietuk:
>  Also, a question is where we put the GATE macros. etc. It feels strange to me to have them in recon_test_pack, as that is currently standalone (scripts and data, used on "installed STIR"). I note that files for recon_test_pack/run_root_GATE.sh sit in examples/GATE_test_macro (even that's incomplete). That is of course a bit confusing as well. Depends a bit how independent the test is from these files. (We could have them in examples but include them in the Zenodo upload for instance)

I think this is part of the final step now. Where do these macro files belong? Secondly, how do we run the test if the data is not present?

KrisThielemans:
> Where do these macro files belong?

I wouldn't mind putting them in `examples/GATE_point_source_test_macros`. As long we have a `README.md` explaining things.

> Secondly, how do we run the test if the data is not present?

I suggest that the download is part of the test. Two options:

- a script to do this. Could re-use https://github.com/SyneRBI/SIRF-Exercises/blob/master/scripts/download_data.sh#L73-L179.
- use [CbMake script mode](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#id12) , where all the above is a one liner https://github.com/SyneRBI/SIRF/blob/6713bab5838926e07d264f757bed5f8136b76db6/src/xGadgetron/cGadgetron/tests/CMakeLists.txt#L26

The 2nd option seems far simpler, even though it relies on CMake being present at test-time (which could be a minor problem with the conda recipe, but we can fix that).

robbietuk (commented):
Suggested fixes to simplify, speed up, and reduce data size for the test. Some future (TOF) proofing too.

recon_test_pack/ROOT_STIR_consistency/Gate_macros/geometry_D690.mac
```diff
@@ -0,0 +1,117 @@
...
+
+#	C R Y S T A L
+
+/gate/module/daughters/name                    crystal
+/gate/module/daughters/insert                  box
+/gate/crystal/placement/setTranslation        0.0 0.0 0.0 cm
+/gate/crystal/geometry/setXLength             25 mm
+/gate/crystal/geometry/setYLength             4.44 mm
+/gate/crystal/geometry/setZLength             6.54 mm           # Filled the gaps between the crystals for convenience
+/gate/crystal/setMaterial                     LYSO
```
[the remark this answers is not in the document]
robbietuk: Makes significant difference to the simulations with shorter cystals (XLength=5mm) with LYSO=`155` and BGO=`2404` events recorded.

recon_test_pack/ROOT_STIR_consistency/Gate_macros/geometry_D690.mac
```diff
@@ -0,0 +1,117 @@
+#	CYLINDRICAL
```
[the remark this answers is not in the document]
robbietuk: Gate geometry has been updated to better reflect the blocks and buckets. We could continue to modify this.

robbietuk:
With the latest changes to the GATE geometry and crystal material, the ROOT files are ~400kB (each so ~5MB total) with 2000-2400 coincidence events in each. These tests are all passing on my machine. Is this still too large to include on GitHub and do we still need Zenodo?

As alluded to in https://github.com/UCL/STIR/issues/938#issuecomment-1079977343, this GATE scanner is still 90 degrees rotated so before an upload of the root files we may want to address that issue.

Also not very happy with the verbosity of the test, as seen in the previous comment: https://github.com/UCL/STIR/pull/991#issuecomment-1043283911

documentation/release_5.0.htm
```diff
@@ -382,13 +382,17 @@ <h3>New functionality</h3>
...
+<li>Added <code>test_view_offset_root</code> to ensure reliability between GATE and STIR voxel positions by
+    comparing center of mass of ROOT data LORs with the original STIR position.</li>
 </ul>
 
 
 <h3>Other code changes</h3>
 <ul>
     <li>store data in <code>ProjDataInMemory</code> in the same order as what is used by <code>copy_to</code> and <code>fill_from</code>. This enabled using straight-forward copy. (This change should not affect anyone, except if you relied on a specific order in the buffer before.)
     </li>
+    <li>Modification to <code>generate_image</code> by moving the <code>GenerateImage</code> class to it's own
```
KrisThielemans, line 394:
```suggestion
    <li>Modification to <code>generate_image</code> by moving the <code>GenerateImage</code> class to its own
```

examples/ROOT_files/ROOT_STIR_consistency/run_pretest_script.sh
```diff
@@ -0,0 +1,54 @@
...
+do
+	echo ""
+	echo "Generating data for test${I}..."
+	
+	# Generate images and move to the pretest output directory
+	echo "Generating STIR image..."
+	generate_image SourceFiles/generate_image${I}.par
+	mv stir_image${I}.ahv ${PRE_TEST_OUTPUT_DIR}
+	mv stir_image${I}.hv ${PRE_TEST_OUTPUT_DIR}
+	mv stir_image${I}.v ${PRE_TEST_OUTPUT_DIR}
```
KrisThielemans, line 35:
```suggestion
	mv stir_image${I}.*v ${PRE_TEST_OUTPUT_DIR}
```
simpler and more future proof

people/eliseemond/.gitignore
```diff
@@ -0,0 +1,39 @@
+# Compiled source #
```
KrisThielemans, line 1: this file needs to be removed

KrisThielemans (changes requested):
tiny clean-up comments

KrisThielemans:
> do we still need Zenodo?

On the one hand, Zenodo/github often creates time-outs, but on the other, test data size can only grow as you say. I therefore still think we should do that.

I do think we have to remove the `quarter` stuff first (and therefore adjust the GATE scanner).

Do you have some plots to show?

robbietuk:
> I therefore still think we should do that.

Okay

> I do think we have to remove the quarter stuff first (and therefore adjust the GATE scanner).

Agreed. This PR or a seperate with local testing for now?

______
> Do you have some plots to show?

No plots needed I think. The output of the python script is below. Changing the GATE scanner geometry, crystal size, and crystal material appears to have alligned the tests fairly well and simplified a lot.

Less than 5% error for each point source.
```
Information regarding pass fail rate of the view offset test
 ID || The Number of Events  ||  The Number of Failed Events  ||  Percentage Error
-----------------------------------------------------------------------------------
1  ||           2404         ||              72                ||      3.0      
2  ||           2407         ||              72                ||      2.99      
3  ||           2409         ||              76                ||      3.15      
4  ||           2384         ||              73                ||      3.06      
5  ||           2074         ||              9                ||      0.434      
6  ||           2089         ||              8                ||      0.383      
7  ||           2408         ||              62                ||      2.57      
8  ||           2309         ||              60                ||      2.6      
9  ||           2406         ||              81                ||      3.37      
10  ||           2451         ||              74                ||      3.02      
11  ||           2011         ||              9                ||      0.448      
```


No real offset in any axis. The `TOTAL BIAS` test is to ensure there isn't a general shift in the axis.
```
Mean offset in each axis for each source position
 ID ||   x   ||   y   ||   z
------------------------
1  || -2.1  || 0.61  || 0.36  
2  || 2.0  || -0.52  || 0.26  
3  || -0.54  || -2.0  || 0.33  
4  || 0.54  || 2.0  || 0.3  
5  || -0.72  || -0.3  || 0.42  
6  || -0.24  || 0.81  || 0.45  
7  || -2.0  || 0.54  || -0.41  
8  || 2.0  || -0.61  || -0.45  
9  || -0.64  || -2.1  || -0.43  
10  || 0.58  || 2.0  || -0.42  
11  || -0.66  || -0.36  || -0.65  
TOTAL BIAS
   ||   x    ||   y    ||   z
   || -0.16   || 0.011   || -0.021
Done
```

examples/ROOT_files/ROOT_STIR_consistency/DebugScripts/debug_view_offset_consistency.py
```diff
@@ -0,0 +1,188 @@
...
+        num_events = point_sources_data[key].get_num_events()
+        num_failed_events = point_sources_data[key].get_num_failed_events()
+        percentage = num_failed_events / num_events * 100
+        if percentage > allowed_fraction_of_failed_events * 100:
+            warning_msg = "HIGH VALUE WARNING!"
+        else:
+            warning_msg = ""
+        print(f"{key}  ||           {num_events}         "
+              f"||              {num_failed_events}                "
+              f"||      {round_sig(percentage, 3)}      {warning_msg}")
```
KrisThielemans, line 132: if you happen to know how to put consistent width in printing in Python, would be good to add it here

examples/ROOT_files/ROOT_STIR_consistency/DebugScripts/debug_view_offset_consistency.py
```diff
@@ -0,0 +1,188 @@
...
+        mean_err_x = np.mean(point_sources_data[key].err_x)
+        mean_err_y = np.mean(point_sources_data[key].err_y)
+        mean_err_z = np.mean(point_sources_data[key].err_z)
+        total_bias_x += mean_err_x
+        total_bias_y += mean_err_y
+        total_bias_z += mean_err_z
+        print(f"{key}  "
+              f"|| {round_sig(mean_err_x)}  "
+              f"|| {round_sig(mean_err_y)}  "
+              f"|| {round_sig(mean_err_z)}  ")
```
KrisThielemans, line 157: if you happen to know how to put consistent width in printing, precision and sign in Python, would be good to add it here

KrisThielemans:
Less than 5% is great, but if some of these are 10cm off, it's still weird. I guess a histogram plot would reassure me (truncated in vertical axis as it'll be dominated by the peak).

KrisThielemans (changes requested):

src/recon_test/test_view_offset_root.cxx
```diff
@@ -0,0 +1,274 @@
...
+              failed_events += 1;
+            }
+            tested_events += 1;
+          }
+        }
+      }
+    }
+
+  cerr << "\nNUMBER OF FAILED EVENTS = " << failed_events << "\t NUMBER OF TESTED EVENTS = " << tested_events << std::endl;
+  check_if_less(failed_events, 0.5 * tested_events,
```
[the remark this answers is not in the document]
robbietuk: We can tighten now. The largest fail rate is 3.37%. Shall we change this to 5% to give us some tolerance margin?

robbietuk:
> we could merge it now, but then the test will have to be disabled in CMakeLists.txt as it will fail due to no data presumably.

I think we merge now and this test can be used on seperate branch to try to fix #938

PR #992, opened 2022-02-19
CListEvent changes for BlocksOnCylindrical

VietAnhDao:
1. Adjusted CListEvent for root files.
 
- Stored ProjDataInfo instead of ProjDataInfoCylindrical.

- Downcast ProjDataInfo to correct geometry when needed.

### DIFF
--- a/src/buildblock/DetectorCoordinateMap.cxx
+++ b/src/buildblock/DetectorCoordinateMap.cxx
@@ -114,7 +114,7 @@ void DetectorCoordinateMap::set_detector_map( const DetectorCoordinateMap::det_p
               ", axial " +std::to_string(num_axial_coords) + ", radial " + std::to_string(num_radial_coords) +
               "\nOveral  size: " + std::to_string(coord_map.size()));
 
-    std::sort(coords_to_be_sorted.begin(), coords_to_be_sorted.end());
+//    std::sort(coords_to_be_sorted.begin(), coords_to_be_sorted.end());
     stir::DetectionPosition<> detpos(0,0,0);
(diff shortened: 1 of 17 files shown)


KrisThielemans:
It seems this breaks the listmode recon, `run_test_listmode_recon.sh` crashes out with a segfault. https://github.com/UCL/STIR/runs/5259249760?check_suite_focus=true#step:9:321

KrisThielemans:
@VietAnhDao I believe I fixed the problem. Unfortunately I cannot write to your branch. Normally with a PR, maintainers can write to the branch (I believe that's the default in GitHub), but possibly it doesn't work here because you did this work on your `master`. Never a good idea...

I've created #994 where I've squashed some of your commits, and added mine on top. Have a look. Could you check if that works for you?

By the way, as I squashed your commits, more than likely you'll have trouble with your `master` later. I'd recommended making it identical to `UCL/master` again once this is merged. See e.g. https://stackoverflow.com/a/1628334/3467846

VietAnhDao:
@KrisThielemans works for me 👍. PR close.

KrisThielemans:
sorry, did you mean that  #994 works for you?

VietAnhDao:
sorry about the ambiguous wording of "works" but yes just tested the build and the conversion definitely works for 'Cylindrical' and seems to work for 'BlocksOnCylindrical'. The main take away is:

- The sinogram of a point source has the characteristic kink instead of a smooth sine wave(pic attached).
- The output interfile says 'Cylindrical'... very 

Btw should we move this conversation to #994 ?

test_root_point_source

KrisThielemans:
Yes please. Comment on the new PR and complete the sentence in the 2nd bullet please

PR #994, opened 2022-02-20
CListEvent adaption for BlocksOnCylindrical

KrisThielemans:
replaces #992 due to write permissions

### DIFF
--- a/src/buildblock/Scanner.cxx
+++ b/src/buildblock/Scanner.cxx
@@ -809,8 +809,6 @@ void Scanner::set_up()
     }
   else
     {
-      if (crystal_map_file_name != "")
-        error("Scanner: use scanner_geometry=Generic when specifying a crystal map");
       if (scanner_geometry == "BlocksOnCylindrical")
         this->detector_map_sptr.reset(new GeometryBlocksOnCylindrical(*this));
       else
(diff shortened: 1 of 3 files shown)


VietAnhDao:
build and tested by converting root point source to projection data. Main points:

- Conversion using 'Cylindrical' and 'BlocksOnCylindrical' made no different (pic attached)
- Output projection data, with 'BlocksOnCylindrical' template, has 'Cylindrical' for geometry in header.

This probably means conversion from root to projection data used 'Cylindrical' geometry and doesn't switch based on projdata template (attached log).

test_root_point_source_blk

test_root_point_source_cyl

[blk.log](https://github.com/UCL/STIR/files/8104612/blk.log)

[cyl.log](https://github.com/UCL/STIR/files/8104616/cyl.log)

KrisThielemans:
Your `blk.log` write quite a few warnings at the start, and finds as scanner
```
Scanner geometry (BlocksOnCylindrical/Cylindrical/Generic)  := Cylindrical
```
In fact, `CListModeDataROOT.cxx` doesn't know yet about blocks/generic. We'll need some changes there as well. Sadly, the `.hroot` parsing and `.hs` parsing is currently independent.

I'm confused though what you're doing different in the 2 cases. Different template?

VietAnhDao:
Exactly same template but different geometry and nothing else changed.

KrisThielemans:
These changes are now obsolete after #1222

PR #996, opened 2022-02-21
Remove overriding redefinitions in SymmertryOperations_PET_CartesianGrid

### LINKED ISSUE #995: `SymmetryOperations_PET_CartisianGrid` have significant code duplication
[`SymmetryOperations_PET_CartisianGrid.cxx` ](https://github.com/UCL/STIR/blob/master/src/recon_buildblock/SymmetryOperations_PET_CartesianGrid.cxx) is entirely duplication of the parent classes methods [`SymmetryOperation.cxx`](https://github.com/UCL/STIR/blob/master/src/recon_buildblock/SymmetryOperation.cxx)

Except the `SymmetryOperation_PET_CartesianGrid_swap_zq::transform_proj_matrix_elems_for_one_bin()` for some reason, which misses some preprocessing.
https://github.com/UCL/STIR/blob/06963f9292d7e9cbfe5b3ffa8d49eb7a533ec6eb/src/recon_buildblock/SymmetryOperations_PET_CartesianGrid.cxx#L359-L372
______
We can probably just remove the redefinitions of the methods and rely on those defined in `SymmetryOperation`? I can do this for the TOF symmetry changes or in STIR master?

robbietuk:
Fixes #995. Why? Improve the debugging experience...


robbietuk:
Okay, I read the documentation. I guess it explains why this is done:
https://github.com/UCL/STIR/blob/06963f9292d7e9cbfe5b3ffa8d49eb7a533ec6eb/src/include/stir/recon_buildblock/SymmetryOperation.h#L37-L61
Is this still relevant? This was written 22 years ago...

KrisThielemans:
another bit of documentation
https://github.com/UCL/STIR/blob/06963f9292d7e9cbfe5b3ffa8d49eb7a533ec6eb/src/recon_buildblock/SymmetryOperations_PET_CartesianGrid.cxx#L10-L15

> Is this still relevant? This was written 22 years ago... Is this a legacy work around we should be scared of removing? It still uses the correct inline methods for `transform_bin_coordinates` and `transform_image_coordinates` but I am unaware of any efficency impact.

well, my C++ knowledge is 22 years old, so... caveats apply.

It seems clear to me that if `transform_image_coordinates` is virtual, the compiler cannot inline it in the calls to `transform_proj_matrix_elems_for_one_bin`, which means ~200x calling a virtual function vs ~200x doing some (almost trivial) inline stuff.

I guess this made a difference 20 years ago, not sure if it still does. I know that the caching+symmetry stuff is now inefficient, as in many cases it's faster to just switch the caching off and recompute the whole LOR probabilities. (This wasn't the case 20 years ago.) Therefore, slowing down the symmetry operations seems like a bad strategy.

Therefore, although I like code simplfication, you should do some timings in relatively realistic circumstances (and I guess with recent gcc as well as clang compilers) to convince me before I'll accept this. Or just keep it of course.

robbietuk:
> Therefore, although I like code simplfication, you should do some timings in relatively realistic circumstances (and I guess with recent gcc as well as clang compilers) to convince me before I'll accept this. Or just keep it of course.

Maybe it is something that can be look at in future but the benifit of this PR seems insignificant w.r.t the testing. I think we abandon this, maybe leave the issue up?

KrisThielemans:
indeed, too much effort for hardly any benefit.

PR #998, opened 2022-02-22
update of ML_norm.cxx and ML_norm.h to take into account BlocksOnCyli…

emikhay:
…ndrical scanner geometry

### DIFF
--- a/src/buildblock/ML_norm.cxx
+++ b/src/buildblock/ML_norm.cxx
@@ -129,27 +129,54 @@ void make_det_pair_data(DetPairData& det_pair_data,
 			const int segment_num,
 			const int ax_pos_num)
 {
-  const ProjDataInfoCylindricalNoArcCorr& proj_data_info =
-    dynamic_cast<const ProjDataInfoCylindricalNoArcCorr&>(proj_data_info_general_type);
+  if(proj_data_info_general_type.get_scanner_ptr()->get_scanner_geometry()== "BlocksOnCylindrical")
+    {
+      const ProjDataInfoBlocksOnCylindricalNoArcCorr& proj_data_info =
+        dynamic_cast<const ProjDataInfoBlocksOnCylindricalNoArcCorr&>(proj_data_info_general_type);
 
-  const int num_detectors = 
-    proj_data_info.get_scanner_ptr()->get_num_detectors_per_ring();
-  const int fan_size = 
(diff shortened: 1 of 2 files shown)


emikhay:
@KrisThielemans @danieldeidda  @VietAnhDao

src/buildblock/ML_norm.cxx
```diff
@@ -129,68 +129,135 @@ void make_det_pair_data(DetPairData& det_pair_data,
...
-    det_pair_data.get_num_detectors();
-  const ProjDataInfoCylindricalNoArcCorr& proj_data_info =
-    dynamic_cast<const ProjDataInfoCylindricalNoArcCorr&>(*proj_data.get_proj_data_info_sptr());
-
-  shared_ptr<Sinogram<float> > 
-    pos_sino_ptr(new Sinogram<float>(proj_data.get_sinogram(ax_pos_num,segment_num)));
-  shared_ptr<Sinogram<float> > neg_sino_ptr;
-  if (segment_num == 0)
-    neg_sino_ptr = pos_sino_ptr;
+  if(proj_data.get_proj_data_info_sptr()->get_scanner_ptr()->get_scanner_geometry()== "BlocksOnCylindrical")
```
KrisThielemans, line 187: so here we do need the cast, as we use `get_det_num_pair_for_view_tangential_pos_num`. Still, it's very ugly to have to duplicate the whole code. It's of course a consequence of our "faulty" hierarchy. What about creating a small function
```c++
void get_det_num_pair_for_view_tangential_pos_num(const ProjDataInfo proj_data_info&, ...)
{
  if (auto ptr = dynamic_cast<const ProjDataInfoCylindricalNoArcCorr const *>(&proj_data_info))
    ptr->get_det_num_pair_for_view_tangential_pos_num(...);
  else if (auto ptr = dynamic_cast<const ProjDataInfoGenericNoArcCorr const *>(&proj_data_info))
    ptr->get_det_num_pair_for_view_tangential_pos_num(...);
}
```
and just call that. No need for an `if` in the main code then. I think we use this trick elsewhere.

KrisThielemans (changes requested):
Unfortunately, this PR introduces quite a lot of duplication. I think it should be possible to get rid of most of that by using some extra functions and/or templates. We did something similar in #994 (although still too much duplication there).

KrisThielemans:
superseded by #1085

PR #1001, opened 2022-02-24
Symmetry fixes

### LINKED ISSUE #999: `SymmetryOperation_PET_CartesianGrid` are not correctly reverting `basic_bin` back to the original `bin`
To summarise, 

1. The symmetry methodology takes a `bin`, finds the basic bin with [`find_basic_bin()`](https://github.com/UCL/STIR/blob/06963f9292d7e9cbfe5b3ffa8d49eb7a533ec6eb/src/include/stir/recon_buildblock/DataSymmetriesForBins_PET_CartesianGrid.inl#L474-L666) and the corresponding [symmetry operation](https://github.com/UCL/STIR/blob/06963f9292d7e9cbfe5b3ffa8d49eb7a533ec6eb/src/include/stir/recon_buildblock/DataSymmetriesForBins_PET_CartesianGrid.inl#L144-L435) to revert `basic_bin` back to `bin`.
 
2. Generally `basic_bin` is used to check against the cache or the `probabilities`/lor is computed, see [calculate_proj_matrix_elems_for_one_bin()](https://github.com/UCL/STIR/blob/06963f9292d7e9cbfe5b3ffa8d49eb7a533ec6eb/src/include/stir/recon_buildblock/ProjMatrixByBin.inl#L41-L104)

3. The symmetry operation (found in 1.) is used to compute the `probabilities` elements (values on the cartesian grid) that correspond to `bin`s LOR (_this bit works_) and convert the`basic_bin` (stored in the `probabilities` object) back to `bin` (_this bit is broken_). 

___

I added an additional test into the [`test_DataSymmetriesForBins_PET_CartesianGrid`](https://github.com/UCL/STIR/blob/06963f9292d7e9cbfe5b3ffa8d49eb7a533ec6eb/src/recon_test/test_DataSymmetriesForBins_PET_CartesianGrid.cxx#L125-L127) code by replacing the `if` statement 
```
if (!check(elems_no_sym == elems_with_sym, "comparing lors")) 
```
with
```
      if (!check(elems_no_sym == elems_with_sym, "comparing lors") ||
          !check(elems_with_sym.get_bin() == elems_no_sym.get_bin(), "Comparing Bin"))
```
An example set of errors (with some further `std::cerr` modifications) are
```
Error. Comparing Bin
Current bin: 	segment = -2, 	axial pos 13, 	view = 0, 	tangential_pos_num = 3
Symm bin: 	segment = 2, 	axial pos 0, 	view = 0, 	tangential_pos_num = 3

Error. Comparing Bin
Current bin: 	segment = -2, 	axial pos 13, 	view = 0, 	tangential_pos_num = 6
Symm bin: 	segment = 2, 	axial pos 0, 	view = 0, 	tangential_pos_num = 6

Error. Comparing Bin
Current bin: 	segment = -2, 	axial pos 0, 	view = 1, 	tangential_pos_num = 0
Symm bin: 	segment = 2, 	axial pos 0, 	view = 1, 	tangential_pos_num = 0

```

As far as I can tell, there are only issues with the flipping of `segments` and `axial_positions` offsets. 

___
This issue was observed in https://github.com/NikEfth/STIR/pull/25 but this confirms that it exists in the non-TOF STIR/master.

robbietuk:
An attempt to resolve #999 and others.

1. Improve the symmetry tests but checking if probabilities bin is equal to the original.

2. I had identified a issue in bin transform with `SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy` with `axial_pos` when `shift_z = false`
```
Current bin: 	segment = -5, 	axial pos 1, 	view = 0, 	tangential_pos_num = -3
Symm bin: 	segment = -5, 	axial pos 2, 	view = 0, 	tangential_pos_num = -3
```
This has been fixed with `const int axial_pos_shift = do_symmetry_shift_z ? axial_pos_num : 0;` 

3. Added symmetry configuration tests. We need to see that everything is okay.


KrisThielemans:
looks good. In #999 you also had `segment_num` stuff. I don't see you change that here. Was that fixed by #1000?

robbietuk:
> Was that fixed by https://github.com/UCL/STIR/pull/1000?
Maybe? I guess but I am not 100% sure how. Still working on a zero tilt scanner for view symmetry tests.

KrisThielemans:
ok. well, that `swap_zq` symmetry certainly swaps segments, which it didn't do.

src/recon_test/test_DataSymmetriesForBins_PET_CartesianGrid.cxx
```diff
@@ -124,14 +124,24 @@ run_tests_2_proj_matrices_1_bin(const ProjMatrixByBin& proj_matrix_no_symm,
...
+           << ", \tview = " << elems_with_sym.get_bin().view_num()
+           << ", \ttangential_pos_num = " << elems_with_sym.get_bin().tangential_pos_num() << "\n\n";
+
+        if (elems_no_sym != elems_with_sym)
+        {
+        std::cerr << "No Symmetries Iterator || Symmetries Iterator " << std::endl;
+	    ProjMatrixElemsForOneBin::const_iterator no_sym_iter= elems_no_sym.begin();
 	    ProjMatrixElemsForOneBin::const_iterator with_sym_iter = elems_with_sym.begin();
 	    while (no_sym_iter!= elems_no_sym.end() || with_sym_iter!=elems_with_sym.end())
 	      {
```
robbietuk, line 147: This check, and while loop, can give a error due to the first element of the `elems_with_sym` not being in the `elems_no_sym` list, but is close (this is noticed when I am trying a scanner with 0 tilt). An example error is when `{2, 5, -8}` exists in `elems_with_sym`(right) but is never matched in the `elems_no_sym`? ` {2, 4, -8}`, second right entry, does exist in `elems_no_sym`, as do the vast majority of tested cases.

```
Error. comparing lors
Current bin: 	segment = -1, 	axial pos 0, 	view = 112, 	tangential_pos_num = 0
Symm bin: 		segment = -1, 	axial pos 0, 	view = 112, 	tangential_pos_num = 0

No Symmetries Iterator || Symmetries Iterator 
{2, -4, 6}:0.50568    ||   
{2, -4, 7}:1.15499    ||   
{2, -4, 8}:0.339834    ||   
{2, -3, 4}:0.1962    ||   
{2, -3, 5}:1.15499    ||   
{2, -3, 6}:0.649314    ||   
{2, -2, 3}:1.04171    ||   
{2, -2, 4}:0.958793    ||   
{2, -1, 1}:0.732236    ||   
{2, -1, 2}:1.15499    ||   
{2, -1, 3}:0.113278    ||   
{2, 0, -1}:0.422757    ||   
{2, 0, 0}:1.15499    ||   
{2, 0, 1}:0.422757    ||   
{2, 1, -3}:0.113278    ||   
{2, 1, -2}:1.15499    ||   
{2, 1, -1}:0.732236    ||   
{2, 2, -4}:0.958792    ||   
{2, 2, -3}:1.04172    ||   
{2, 3, -6}:0.649313    ||   
{2, 3, -5}:1.15499    ||   
{2, 3, -4}:0.196201    ||   
{2, 4, -8}:0.339833    ||   
{2, 4, -7}:1.15499    ||   
{2, 4, -6}:0.50568    ||   
                       ||   {2, 5, -8}:0.815161
                       ||   {2, 4, -8}:0.339832
                       ||   {2, 4, -7}:1.15499
                       ||   {2, 4, -6}:0.505682
                       ||   {2, 3, -6}:0.649312
                       ||   {2, 3, -5}:1.15499
                       ||   {2, 3, -4}:0.196202
                       ||   {2, 2, -4}:0.958791
                       ||   {2, 2, -3}:1.04172
                       ||   {2, 1, -3}:0.113277
                       ||   {2, 1, -2}:1.15499
                       ||   {2, 1, -1}:0.732237
                       ||   {2, 0, -1}:0.422756
                       ||   {2, 0, 0}:1.15499
                       ||   {2, 0, 1}:0.422758
                       ||   {2, -1, 1}:0.732235
                       ||   {2, -1, 2}:1.15499
                       ||   {2, -1, 3}:0.113279
                       ||   {2, -2, 3}:1.04171
                       ||   {2, -2, 4}:0.958794
                       ||   {2, -3, 4}:0.196199
                       ||   {2, -3, 5}:1.15499
                       ||   {2, -3, 6}:0.649315
                       ||   {2, -4, 6}:0.505678
                       ||   {2, -4, 7}:1.15499
                       ||   {2, -4, 8}:0.339836
                       ||   {2, -5, 8}:0.815157
```

robbietuk (commented):
See in code comment. 

This appears to be a minor problem but does inidcate that the symmetry LORs are not perfectly equal to the no symmetry LORs. A potential fix for this is to provide some tolerance to the coordinates because a single coordinate in the LOR being non-zero is not a big deal?

I propose that we add a tolerance. If a coordinate and value are not matched by the current while loop checks, check if the coordinate does exist in the other LOR (just out of order) and proceed. Count how many coordinates are unmatched between the LORs (`u`). If `u/T < t%` (e.g. `t=5` and `T` is total number of LORs coordinates) of the total number of coordinates, we mark failed test?

src/recon_test/test_DataSymmetriesForBins_PET_CartesianGrid.cxx
KrisThielemans: ok. pretty weird that this happens now and not before but fine. There is a difficulty with symmetries for the first/last voxel, which I thought we had ok, but maybe not. Could explain some of the "circular FOV" problems.

In any case, this test would first need to sort the rows (luckily, we do have a `sort` method). Can you paste some after sorting? Ideally it's only the end bits.

If that's the case, I'd be interested in seeing what happens if you give it an odd-sized images.

(apologies, this is turning into a project on its own)

KrisThielemans: I'm confused as we do call sort() higher up. certainly doesn't look like it

robbietuk: > pretty weird that this happens now and not before but fine. 

Currently view mashing > 1 (so tilt). Set view mashing to 1 involves a lot more views/angles (error occurs at view 112) instead of the 8 currently tested.

The `sort` method is used above. It is working correctly but `{2, 5, -8}` exists on the right and not the left. Some more debugging is showing that there are 11 unmatched elements for symmetry but only 9 for no symmetry. All no symmetry coords exist in the symmetries, which happens to have an additional two.

robbietuk: To clairify, unmatched in the current test although there is significant overlap between the sets. Just a few outliers.

KrisThielemans: I'm sorry. I seem to be very stupid today. The left column (no sym) increases in 'y' (2nd coord), the right column (sym) decreases in 'y'. how can that be with the `sort`?

> there are 11 unmatched elements for symmetry but only 9 for no symmetry. All no symmetry coords exist in the symmetries, which happens to have an additional two.

completely lost. sorry. I guess you mean in 11 unmatches in the right column (use the matrix with symmetries) and 9 in the left (matrix without symmetries). but then you say 9 of them exist, so they match anyway?

robbietuk: Sorry. Let me try to explain again. There is a `sort` that arranges them somewhat such that the coordinate order makes some sense. As it loops through the vectors, it tests if the iterates are equal. However, in the example above there is an additional entry `{2, 5, -8}`. Given the current `while` loop, this then stalls the iterate on the RHS from progressing until LHS has reached `end()`. 

Could use is something like [`std::set_difference`](https://www.cplusplus.com/reference/algorithm/set_difference) and look for discrepancies? However, this requires C++ 17/20 and two uses (A-B) and (B-A).

KrisThielemans: still, the sort didn't quite sort. Obviously, this doesn't have anything to do with `2,5,-8` missing, but it seems to point to yet another problem. either with how I think `sort` works
https://github.com/UCL/STIR/blob/5daf0d6d2684bd6d1b14b4d2f6eaede92cdc283f/src/recon_buildblock/ProjMatrixElemsForOneBin.cxx#L150
or
https://github.com/UCL/STIR/blob/877511b607a90cbc3c3faebf9fa4e2dc3ea72f1b/src/include/stir/recon_buildblock/ProjMatrixElemsForOneBinValue.inl#L127-L131

KrisThielemans: I'd like to understand the `sort` problem more, because in some places we do rely on it (e.g. in `merge` I believe). Possibly I'm just missing something obvious.

robbietuk: I see what you mean now. Okay lets take sometime to talk about this tomorrow.

robbietuk: Resolving and moving this to an issue https://github.com/UCL/STIR/issues/1002

robbietuk:
Ready to merge (on GHA etc results). This is now
-  an `axial_position` fix with `do_symmetry_shift_z`, 
- some more unique symmetry configuration tests, and 
- an additional test checking the bin in the symmetry `ProjMatrixElemsForOneBin` has been reverted back to the original bin by the symmetry operator.

Issue #495, opened 2020-04-14
(potential) axial shift in scatter simulations


KrisThielemans:
While reviewing #44, I discovered a problem in the original scatter code, existing probably since 2005. 

There are 3 images passed to the scatter simulation code: the activity image, the attenuation image and the "zoomed" attenuation image used for finding the scatter points. The issue occurs when any of these images do not have the same "size" in z-direction as the original data (see below). However, this is in practice always the case, especially for the "scatter-point" image.

#### Projectors
For historical reasons, STIR projectors currently "centre" the image with respect to the scanner, i.e. it puts `0+(max_index.z()+min_index.z())/2.F * voxel_size.z()`as the centre (i.e. the image is centred if `origin.z()==0`), see e.g. 
https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/recon_buildblock/ProjMatrixByBinUsingRayTracing.cxx#L655

This convention causes/d major headaches to @ashgillman in https://github.com/UCL/STIR/projects/1. It is dangerous, e.g. when using `generate_image`, as the location w.r.t. the scanner will change depending on the number of planes. Hence the recommendation to always use the "standard" number of planes.

#### Scatter
This attempts to follow the same convention. However, the scatter simulation code uses downsampled images...

In the scatter code, relevant lines are https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/scatter_buildblock/sample_scatter_points.cxx#L65-L67 and https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/scatter_buildblock/single_scatter_integrals.cxx#L100-L102

Side note: the scatter code also uses coordinates for the detectors (stored in `detector_points_vector`) which are in the "scanner-centred" coordinate system (used by `ProjDataInfo::get_m()` etc).  (By the way, it still does this by using the out-dated  `find_cartesian_coordinates_of_detection` but shifting to the centre, see [here](https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/scatter_buildblock/scatter_detection_modelling.cxx#L70-L75) and [here for the shift](https://github.com/UCL/STIR/blob/29133a3ffa690b444894b4392aa710e8bb33a42e/src/scatter_buildblock/ScatterEstimationByBin.cxx#L303-L304). See #105. This is however not a problem here).

### The problem

[...]

PR #1006, opened 2022-03-16
making scatter work with blocks

danieldeidda:


src/include/stir/scatter/ScatterEstimation.h
```diff
@@ -5,6 +5,7 @@
 /*
     Copyright (C) 2018 - 2019 University of Hull
     Copyright (C) 2016,2020 University College London
+    Copyright (C) 2022 Nationcal Physical Laboratory
```
KrisThielemans, line 8: I guess currently no changes in this file from you, so I'd remove this (and the line below), unless you will change it

src/scatter_buildblock/ScatterSimulation.cxx
```diff
@@ -831,13 +831,30 @@ ScatterSimulation::downsample_scanner(int new_num_rings, int new_num_dets)
 
     const Scanner *const old_scanner_ptr = this->proj_data_info_cyl_noarc_cor_sptr->get_scanner_ptr();
     shared_ptr<Scanner> new_scanner_sptr( new Scanner(*old_scanner_ptr));
-
+
```
KrisThielemans, line 834: remove white space change

KrisThielemans (changes requested):
some comments as discussed. Also probably drop changes to `inverse_SSRB` as discussed (although we probably want to do that nearer to the end of this PR)

danieldeidda:
I am a bit confused by the following:
https://github.com/UCL/STIR/blob/8ef0d1084af169d96f2898ed541837be293e362b/src/scatter_buildblock/ScatterSimulation.cxx#L93

Are we meant to provide a template which already describes the downsample scanner?

KrisThielemans:
`process_data()` requires `set_up()` first. This has logic to create the template via downsampling (if you didn't give one yet I guess).

danieldeidda:
> `process_data()` requires `set_up()` first. This has logic to create the template via downsampling (if you didn't give one yet I guess).

OK so it should work even if we don't give a template in input? In my mind the template was the original scanner that then is downsampled

danieldeidda:
> for cylindrical, the template is optional (as far as I recall), i.e. it will downsample then according to (optional) parameters.
> 
> It'd be nice to end up there in that situation for blocksoncylindrical as well, which I think you nearly had it.
> 
> Before @NikEfth created the downsampling functionality, we had to create the template by hand. Now it should be optional.

It seems that if I don't give a template name then post_processing() which calls set_output_proj_data(const std::string& filename)  throws the error:

```
set_output_proj_data(const std::string& filename)
{

    if(is_null_ptr(this->proj_data_info_cyl_noarc_cor_sptr))
    {
        error("ScatterSimulation: Template ProjData has not been set. Aborting.");
    }
```

KrisThielemans:
Possibly that's a bug.then. I believe we do run `ScatterSimulation` without template in `ScatterEstimation`. Might need some debugging. sorry

danieldeidda:
> Possibly that's a bug.then. I believe we do run `ScatterSimulation` without template in `ScatterEstimation`. Might need some debugging. sorry

OK after some debugging I think for the Scatterestimation it makes sens to allow no template because the info about the scanner are provide in that class and post_processing is not called, therefore we don't encounter that error. However when we run the simulate_scatter utility on its own we need a template otherwise it does have any clue about scanner geometry. I think the easiest solution would be to re-set  output_proj_data after the downsampling in this way the everything runs and the simulated scatter looks something reasonably alike to a cylinder

KrisThielemans:
> for the ScattereEstimation it makes sense to allow no template because the info about the scanner are provided in that class and post_processing is not called, therefore we don't encounter that error. However, when we run the simulate_scatter utility on its own we need a template otherwise it does not have any clue about scanner geometry.

yes, makes sense.

> re-set output_proj_data after the downsampling in this way everything runs and the simulated scatter looks something reasonably alike to a cylinder

I'm 100% lost. reset where? `ScatterEstimation` ,but that works already? `ScatterSimulation`, but you just said it cannot work.

danieldeidda:
> 
> I'm 100% lost. reset where? `ScatterEstimation` ,but that works already? `ScatterSimulation`, but you just said it cannot work.

In ScatterSimulation, we give a template which is the original scanner (out_projdata is set equal to that in post_processing()), then we downsample this scanner and change out_projdata info with the downsampled one.

KrisThielemans:
ok. I think it does make sense to (optionally?) let ScatterSimulation do the downsampling, but then I also think it has to do the upsampling, i.e. output has to be the same size as the template. This would be good, but obviously out of scope for this PR.

danieldeidda:
This seems to run OK. However, if we look at the downsample scatter and the upsample one in the figure below, we can see that the downsample seem to have some empty part which I am not sure is correct

image

KrisThielemans:
who knows how Amide displays output of `extract_segments`. "image" dimensions might not make sense anyway. A better test is to compute the scatter with full sampling (in `m`) and compare that to the upsampled sino. Clearly, you will have to use something with asymmetry in `z`

danieldeidda:
> who knows how Amide displays output of `extract_segments`. "image" dimensions might not make sense anyway. A better test is to compute the scatter with full sampling (in `m`) and compare that to the upsampled sino. Clearly, you will have to use something with asymmetry in `z`

I am running simulate_scatter with original scanner. It has been running for 25h with openMP on. does this seem normal?

KrisThielemans:
it will take a long time yes, certainly if you have it in full 3D. That's the whole reason we do the downsampling... You should be able to monitor progress in the log files. It'll give an estimate of run-time, which initially is too pessimistic as it ignores the benefit of the caching of "scatter-point <-> detector" lines, but it would still give you some idea.

danieldeidda:
> it will take a long time yes, certainly if you have it in full 3D. That's the whole reason we do the downsampling... You should be able to monitor progress in the log files. It'll give an estimate of run-time, which initially is too pessimistic as it ignores the benefit of the caching of "scatter-point <-> detector" lines, but it would still give you some idea.

I think this look weird, is it possible that the z_shift needs to be applied again?

image

KrisThielemans:
I guess it'd be good to see coronal slice of the activity/attenuation images.

Might need a hunt through the `ScatterSimulation` code how it gets the coordinates, and in particular if there's something checks on cylindrical or not.

danieldeidda:
> I guess it'd be good to see coronal slice of the activity/attenuation images.
> 
> Might need a hunt through the `ScatterSimulation` code how it gets the coordinates, and in particular if there's something checks on cylindrical or not.

I had put this check for Cylindrical only but pheraps needs to be there  and in fact detector_coord_A.z()  (and  B) is different than 0 
https://github.com/UCL/STIR/blob/bdaa4663226d40c44151f847d0757a3a723dab70/src/scatter_buildblock/ScatterSimulation.cxx#L385
https://github.com/UCL/STIR/blob/bdaa4663226d40c44151f847d0757a3a723dab70/src/scatter_buildblock/ScatterSimulation.cxx#L386
https://github.com/UCL/STIR/blob/bdaa4663226d40c44151f847d0757a3a723dab70/src/scatter_buildblock/ScatterSimulation.cxx#L387

before this line find_cartesian_coordinates_of_detection(detector_coord_A, detector_coord_B, Bin(0, 0, 0, 0)); is called which adds a z_shift further away from zero

danieldeidda:
> Might need a hunt through the `ScatterSimulation` code how it gets the coordinates, and in particular if there's something checks on cylindrical or not.

So I had a look at the code and there is a shift in the following:
https://github.com/UCL/STIR/blob/bdaa4663226d40c44151f847d0757a3a723dab70/src/scatter_buildblock/ScatterSimulation.cxx#L404-L405

when I change the sign of this shift I get something that looks more like the equivalent in the cylindrica geometry

image

KrisThielemans:
unfortunately that makes little sense to me. You did a lot of hard work to make sure that `find_cartesian*` and `get_m` give the same results for cylindrical and block. That then implies that the shift has to be the same in both cases. So I'm not sure what's going on.

Possibly you could enable the debug code above, removing lines 
https://github.com/UCL/STIR/blob/bdaa4663226d40c44151f847d0757a3a723dab70/src/scatter_buildblock/ScatterSimulation.cxx#L385 and https://github.com/UCL/STIR/blob/bdaa4663226d40c44151f847d0757a3a723dab70/src/scatter_buildblock/ScatterSimulation.cxx#L394 as those were appropriate before we shifted everything back.

danieldeidda:
> unfortunately that makes little sense to me. You did a lot of hard work to make sure that `find_cartesian*` and `get_m` give the same results for cylindrical and block. That then implies that the shift has to be the same in both cases. So I'm not sure what's going on.
> 
> Possibly you could enable the debug code above, removing lines
> 
> https://github.com/UCL/STIR/blob/bdaa4663226d40c44151f847d0757a3a723dab70/src/scatter_buildblock/ScatterSimulation.cxx#L385
> 
> and
> https://github.com/UCL/STIR/blob/bdaa4663226d40c44151f847d0757a3a723dab70/src/scatter_buildblock/ScatterSimulation.cxx#L394
> 
> as those were appropriate before we shifted everything back.
L394 I have already removed and works fine but the 385 will fail. the z position appears to be -axial_length

danieldeidda:
> > unfortunately that makes little sense to me. You did a lot of hard work to make sure that `find_cartesian*` and `get_m` give the same results for cylindrical and block. That then implies that the shift has to be the same in both cases. So I'm not sure what's going on.
> > Possibly you could enable the debug code above, removing lines
> > https://github.com/UCL/STIR/blob/bdaa4663226d40c44151f847d0757a3a723dab70/src/scatter_buildblock/ScatterSimulation.cxx#L385
> > 
> > and
> > https://github.com/UCL/STIR/blob/bdaa4663226d40c44151f847d0757a3a723dab70/src/scatter_buildblock/ScatterSimulation.cxx#L394
> > 
> > as those were appropriate before we shifted everything back.
> > L394 I have already removed and works fine but the 385 will fail. the z position appears to be -axial_length

or pheraps we could have a check like in L385 and if is not zero calculate the difference and force it zero? I know it sounds ugly even while I am writing :)

KrisThielemans:
yup. ugly :grin:. My interpretation is that this check has to be satisfied, as otherwise there is something wrong with the coordinate systems. No?

By the way, I feel there's no need to quote whole messages if the reply is "in sequence".

danieldeidda:
This now seems to be working. Although I am a bit confused with the sampling in particular the line below:

https://github.com/UCL/STIR/blob/0b9c81d567d395d11aa1351064f25fff58d0901b/src/include/stir/ProjDataInfoGeneric.inl#L96

to me this should not be divided by 2

danieldeidda:
@emikhay could you try to reconstruct you data with this branch? Pheraphs you could try to do the scatter correction with BlocksOnCylindrical

KrisThielemans:
> Although I am a bit confused with the sampling in particular the line below:

that is indeed a bug (repeated a few times). Maybe you could correct that in a separate PR as I hope we can release 5.0.1 with bug corrections early next week or even earlier.

danieldeidda:
Ok yes. I'll do that

emikhay:
> @emikhay could you try to reconstruct you data with this branch? Pheraphs you could try to do the scatter correction with BlocksOnCylindrical

Yes, I will try it between today and tomorrow.

danieldeidda:
I am running estimate_scatter with mediso data using Blocks on cylindrical. I am getting this error in different places:
ProjDataInfoCylindrical::get_ring_pair_for_segment_axial_pos_num does not work for data with axial compression

the first place was when is trying to apply SSRB to the attenuation which I fixed with the reduce_segment... but now also in the normalisation step which is not really calling SSRB.
this always happen when calling create_new_proj_data(

KrisThielemans:
does your mediso data (or a template) have axial compression?

danieldeidda:
> does your mediso data (or a template) have axial compression?

no i doesn't. the interesting thing is that if I change the create_new_proj() with .reset(new ProjDataInMemory()) it seems to be OK

KrisThielemans:
so I guess you have `run_debug_mode=true` such that it creates a `ProjDataInterfile` [here](https://github.com/danieldeidda/STIR/blob/0b9c81d567d395d11aa1351064f25fff58d0901b/src/scatter_buildblock/ScatterEstimation.cxx#L1395-L1399)

I guess somewhere a test that is too stringent. Only way to resolve that seems to be to by debugging and doing a stack trace.

weird

danieldeidda:
Yes debug is on and I am following the errors as they come. These all happens when creating 2d projdata. altough before I was running in release mode and I still got that error, however I suspect that was because of the SSRB

danieldeidda:
also I have bin printing the max_ring_diff and min_ring_diff  per segment and the error always happens with these values:
-37  37  for segment 0.  None of the interfiles have such values

danieldeidda:
another thing is when  there are gaps axially  and do the following:
` shared_ptr<ProjDataInfo> proj_data_info_2d_sptr(
                    SSRB(*this->input_projdata_sptr->get_proj_data_info_sptr(),
                          this->input_projdata_sptr->get_num_segments(), 1, false));`
we get this error
`const float number_of_ms =
	  (max_m - min_m)/out_proj_data_info_sptr->get_axial_sampling(out_segment_num)+1;
	if (fabs(round(number_of_ms)-number_of_ms) > 1.E-3)
	  error("SSRB: number of axial positions to be found in out_segment %d is non-integer %g\n",
		out_segment_num, number_of_ms);`

src/scatter_buildblock/upsample_and_fit_scatter_estimate.cxx
```diff
@@ -53,12 +54,21 @@ upsample_and_fit_scatter_estimate(ProjData& scaled_scatter_proj_data,
 {
   shared_ptr<ProjDataInfo> 
     interpolated_direct_scatter_proj_data_info_sptr(emission_proj_data.get_proj_data_info_sptr()->clone());
-  interpolated_direct_scatter_proj_data_info_sptr->reduce_segment_range(0,0);
```
KrisThielemans, line 56: sadly this seems wrong.  your change does the reduction after the construction `interpolated_direct_scatter`.

KrisThielemans (changes requested):
some comments, not related to your problem

KrisThielemans:
> Another thing is when there are gaps axially and do the following:

I'm confused: gaps axially => blocksOnCylindrical => no SSRB but reduce_segments?

danieldeidda:
> > Another thing is when there are gaps axially and do the following:
> 
> I'm confused: gaps axially => blocksOnCylindrical => no SSRB but reduce_segments?

so we don't need this bit?
shared_ptr<ProjDataInfo> proj_data_info_2d_sptr( SSRB(*this->input_projdata_sptr->get_proj_data_info_sptr(), this->input_projdata_sptr->get_num_segments(), 1, false));
 we have it before the check on the geometry

KrisThielemans:
the SSRB needs to be inside the `if`, in `else` you call `reduce_segment`

danieldeidda:
> the SSRB needs to be inside the `if`, in `else` you call `reduce_segment`

I think that was part of the problem then, now it complains in the forward_project() line:
`forw_projector_sptr->forward_project(*mask_projdata, *this->mask_image_sptr);`

the error is saying there is an  out of range. Should this be following the downsampled_scanner sizes?

danieldeidda:
Also I would expect max_axial_pos_per segments to have only 1 element for a direct sino? mask_projdata has 15 elements

src/scatter_buildblock/upsample_and_fit_scatter_estimate.cxx
danieldeidda: are you saying I need to construct it with the reduction? 
`interpolated_direct_scatter_proj_data_info_sptr(emission_proj_data.get_proj_data_info_sptr()->reduce_segment_range(0,0))`

src/scatter_buildblock/ScatterSimulation.cxx
```diff
@@ -861,6 +902,7 @@ ScatterSimulation::downsample_scanner(int new_num_rings, int new_num_dets)
 	 % templ_proj_data_info_sptr->parameter_info(),
 	 3);
     this->set_template_proj_data_info(*templ_proj_data_info_sptr);
+    this->set_output_proj_data(this->output_proj_data_filename);
```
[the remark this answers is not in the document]
danieldeidda: If I don't use it I get an error saying the data is incompatible

src/buildblock/interpolate_axial_position.cxx
```diff
@@ -0,0 +1,80 @@
...
+          Bin bin(segment,0,axial_pos,0);
+          int relative_pos=0;
+          float m = proj_data_out_info.get_m(bin);
+          float in_m_sampling= proj_data_in_info.get_sampling_in_m(bin)*2;
+          relative_pos=abs(round((m -m_offset_in)/ in_m_sampling));
+           Sinogram<float> sino= proj_data_out.get_empty_sinogram(axial_pos,segment);
+          for (int view=proj_data_out.get_min_view_num(); view<=proj_data_out.get_max_view_num();view++)
+              for (int tan=proj_data_out.get_min_tangential_pos_num(); tan<=proj_data_out.get_max_tangential_pos_num();tan++)
+              {
+                  sino[view][tan]=proj_data_in.get_sinogram(relative_pos,0)[view][tan]; //proj_data_interpolator(relative_pos));
```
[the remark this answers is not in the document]
danieldeidda: I tried that but complains because projdata_in has different projdata_info. perhaps I could create a copy of of proj_data_in with the output_projdata_info

src/buildblock/ProjDataInfoBlocksOnCylindricalNoArcCorr.cxx
```diff
@@ -120,8 +120,8 @@ find_scanner_coordinates_given_cartesian_coordinates(int& det1, int& det2, int&
 
   DetectionPosition<> det_pos1;
   DetectionPosition<> det_pos2;
-  if (get_scanner_ptr()->find_detection_position_given_cartesian_coordinate(det_pos1, c1-this->z_shift)==Succeeded::no ||
-      get_scanner_ptr()->find_detection_position_given_cartesian_coordinate(det_pos2, c2-this->z_shift)==Succeeded::no)
+  if (get_scanner_ptr()->find_detection_position_given_cartesian_coordinate(det_pos1, c1+this->z_shift)==Succeeded::no ||
+      get_scanner_ptr()->find_detection_position_given_cartesian_coordinate(det_pos2, c2+this->z_shift)==Succeeded::no)
```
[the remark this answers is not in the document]
danieldeidda: because for the test the ax_pos for the shift was calculated on the last ring instead of the first and I had based the shift on the test but for direct sino the point A and B of the LOR have the same z which is negative

danieldeidda:
OK there seems to be a problem with m_offset for 2d projdata.
https://github.com/UCL/STIR/blob/0b9c81d567d395d11aa1351064f25fff58d0901b/src/buildblock/ProjDataInfoCylindrical.cxx#L291-L293

for the case: seg=0 ring1_plus_ring2=4 instead it shoul be 2 i guess

danieldeidda:
or actually it should start from 0

KrisThielemans:
hmmm. I thought we never used `m_offset` etc in the blocks case. Or indeed `axial_sampling`. It doesn't make sense for the general case. Seems we need more modifications.

Why do we see this stuff only now? Possibly because the BPET test scanner has in axial direction only one block per bucket  and only one bucket and therefore no gaps?

danieldeidda:
I am not sure. We have a test with 1 ring and one with full scanner so in theory I would expect this happen there two. The only difference here is that we expressly reduce the projdata to only direct LOR

KrisThielemans:
ah. sorry. I forgot to merge #1021. done now.

Nevertheless, we really shouldn't be using `m_offset` and `sampling_distance` for this stuff.

danieldeidda:
OK another issue happens in 
https://github.com/UCL/STIR/blob/0b9c81d567d395d11aa1351064f25fff58d0901b/src/scatter_buildblock/ScatterSimulation.cxx#L416
I am reading now the issue #495

danieldeidda:
This is now working! but I need to have a look at some of @KrisThielemans  comments

src/buildblock/interpolate_axial_position.cxx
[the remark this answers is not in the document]
danieldeidda: @KrisThielemans any thought on this?

src/scatter_buildblock/upsample_and_fit_scatter_estimate.cxx
danieldeidda: sorry i don't see why this is wrong though

KrisThielemans: ok.my mistake. ignore

src/buildblock/interpolate_axial_position.cxx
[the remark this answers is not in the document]
danieldeidda: >if view and tang_pos ranges are the same, the following should work as well
sino.fill(sino_in)

this doesn't work fill() usually gets a scalar and set every element to that

danieldeidda:
sadly another issue in scatter estimate: the scatter estimation is producing empty sinograms both with simulation and mediso data I did not realise that as I was looking at the additive sino. Anyway one thing that seem strange is the mask_projdata attached

image

danieldeidda:
in fact there is also this message
0 bins in mask for sinogram at segment 0, axial_pos 1
0 bins in mask for sinogram at segment 0, axial_pos 2
0 bins in mask for sinogram at segment 0, axial_pos 3
0 bins in mask for sinogram at segment 0, axial_pos 4
0 bins in mask for sinogram at segment 0, axial_pos 5
0 bins in mask for sinogram at segment 0, axial_pos 6
0 bins in mask for sinogram at segment 0, axial_pos 7
0 bins in mask for sinogram at segment 0, axial_pos 8
0 bins in mask for sinogram at segment 0, axial_pos 9
0 bins in mask for sinogram at segment 0, axial_pos 
10
0 bins in mask for sinogram at segment 0, axial_pos 11
0 bins in mask for sinogram at segment 0, axial_pos 12
0 bins in mask for sinogram at segment 0, axial_pos 13
0 bins in mask for sinogram at segment 0, axial_pos 14
0 bins in mask for sinogram at segment 0, axial_pos 15

KrisThielemans:
will need some hunting for the bug in the construction of the mask. Any log file output?

danieldeidda:
[estimateSC.log](https://github.com/UCL/STIR/files/8603417/estimateSC.log)

here is the log file. the mask image is OK the FP mask is weird. I am suspecting some sort of axial shifting that we did not consider?

KrisThielemans (changes requested):
I've had a look but am not sure if this the version you're running at the moment. At first sight, this is incomplete as data isn't filled-in in the blocks case as far as I can see.

There's a lot of `if` statements here which make the code even worse than it was already. I wonder if we couldn't move the `if`s into separate functions such as
```
  shared_ptr<ProjDataInfo>make_2D(ProjDataInfo const&);
  void make_2D(ProjData out_2D&, ProjData const in_3D&);
```
maybe similar on the "inverse" side, although there it might be possible to simply let `interpolate_projdata` do an `inverse_SSRB` in the cylindrical case.

all a bit confusing...

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -439,15 +441,21 @@ set_up()
...
+            this->input_projdata_2d_sptr.reset(new ProjDataInMemory(this->input_projdata_sptr->get_exam_info_sptr(),
+                                                                    proj_data_info_2d_sptr));
+            
+            SSRB(*this->input_projdata_2d_sptr,
+                 *input_projdata_sptr,false);
+        }
+        else{ 
+            proj_data_info_2d_sptr->reduce_segment_range(0,0);
+            this->input_projdata_2d_sptr.reset(new ProjDataInMemory(this->input_projdata_sptr->get_exam_info_sptr(),
+                                                                    proj_data_info_2d_sptr));
```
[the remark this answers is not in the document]
danieldeidda: OK that's it then. Naively I thought reduce_seg_range() would do everything

src/scatter_buildblock/upsample_and_fit_scatter_estimate.cxx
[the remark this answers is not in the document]
danieldeidda: for blocks interpolate_projdata basically goes directly to interpolate_axial_position

src/scatter_buildblock/ScatterEstimation.cxx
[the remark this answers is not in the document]
danieldeidda: I could add something like:
`SegmentBySinogram<float> segment=input_projdata_sptr->get_segment_by_sinogram(0);`
` input_projdata_2d_sptr->set_segment(segment);`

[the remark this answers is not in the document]
KrisThielemans: yes, but this occurs many times, so needs to be in a function.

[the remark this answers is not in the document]
danieldeidda: yes definetely

danieldeidda:
with this commit the 2d projdata are not empty however the scatter estimate is still empty. I am not sure anymore that the mask is the problem because with the mediso data the mask is zero where the pantom is and 1 outside. Which I guess it makes sens for the tail fitting

danieldeidda:
unscaled_est_projdata_sptr in the following is used for the upsampled:
https://github.com/UCL/STIR/blob/852c10d86beaac4892a2897922fd398911667e8c/src/scatter_buildblock/ScatterEstimation.cxx#L936-L943

but I don't see anywhere where it is filled and in fact the file is empty. I can see it is reset () here but not filled
https://github.com/UCL/STIR/blob/852c10d86beaac4892a2897922fd398911667e8c/src/scatter_buildblock/ScatterEstimation.cxx#L841-L843

danieldeidda:
using the following before L936 seems to fix it

`unscaled_est_projdata_sptr(new ProjDataInMemory(*this->scatter_simulation_sptr->get_output_proj_data_sptr()))`

src/buildblock/Scanner.cxx
```diff
@@ -1038,9 +1038,9 @@ check_consistency() const
               get_transaxial_crystal_spacing(), get_num_transaxial_crystals_per_block(), get_transaxial_block_spacing());
           return Succeeded::no;
         }
-  
-        if (get_transaxial_block_spacing()*get_num_transaxial_blocks_per_bucket()
-            < round (2*inner_ring_radius*tan(_PI/get_num_transaxial_blocks()/get_num_transaxial_blocks_per_bucket())*1000.0)/1000.0)
+        
+        if (round(get_transaxial_block_spacing()*get_num_transaxial_blocks_per_bucket())
+            < round (2*inner_ring_radius*tan(_PI/get_num_transaxial_buckets()/get_num_transaxial_blocks_per_bucket())*1000.0)/1000.0)
```
KrisThielemans, line 1043:
```suggestion
        if (round(get_transaxial_block_spacing()*get_num_transaxial_blocks_per_bucket()*1000.0)/1000.0
            < round (2*inner_ring_radius*tan(_PI/2/get_num_transaxial_buckets())*1000.0)/1000.0)
```

KrisThielemans:
RE the problem with the mask

we reviewed this and it turns out that in the simulation, the object is slightly too large, i.e. larger than the FOV for the scanner used. Therefore, there are no "tails" in the sinograms where the object is present.

The tail-fitting will then just use scale factor 1 for those sinograms (i.e. assume there is only single scatter), unless the min_scalefactor is set higher of course. This seems to be the best it can do. So this is not a bug in the code, but arguably one in the simulation set-up.

src/scatter_buildblock/ScatterSimulation.cxx
```diff
@@ -826,18 +832,52 @@ ScatterSimulation::downsample_scanner(int new_num_rings, int new_num_dets)
...
+            new_num_dets=proj_data_info_cyl_noarc_cor_sptr->get_scanner_ptr()->get_num_detectors_per_ring();
     }
 
     const Scanner *const old_scanner_ptr = this->proj_data_info_cyl_noarc_cor_sptr->get_scanner_ptr();
     shared_ptr<Scanner> new_scanner_sptr( new Scanner(*old_scanner_ptr));
-
-    // preserve the length of the scanner
-    float scanner_length = new_scanner_sptr->get_num_rings()* new_scanner_sptr->get_ring_spacing();
+    // preserve the length of the scanner the following includes gaps
+    float scanner_length = new_scanner_sptr->get_num_axial_buckets()*
```
danieldeidda, line 841: the problem with the test is here. Cylindrical geometry has axial_block_spacing set to -1. I guess we should have a different length calculation for Cylindrical?

danieldeidda: test runs OK with this modification

KrisThielemans:
`./run_scatter_tests.sh` is causing time-outs (run time of more than 5 hours)

KrisThielemans:
The "2d" sino output is not SSRB'd.

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -185,6 +187,62 @@ ScatterEstimation(const std::string& parameter_filename)
...
+shared_ptr<ProjData> 
+ScatterEstimation::
+make_2D_projdata(const shared_ptr<ProjData> in_3d_sptr)
+{
+    shared_ptr<ProjDataInfo> out_info_2d_sptr(in_3d_sptr->get_proj_data_info_sptr()->create_shared_clone());
+    shared_ptr<ProjData> out_2d_sptr;
+    if (in_3d_sptr->get_proj_data_info_sptr()->get_scanner_sptr()->get_scanner_geometry()=="Cylindrical")
+    {
+        out_2d_sptr.reset(new ProjDataInMemory(in_3d_sptr->get_exam_info_sptr(),
+                                               out_info_2d_sptr));
```
KrisThielemans, line 200: as `out_info_2d_sptr` is set above to be the original, it isn't 2D. Therefore `SSRB` won't do anything.

Instead, `out_info_2d_sptr` needs to be set inside this `if` via a call to `SSRB` (as in the original code)

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -185,6 +187,62 @@ ScatterEstimation(const std::string& parameter_filename)
...
+
+shared_ptr<ProjData> 
+ScatterEstimation::
+make_2D_projdata(const shared_ptr<ProjData> in_3d_sptr, string template_filename)
+{
+    shared_ptr<ProjDataInfo> out_info_2d_sptr(in_3d_sptr->get_proj_data_info_sptr()->create_shared_clone());
+    shared_ptr<ProjData> out_2d_sptr;
+    if (in_3d_sptr->get_proj_data_info_sptr()->get_scanner_sptr()->get_scanner_geometry()=="Cylindrical")
+    {
+        out_2d_sptr = create_new_proj_data(template_filename,
```
KrisThielemans, line 226: same problem as above

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -439,15 +497,7 @@ set_up()
...
-        shared_ptr<ProjDataInfo> proj_data_info_2d_sptr(
-                    SSRB(*this->input_projdata_sptr->get_proj_data_info_sptr(),
-                          this->input_projdata_sptr->get_num_segments(), 1, false));
-
-	this->input_projdata_2d_sptr.reset(new ProjDataInMemory(this->input_projdata_sptr->get_exam_info_sptr(),
-                                                                 proj_data_info_2d_sptr));
-
-        SSRB(*this->input_projdata_2d_sptr,
-             *input_projdata_sptr,false);
+        this->input_projdata_2d_sptr.reset(new ProjDataInMemory(*make_2D_projdata(this->input_projdata_sptr)));
```
KrisThielemans, line 500: now we're making a copy of a copy... instead
```suggestion
        this->input_projdata_2d_sptr = make_2D_projdata(this->input_projdata_sptr));
```

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -595,20 +645,16 @@ set_up_iterative(shared_ptr<IterativeReconstruction<DiscretisedDensity<3, float>
...
         info("ScatterEstimation: Running SSRB on attenuation correction coefficients ...");
 
         std::string out_filename = "tmp_atten_sino_2d.hs";
-
-        atten_projdata_2d_sptr = create_new_proj_data(out_filename, this->input_projdata_2d_sptr->get_exam_info_sptr(),
-                             this->input_projdata_2d_sptr->get_proj_data_info_sptr()->create_shared_clone());
-
-        SSRB(*atten_projdata_2d_sptr,
-             *tmp_atten_projdata_sptr, true);
+        atten_projdata_2d_sptr.reset(new ProjDataInMemory(*make_2D_projdata(tmp_atten_projdata_sptr, out_filename)));
```
KrisThielemans, line 656: as above

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -649,8 +695,7 @@ set_up_iterative(shared_ptr<IterativeReconstruction<DiscretisedDensity<3, float>
 
               info("ScatterEstimation: Performing SSRB on efficiency factors ...");
 
-              SSRB(*norm_projdata_2d_sptr,
-                   *inv_projdata_3d_sptr,false);
+              norm_projdata_2d_sptr.reset(new ProjDataInMemory(*make_2D_projdata(inv_projdata_3d_sptr)));
```
KrisThielemans, line 698: as above

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -690,11 +735,10 @@ set_up_iterative(shared_ptr<IterativeReconstruction<DiscretisedDensity<3, float>
         {
             info("ScatterEstimation: Running SSRB on the background data ...");
 
-	    this->back_projdata_2d_sptr.reset(new ProjDataInMemory(this->input_projdata_2d_sptr->get_exam_info_sptr(),
-								   this->input_projdata_2d_sptr->get_proj_data_info_sptr()->create_shared_clone()));
+//	    this->back_projdata_2d_sptr.reset(new ProjDataInMemory(this->input_projdata_2d_sptr->get_exam_info_sptr(),
+//								   this->input_projdata_2d_sptr->get_proj_data_info_sptr()->create_shared_clone()));
```
KrisThielemans, line 739:
```suggestion
```

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -690,11 +735,10 @@ set_up_iterative(shared_ptr<IterativeReconstruction<DiscretisedDensity<3, float>
...
 
-            SSRB(*this->back_projdata_2d_sptr,
-                 *this->back_projdata_sptr, false);
+            this->back_projdata_2d_sptr.reset(new ProjDataInMemory(*make_2D_projdata(back_projdata_sptr)));
```
KrisThielemans, line 741: as above

src/scatter_buildblock/upsample_and_fit_scatter_estimate.cxx
```diff
@@ -53,12 +54,22 @@ upsample_and_fit_scatter_estimate(ProjData& scaled_scatter_proj_data,
...
+  {
+      bool remove_interleaving_block=false;
+      interpolated_direct_scatter_proj_data_info_sptr->reduce_segment_range(0,0);
+      interpolate_projdata(interpolated_direct_scatter, scatter_proj_data, spline_type, remove_interleaving_block);
+  }
+  else
+  {
+      interpolated_direct_scatter_proj_data_info_sptr->reduce_segment_range(0,0);
+      interpolate_projdata(interpolated_direct_scatter, scatter_proj_data, spline_type, remove_interleaving);
+  }
```
KrisThielemans, line 72: can't this be done simpler? do we need a different `bool`? I guess we could force it ` false` in `set-up`

src/scatter_buildblock/upsample_and_fit_scatter_estimate.cxx
```diff
@@ -67,7 +78,13 @@ upsample_and_fit_scatter_estimate(ProjData& scaled_scatter_proj_data,
     {
       ProjDataInMemory interpolated_scatter(emission_proj_data.get_exam_info_sptr(),
 					    emission_proj_data.get_proj_data_info_sptr()->create_shared_clone());
-      inverse_SSRB(interpolated_scatter, interpolated_direct_scatter);
+      if (emission_proj_data.get_proj_data_info_sptr()->get_scanner_sptr()->get_scanner_geometry()=="BlocksOnCylindrical")
```
KrisThielemans, line 81: check on cylindrical, such that maybe this'd work with generic

src/scatter_buildblock/upsample_and_fit_scatter_estimate.cxx
```diff
@@ -135,7 +152,10 @@ upsample_and_fit_scatter_estimate(ProjData& scaled_scatter_proj_data,
     }
   else // min/max_scale_factor equal to 1 and no norm
     {
-      inverse_SSRB(scaled_scatter_proj_data, interpolated_direct_scatter);
+      if (emission_proj_data.get_proj_data_info_sptr()->get_scanner_sptr()->get_scanner_geometry()=="BlocksOnCylindrical")
```
KrisThielemans, line 155: as above

KrisThielemans (changes requested):

src/scatter_buildblock/ScatterEstimation.cxx
KrisThielemans: yup. that's what it was before

KrisThielemans:
looks like it's still taking too much time. Something else I guess. Did you run the test locally?

danieldeidda:
not yet. Still waiting for my computer

src/scatter_buildblock/upsample_and_fit_scatter_estimate.cxx
danieldeidda: in set-up?

src/scatter_buildblock/ScatterSimulation.cxx
```diff
@@ -230,7 +232,7 @@ ScatterSimulation::set_defaults()
     this->zoom_size_xy = -1;
     this->zoom_size_z = -1;
     this->downsample_scanner_bool = false;
-    this->downsample_scanner_dets = 64;
+    this->downsample_scanner_dets = -1;
```
danieldeidda, line 235: could this be the reason of the slow-down? we may need to force it to 64 in the parfile

danieldeidda: it does make the test way faster

KrisThielemans: It seems to point to an error in the code where you set it to 64 if it is -1. I wouldn't want to have a default that is essentially unworkable.

danieldeidda: sorry not sure I follow

KrisThielemans: resolved in offline discussions

src/scatter_buildblock/upsample_and_fit_scatter_estimate.cxx
KrisThielemans:
```suggestion
  bool actual_remove_interleaving = remove_interleaving;
  if (remove_interleaving && emission_proj_data.get_proj_data_info_sptr()->get_scanner_sptr()->get_scanner_geometry()!="Cylindrical")
  {
     warning("upsample_and_fit_scatter_estimate: forcing remove_interleaving to false as non-cylindrical projdata");
     actual_remove_interleaving = false;
  }
  interpolated_direct_scatter_proj_data_info_sptr->reduce_segment_range(0,0);
  interpolate_projdata(interpolated_direct_scatter, scatter_proj_data, spline_type, actual_remove_interleaving);
```

KrisThielemans: However, the line
```
interpolated_direct_scatter_proj_data_info_sptr->reduce_segment_range(0,0);
```
should happen BEFORE the construction of `interpolated_direct_scatter` (as it was before). Now it will have no effect

src/scatter_buildblock/ScatterEstimation.cxx
KrisThielemans: why was this resolved? The code is unchanged. (sorry, I probably lost track). There's a few more like this below.

src/buildblock/ProjDataInfoCylindrical.cxx
```diff
@@ -291,7 +294,10 @@ initialise_ring_diff_arrays() const
         const int ring1_plus_ring2 =
           round(ring1_plus_ring2_float);
         // check that it was integer
+        if (get_scanner_sptr()->get_scanner_geometry()=="Cylindrical")
+        {
         assert(fabs(ring1_plus_ring2 - ring1_plus_ring2_float) < 1E-4) ;
+        }
         segment_axial_pos_to_ring1_plus_ring2[s_num][ax_pos_num] = ring1_plus_ring2;
```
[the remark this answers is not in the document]
KrisThielemans: To clarify this comment. We skip the `assert` in the non-cylindrical case, but still fill-in the array (in fact `m_offset` makes no sense), resulting in potentially unsafe usage. I was wondering if we can make sure that the variable is not used in the non-cylindrical case.

Checking the code now, it looks like `sampling_corresponds_to_physical_rings` should be set to `false` such that we never get into this loop. If it is `true`, then the `assert` should remain. (ok, in fact I think we should change the `assert` to an explicit `if` statement with call to `error`, as an `assert` will only be checked in Debug modem but let's leave that for another day).

Summary:
- make sure that `sampling_corresponds_to_physical_rings` is set to `false` in the non-cylindrical case (in the constructor)
- do not grow/fill `m_offset` in the non-cylindrical case either

src/scatter_buildblock/ScatterEstimation.cxx
KrisThielemans: looks like you did. GitHub tripped me up by going to an old commit. sorry.

danieldeidda: Ah OK

src/scatter_buildblock/upsample_and_fit_scatter_estimate.cxx
danieldeidda: so shall I move it there?

src/buildblock/ProjDataInfoCylindrical.cxx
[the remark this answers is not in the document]
KrisThielemans: yes. if non-cylindrical, set it to `false`.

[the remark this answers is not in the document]
KrisThielemans: > Sorry I don't see where m_offset is incremented

it is initialised in the lines above, but is never used (and should never be used) in the non-cylindrical case (as `get_m` is overloaded)

src/buildblock/ProjDataInfoCylindrical.cxx
```diff
@@ -202,16 +204,19 @@ initialise_ring_diff_arrays() const
...
-      if (fabs(ax_pos_num_offset[segment_num] -
-	       ((num_rings-1) - 2*m_offset[segment_num]/ring_spacing)) > 1E-4)
-       {
-	 error("ProjDataInfoCylindrical: in segment %d, the axial positions\n"
-	       "do not correspond to the usual locations between physical rings.\n"
-	       "This is suspicious and can make things go wrong in STIR, so I abort.\n"
-	       "Check the number of axial positions in this segment.",
-	       segment_num);
-       }
+      if (get_scanner_sptr()->get_scanner_geometry()=="Cylindrical")
```
KrisThielemans, line 207: not necessary, as we never get here in the non-cylindrical case

KrisThielemans: I'm fixing this

KrisThielemans:
My suggestions cause the tests to fail with
```
ERROR: ProjDataInfoCylindrical::get_ring_pair_for_segment_axial_pos_num does not work for this type of sampled data
```
which is thrown in `get_ring_pair_for_segment_axial_pos_num` https://github.com/UCL/STIR/blob/aaa0f8e828cd60e7c6dc541696d62de0cf07e6a7/src/buildblock/ProjDataInfoCylindrical.cxx#L338-L339

So now I'm confused. This code needs `segment_axial_pos_to_ring1_plus_ring2` which previously we did initialise, but potentially wrongly, as it used `m_offset` etc, which does not necessarily make sense for the block scanner... There seems to be something deeper that needs fixing.

danieldeidda:
I am kind of lost here :D

src/buildblock/ProjDataInfoCylindrical.cxx
KrisThielemans: we will leave this for later

KrisThielemans:
We will fix the initialisation of the ring diff arrays in a future PR.

Issue #1008, opened 2022-03-24
Uninitialised value in ROOT processing  causes a segfault in LmToProjData::process_data()


jvavrek:
I'm encountering some segfaults when trying to read a ROOT file into Python as per #1005. A small fraction of the time, my code produces the following segfault output:
```bash
$ python tof_pet_recon.py 
Scanner parameters:= 
Scanner type := myScanner
Number of rings                          := 48
Number of detectors per ring             := 666
Inner ring diameter (cm)                 := 87
Average depth of interaction (cm)        := 0.84
Distance between rings (cm)              := 0.85
Default bin size (cm)                    := 0.197
View offset (degrees)                    := 0
Maximum number of non-arc-corrected bins := 672
Default number of arc-corrected bins     := 336
Number of blocks per bucket in transaxial direction         := 1
Number of blocks per bucket in axial direction              := 1
Number of crystals per block in axial direction             := 1
Number of crystals per block in transaxial direction        := 1
Number of detector layers                                   := 1
Number of crystals per singles unit in axial direction      := 48
Number of crystals per singles unit in transaxial direction := 1
Scanner geometry (BlocksOnCylindrical/Cylindrical/Generic)  := Cylindrical
Scanner orientation (X or Y)                                := X
end scanner parameters:=

Constructing proj_data_info
Done constructing proj_data_info
root_header.name =  tmp_header.hroot

INFO: InputStreamFromROOTFile: Processing data with the following event type inclusions:
  Unscattered from same eventID:        1
  Unscattered from different eventIDs:  1
  Scattered from same eventID:          1
  Scattered different eventIDs:         1
/bin/bash: /home/jvavrek/anaconda3/lib/libtinfo.so.6: no version information available (required by /bin/bash)

WARNING: CListModeDataROOT: Trying to figure out the scanner geometry from the information given in the ROOT header file.

INFO: CListModeDataROOT: used ROOT file /home/jvavrek/5Mevents.root
Processing listmode... set_up:
Processing listmode... process_data:

Processing time frame 1

 *** Break *** segmentation violation



===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================
#0  0x00007f39f8cc7dba in __GI___wait4 (pid=2588745, stat_loc=stat_loc
entry=0x7ffc9e4e38e8, options=options
entry=0, usage=usage
entry=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:27
[...]

PR #1012, opened 2022-03-25
Set default value for half_block = 0 when reading ROOT files

robbietuk:
Fix for #1008?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,6 +141,7 @@ endif()
 # Legacy options
 option(STIR_PROJECTORS_AS_V3 "Enable STIR version 3 legacy code" OFF)
 option(STIR_LEGACY_IGNORE_VIEW_OFFSET "Ignore using scanner view-offset (or intrinsic azimuthal tilt), as in STIR 4 or earlier" OFF)
+option(STIR_ROOT_ROTATION_AS_V4 "Ignore the changes to the ROOT rotation introduced in STIR version 5" OFF)
 
 # NiftyPET projector
 if(NOT DISABLE_NiftyPET_PROJECTOR OR NOT DISABLE_Parallelproj_PROJECTOR)
--- a/documentation/release_5.0.htm
+++ b/documentation/release_5.0.htm
@@ -340,7 +340,11 @@ from the above):</H2>
       You will get a run-time exception if you forgot to do this.
     </li>
     <li>
-        Rotational changes to STIR's interface with ROOT files as subtraction of a <code>half_block</code> has been remove the from the determination of the crystalIDs.
+        Rotational changes to STIR's interface to ROOT files, <a href=https://github.com/UCL/STIR/pull/1012>PR #1012</a>. In previous versions (roughly) half a bucket,
+        <code>(module_repeater_y * submodule_repeater_y * crystal_repeater_y) /2 - 1)</code>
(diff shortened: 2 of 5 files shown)


KrisThielemans:
I suggest to have an option for compatibility with STIR 4, setting it to the old version. We should have given people notice about this change, so this way they can at least keep the same results. Obviously I'd default it to off (i.e. `half_block=0`). Not sure how to call it. feel free to suggest something

robbietuk:
Yes. However, it still needs a test because now I am concerned about rotations.

KrisThielemans:
only now? 😄

CMakeLists.txt
```diff
@@ -141,6 +141,7 @@ endif()
 # Legacy options
 option(STIR_PROJECTORS_AS_V3 "Enable STIR version 3 legacy code" OFF)
 option(STIR_LEGACY_IGNORE_VIEW_OFFSET "Ignore using scanner view-offset (or intrinsic azimuthal tilt), as in STIR 4 or earlier" OFF)
+option(STIR_ROOT_AS_V4 "Ignore the changes to the ROOT rotation introduced in STIR version 5" OFF)
```
robbietuk, line 144: Could do `STIR_ROOT_ROTATION_AS_V4`?

src/IO/InputStreamFromROOTFileForCylindricalPET.cxx
```diff
@@ -146,6 +148,13 @@ InputStreamFromROOTFileForCylindricalPET::set_defaults()
     module_repeater_y = -1;
     module_repeater_z = -1;
     rsector_repeater = -1;
+#if STIR_ROOT_AS_V4
+    half_block = module_repeater_y * submodule_repeater_y * crystal_repeater_y / 2  - 1;
+    if (half_block < 0 )
+      half_block = 0;
+#else
+    half_block = 0;
+#endif
```
robbietuk, line 157: Keeping `half_block` for now. Can be removed when `STIR_ROOT_AS_V4=false` later.

robbietuk (commented):
Forgot to commit before a forcepush... Apparently that closes the PR, who knew.

CMakeLists.txt
KrisThielemans: Ok for me

robbietuk:
Changed `STIR_ROOT_AS_V4` to `STIR_ROOT_ROTATION_AS_V4` as it will be more useful later. I think this is ready to go and we can test later? It will likely mean that a number of older GATE geometries are now rotated in version5 but we are going to fix that later?

documentation/release_5.0.htm
```diff
@@ -339,6 +339,11 @@ <h3>Backward incompatibities</h3>
       scanners, this will fill in the detector maps etc.
       You will get a run-time exception if you forgot to do this.
     </li>
+    <li>
+        Rotational changes to STIR's interface with ROOT files as subtraction of a <code>half_block</code> has been remove the from the determination of the crystalIDs.
```
KrisThielemans, line 343:
```suggestion
        Rotational changes to STIR's interface to ROOT files. In previous versions (roughly) half a bucket,
        <code>(module_repeater_y * submodule_repeater_y * crystal_repeater_y) /2 - 1)</code>
        in the <tt>Cylindrical</tt> case, was subtracted from the crystalIDs, in an attempt to rotate
        the scanner. However, this was incompatible with the STIR code to find normalisation factors,
        and could be surprising for odd numbers etc.
```

PR #1016, opened 2022-03-30
Root/scanner rotation changes

robbietuk:
Fix for #938 and extension of #991

- [x] enable `test_view_offset_root` test
- [x] Removes the addition of `quarter_of_detectors` from deterctor position when reading of ROOT files in STIR
- [x] Rotate the Gate CylindricalPET scanner by -90 degrees 
- [x] Remove interfile parameter `offset (num of detectors)` from all ROOT headers
- [x] Update users guide
- [x] Reduce number of test
- [x] Upload data to Zenodo (see https://zenodo.org/record/6410309)
- [ ] Automate download of data in CTests


src/include/stir/IO/InputStreamFromROOTFile.inl
```diff
@@ -135,11 +135,13 @@ InputStreamFromROOTFile::set_exclude_random_events(bool val)
     exclude_randoms = val;
 }
 
+#if STIR_ROOT_ROTATION_AS_V4
 void
 InputStreamFromROOTFile::set_detectors_offset(int val)
 {
     offset_dets = val;
 }
+#endif
```
robbietuk, line 144: Nesting the method inside of `#if STIR_ROOT_ROTATION_AS_V4` is okay?

src/IO/InputStreamFromROOTFile.cxx
```diff
@@ -86,7 +86,9 @@ InputStreamFromROOTFile::initialise_keymap()
     this->parser.add_key("exclude unscattered events", &this->exclude_unscattered);
     this->parser.add_key("exclude random events", &this->exclude_randoms);
     this->parser.add_key("check energy window information", &this->check_energy_window_information);
+#if STIR_ROOT_ROTATION_AS_V4
     this->parser.add_key("offset (num of detectors)", &this->offset_dets);
+#endif
```
robbietuk, line 91: Looks a little ugly...?

robbietuk (commented):
The removal of `offset (num of detectors)` and `quarter_of_detectors` works in the tests now with the rotations in GATE geometry. All our GATE geometries should be updated (thinking STIR-GATE-Connection).

src/IO/InputStreamFromROOTFile.cxx
KrisThielemans: why?

src/include/stir/IO/InputStreamFromROOTFile.inl
KrisThielemans: sure

examples/ROOT_files/ROOT_STIR_consistency/Gate_macros/geometry.mac
```diff
@@ -31,6 +31,12 @@
 /gate/cylindricalPET/geometry/setRmin         40.5 cm
 /gate/cylindricalPET/geometry/setHeight       15.716 cm
 /gate/cylindricalPET/setMaterial              Air
+
+# Allign STIR and Gate geometries first cystals on y axis 
+# by rotate the cylindricalPET -90 degrees around z
```
KrisThielemans, line 36: There's some spelling problems here, but it is also not quite accurate. I suggest something like this
```suggestion
# Rotate Gate geometry such that the first module will be on "top" of the scanner (i.e. x=0, y>0)
# by rotating the cylindricalPET -90 degrees around z.
# Note that STIR still needs view-offset according to half the module size.
```

documentation/release_5.0.htm
```diff
@@ -340,12 +340,24 @@ <h3>Backward incompatibities</h3>
...
+            <li>Similarly, the detector numbers of each ROOT event were offset by <code>quarter_of_detectors</code>.
+            This was because of the assumption that GATE the 0th crystal is on the on the x axis (STIR's is on y).
+            However, the usage of <code>quarter_of_detectors</code> requires the number of scanner
+            crystals/detectors to be factorisable by 4, which may not always be the case.
+            This addition of <code>quarter_of_detectors</code> has been removed and GATE scanners are expected
+            to be rotated by -90 degrees so the 0th crystal is on the on the y axis. </li>
+            <li>The <code>offset (num of detectors)</code> parameter has been removed from ROOT header files and
+            and all underlying methodology removed.</li>
+        </ul>
+        Rotations should use the new <code>View offset (degrees)</code> support described above.
```
KrisThielemans, line 360:
```suggestion
        Pre STIR 5..0, STIR renumbered crystals obtained from the ROOT file to rotate the reconstructed image.
        This was because GATE macros that we used placed the first bucket (i.e. Gate module) on the <i>right</i>
        of the scanner, while STIR by default places the 0th crystal is on the <i>top</i>.
        This was done by adding various offsets to the crystal ID. However, this was incompatible with the 
        STIR code to find normalisation factors, which assumes that a bucket starts at crystal 0.
        Moreover, the offsets used integer division which could be surprising for odd numbers.
        Specific offsets that have been removed:
        <ul>
            <li> (roughly) half a bucket,
            <code>(module_repeater_y * submodule_repeater_y * crystal_repeater_y) /2 - 1)</code>
            in the <tt>Cylindrical</tt> case, was subtracted from the crystalIDs</li>
            <li>(roughly) 90 degrees, <code>quarter_of_detectors=get_num_detectors_per_ring()/4</code>.</li>
            <li>An optional<code>offset (num of detectors)</code> parameter in the `.hroot` and
            and all underlying methodology has been removed.</li>
        </ul>
        Rotations should use the new <code>View offset (degrees)</code> support described above.
        See also the STIR User's guide.
```

KrisThielemans (changes requested):
Currently using 
```
#if STIR_ROOT_ROTATION_AS_V4
```
this is non-standard.  Replace with using `#ifdef` as usual. This obviously means that it shouldn't be set by default (I guess you currently set it to 0).

I think the user's guide need some more tweaks but first want to understand Gate macro options better (see comment in the `geometry.mac`).

You've enabled the test, but it cannot work yet as no data?

KrisThielemans:
Forgot to write my question on the geometry.

You've used rotation of the GATE scanner
https://github.com/UCL/STIR/blob/cccb606b67d4e25d44a5c6e24d929d97f6bed255/examples/ROOT_files/ROOT_STIR_consistency/Gate_macros/geometry.mac#L38
This is nice, but I do wonder if an alternative solution isn't to modify the placement
https://github.com/UCL/STIR/blob/cccb606b67d4e25d44a5c6e24d929d97f6bed255/examples/ROOT_files/ROOT_STIR_consistency/Gate_macros/geometry.mac#L50
I haven't checked of course.

robbietuk:
@francescaleek If you fancy taking a look that would be useful 😄

KrisThielemans:
I've updated the documentation a bit more. @robbietuk please check if you're happy with it.

This seems good to merge. The test is enabled though, which means we'd have to download the data, otherwise ctest/GitHub Actions will fail.

zenodo is one option. Adding it to git is another but it'd have to be quite small (preferably less than 2MB). I guess you could see if you `tar/gzip -9/bzip` it, if size reduces. Then you'd have to uncompress in CMake, see https://stackoverflow.com/a/7052049/3467846 for the opposite problem (creating one).. If you choose that option, make sure it's classified as binary via `.gitattributes`.

robbietuk:
The data is uploaded to https://zenodo.org/record/6410309 I am working on CMAKE downloads.

robbietuk:
Okay, I am stuck because of my lack of Cmake knowledge. I think the tools are all here but I cant get `ExternalData` CTests to work... 😢

KrisThielemans:
I don't know about the `ExternalData` stuff. sounds interesting, but I have no time to look at it.

In SIRF, we use something which is simply, but also a bit of a pain in that it downloads the data even if you don't test. Still, I think putting that into place should be quite easy. We can presumably replace it later if there's smarter stuff around.

have a look at
https://github.com/SyneRBI/SIRF/blob/843eb25760ae8e1e15bc816748d8b80d9c34f7d2/src/Synergistic/tests/CMakeLists.txt#L19

I guess we might have to unzip, but CMake can do that as well https://cmake.org/cmake/help/latest/command/file.html#archive-extract but thi sseems to require CMAke 3.18, so I guess better to download one by one if that's possible.

src/recon_test/CMakeLists.txt
```diff
@@ -32,13 +41,83 @@ set(${dir_INVOLVED_TEST_EXE_SOURCES}
...
+        get_zenodo_file("${ROOT_STIR_consistency_data_dir}/root_data_test8.root"
+                        "${ROOT_DATA_URL}/root_data_test8.root?download=1"
+                        "24150ce5d655212dc975989b3dbe4b33")
+        get_zenodo_file("${ROOT_STIR_consistency_data_dir}/root_header_test8.hroot"
+                        "${ROOT_DATA_URL}/root_header_test8.hroot?download=1"
+                        "a821a3eac0000544f000cd5c74150023")
+    endif()
+
+    # Add test if the data is downloaded (only checking root_header_test1.hroot)
+    if(EXISTS "${ROOT_STIR_consistency_data_dir}/root_header_test1.hroot")
```
KrisThielemans, line 112: are you sure this works ok? i.e. is the `if(EXISTS)` evaluated *after* the download or before? I guess you'll need to check if this test is run once you deleted the data.

KrisThielemans:
we're not currently installing `ROOT` in GitHub Actions (nor Appveyor), so I guess this is never tested. Sadly, installing `ROOT` might not be so trivial as there's no `APT` package.

if you tell me that this works on your system, we could merge this, and try to get CI going in a separate PR.

src/recon_test/CMakeLists.txt
robbietuk: Configuring downloads the data and then checks if `root_header_test1.hroot` exists. Then it adds the test to the list of tests. 

If you download, configure and build, then delete the data, the test will be run and fail.

robbietuk:
I forgot there were no GHA tests for root, at least it works without ROOT 😄 

It works nicely on my machine.

KrisThielemans (approved):

KrisThielemans:
worked on mine!

PR #1019, opened 2022-04-12
[SWIG] split stir.i and change order of declarations

### LINKED ISSUE #1018: BinNormalisation is not correctly configured for python
The following exists in the `stir.i` file
https://github.com/UCL/STIR/blob/b8956388b5110b47667c2ea499321ae31748d082/src/swig/stir.i#L1932-L1952
but the `stir.BinNormalisation` class is not exposed.

The following code returns an object where `norm` is type `<Swig Object of type 'BinNormalisation *' at 0x10f8239b0>` but is uninteractable.
```
import stir
poissonobj = stir.PoissonLogLikelihoodWithLinearModelForMeanAndProjData3DFloat()
norm = poissonobj.get_normalisation()
```

KrisThielemans:
The `GeneralisedObjectiveFunction` hierarchy was defined before normalisation and projectors, leading to problems setting/getting corresponding members.

As this was very messy, I've split `stir.i` in several files. It still leads to one huge `*wrap.cxx` but at least reading `stir.i` is now a bit easier. (still work to do).

This commit also fixes a line in SWIG for `ProjData::get_subset`, but that function is still ignored, so it doesn't change the actual Python interface.

Fixes #1018

### DIFF
--- a/documentation/release_5.0.htm
+++ b/documentation/release_5.0.htm
@@ -73,16 +73,22 @@ improvements to the documentation.
   Markus Jehl (Positrigo), Gefei Chen (Univ of Macao)
 </p>
 
-    <h2>Patch release info</h2>
+<h2>Patch release info</h2>
+<ul>
+  <li> 5.0.0 released 23/03/2022</li>
+  <li> 5.0.1 released ???
     <ul>
-      <li> 5.0.0 released 23/03/2022</li>
-      <!--
-      <li> 4.0.1 released 28/04/2020
-        <ul>
-          <li><a href=https://github.com/UCL/STIR/pull/513>PR 513</a>  (suppress warnings with clang)</li>
-          </ul>
-          -->
+      <li><a href=https://github.com/UCL/STIR/pull/1019>PR 1019</a>: fixes for SWIG and hence Python interface.
+      </li>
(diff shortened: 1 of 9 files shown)


KrisThielemans:
@markus-jehl can you please check this? Also, because of the reordering, the FBP problem tha tyou had might be resolved, so you could try to enable the "instantiations" (now in `stir_reconstructors.i`)

markus-jehl:
Sure, I'll have a look if this still works with my reconstruction pipeline and if that solves the FBP problem! Will probably be next week, though.

robbietuk (approved):
I like the simplification of `stir.i`. I was able to compile this PR and I now have access to `stir.BinNormalisationFromProjData()`, which addresses https://github.com/UCL/STIR/issues/1018

markus-jehl:
I just checked it and it all works fine for me as well. Defining the FBP templates no longer crashes everything, so that's a plus in my view :-) I still can't define the FBP templates in a way that exposes their member functions, but this is not just about working out how this needs to be done in SWIG.

KrisThielemans:
thanks. can you create a new issue for the FBP problem (maybe there's one already..)

PR #1021, opened 2022-04-13
 fixing sampling calculation for ProjDataInfoGeneric

danieldeidda:


KrisThielemans:
thanks. Looks like we need to add a test for these sampling functions really (like `get_m(nextbin)-get_m(bin))`. Feel like having a go or postpone for later?

PR #1023, opened 2022-04-14
Download Zenodo test data off by default

### LINKED ISSUE #1022: STIR configuration is slow because of downloading `test_view_offset_root` data
With #991 and #1016 adding `test_view_offset_root`, configuration time is now very slow and can fail without an internet connection when `DISABLE_CERN_ROOT=OFF`. 

Adding a cmake variable e.g., `DOWNLOAD_TEST_DATA`, with a default as `OFF`, and using this as a condition for running the test would help. Most users will not need to run this test, only GHA really do.

robbietuk:
Fixes #1022
Running to test GHA


robbietuk:
Getting hash errors on the [configuration Zenodo download](https://github.com/UCL/STIR/runs/6018444711?check_suite_focus=true#step:6:150). Otherwise the change worked.

```
-- Downloading ROOT consistency test data [~3MB] (if not already done)... 
-- Downloading https://zenodo.org/record/6410309/files/root_data_test1.root?download=1 ...
CMake Error at src/recon_test/CMakeLists.txt:14 (file):
-- Downloading https://zenodo.org/record/6410309/files/root_header_test1.hroot?download=1 ...
  file DOWNLOAD HASH mismatch

    for file: [/home/runner/work/STIR/STIR/examples/ROOT_files/ROOT_STIR_consistency/pretest_output/root_data_test1.root]
      expected hash: [0195985779dcd0cc238ba1cf73ec734b]
        actual hash: [d41d8cd98f00b204e9800998ecf8427e]
             status: [22;"HTTP response code said error"]

Call Stack (most recent call first):
  src/recon_test/CMakeLists.txt:56 (get_zenodo_file)
```

KrisThielemans:
Download of the data should really only happen in the first configuration. After that there should be essentially no penalty.

Of course, it is fine to prevent the download for most people. However, what when will then happen with the test? (I didn't check)

robbietuk:
The data is now downloaded `if (HAVE_CERN_ROOT)` AND `if (DOWNLOAD_ZENODO_TEST_DATA)`

The test is added to the test suite
`if (HAVE_CERN_ROOT)` AND `if(EXISTS "${ROOT_STIR_consistency_data_dir}/root_header_test1.hroot")`

This PR will set `DOWNLOAD_ZENODO_TEST_DATA=OFF` by default that means the test will not be added to the test suite.

PR #1026, opened 2022-04-19
Radionuclide code improvements

KrisThielemans:
I created a test for Radionuclide code and then had to fix various things to get it to work/understand.

The test fails locally because the JSON file has `PET` and `nucmed` but `ImagingModality` uses `PT` and `NM` as strings. We could fix this in the code or the JSON file. Not sure what the best option is. @danieldeidda ?

In addition, I think this will fail on Github Actions etc as the test code is run before install, so it won't find the JSON files.

### DIFF
--- a/recon_test_pack/run_test_zoom_image.sh
+++ b/recon_test_pack/run_test_zoom_image.sh
@@ -168,7 +168,7 @@ if [ $? -ne 0 ]; then
   exit 1
 fi
 
-org_sum=`list_projdata_info --sum my_prompts.hs | awk -F: '{ print $2}'`
+org_sum=`list_projdata_info --sum my_prompts.hs | awk -F: '/sum/{ print $2}'`
 
 ./simulate_data.sh my_zoom_test4.hv my_atten_image.hv ${template_sino} 0
 if [ $? -ne 0 ]; then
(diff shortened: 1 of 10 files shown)


KrisThielemans:
It seems that I've been a bit too enthusiastic with adding `error` statements as we now have a lot of failures with
```
ERROR: RadioNuclideDB::get_radionuclide: unknown modality
```
even for the tests where `RadioNuclide` isn't used. Not sure where/why we're calling this but apparently more than I hoped.

Appveyor (i.e. no JSON) is failing with https://ci.appveyor.com/project/KrisThielemans/stir/builds/43273181/job/r0j359avdtqrl8iw#L133
```
c:\projects\stir\src\buildblock\radionuclidedb.cxx(196): error C4716: 'stir::RadionuclideDB::get_radionuclide_from_json': must return a value [C:\projects\stir\build\src\buildblock\buildblock.vcxproj]
```
and also has warnings about unused variables.

danieldeidda (approved):

KrisThielemans:
@danieldeidda I've changed the database format rather drastically. I think this makes more sense and is more extendible. Obviously that needed code changes. Could you have a look please? (I've added a test on Y90, so that seems to work ok).

danieldeidda (commented):
This looks good I guess I can try this to double check

KrisThielemans:
@danieldeidda let me know if you tested it. I'd like to merge this tomorrow or the day after.

KrisThielemans:
Merging as tests work, @danieldeidda please create a new issue etc if you discover any problems.

PR #1028, opened 2022-04-26
Support for input from PENNPET Explorer

NikEfth:
Successful unlisting of UPENN acquired data
Support for the 5-ring variant (with gaps)

PR #1029, opened 2022-04-26
Speed up lm recon

NikEfth:
Distribute sensitivity calculation


KrisThielemans:
this has all the UPENN stuff in here as well. I won't review it. Possibly it needs to be rebased after #1028  is merged. Or just create a new PR with only the sensitivity?

Issue #1036, opened 2022-05-16
GHA: MacOS ctest failures with OpenMP


KrisThielemans:
Our MacOS test job now fails consistently, `test_PoissonLogLikelihoodWithLinearModelForMeanAndProjData`, `test_blocks_on_cylindrical_projectors`, `test_FBP2D`, `test_DataSymmetriesForBins_PET_CartesianGrid` etc now all fail really soon with
```
Subprocess aborted***Exception:   0.62 sec
```
see e.g. https://github.com/UCL/STIR/runs/6446997086?check_suite_focus=true#step:8:19

However, this occurs also after simply re-running the same job, e.g. the above job is from a re-run of the 5.0.1 commit, which did work originally.

so, this is no a STIR code change, but something else.

KrisThielemans:
I've used `tmate` to ssh into the process. It turns out that this is due to `STIR_CONFIG_DIR` problems. After setting it to the location in the `src/config` dir, it works fine.

I thought this was addressed in #1026 and in particular 5af1f0defc762f01c47908eb3b7876100e6e416f5af1f0defc762f01c47908eb3b7876100e6e416f (and indeed it was).

I find that on my Windows system, I have the same problem. It's running CMake 3.22.3. I think MacOS GHA is now running CMake 3.22.1. 

a CMake problem?

KrisThielemans:
I've tested this in a tiny repo, and there everything works
https://github.com/KrisThielemans/GHA-tests/commit/c77bc32deabc6360cc1d83a4d2fdaa252b930afe

KrisThielemans:
turns out I forgot to  add `STIR_CONFIG_DIR` to the "manual" tests, https://github.com/UCL/STIR/blob/faf0f70fd8bcaeea9d9695bbde8d52afcc9e9906/src/recon_test/CMakeLists.txt#L136
this is true for other tests as well.
see e.g. https://github.com/UCL/STIR/runs/6478973644?check_suite_focus=true#step:7:589

PR in progress

KrisThielemans:
While the above problem with `STIR_CONFIG_DIR` is true, it is unrelated to the crashes.

It also turns out that crashes do not occur in debug mode, see the `test_OSMAPOSL` run passing here https://github.com/KrisThielemans/STIR-1/runs/6484447529?check_suite_focus=true#step:9:105

[...]

PR #1040, opened 2022-05-21
Small fixes to listmode recon

NikEfth:
* Set has_add from par file
* Use cache from par file


KrisThielemans:
```
/home/runner/work/STIR/STIR/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx:352:7: error: use of undeclared identifier 'has_add'
      has_add = true;
```

KrisThielemans:
First 2 jobs fail with
```
=== Reconstruct listmode data without cache
Error: Process completed with exit code 134.
```
The MacOS job will always fail #1036. I guess I should disable it such that we can get at the log files (it doesn't seem to store them if we cancel the workflow).

NikEfth:
I am trying to find what error 134 is, but it doesn't look common. 
I have already tried this PR in 2 machines and looks good on myside.

KrisThielemans:
I ran this on mine and see several problems in OSMAPOSL_test_lm_1.log
```
WARNING: KeyParser warning: unrecognized keyword: projector pair type
WARNING: KeyParser warning: unrecognized keyword: projector pair using matrix parameters
WARNING: KeyParser warning: unrecognized keyword: end projector pair using matrix parameters
```
indicate problems with the .par file which should be fixed, but doesn't cause failures. but then
```
WARNING: Please set the max cache size for the listmode file
INFO: Reading cache from disk...
ERROR: FilePath: Cannot access /home/sirfuser/devel/STIR2/recon_test_pack/my_CACHE00.bin.
terminate called after throwing an instance of 'std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >'
Aborted (core dumped)
```
so I guess you ran with an existing cache file and therefore didn't notice this error.

another problem in the test script is that 3 calls to `compare_image` write to the same log file

NikEfth:
I am looking into this. Indeed my tests were with an already calculated cache. But I am also preparing the exercises, and that works.

NikEfth:
Hopefully, this will do the trick.

KrisThielemans:
looks like it. great.

can you please fix the test script and par file as well (delete the myCACHE.bin at the start for instance)

KrisThielemans:
thanks. I'll merge, but please do another PR to fix the .par file

KrisThielemans:
Merged via fb971fe51939c3c91ed7e211e9c714f6c97e10a1 (squashed listmode-related commits). I've kept the GHA commit separately.

PR #1042, opened 2022-05-22
fix leaks when using SPECTUB matrix and openmp

KrisThielemans:
The SPECTUB matrix set_up() currently sets num_threads to 1 under certain conditions. However, the (base class) projectors create temp data depending on number of threads. Therefore, call set_up of the matrix before the one from the projector.

Might fix leak observed in https://github.com/SyneRBI/SIRF/issues/1101#issuecomment-1116006149

### DIFF
--- a/src/recon_buildblock/BackProjectorByBinUsingProjMatrixByBin.cxx
+++ b/src/recon_buildblock/BackProjectorByBinUsingProjMatrixByBin.cxx
@@ -100,8 +100,8 @@ set_up(const shared_ptr<const ProjDataInfo>& proj_data_info_ptr,
        const shared_ptr<const DiscretisedDensity<3,float> >& image_info_ptr)
 
 {
-  BackProjectorByBin::set_up(proj_data_info_ptr, image_info_ptr);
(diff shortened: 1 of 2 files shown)


KrisThielemans:
@samdporter maybe you could check?

danieldeidda:
> 

just to check. doesn't the matrix depend on that projector?

KrisThielemans:
no. The base class doesn't know yet what projector it's going to use.

Arguably it's safer anyway to first set-up the matrix, and then the projector. In practice it shouldn't have made any difference, but in the current case, the matrix will change the number of threads, which the base class has just used

KrisThielemans:
main thing is to run this on some data with a high number of core system. (ideally with view-caching on and off)

danieldeidda:
with view caching off it runs OK but it got killed with view caching on.

danieldeidda:
If I use 2D PSF it runs OK

KrisThielemans:
> with view caching off it runs OK but it got killed with view caching on.

more than likely ran out of memory. could you just check with a system monitor?

Issue #1043, opened 2022-05-25
BlocksOnCylindrical ForwardProjection results in flipped first view


markus-jehl:
When forward projecting an image with a BPET scanner and BlocksOnCylindrical, the first view appears to be flipped (or rotated by 180°).

Additionally, the last slice is homogeneous 1 instead of actual projection data, but this could be unrelated.

danieldeidda:
a fix is coming. There was a condition moving lors around in the get_phi() function

KrisThielemans:
fixed in 40ede282ad39f87faaef015c92901db25796aa6c (i.e. #1045 merged back to `master`)

markus-jehl:
Brilliant, thank you!

danieldeidda:
should be OK

KrisThielemans:
final fix in #1050

PR #1045, opened 2022-05-25
fixing issue #1043

danieldeidda:


KrisThielemans:
an idea why our tests didn't catch this?

danieldeidda:
> an idea why our tests didn't catch this?

Yes, it was probably failing but we introduced some condition for phi with different checks and tolerance

PR #1052, opened 2022-06-22
 make compute_kernelised_image() pubblic for SIRF

danieldeidda:


src/include/stir/KOSMAPOSL/KOSMAPOSLReconstruction.h
```diff
@@ -218,6 +218,11 @@ class KOSMAPOSLReconstruction:
   virtual void initialise_keymap();
   virtual bool post_processing();
 
+
+  /*! choose between compact_compute_kernelised_image() and  full_compute_kernelised_image()*/
```
KrisThielemans, line 222: ok to move, but this doc-string makes no sense to me

danieldeidda: I agree

PR #1061, opened 2022-07-26
interpolate_axial: fixing rounding issue causing out of range

danieldeidda:


KrisThielemans:
I'm afraid I don't understand. Without `round` it is going to truncate (i.e. equivalent of `floor`). Is that really what you want?

With `round`, it is doing nearest neighbour interpolation I guess. With `floor`, it seems to me you could have half a plane offset. An alternative would be to use linear interpolation or overlap interpolation.

In any case, I think boundary conditions need to be handled separately.

KrisThielemans:
In fact, the use of offsets will be broken for blocksOnCylindrical with gaps in axial direction.

To implement nearest neighbour interpolation, I guess I'd do something like
1. construct `VectorWithOffset` with `m_in`
2. construct `VectorWithOffset` with `m_out`
3. for out_axial_pos: find closest `m_in`, and store corresponding sinogram

danieldeidda:
> I'm afraid I don't understand. Without `round` it is going to truncate (i.e. equivalent of `floor`). Is that really what you want?
> 

Yes that was what I wanted to do becaues the relative_pos needs to be within 0 to num_rings-1 and round would give me an index of 1 everytime the float part is >= .5;  But I was probably being naive, though I am not sure why the use of offset would be broken

danieldeidda:
> In fact, the use of offsets will be broken for blocksOnCylindrical with gaps in axial direction.
> 
> To implement nearest neighbour interpolation, I guess I'd do something like
> 
> 1. construct `VectorWithOffset` with `m_in`
> 2. construct `VectorWithOffset` with `m_out`
> 3. for out_axial_pos: find closest `m_in`, and store corresponding sinogram

OK but to find the the axial position we still need to apply the following no?
relative_pos=abs(((m -m_offset)/ m_sampling));

KrisThielemans:
> to find the the axial position we still need to apply the following no?
relative_pos=abs(((m -m_offset)/ m_sampling));

no. that would defeat the purpose. why not use `get_m(Bin(....,axial_pos,...))` to create the table I mentioned above?

danieldeidda:
yeah exactly.  now that i think about it the index of the m vector would actually correspond to the axial_position right?

KrisThielemans:
indeed!

src/buildblock/interpolate_axial_position.cxx
```diff
@@ -59,22 +59,70 @@ interpolate_axial_position(ProjData& proj_data_out,
...
+  {
+      m_in.at(segment).resize(proj_data_in.get_min_axial_pos_num(segment),proj_data_in.get_max_axial_pos_num(segment));
+
+#ifdef STIR_OPENMP
+#  if _OPENMP <201107
+                      #pragma omp parallel for
+#  else
+                      #pragma omp parallel for schedule(dynamic)
+#  endif
+#endif
```
KrisThielemans, line 76:
```suggestion
```
simpler and probably faster

src/buildblock/interpolate_axial_position.cxx
```diff
@@ -59,22 +59,70 @@ interpolate_axial_position(ProjData& proj_data_out,
...
+  {
+      m_out.at(segment).resize(proj_data_out.get_min_axial_pos_num(segment),proj_data_out.get_max_axial_pos_num(segment));
+
+#ifdef STIR_OPENMP
+#  if _OPENMP <201107
+                      #pragma omp parallel for
+#  else
+                      #pragma omp parallel for schedule(dynamic)
+#  endif
+#endif
```
KrisThielemans, line 94:
```suggestion
```
simpler and probably faster

src/buildblock/interpolate_axial_position.cxx
```diff
@@ -59,22 +59,70 @@ interpolate_axial_position(ProjData& proj_data_out,
...
-          else
-              relative_pos=abs(round((m -m_offset_in)/ in_m_sampling));
+          else{
+#ifdef STIR_OPENMP
+#  if _OPENMP <201107
+                      #pragma omp parallel for
+#  else
+                      #pragma omp parallel for schedule(dynamic)
+#  endif
+#endif
```
KrisThielemans, line 118:
```suggestion
```
simpler and probably faster

KrisThielemans (changes requested):

KrisThielemans (changes requested):
great. let's rename `relative_pos` to `sino_in` then, as that seems more logical.
Possibly also update the release notes?

danieldeidda:
mh we already have sino_in which is the input sinogram while relative_pos is an index:
sino_in=proj_data_in.get_sinogram(relative_pos,0);

KrisThielemans:
oops. `axial_pos_in`?

danieldeidda:
> but they seem to have gone now, which would be good 😄

ah OK glad they are gone 😆 
axial_pos_in sonds good

KrisThielemans:
great. should be good. Please add something to documentation/release_5.1.htm (with `[ci skip]`).

@emikhay this should be ready for testing now

KrisThielemans:
first 2 Appveyor jobs for previous commit were green. I've cancelled that job as it takes 2 hours, and just let the last one run.

KrisThielemans:
Great. I think we can merge this now as we know the current `master` is buggy anyway. @emikhay can then test from `master`.

@danieldeidda ok to squash merge?

danieldeidda:
yes fine for me!

KrisThielemans (approved):

PR #1065, opened 2022-08-10
Update run_scatter_tests.sh

mastergari:
Typo in path for scatter_pardir

### DIFF
--- a/recon_test_pack/run_scatter_tests.sh
+++ b/recon_test_pack/run_scatter_tests.sh
@@ -34,7 +34,7 @@ export LC_ALL
 
 # find location of scatter parameter files
 script_name=`type -p $0`
-scatter_pardir=`dirname $script_name`/../examples/samples/scatter_estimation_par_files
+scatter_pardir=`dirname $script_name`../examples/samples/scatter_estimation_par_files
 echo "Using scatter parameter files from $scatter_pardir"
 
 ./simulate_PET_data_for_tests.sh
--- a/recon_test_pack/run_scatter_tests.sh
+++ b/recon_test_pack/run_scatter_tests.sh
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-# (note: need bash for "type -p" below)
 # A script to check to see if scatter simulation and estimation give the expected result.
 #
 #  Copyright (C) 2011, Kris Thielemans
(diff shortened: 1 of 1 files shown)


KrisThielemans:
Are you sure? `dirname` normally doesn't append the trailing slash, so we need to add it here. What was your test case?

KrisThielemans:
OK, this is a `bash` script, as indicated in the first line. (`sh` doesn't support `type -p`). therefore either do `bash run_scatter_tests.sh` or (recommended) `./run_scatter_tests.sh`

mastergari:
Okay, then should the README.txt be updated? It instructs users to call the scatter test as `sh run_scatter_tests.sh`. Calling `bash run_scatter_tests.sh` yields the same `dirname` issue, but `./run_scatter_tests.sh` is successful.

KrisThielemans:
ok. This took me a while to understand. It turns out that when running `bash run_scatter_tests.sh`, 
https://github.com/UCL/STIR/blob/4d087f3e7cf3c6d2d95b87f64e9e68bf9f1b186e/recon_test_pack/run_scatter_tests.sh#L36
returns an empty string, while when running either `./run_scatter_tests.sh` or `bash ./run_scatter_tests.sh` it returns `./run_scatter_tests.sh`. I think this is because without explicit pathname, `type -p` will go hunt for its argument in the `PATH`, and doesn't find it (as the current directory isn't in the `PATH` usually).

So we definitely need to update the `README.txt`. We could presumably also try and fix this by adding some logic that tries to detect this (e.g. if `$script_name` is empty, set it to `$0`). Want to give that a go?

KrisThielemans:
> I can't find any information on the `type -p` option

in bash, do `type --help`

Using `type -p` might be overcomplicating things as you say. It was designed to be able to run the script from elsewhere, but possibly was never done and therefore the rest fails anyway. (I can't qutie remember why I went that far).

Note that the `conda` scripts test the recon_test_pack without having the complete source tree available (see [here](https://github.com/conda-forge/stir-feedstock/blob/7ef9fc6ef3e1a3694904d1da48c99b44f7bd0152/recipe/meta.yaml#L44), so  we'd have to be sure any mods don't break that. I think yours would be fine, but I haven't checked properly.

In fact, this whole `pardir` thing is a bit outdated. we should be using the installed location of the example files anyway. At present, they can be found via `stir_config --config-dir` and then going up `../../../doc/stir/examples/...`. But doing that is quite inconvenient at the moment as `stir_config` outputs too much. Ideally we'd give it an option `stir_config --example-dir` which outputs ONLY the directory. I certainly don't have the time to look into that sadly.

KrisThielemans:
obsolete due to using `stir_config` in #1118

PR #1068, opened 2022-08-31
Set CXX14 (min) if ROOT Version >= 6.24

### LINKED ISSUE #1067: ROOT Version: 6.26 requires at least c++14
I was building with c++11 but I got the following error:
```
In file included from /usr/local/Cellar/root/6.26.04_1/include/root/RtypesCore.h:23:
/usr/local/Cellar/root/6.26.04_1/include/root/ROOT/RConfig.hxx:48:4: error: "ROOT requires support for C++14 or higher."
#  error "ROOT requires support for C++14 or higher."
   ^
/usr/local/Cellar/root/6.26.04_1/include/root/ROOT/RConfig.hxx:50:5: error: "Pass `-std=c++14` as compiler argument."
#   error "Pass `-std=c++14` as compiler argument."
```

This is documented https://root.cern/install/build_from_source/#setting-the-c-standard

Fix PR incoming

robbietuk:
Fixes #1067

Newer ROOT version require min of C++14 - https://root.cern/install/build_from_source/#setting-the-c-standard

Local CMake log 

```
...
-- Using C++ version 11
/Users/rts/devel/STIR/devel1/release_install/share/stir/config defined
-- LLN matrix library not found. If you do have it, set the missing variables (missing: LLN_LIBRARIES LLN_INCLUDE_DIRS) 
-- ROOT Version: 6.26.04
-- ROOT Version is >= 6.24. Setting the minimum CXX version to 14.
-- Using C++ version 14
...
```
where `-- Using C++ version ##` are logged from `UseCXX(##)`


CMakeLists.txt
```diff
@@ -115,7 +115,11 @@ endif()
 if(NOT DISABLE_CERN_ROOT)
   find_package(CERN_ROOT)
   if (CERN_ROOT_FOUND)
-    message(STATUS "ROOT Version: ${CERN_ROOT_VERSION}")
+    message(STATUS "ROOT Version: ${CERN_ROOT_VERSION_MAJOR}.${CERN_ROOT_VERSION_MINOR}.${CERN_ROOT_VERSION_PATCH}")
```
KrisThielemans, line 118: This should work
```cmake
if (${CERN_ROOT_VERSION} VERSION_GREATER 6.23)
...
```
So no need for the parsing in the `Find*cmake`.
(`VERSION_GREATER_EQUAL` was introduced in 3.8 or so, while we only require 3.1).

KrisThielemans:
it is surprising that we need this. Normally CMake should "transition" such requirements. In any case, the strategy is fine for me.

robbietuk:
Excelent, I didnt know `VERSION_GREATER` could handle strings. It passes on my machine, let's wait for GHA

robbietuk:
> it is surprising that we need this. Normally CMake should "transition" such requirements. In any case, the strategy is fine for me.

It might have been a mac/brew issue? In the configure log of https://github.com/UCL/STIR/runs/7283929344?check_suite_focus=true it uses ROOT version `6.24.02` but doesn't appear to upgrade the CXX version from 11 after... see 
```
...
-- Using C++ version 11
-- Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.71.0/BoostConfig.cmake (found suitable version "1.71.0", minimum required is "1.36.0")  
-- LLN matrix library not found. If you do have it, set the missing variables (missing: LLN_LIBRARIES LLN_INCLUDE_DIRS) 
-- Found CERN_ROOT: 6.24.02  
-- ROOT Version: 6.24.02
-- AVW matrix library not found. If you do have it, set the missing variables (missing: AVW_LIBRARIES AVW_INCLUDE_DIRS) 
...
```

KrisThielemans:
it wouldn't upgrade to C++-14 for whole of STIR, but it should have for those files that depend on ROOT. But of course, that might be hard to get correct, so this seems better.

Issue #1070, opened 2022-09-02
Parallelproj support issues linked to scanner/FOV radius


markus-jehl:
In the parallelproj wrapper in STIR there still seem to be a couple of issues. Currently, there are artifacts on either side of the sinogram when forward-projecting a sphere (or a voxel) that is positioned towards the edge of the FOV (left: Cylindrical, right: BlocksOnCylindrical):
![image](https://user-images.githubusercontent.com/87262701/188109342-7f39a091-f97d-4a37-9ef3-8cec78a4e317.png)
This can be fixed by increasing the radius in ParallelprojHelper.cxx (which is currently set to the inner ring radius of the scanner). However, when increasing this radius a lot, it not only fixes these horizontal artifacts, but it also shrinks the extent of the activity curve in the sinograms, which should not happen (left: inner_radius + 5, right: inner_radius + 20):
![image](https://user-images.githubusercontent.com/87262701/188109934-cc7b79fe-b2a5-4bf5-909f-b560b530be1a.png)

markus-jehl:
Kris: "The ray really shouldn’t change if we ask it to compute intersections with a larger FOV. That might be another bug on how we call parallelproj (although our code is quite small). I think we need to add a few test cases to the “compute intersections” for the LOR code. Stuff like checking that if we give an LORAs2Points, we get a line that still contains the 2 original points."

KrisThielemans:
At present we see the only relevant code is https://github.com/UCL/STIR/blob/4142d5969e4320e32ed9dbe463b0fd50f36fb3f6/src/recon_buildblock/Parallelproj_projector/ParallelprojHelper.cxx#L91

We should have a test function for `get_intersections_with_cylinder` just to exclude this is where the problem sits.

markus-jehl:
![image](https://user-images.githubusercontent.com/87262701/193082387-870d7ce5-f078-420c-be2c-858503a5ee51.png)
The "squashing" of the sinograms was caused by a bug in the code that computes intersections with a cylinder.

markus-jehl:
Should this be closed and a new one created that refers specifically to investigating methods to replicate tangential lines of response, @KrisThielemans ?

PR #1072, opened 2022-09-29
Parallelproj support: bugfix for computing LOR intersection with cylinder

markus-jehl:
Fixes part of the issues mentioned in issue: https://github.com/UCL/STIR/issues/1070.

Previously, the LOR endpoints were radially projected outwards, instead of extended and intersected with cylinder. As part of this, the code computing the intersection was slightly modified to align with the equations in the comments.

### DIFF
--- a/documentation/release_5.1.htm
+++ b/documentation/release_5.1.htm
@@ -39,7 +39,9 @@ improvements to the documentation.
 <h2> Summary for end users (also to be read by developers)</h2>
 
 <h3>Bug fixes</h3>
-none
+<ul>
+  <li>Fix of find_LOR_intersections_with_cylinder function, see <a href="https://github.com/UCL/STIR/pull/1072/">PR #1072</a>.
+</ul>
 
 <h3>New functionality</h3>
 <li>Improvements to listmode reconstruction by Nikos Efthimiou, see <a href="https://github.com/UCL/STIR/pull/1030/">PR #1030</a>. 
--- a/documentation/release_5.1.htm
+++ b/documentation/release_5.1.htm
@@ -40,7 +40,7 @@ improvements to the documentation.
 
 <h3>Bug fixes</h3>
 <ul>
-  <li>Fix of find_LOR_intersections_with_cylinder function, see <a href="https://github.com/UCL/STIR/pull/1072/">PR #1072</a>.
+  <li>Fix of <code>find_LOR_intersections_with_cylinder</code> functions in the LOR hierarchy. This fixes the use of the <tt>parallelproj<tt> projector for non-cylindrical geometries. See <a href="https://github.com/UCL/STIR/pull/1072/">PR #1072</a>.</li>
 </ul>
 
 <h3>New functionality</h3>
--- a/src/include/stir/LORCoordinates.h
+++ b/src/include/stir/LORCoordinates.h
@@ -143,8 +143,9 @@ class LORCylindricalCoordinates_z_and_radius
     {
       check_state();
       assert(new_radius>0);
-      _z1 *= new_radius/_radius;
-      _z2 *= new_radius/_radius;
+      const auto mid_point = 0.5 * (_z1 + _z2);
+      _z1 = mid_point + (_z1 - mid_point) * new_radius / _radius;
+      _z2 = mid_point + (_z2 - mid_point) * new_radius / _radius;
       _radius = new_radius;
     }
   coordT _radius;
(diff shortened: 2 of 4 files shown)


src/include/stir/LORCoordinates.inl
```diff
@@ -333,15 +333,18 @@ find_LOR_intersections_with_cylinder(LORAs2Points<coordT1>& intersection_coords,
...
      R^2*(d.x^2+d.y^2)-(d.x*c1.y-d.y*c1.x)^2
   */
-  const double dxy2 = (square(d.x())+square(d.y()));
-  const double argsqrt=
-    (square(radius)*dxy2-square(d.x()*c1.y()-d.y()*c1.x()));
-  if (argsqrt<=0)
+  const double a = square(d.x()) + square(d.y());
+  const double b = d.x() * c1.x() + d.y() * c1.y();
+  const double e = square(c1.x()) + square(c1.y()) - square(radius);
+  double argsqrt = square(b) - a * e;
```
KrisThielemans, line 339: this is just a change to fit the comment above, correct?

src/include/stir/LORCoordinates.h
```diff
@@ -143,8 +143,9 @@ class LORCylindricalCoordinates_z_and_radius
     {
       check_state();
       assert(new_radius>0);
-      _z1 *= new_radius/_radius;
-      _z2 *= new_radius/_radius;
+      const auto mid_point = 0.5 * (_z1 + _z2);
+      _z1 = mid_point + (_z1 - mid_point) * new_radius / _radius;
+      _z2 = mid_point + (_z2 - mid_point) * new_radius / _radius;
```
KrisThielemans, line 148: I guess you're saying that this fix isn't enough, as need to take angles into account?

If so, it is probably better to not adjust `_z1` and `_z2`, and adjust the documentation that `set_radius*` does *not* preserve the LOR.

src/include/stir/LORCoordinates.h
```diff
@@ -192,8 +193,9 @@ class LORInCylinderCoordinates : public LOR<coordT>
       _radius*fabs(cos((_p1.psi()-_p2.psi())/2));
     if (new_radius >= min_radius)
       return Succeeded::no;
-    _p1.z() *= new_radius/_radius;
-    _p2.z() *= new_radius/_radius;
+    const auto mid_point = 0.5 * (_p1.z() + _p2.z());
+    _p1.z() = mid_point + (_p1.z() - mid_point) * new_radius / _radius;
+    _p2.z() = mid_point + (_p2.z() - mid_point) * new_radius / _radius;
```
KrisThielemans, line 198: as above

KrisThielemans (commented):
indeed simpler to do it this way, especially as it is now correct :-)

Would it be easy to add some of your tests to this PR?

src/include/stir/LORCoordinates.inl
markus-jehl: correct

src/include/stir/LORCoordinates.h
markus-jehl: I think this fix is important nonetheless. Before, set_radius was also not correctly scaling up the LOR to a new radius and with this fix at least this behaviour is as expected. But I can definitely adjust the documentation to say that set_radius does not preserve the LOR.

KrisThielemans: > with this fix at least this behaviour is as expected. 

well, I expected it to be the correct LOR, but it isn't. Not sure therefore what we achieve with this.

markus-jehl:
Yes, it's fairly easy to add the test for the intersections. Will do!

KrisThielemans:
If we do want to preserve the LOR, my feeling is that
- `phi` remains the same
- `s` remains the same
- therefore `R1 sin(beta1) = R2 sin(beta2)`, hence `beta2 = arcsin( R1/R2 sin(beta1))`

KrisThielemans:
MacOS-Debug job failed due to #1073. This has nothing to do with this PR.

src/include/stir/LORCoordinates.h
markus-jehl: Ok, I wasn't sure what the expected behaviour was meant to be :-) in this case I'll remove this fix from the PR.

KrisThielemans (changes requested):
great. now just an update to the release notes and we're good

documentation/release_5.1.htm
```diff
@@ -39,7 +39,9 @@ <h2>Patch release info</h2>
 <h2> Summary for end users (also to be read by developers)</h2>
 
 <h3>Bug fixes</h3>
-none
+<ul>
+  <li>Fix of find_LOR_intersections_with_cylinder function, see <a href="https://github.com/UCL/STIR/pull/1072/">PR #1072</a>.
```
KrisThielemans, line 43:
```suggestion
  <li>Fix of <code>find_LOR_intersections_with_cylinder</code> functions in the LOR hierarchy. This fixes the use of the <tt>parallelproj<tt> projector for non-cylindrical geometries. See <a href="https://github.com/UCL/STIR/pull/1072/">PR #1072</a>.</li>
```

KrisThielemans (changes requested):
I'll do this myself

KrisThielemans:
Ready. ok to squash-merge? (Fine to merge as well)

PR #1076, opened 2022-10-04
Issue/1070 second iteration changes for parallelproj interface

markus-jehl:
Added a call to get the max_FOV_radius of the scanner, to fix the parallelproj interface for non-cylindrical scanners.

### DIFF
--- a/src/buildblock/Scanner.cxx
+++ b/src/buildblock/Scanner.cxx
@@ -910,7 +910,7 @@ void Scanner::set_up()
             error("Scanner::scanner_geometry needs to be one of Cylindrical, BlocksOnCylindrical, Generic");
         }
     }
-  set_outer_FOV_radius();
+  set_max_FOV_radius();
   _already_setup = true;
 }
 
@@ -927,23 +927,23 @@ set_detector_map( const DetectorCoordinateMap::det_pos_to_coord_type& coord_map 
 
 void
 Scanner::
-set_outer_FOV_radius() {
+set_max_FOV_radius() {
   if (!this->detector_map_sptr)
   {
     // for cylindrical scanners, all detectors have the same distance from the rotational axis
-    outer_FOV_radius = inner_ring_radius;
+    max_FOV_radius = inner_ring_radius;
   }
   else
   {
     // for other geometries, loop through all detectors and set the radius to the largest found distance
-    outer_FOV_radius = inner_ring_radius;
+    max_FOV_radius = inner_ring_radius;
     for (auto tangential_pos = 0; tangential_pos < detector_map_sptr->get_num_tangential_coords(); tangential_pos++)
       for (auto axial_pos = 0; axial_pos < detector_map_sptr->get_num_axial_coords(); axial_pos++)
         for (auto radial_pos = 0; radial_pos < detector_map_sptr->get_num_radial_coords(); radial_pos++)
     {
       auto coord = detector_map_sptr->get_coordinate_for_det_pos(stir::DetectionPosition<>(tangential_pos, axial_pos, radial_pos));
       const auto detector_radius = sqrt(coord.x() * coord.x() + coord.y() * coord.y()) - average_depth_of_interaction;
-      outer_FOV_radius = fmax(outer_FOV_radius, detector_radius);
+      max_FOV_radius = fmax(max_FOV_radius, detector_radius);
     }
   }
 }
--- a/src/buildblock/Scanner.cxx
+++ b/src/buildblock/Scanner.cxx
@@ -910,6 +910,7 @@ void Scanner::set_up()
             error("Scanner::scanner_geometry needs to be one of Cylindrical, BlocksOnCylindrical, Generic");
         }
     }
+  set_outer_FOV_radius();
   _already_setup = true;
 }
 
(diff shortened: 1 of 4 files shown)


markus-jehl:
Kris (https://github.com/UCL/STIR/pull/1074): "I suspect it isn't the image radius though. Could you check what parallelproj does if you give it a small radius?
Instead I think we should check the return value of get_intersections_with_cylinder such that we don't ever use intersections with a cylinder where there aren't any,"

markus-jehl:
I tested it, and parallelproj works as expected for smaller radii. I now also added handling of the return value of get_intersections_with_cylinder. The remaining things to be done are:

1. Add parallelproj sections to existing tests and confirm they pass.
2. Run Positrigo reconstruction pipeline to check there are no obvious bugs anymore.

src/include/stir/Scanner.h
```diff
@@ -506,6 +509,7 @@ class Scanner
   shared_ptr<DetectorCoordinateMap> detector_map_sptr;
 
   void set_detector_map( const DetectorCoordinateMap::det_pos_to_coord_type& coord_map );
+  void set_max_FOV_radius();
```
KrisThielemans, line 512: Possibly this naming is confusing. `set*` functions normally take a value. We could do that here as well, and if it's negative or so, use the current code. It might get a bit messy though to know if it has been done or not. Probably simpler for now is to rename this private function to  `initialise_max_FOV_radius` and add a `set` function when someone asks for it.

KrisThielemans (changes requested):

src/buildblock/Scanner.cxx
```diff
@@ -924,6 +925,29 @@ set_detector_map( const DetectorCoordinateMap::det_pos_to_coord_type& coord_map
...
+  else
+  {
+    // for other geometries, loop through all detectors and set the radius to the largest found distance
+    max_FOV_radius = inner_ring_radius;
+    for (auto tangential_pos = 0; tangential_pos < detector_map_sptr->get_num_tangential_coords(); tangential_pos++)
+      for (auto axial_pos = 0; axial_pos < detector_map_sptr->get_num_axial_coords(); axial_pos++)
+        for (auto radial_pos = 0; radial_pos < detector_map_sptr->get_num_radial_coords(); radial_pos++)
+    {
+      auto coord = detector_map_sptr->get_coordinate_for_det_pos(stir::DetectionPosition<>(tangential_pos, axial_pos, radial_pos));
+      const auto detector_radius = sqrt(coord.x() * coord.x() + coord.y() * coord.y()) - average_depth_of_interaction;
```
KrisThielemans, line 945: hmmm. are we sure about this? is the detector-map giving "internal" points or surface points? This could be important for the blocks case, where we're suppose to pass the internal "radius".

src/recon_buildblock/Parallelproj_projector/ParallelprojHelper.cxx
```diff
@@ -73,7 +73,7 @@ detail::ParallelprojHelper::ParallelprojHelper(const ProjDataInfo& p_info, const
   Bin bin;
   LORInAxialAndNoArcCorrSinogramCoordinates<float> lor;
   LORAs2Points<float> lor_points;
-  const float radius = p_info.get_scanner_sptr()->get_inner_ring_radius();
+  const float radius = p_info.get_scanner_sptr()->get_max_FOV_radius() + p_info.get_scanner_sptr()->get_average_depth_of_interaction();
```
KrisThielemans, line 76: I think still to remove the DOI

KrisThielemans (changes requested):

KrisThielemans:
Do you intend to add the tests in this PR? Maybe best to finish this one and squash-merge it first.

src/include/stir/Scanner.h
markus-jehl: I like the "initialise_max_FOV_radius" - will change!

src/buildblock/Scanner.cxx
markus-jehl: Yes, I was also confused by this, but it really is the internal position including DOI. Does this mean there might be a problem in the BlocksOnCylindrical implementation?

src/recon_buildblock/Parallelproj_projector/ParallelprojHelper.cxx
markus-jehl: If I remove the DOI, it will crop parts of the sinogram though - is that what we want?

markus-jehl:
Initially, I was thinking of doing it as part of this PR - but I agree it would be nice to get this merged in sooner rather than later!

src/buildblock/Scanner.cxx
markus-jehl: https://github.com/UCL/STIR/blob/2945b72f9974a658f866fc2f47bb07ea869e0182/src/buildblock/GeometryBlocksOnCylindrical.cxx#L93
This is where the radius for the detector-map is set for BoC: it uses get_effective_ring_radius, not get_inner_ring_radius.

KrisThielemans: ok. Let's document this then in `Scanner`: detector map stores "effective detection positions"  (i.e. approximately including DOI) for each crystal.

src/recon_buildblock/Parallelproj_projector/ParallelprojHelper.cxx
KrisThielemans: well, it will crop the part where they will never be any activity. no? (no activity inside the detector)

markus-jehl: image
image
Yes, I agree. But strangely enough it does impact the way the resulting sinograms look (first: with DOI; second: without DOI). I now removed the "+ DOI" in the code, but would still be good to understand why the sinograms differ.

src/buildblock/Scanner.cxx
markus-jehl: added a comment in the scanner.h file where the detector map is defined.

src/include/stir/Scanner.h
markus-jehl: done

src/recon_buildblock/Parallelproj_projector/ParallelprojHelper.cxx
markus-jehl: This has been discussed, and it is actually caused by LOR's along one detector module that remain behind the front edge of the crystal. Without modelling DOI they are obviously cropped, whereas they are included in the outermost parts of the sinogram when including the DOI for the definition of the LOR's. Since these LOR's are not relevant for the image reconstruction, they can be set to zero as in the second plot.

markus-jehl:
IMPORTANT: currently, parallelproj truncates the resulting images to a cylinder with the inner ring radius of the scanner. This means any activity in the corners of a BlocksOnCylindrical scanner will be removed.

KrisThielemans:
AppVeyor time-outs

PR #1085, opened 2022-10-21
 template normalisation and random functionality for BlocksOnCylindrical

danieldeidda:
…lockOnCylindrical

### DIFF
--- a/src/buildblock/ML_norm.cxx
+++ b/src/buildblock/ML_norm.cxx
@@ -3,6 +3,7 @@
     Copyright (C) 2016, 2020, 2021, University College London
     Copyright (C) 2016-2017, PETsys Electronics
     Copyright (C) 2021, Gefei Chen
+    Copyright (C) 2022, National Physical Laboratory
     This file is part of STIR.
 
     SPDX-License-Identifier: Apache-2.0
@@ -19,6 +20,7 @@
   \author Kris Thielemans
   \author Tahereh Niknejad
   \author Gefei Chen
+  \author daniel deidda
 
 */
 
@@ -124,19 +126,17 @@ int DetPairData::get_num_detectors() const
   return num_detectors;
 }
 
-void make_det_pair_data(DetPairData& det_pair_data,
-			const ProjDataInfo& proj_data_info_general_type,
-			const int segment_num,
-			const int ax_pos_num)
+template <class TProjDataInfo>
+static void make_det_pair_data_help(DetPairData& det_pair_data,
+                        const TProjDataInfo& proj_data_info_general_type,
+                        const int segment_num,
+                        const int ax_pos_num)
 {
-  const ProjDataInfoCylindricalNoArcCorr& proj_data_info =
-    dynamic_cast<const ProjDataInfoCylindricalNoArcCorr&>(proj_data_info_general_type);
-
-  const int num_detectors = 
-    proj_data_info.get_scanner_ptr()->get_num_detectors_per_ring();
-  const int fan_size = 
-    2*max(proj_data_info.get_max_tangential_pos_num(),
-          -proj_data_info.get_min_tangential_pos_num()) + 1;
+  const int num_detectors =
+    proj_data_info_general_type.get_scanner_ptr()->get_num_detectors_per_ring();
+  const int fan_size =
+    2*max(proj_data_info_general_type.get_max_tangential_pos_num(),
+          -proj_data_info_general_type.get_min_tangential_pos_num()) + 1;
   // fan will range from -half_fan_size to +half_fan_size (i.e. an odd number of elements)
   const int half_fan_size = fan_size/2;
 
@@ -144,7 +144,7 @@ void make_det_pair_data(DetPairData& det_pair_data,
   fan_indices.grow(0,num_detectors-1);
   for (int a = 0; a < num_detectors; ++a)
   {
-    fan_indices[a] = 
+    fan_indices[a] =
       IndexRange<1>(a+num_detectors/2-half_fan_size,
                     a+num_detectors/2+half_fan_size);
   }
@@ -153,20 +153,42 @@ void make_det_pair_data(DetPairData& det_pair_data,
 }
 
 void make_det_pair_data(DetPairData& det_pair_data,
-			const ProjData& proj_data,
-			const int segment_num,
-			const int ax_pos_num)
+                        const ProjDataInfo& proj_data_info_general_type,
+                        const int segment_num,
+                        const int ax_pos_num)
+{
+    if (proj_data_info_general_type.get_scanner_ptr()->get_scanner_geometry()=="Cylindrical")
+    {
+        auto proj_data_info =
+                dynamic_cast<const ProjDataInfoCylindricalNoArcCorr&>(proj_data_info_general_type);
+
+        make_det_pair_data_help(det_pair_data, proj_data_info, segment_num, ax_pos_num);
+    }
+    else
+    {
+        auto proj_data_info =
+                dynamic_cast<const ProjDataInfoBlocksOnCylindricalNoArcCorr&>(proj_data_info_general_type);
+
+        make_det_pair_data_help(det_pair_data, proj_data_info, segment_num, ax_pos_num);
+    }
+
+}
+
+template <class TProjDataInfo>
+static void make_det_pair_data_help(DetPairData& det_pair_data,
+                        const TProjDataInfo& proj_data_info,
+                        const ProjData& proj_data,
+                        const int segment_num,
+                        const int ax_pos_num)
 {
   make_det_pair_data(det_pair_data,
-		     *proj_data.get_proj_data_info_sptr(),
-		     segment_num,
-		     ax_pos_num);
-  const int num_detectors = 
+                     *proj_data.get_proj_data_info_sptr(),
+                     segment_num,
+                     ax_pos_num);
+  const int num_detectors =
     det_pair_data.get_num_detectors();
-  const ProjDataInfoCylindricalNoArcCorr& proj_data_info =
-    dynamic_cast<const ProjDataInfoCylindricalNoArcCorr&>(*proj_data.get_proj_data_info_sptr());
 
-  shared_ptr<Sinogram<float> > 
+  shared_ptr<Sinogram<float> >
     pos_sino_ptr(new Sinogram<float>(proj_data.get_sinogram(ax_pos_num,segment_num)));
   shared_ptr<Sinogram<float> > neg_sino_ptr;
   if (segment_num == 0)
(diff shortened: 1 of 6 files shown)


src/buildblock/ML_norm.cxx
```diff
@@ -1076,35 +1168,61 @@ void make_fan_data_remove_gaps(FanProjData& fan_data,
...
+                                int fan_size,
+                                const TProjDataInfo& proj_data_info,
+                                const FanProjData& fan_data,
+                                const float gap_value=0.F)
+{
     const int half_fan_size = fan_size/2;
-    //assert(num_rings == fan_data.get_num_rings());
-    //assert(num_detectors_per_ring == fan_data.get_num_detectors_per_ring());
+    //assert(num_rings == num_rings);
+    //assert(num_detectors_per_ring == num_detectors_per_ring);
```
KrisThielemans, line 1212: obviously that doesn't make sense, so let's cut them
```suggestion
```

src/buildblock/ML_norm.cxx
```diff
@@ -1336,47 +1481,73 @@ void make_fan_sum_data(Array<2,float>& data_fan_sums, const FanProjData& fan_dat
...
 
+void make_fan_sum_data(Array<2,float>& data_fan_sums,
+               const ProjData& proj_data)
+{
+
+    int num_rings;
+    int num_detectors_per_ring;
+    int fan_size;
+    int max_delta;
+    const ProjDataInfo& proj_data_info = *(get_fan_info(num_rings, num_detectors_per_ring, max_delta, fan_size,
```
KrisThielemans, line 1532: very dangerous. a reference to effectively a returned shared_ptr. This could just as well be cleaned up immediately, and therefore the object pointed to could disappear.

src/buildblock/ML_norm.cxx
```diff
@@ -1668,6 +1839,40 @@ double KL(const FanProjData& d1, const FanProjData& d2, const double threshold)
...
+
+template void set_fan_data_add_gaps_help
+(ProjData&, int, int, int, int, const ProjDataInfoBlocksOnCylindricalNoArcCorr&, const FanProjData&, const float );
+template void set_fan_data_add_gaps_help
+(ProjData&, int, int, int, int, const ProjDataInfoCylindricalNoArcCorr&, const FanProjData&, const float );
+
+template void make_fan_data_remove_gaps_help
+(FanProjData&, int, int, int, int,  const ProjDataInfoBlocksOnCylindricalNoArcCorr&, const ProjData&);
+template void make_fan_data_remove_gaps_help
+(FanProjData&, int, int, int, int, const ProjDataInfoCylindricalNoArcCorr&, const ProjData&);
```
KrisThielemans, line 1876: These declarations serve no purpose. We'd need them if you would have used them BEFORE defining them, but you don't.
```suggestion
```

src/buildblock/ML_norm.cxx
```diff
@@ -1076,35 +1168,61 @@ void make_fan_data_remove_gaps(FanProjData& fan_data,
...
     int num_rings;
     int num_detectors_per_ring;
     int fan_size;
     int max_delta;
-    shared_ptr<const ProjDataInfoCylindricalNoArcCorr> proj_data_info_ptr =
-    get_fan_info(num_rings, num_detectors_per_ring, max_delta, fan_size,
-                 *proj_data.get_proj_data_info_sptr());
+    auto fan_info_sptr = (get_fan_info(num_rings, num_detectors_per_ring, max_delta, fan_size,
+                                       *proj_data.get_proj_data_info_sptr()));
+    const ProjDataInfo& proj_data_info =*fan_info_sptr.get();
```
KrisThielemans, line 1180: as far as I can see, `*fan_info_sptr` is equal to `*proj_data.get_proj_data_info_sptr()`. I think therefore the following is much clearer
```suggestion
    const ProjDataInfo& proj_data_info =*proj_data.get_proj_data_info_sptr();
    get_fan_info(num_rings, num_detectors_per_ring, max_delta, fan_size, proj_data_info);
```

src/buildblock/ML_norm.cxx
```diff
@@ -1155,14 +1273,41 @@ void set_fan_data_add_gaps(ProjData& proj_data,
...
+                           const FanProjData& fan_data,
+                           const float gap_value)
+{
+    int num_rings;
+    int num_detectors_per_ring;
+    int fan_size;
+    int max_delta;
+    auto fan_info_sptr = (get_fan_info(num_rings, num_detectors_per_ring, max_delta, fan_size,
+                                       *proj_data.get_proj_data_info_sptr()));
+    const ProjDataInfo& proj_data_info =*fan_info_sptr.get();
```
KrisThielemans, line 1294: as above

KrisThielemans (changes requested):
detailed comments, but overall ok.

Make all `*_help` function `static`. That means they cannot be called by anyone else (i.e. they have static linkage for this object file).

I think though that we don't need the return value of `get_fan_info` at all (see comments), in which case we don't need `get_fan_info_help` anymore. Then we can keep the call to `get_fan_info` in the original place (i.e. now the templated functions), making this a bit smaller.

I'd take the opportunity to remove `get_fan_info` from `ML_norm.h` and make it a `static` function. (It's just a convenience function here. shouldn't be used by anybody else).

Add a test case to `test_ML_norm.cxx`. should be trivial

KrisThielemans:
by the way, please don't commit suggestion one-by-one, but batch them up. Less global warming...

KrisThielemans:
I see a warning on AppVeyor
```
C:\projects\stir\src\buildblock\multiply_crystal_factors.cxx(95): warning C4067: unexpected tokens following preprocessor directive - expected a newline
```
This is a mistake of mine I guess
https://github.com/UCL/STIR/blob/b480282099055080a3efd11f4fe098f4803ed241/src/buildblock/multiply_crystal_factors.cxx#L95
can you replace `and` with `&&`?

src/buildblock/ML_norm.cxx
danieldeidda: so wat is the safe way to pass those info?

KrisThielemans: would have to be
```c++
auto proj_data_info_sptr = get_fan_info(...);
```
However, as I write below, I don't think we need the return value at all.

KrisThielemans:
> sorry it doesn't let me use the botton "add suggestion to batch"

also not from the Files tab?

danieldeidda:
> I'd take the opportunity to remove `get_fan_info` from `ML_norm.h` and make it a `static` function. (It's just a convenience function here. shouldn't be used by anybody else).

get_fan_info is called by find_ML_singles... though

danieldeidda:
also if get_fan_info goes back to original we will trigger this:
```

if (proj_data_info_ptr == 0)
        {
          error("Can only process not arc-corrected data\n");
        }
```

KrisThielemans:
> > I'd take the opportunity to remove `get_fan_info` from `ML_norm.h` and make it a `static` function. (It's just a convenience function here. shouldn't be used by anybody else).
> 
> get_fan_info is called by find_ML_singles... though

ok true. that should be revamped a bit, but I'm not up for that now. However, that doesn't use its return value, so making it a `void` will still work.

> also if get_fan_info goes back to original we will trigger this:
> 
> ```
> 
> if (proj_data_info_ptr == 0)
>         {
>           error("Can only process not arc-corrected data\n");
>         }
> ```

Those conditions should be removed there, and in the other functions, I'd replace
```c++
    if (proj_data.get_proj_data_info_sptr()->get_scanner_ptr()->get_scanner_geometry()=="Cylindrical")
    {
        auto proj_data_info_ptr =
                dynamic_cast<const ProjDataInfoCylindricalNoArcCorr * const>(&proj_data_info);

        make_fan_data_remove_gaps_help(fan_data, num_rings, num_detectors_per_ring, max_delta, fan_size, *proj_data_info_ptr, proj_data);
    }
    else
    {
        auto proj_data_info_ptr =
                dynamic_cast<const ProjDataInfoBlocksOnCylindricalNoArcCorr * const>(&proj_data_info);

        make_fan_data_remove_gaps_help(fan_data, num_rings, num_detectors_per_ring, max_delta, fan_size, *proj_data_info_ptr, proj_data);
    }
```
with
```
    if (auto proj_data_info_ptr =
                dynamic_cast<const ProjDataInfoCylindricalNoArcCorr * const>(proj_data.get_proj_data_info_sptr()))
   {
        make_fan_data_remove_gaps_help(fan_data, num_rings, num_detectors_per_ring, max_delta, fan_size, *proj_data_info_ptr, proj_data);
    }
    else if (auto proj_data_info_ptr =
                dynamic_cast<const ProjDataInfoBlocksOnCylindricalNoArcCorr * const>(proj_data.get_proj_data_info_sptr()))

        make_fan_data_remove_gaps_help(fan_data, num_rings, num_detectors_per_ring, max_delta, fan_size, *proj_data_info_ptr, proj_data);
    }
   else
  {
     error("ML_norm functions only support non-arccorrected data (cylindrical or blocks-on-cylindrical) at the moment.");
  }
```
or something like that.

`get_fan_info` will need to do a `dynamic_cast` to `ProjDataInfoCylindrical` at the moment to get at `get_max_ring_difference`, but that should be it.

...maybe....

danieldeidda:
those condition were already removed from get_fan_info but 
 the error is triggered when doing the bit the was originally there:

```
proj_data_info_ptr =
                dynamic_cast<const ProjDataInfoCylindricalNoArcCorr * const>(&proj_data_info);
```
```

        if (proj_data_info_ptr == 0)
        {
          error("Can only process not arc-corrected data\n");
        }
```

are you saying we can remove this check?

KrisThielemans:
yes, we need to check on `ProjDataInfoCylnidrical`, not more

danieldeidda:
> Add a test case to `test_ML_norm.cxx`. should be trivial

you mean the same as 
`test_proj_data_info(shared_ptr<const ProjDataInfoCylindricalNoArcCorr> proj_data_info_sptr)`

but for blocksOnCylindrical?

KrisThielemans:
yes.

src/buildblock/ML_norm.cxx
```diff
@@ -1529,8 +1489,9 @@ void make_fan_sum_data(Array<2,float>& data_fan_sums,
     int num_detectors_per_ring;
     int fan_size;
     int max_delta;
-    const ProjDataInfo& proj_data_info = *(get_fan_info(num_rings, num_detectors_per_ring, max_delta, fan_size,
-                                                        *proj_data.get_proj_data_info_sptr())).get();
+    get_fan_info(num_rings, num_detectors_per_ring, max_delta, fan_size,
+                 *proj_data.get_proj_data_info_sptr());
+    const ProjDataInfo& proj_data_info =*proj_data.get_proj_data_info_sptr();
```
KrisThielemans, line 1494: not used?

src/buildblock/ML_norm.cxx
```diff
@@ -1289,9 +1249,9 @@ void set_fan_data_add_gaps(ProjData& proj_data,
     int num_detectors_per_ring;
     int fan_size;
     int max_delta;
-    auto fan_info_sptr = (get_fan_info(num_rings, num_detectors_per_ring, max_delta, fan_size,
-                                       *proj_data.get_proj_data_info_sptr()));
-    const ProjDataInfo& proj_data_info =*fan_info_sptr.get();
+    get_fan_info(num_rings, num_detectors_per_ring, max_delta, fan_size,
+                                       *proj_data.get_proj_data_info_sptr());
+    const ProjDataInfo& proj_data_info = *proj_data.get_proj_data_info_sptr();
```
KrisThielemans, line 1254: not used?

src/buildblock/ML_norm.cxx
```diff
@@ -950,17 +944,19 @@ std::istream& operator>>(std::istream& s, FanProjData& fan_data)
...
-                  int& max_ring_diff, int& fan_size,
-                  const TProjDataInfo& proj_data_info)
+void
+get_fan_info(int& num_rings, int& num_detectors_per_ring,
+         int& max_ring_diff, int& fan_size,
+         const ProjDataInfo& proj_data_info)
 {
-    if (proj_data_info.get_view_mashing_factor()>1)
+    auto proj_data_info_ptr =
+            dynamic_cast<const ProjDataInfoCylindricalNoArcCorr * const>(&proj_data_info);
```
KrisThielemans, line 953:
```suggestion
            dynamic_cast<const ProjDataInfoCylindrical * const>(&proj_data_info);
```
but also add a check if not `nullptr`

src/test/test_ML_norm.cxx
```diff
@@ -45,6 +45,7 @@ class ML_normTests: public RunTests
   void run_tests();
 protected:  
   void test_proj_data_info(shared_ptr<const ProjDataInfoCylindricalNoArcCorr> proj_data_info_sptr);
+  void test_proj_data_info(shared_ptr<const ProjDataInfoBlocksOnCylindricalNoArcCorr> proj_data_info_sptr);
```
KrisThielemans, line 48:
```suggestion
  template <class ProjDataInfoT>
  void test_proj_data_info(shared_ptr<const ProjDataInfoT> proj_data_info_sptr);
```

src/test/test_ML_norm.cxx
```diff
@@ -60,6 +61,7 @@ ML_normTests::run_tests()
                                                                                        /*tang_pos*/64, 
                                                                                        /*arc_corrected*/ false));
     test_proj_data_info(dynamic_pointer_cast<ProjDataInfoCylindricalNoArcCorr>(proj_data_info_sptr));
+    test_proj_data_info(dynamic_pointer_cast<ProjDataInfoBlocksOnCylindricalNoArcCorr>(proj_data_info_sptr));
```
KrisThielemans, line 64:
```suggestion
```
instead, create a separate SAFIR scanner

src/test/test_ML_norm.cxx
```diff
@@ -71,6 +73,7 @@ ML_normTests::run_tests()
                                                                                         /*tang_pos*/64, 
                                                                                         /*arc_corrected*/ false));
     test_proj_data_info(dynamic_pointer_cast<ProjDataInfoCylindricalNoArcCorr>(proj_data_info_sptr));
+    test_proj_data_info(dynamic_pointer_cast<ProjDataInfoBlocksOnCylindricalNoArcCorr>(proj_data_info_sptr));
```
KrisThielemans, line 76:
```suggestion
```

src/utilities/apply_normfactors.cxx
```diff
@@ -137,19 +137,17 @@ int main(int argc, char **argv)
...
-
-        set_det_pair_data(*out_proj_data_ptr,
-                                                            det_pair_data,
-                                                            segment_num,
-                                                            ax_pos_num);
+    set_det_pair_data(*out_proj_data_ptr,
+                      det_pair_data,
+                      segment_num,
+                      ax_pos_num);
       }
```
KrisThielemans, line 151: undo white-space change

KrisThielemans (changes requested):

KrisThielemans:
This is now fine. I've cleaned-up the history a bit, and undid white-space changes in `apply_normfactors.cxx`. I'm afraid I force pushed to your branch. sorry. That was a bad idea!

KrisThielemans:
@emikhay @markus-jehl we believe this should allow randoms and norm without having to switch to cylindrical. Let us know if not.

markus-jehl:
This works, thank you!

PR #1096, opened 2022-11-13
Swig improvements

KrisThielemans:
improvements related to `ExamInfo`, but also clean-up.

@markus-jehl maybe you can have a look.


markus-jehl:
These are useful changes/additions, thanks you!

I have noticed that I have forgotten to rename the set_output_projdata_sptr function I added to LmToProjData in SWIG, so maybe you could add it to this PR?

KrisThielemans:
@markus-jehl This is as far as I intend to go with this. Maybe you can check if it breaks your pipeline? Currently we don't have enough Python tests really.

markus-jehl:
Thanks! Just ran the whole reconstruction pipeline and it runs through without any hiccups.

PR #1098, opened 2022-11-17
fix bugs in Hessians Prior code

KrisThielemans:
code contained `dz == dy == dx == 0`, but this is surprising in C++. it probably translates to something like
`(dz == ( dy == (dx == 0)))`, which be what we want, but probably not.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -86,6 +86,7 @@ jobs:
           CMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/install
           case ${{matrix.os}} in
             (ubuntu*)
+              sudo apt update
               # install compiler
               if test 'XX${{ matrix.compiler }}' = 'XXclang'; then
                   sudo apt install $CC # package is called clang
--- a/src/recon_buildblock/LogcoshPrior.cxx
+++ b/src/recon_buildblock/LogcoshPrior.cxx
@@ -588,6 +588,8 @@ accumulate_Hessian_times_input(DiscretisedDensity<3,elemT>& output,
             for (int dx=min_dx;dx<=max_dx;++dx)
             {
               elemT current = weights[dz][dy][dx];
+              if (current == elemT(0))
+                continue;
               if ((dz == 0) && (dy == 0) && (dx == 0)) {
                 // The j == k case
                 current *= derivative_20(current_estimate[z][y][x],
--- a/src/recon_buildblock/LogcoshPrior.cxx
+++ b/src/recon_buildblock/LogcoshPrior.cxx
@@ -588,7 +588,7 @@ accumulate_Hessian_times_input(DiscretisedDensity<3,elemT>& output,
             for (int dx=min_dx;dx<=max_dx;++dx)
             {
               elemT current = weights[dz][dy][dx];
-              if (dz == dy == dz == 0) {
+              if ((dz == 0) && (dy == 0) && (dx == 0)) {
                 // The j == k case
                 current *= derivative_20(current_estimate[z][y][x],
                                          current_estimate[z + dz][y + dy][x + dx]) * input[z][y][x];
--- a/src/recon_buildblock/QuadraticPrior.cxx
+++ b/src/recon_buildblock/QuadraticPrior.cxx
@@ -736,7 +736,7 @@ accumulate_Hessian_times_input(DiscretisedDensity<3,elemT>& output,
                     for (int dx=min_dx;dx<=max_dx;++dx)
                       {
                         elemT current = weights[dz][dy][dx];
-                        if (dz == dy == dz == 0) {
+                        if (dz == 0 && dy == 0 && dx == 0) {
                           // The j == k case
                           current *= derivative_20(current_estimate[z][y][x],
                                                    current_estimate[z + dz][y + dy][x + dx]) * input[z][y][x];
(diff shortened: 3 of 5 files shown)


KrisThielemans:
@robbietuk this makes the test fail
```
----- test RDP_no_kappa_no_eps  --> Hessian-vector product for convexity
Error : 0 is larger than -2.671[47](https://github.com/UCL/STIR/actions/runs/3488026682/jobs/5836342182#step:8:48)e-06, 

INFO: FAIL: Computation of x^T H x = -0.000003 < 0 and is therefore NOT convex
test_name=RDP_no_kappa_no_eps
beta=0.010000
input_multiplication=-100.000000
input_addition=-10.000000
current_image_multiplication=100.000000
current_image_addition=10.000000
 >input image max=-10.052037
 >input image min=-109.7740[48](https://github.com/UCL/STIR/actions/runs/3488026682/jobs/5836342182#step:8:49)
 >target image max=0.997741
 >target image min=0.000[52](https://github.com/UCL/STIR/actions/runs/3488026682/jobs/5836342182#step:8:53)0
```
any ideas?

robbietuk:
The PR is good. The previous code was a bug that aided in the test passing.

This failing test validates `x^T H x > 0`, where `x` is `input` and `H` is computed with `current_image`. The `target` is the input image to the function and this function tests variations of `x` and `H` with global scaling + additions. Note, RDP requires that `current_image` is non-negative but input can be negative.

This is failing on the testing the convexity of RDP with no eps. It is not an easy test because if `x=0` and `y->0` the function becomes sensitive to numerical issues. 

_____
Interestingly the tests are failing with different configurations.

1: Failing configuration A
```
beta=0.010000
input_multiplication=-100.000000
input_addition=0.000000                          <-different
current_image_multiplication=0.010000 <-different
current_image_addition=0.000000.         <-different
```

2: Failing configuration B
```
beta=0.010000
input_multiplication=-100.000000
input_addition=-10.000000
current_image_multiplication=100.000000
current_image_addition=10.000000
```

3: Failing configuration B
```
beta=0.010000
input_multiplication=-100.000000
input_addition=-10.000000
current_image_multiplication=100.000000
current_image_addition=10.000000
```

4: Failed on build... Unrelated I think.

Note, these configurations are the first to fail, others might be failing too.
__________

1. The solution could be to reduce the range of configurations. Currently it tests all 900 purturbations of: 
```
float beta_array[] = {0.01, 1, 100};  // Penalty strength should only affect scale
// Modifications to the input image
float input_multiplication_array[] = {-100, -1, 0.01, 1, 100}; // Test negative, small and large values
float input_addition_array[] = {-10, -1, -0.5, 0.0, 1, 10};
// Modifications to the current image (Hessian computation)
float current_image_multiplication_array[] = {0.01, 1, 100};
float current_image_addition_array[] = {0.0, 0.5, 1, 10}; // RDP has constraint that current_image is non-negative
```

2. Alternatively, remove the test of `RDP_no_kappa_no_eps` as it is a known hard case to test.

3. Need to check that the implementation of `accumulate_Hessian_times_input` for RDP is correct. However, as the implementation of `test_Hessian_against_numerical` passes and the error case here (`x^T H x = -0.000003 < 0`) is so small, I believe the implementation to be correct.

robbietuk:
@KrisThielemans This bug also exists for QP
https://github.com/UCL/STIR/blob/8c84f7362d5093216562600232f64800a56bd6d2/src/recon_buildblock/QuadraticPrior.cxx#L739

KrisThielemans:
@robbietuk I haven't reviewed this test yet, but would like to fix this now and release 5.1.0.

I've checked that the Hessian of the RDP "potential" with $\epsilon=0$ is
$8\left[[x^2,-xy],[-xy,y^2]\right]/( x+y+\gamma |x-y|)^3$. So, it goes to infinity when both $x$ and $y$ go to zero. That doesn't seem to be the cause of the current problem therefore (as the rhs is very small).

Relevant lines are
https://github.com/UCL/STIR/blob/f3c3e4b05ea26bfce7fb400f41b81c8cfec54381/src/recon_test/test_priors.cxx#L292-L295
I think this test is too strong. You can always have numerical error giving very small negative numbers. I think `my_sum` should be compared with something like `-10^-9 norm_squared(*input)`, i.e. it can be a tiny bit negative (taking the scale of `input` into account.

Does that sound reasonable to you?

robbietuk:
>I think this test is too strong. You can always have numerical error giving very small negative numbers. I think my_sum should be compared with something like -10^-9 norm_squared(*input), i.e. it can be a tiny bit negative (taking the scale of input into account.
Does that sound reasonable to you?

Agreed, adding some numerical tolerance should work help.

robbietuk:
All priors fail `test_priors` (specifically the `Hessian convexity` test) with the change to `if ((dz == 0) && (dy == 0) && (dx == 0))`, even with the suggested change to compare `my_sum` to `-1e-9 norm_squared(*input)`.

I have attatched [test_priors.log](https://github.com/UCL/STIR/files/10354338/test_priors.log) (with clearer detailing to highlight the issue). Some configurations have significant negitive numbers (`-9.0000` for QP), which indicates another bug. I am expecting this to be in the test code rather than the priors because the `Hessian against numerical` tests still pass.

KrisThielemans:
hmmm. Of course, a question is what is "large" or "small", but using `norm_squared(input)`, we've taken `x` out of that question, but there is still the question for H. However, for the priors, I guess diagonal elements of H are order 1 or so?

hard to see what would be wrong in the test though. Is the `current_image` always positive?  (Although that shouldn't matter for the QP as it doesn't depend on the image).

KrisThielemans:
@robbietuk I don't know what you did in your local version, but for me the previous commits work. Let's see what CI says of course (the 2 Release jobs are already ok).

One thing I did was to compute the inner product using doubles, which would explain smaller numerical errors.

KrisThielemans:
Everything fine here. I've aborted the next 3 AppVeyor jobs as the 1st worked and I'm too impatient (sorry)

robbietuk:
Looks good. Thaks for fixing. I am not sure why my local testing failed so badly.

PR #1110, opened 2022-12-07
Finalize PinholeSPECTUB sample parameter files

mastergari:
Sample parameter files required masking update to remove mask type and allow masking with attenuation map.


KrisThielemans:
first GHA job failed due to unrelated #1111

PR #1113, opened 2022-12-16
Issue/1070 adding parallelproj tests

markus-jehl:
Added parallelproj tests to BlocksOnCylindrical tests, including a comparison between parallelproj and ray tracing on the projection of a single voxel.

### DIFF
--- a/src/recon_test/test_blocks_on_cylindrical_projectors.cxx
+++ b/src/recon_test/test_blocks_on_cylindrical_projectors.cxx
@@ -38,7 +38,6 @@
 #include "stir/ProjDataInfoBlocksOnCylindricalNoArcCorr.h"
 #include "stir/ProjDataInfoGenericNoArcCorr.h"
 #include "stir/ProjDataInfoCylindricalNoArcCorr.h"
-#include "stir/ProjDataInfoCylindrical.h"
 #include "stir/recon_buildblock/ProjMatrixByBinUsingRayTracing.h"
 #include "stir/Sinogram.h"
 #include "stir/Viewgram.h"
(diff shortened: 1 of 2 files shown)


src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -53,6 +54,8 @@
 #include "stir/VoxelsOnCartesianGrid.h"
 #include "stir/recon_buildblock/ForwardProjectorByBin.h"
 #include "stir/recon_buildblock/ForwardProjectorByBinUsingProjMatrixByBin.h"
+#include "stir/recon_buildblock/Parallelproj_projector/ForwardProjectorByBinParallelproj.h"
+#include "stir/recon_buildblock/Parallelproj_projector/BackProjectorByBinParallelproj.h"
```
KrisThielemans, line 58: will have to be guarded by #ifdef

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -38,6 +38,7 @@
 #include "stir/ProjDataInfoBlocksOnCylindricalNoArcCorr.h"
 #include "stir/ProjDataInfoGenericNoArcCorr.h"
 #include "stir/ProjDataInfoCylindricalNoArcCorr.h"
+#include "stir/ProjDataInfoCylindrical.h"
```
KrisThielemans, line 41: implied by previous includes

KrisThielemans (changes requested):
there's more here than I hoped, so I haven't reviewed this properly.

markus-jehl:
Daniel's tests I only modified in that I introduced the templates and reordered them to be consistently ordered, because I was getting a bit confused which one was which. So the only real changes are the templates and the new test.

KrisThielemans (commented):
looks fine, and checks work!

I'm not sure though why you templated all functions in the projector type. As far as I can see, using a `shared_ptr<ForwardProjectorByBin>` would work as well. No templating required. Feel like changing that?

(In fact,there's no reason here to pass a `shared_ptr` and using `ForwardProjectorByBin&` would be slightly better class design here, but that kind of stuff doesn't matter in the test of course).

markus-jehl:
True, didn't think of that! Let me change that quickly.

markus-jehl:
Some images to illustrate results of projections (left: ray tracing; right: parallelproj)
image
image
image
voxel
image

KrisThielemans:
Excellent. Thanks!

There's also `test_view_offset_root.cxx` but feel free to decline fixing that now. (It's more important for the TOF branch later, so we could do that then).

markus-jehl:
That doesn't seem to use projectors, but work straight on `ProjMatrixByBinUsingRayTracing`. I wouldn't know how to test this with parallelproj without digging into the code for a bit (or you telling me ;-) ).

danieldeidda (approved):
looks good

PR #1120, opened 2023-01-02
Listmode "random" renaming to "delayed" for SAFIR and ECAT

### LINKED ISSUE #1064: confusing naming of isRandom in SAFIR listmode class
One of the elements in the SAFIR file is called `isRandom` https://github.com/UCL/STIR/blob/4d087f3e7cf3c6d2d95b87f64e9e68bf9f1b186e/src/include/stir/listmode/CListRecordSAFIR.h#L64
This seems a confusing name as it indicates if it's a delayed coincidendence or not.

@jafische would you agree with this?

KrisThielemans:
Fixes #1064

@jafische @markus-jehl @emikhay maybe you can have a quick look? The example still runs fine.

### DIFF
--- a/documentation/release_5.1.htm
+++ b/documentation/release_5.1.htm
@@ -117,7 +117,11 @@ improvements to the documentation.
 </ul>
 
 <h3>Changed functionality</h3>
-
+<ul>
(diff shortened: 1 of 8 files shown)


markus-jehl:
Looks good to me - the new wording definitely makes more sense.

PR #1122, opened 2023-01-04
adding utility that extracts interfile+binary from SPECT dicom data

danieldeidda:


src/utilities/CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ set(${dir_EXE_SOURCES}
         write_sinogram_to_txt.cxx
         find_sum_projection_of_viewgram_and_sinogram.cxx
         separate_true_from_random_scatter_for_necr.cxx
+        SPECT_dicom_to_interfile.cxx
```
KrisThielemans, line 76: probably within `HAVE_ITK` or so

src/utilities/SPECT_dicom_to_interfile.cxx
```diff
@@ -0,0 +1,505 @@
+#include <iostream>
```
KrisThielemans, line 1: add usual copyright header

KrisThielemans (changes requested):
will need to add it to release notes

KrisThielemans:
@danieldeidda how does this handle multiple energy window data. Is there one DICOM file per energy window?

danieldeidda:
I have never seen a case with multiple energy windows data in the same file in our case we usually have different dicom files but I am pretty sure the dicom file has the information if the acquisition was performed with multiple energy windows

KrisThielemans:
Replaced by #1182

PR #1126, opened 2023-01-07
[GHA] increase ccache size

KrisThielemans:
Debug builds have hardly any cache hits. That's likely because the size is too small.


KrisThielemans:
ubuntu Debug build takes now ~9m as opposed to ~52m

PR #1127, opened 2023-01-08
[Appveyor] attempt to use sccache

KrisThielemans:


KrisThielemans:
This fails due to (unrelated) linking errors with OpenMP. Apparently the linker doesn't get passed the correct flag. This is likely caused by switching to Ninja, which I had to do as the VS Generator ignores the `*COMPILER_LAUNCHER` (I guess I could try `nmake`). But  this is using the `clang` compiler as opposed to `cl`. so lots of things change. 

errors [like](https://ci.appveyor.com/project/KrisThielemans/stir/builds/45871849#L31654)
```
lld-link: error: undefined symbol: __kmpc_dispatch_init_8
>>> referenced by iterative_KOSMAPOSL.lib(KOSMAPOSLReconstruction.cxx.obj):(.omp_outlined.)
>>> referenced by iterative_KOSMAPOSL.lib(KOSMAPOSLReconstruction.cxx.obj):(.omp_outlined..1)
```
and 
```
lld-link: error: undefined symbol: omp_get_num_threads
>>> referenced by recon_buildblock.lib(PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx.obj):(.omp_outlined.)
>>> referenced by recon_buildblock.lib(PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx.obj):(.omp_outlined..1)
```

I don't think this is our problem, but CMake's FindOpenMP.

KrisThielemans:
Stuck due to #1129.

In any case, CMake still doesn't use `ccache` nor `sccache` with the VS Generator.

KrisThielemans:
Obsolete as using `ccache` since #1643

PR #1141, opened 2023-01-15
[CMake] Use an object library for STIR_REGISTRIES

### LINKED ISSUE #83: speed-up building using object libraries
Since v2.8.8, CMake supports "libraries" with object files, see https://cmake.org/Wiki/CMake/Tutorials/Object_Library

This would enable us to prevent recompilation of the registry files for every executable. We might even use this to reduce the number of libraries and therefore solve #6.

It isn't entirely obvious how to do this while being backwards compatible though. We currently rely on the CMake variable `STIR_REGISTRIES` to let anyone add to the list of registries (via the `STIR_LOCAL` mechanism), and that variable is exported as well (while object libraries cannot be exported).

KrisThielemans:
Using an object library means that the registries get compiled only once, therefore speeding up the build process quite a lot.

It's also easier to use when building a project that uses STIR as we just need to depend on `STIR_LIBRARIES`. However, we do set the STIR_REGISTRIES variable empty at export, such that external libraries can still use
```cmake
  add_executable(bla ${STIR_REGISTRIES})
```

Fixes #83


KrisThielemans:
This worked and resulted in great speed-up. Sadly, it doesn't work for SIRF as its executables now do not have the stir registries at all. The reason is that `target_link_libraries` doesn't use transitive behaviour for `OBJECT` libraries. Extensive discussion at https://gitlab.kitware.com/cmake/cmake/-/issues/18090

(I also have some linking errors with the current STIR commit related to HDF5, but not sure if these are related)

KrisThielemans:
The previous commit re-enables the requirement to use `${STIR_REGISTRIES}`  in external projects to avoid problems with the transitive behaviour. `STIR_REGISTRIES is now set (in `STIRConfig.cmake`) to `$<TARGET_OBJECTS:stir_registries>` which works fine on my Linux VM. Fingers crossed.

KrisThielemans:
Currently, the exported `STIRConfig.cmake` will either work for CMAKE < 3.12 or > 3.12, but not both.

KrisThielemans:
This is finished (aside from release notes). @danieldeidda @markus-jehl @robbietuk @NikEfth could you try this one out please? Especially if you have something that uses `find_package(STIR)`.

markus-jehl:
Just tested it and it works fine, as far as I can tell.

PR #1150, opened 2023-02-06
Update ParallelProj minimum version

samdporter:
v1.2.13 has bug fix in https://github.com/gschramm/parallelproj/blob/master/c/include/parallelproj_c.h \\
Bug causes error in STIR compilation.


samdporter:
I guess this parallelproj version is very new. I'm a bit confused as to why this problem hasn't previously occurred

KrisThielemans:
we were using 0.8 before. Only recently switched to 1.x.

I'm surprised that it gives problems on `master` as I didn't think the tof stuff would be included.

samdporter:
I've double checked the cmake options for my SIRF build and STIR_TAG is definitely origin/master

KrisThielemans:
currently failing as we test with
https://github.com/UCL/STIR/blob/593c06909201419716398d699211614eedd93df1/.github/workflows/build-test.yml#L151

gschramm:
Hi Sam, hi Kris,

1. The errors I see in the build logs, seem more like azure connection errors to me. Or did I miss an error before that?
2. The ";" vs "," bug in the header file, was indeed fixed in v1.0.1
3. the API of v1.2.13 should be backward compatible with v1.0.1. I try to stick to semantic version numbers, meaning that if  the API changes, I increase the major version number. After the latest discussions with @markus-jehl, the build with v1.2.13 should be fine.
4. You don't need ```git clone``` anymore to build ```parallelproj``` with the correct version number, since the version number is now read from [a file](https://github.com/gschramm/parallelproj/blob/master/package.json) that gets auto-bumped by ```ver-bump``` before I release (tag) a new version.

KrisThielemans:
@samdporter please change to require 1.0.1. It might be that CMake ignores the patch, but at least it's clearer for the developer what we need.

CMakeLists.txt
```diff
@@ -197,16 +197,16 @@ endif()
 
 # Parallelproj
 if(NOT DISABLE_Parallelproj_PROJECTOR)
-    find_package(parallelproj 1.0 CONFIG)
+    find_package(parallelproj CONFIG)
```
KrisThielemans, line 200: I guess still best to keep 1.0 here. why delete it? If anything, I'd put 1.0.1 here.

KrisThielemans (changes requested):

CMakeLists.txt
samdporter: A misread of CMake find_package documentation. Fixed (I think)

Issue #1151, opened 2023-02-06
Scatter simulation issue when setting max_num_non_arccorrected bins to 0


markus-jehl:
The scatter simulation for BlocksOnCylindrical usually downsamples the scanner in axial direction, e.g. when using a scanner with 30 rings the downsampled projection data contain 7 planes:
![image](https://user-images.githubusercontent.com/87262701/216980242-1e3cd585-4b00-4541-883c-744b3b2e9d01.png)
However, if for the exact same scanner geometry the arccorrected bins variable is set to 0 (arc correction is set to false, so this variable should not be used), then the output of the scatter simulation is massively downsampled in tangential direction as well:
![image](https://user-images.githubusercontent.com/87262701/216980539-067d6090-cb98-4aae-952f-88e6e2b444fc.png)

markus-jehl:
My bad - I read the first "arccorrected" parameter to scanner definition wrongly. It actually refers to non-arc-corrected bins, so shouldn't be set to 0. In discussions, it was decided to slightly modify the way the scanner is downsampled in SimulateScatter anyways: see [PR ](https://github.com/UCL/STIR/pull/1152).

PR #1152, opened 2023-02-06
Fix for scatter simulation with respect to how max_num_non_arccorrected bins are used

markus-jehl:
Fix for scatter simulation. The issue was that the arccorrected bin number was used to initialise the proj_data_info. Instead, the actual proj_data_info tangential bin number should be used.

### DIFF
--- a/documentation/release_5.2.htm
+++ b/documentation/release_5.2.htm
@@ -63,6 +63,9 @@
 
 <h3>Minor (?) bug fixes</h3>
 <ul>
+  <li>Small change in scatter simulation to how non-arccorrected bins are computed. Added a check in the construction of non-arccorrected projdata that the number of tangential bins is not larger than the maximum non-arccorrected bins.
+    <br /><a href="https://github.com/UCL/STIR/pull/1152/">PR #1152</a>.
+  </li>
 </ul>
 
 <h3>Documentation changes</h3>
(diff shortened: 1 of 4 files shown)


markus-jehl:
@danieldeidda Could you have a look at this as well? One of the tests fails with my proposed change, but I'm not sure what's wrong.

markus-jehl:
Related to https://github.com/UCL/STIR/issues/1151.

KrisThielemans:
Note for others: above description is inaccurate, as based on a mis-reading of the name of the variables. I've changed the title of the PR to more accurately reflect what it does.

KrisThielemans:
@markus-jehl can you add a line to the release notes? This change looks innocuous, but anything with scatter is always surprising, so best to make it explicit.

Will you address the `set_num_tangential_poss` at all, in this PR, or a different PR?

markus-jehl:
Let me have a look at it as part of this one - would be nice to fix it in one go!

markus-jehl:
I will just add a check in the right constructors to check for max_num_non_arccorrected_bins. The setting for arccorrected bins does not have a "max" in its name - does it mean that there we should allow setting more tangential positions than arccorrected bins on the scanner?

KrisThielemans:
> I will just add a check in the right constructors to check for max_num_non_arccorrected_bins. 

ok. I think the strategy of making overloading `set_num_tangential_poss` wouldn't work anyway, as I think the `ProjDataInfo` constructor would be calling its own version anyway, see also https://stackoverflow.com/a/8630215/15030207

> The setting for arccorrected bins does not have a "max" in its name - does it mean that there we should allow setting more tangential positions than arccorrected bins on the scanner?

yes, it's a `default` (used for in arc-correction). I guess we could check if it's smaller than the max ring radius, but I think it's overkill as we're hardly using this at all.

PR #1156, opened 2023-02-08
Scatter Estimation / Sinogram interpolation in 3D

NikEfth:
This is in good shape. I have tested with data from various scanners. 
But more testing is needed and updates in the documentation and such. 

@KrisThielemans I am not sure if you would like to change the recon_tests to test this in addition to the existing one.

### DIFF
--- a/src/buildblock/interpolate_projdata.cxx
+++ b/src/buildblock/interpolate_projdata.cxx
@@ -532,338 +532,347 @@ interpolate_projdata_3d(ProjData& proj_data_out,
                         const BSpline::BSplineType& spline_type,
                         const bool use_view_offset)
 {
-//   if (use_view_offset)
-//     warning("interpolate_projdata with use_view_offset is EXPERIMENTAL and NOT TESTED.");
+  //   if (use_view_offset)
+  //     warning("interpolate_projdata with use_view_offset is EXPERIMENTAL and NOT TESTED.");
 
-//   const shared_ptr<const ProjDataInfo> projdata_in_info_sptr = projdata_in_sptr->get_proj_data_info_sptr();
+  //   const shared_ptr<const ProjDataInfo> projdata_in_info_sptr = projdata_in_sptr->get_proj_data_info_sptr();
 
-//   shared_ptr<ProjDataInfo> projdata_in_up_info_sptr(projdata_in_sptr->get_proj_data_info_sptr()->clone());
-//   projdata_in_up_info_sptr->set_num_views(projdata_out->get_num_views());
-//   projdata_in_up_info_sptr->set_num_tangential_poss(projdata_out->get_num_tangential_poss());
+  //   shared_ptr<ProjDataInfo> projdata_in_up_info_sptr(projdata_in_sptr->get_proj_data_info_sptr()->clone());
+  //   projdata_in_up_info_sptr->set_num_views(projdata_out->get_num_views());
+  //   projdata_in_up_info_sptr->set_num_tangential_poss(projdata_out->get_num_tangential_poss());
 
-//   if (projdata_in_info_sptr->get_scanner_sptr()->get_scanner_geometry() == "BlocksOnCylindrical")
-//     {
-//       // interpolate_axial_position(*projdata_out, *projdata_in_sptr);
-//       return Succeeded::yes;
-//     }
+  //   if (projdata_in_info_sptr->get_scanner_sptr()->get_scanner_geometry() == "BlocksOnCylindrical")
+  //     {
+  //       // interpolate_axial_position(*projdata_out, *projdata_in_sptr);
+  //       return Succeeded::yes;
+  //     }
 
-//   // check for the same ring radius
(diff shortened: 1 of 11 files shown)


src/buildblock/interpolate_projdata.cxx
```diff
@@ -313,4 +318,348 @@ interpolate_projdata(ProjData& proj_data_out,
...
+                BSpline::BSplinesRegularGrid<2, float, float> proj_data_interpolator(these_types2);
+                for (int y=extended.get_min_index(); y<= extended.get_max_index(); ++y)
+                {
+                    const int old_min = extended[y].get_min_index();
+                    const int old_max = extended[y].get_max_index();
+                    extended[y].grow(old_min-1, old_max+1);
+                    extended[y][old_min-1] = extended[y][old_min];
+                    extended[y][old_max+1] = extended[y][old_max];
+                }
+                std::cout << "n7" << std::endl;
```
KrisThielemans, line 440: these will have to go

src/buildblock/interpolate_projdata.cxx
```diff
@@ -286,9 +291,9 @@ interpolate_projdata(ProjData& proj_data_out,
     proj_data_interpolator.set_coef(extended);
   }
   else
-  {
-    Array<3,float> extended = 
-      extend_segment_in_views(proj_data_in.get_segment_by_sinogram(0), 2, 2);
+  { //TODO: be removed ...
```
KrisThielemans, line 294: ??

src/buildblock/interpolate_projdata.cxx
```diff
@@ -170,7 +174,8 @@ using namespace detail_interpolate_projdata;
   
 Succeeded 
 interpolate_projdata(ProjData& proj_data_out,
-                     const ProjData& proj_data_in, const BSpline::BSplineType these_types,
+                     const shared_ptr<ProjData> proj_data_in,
```
KrisThielemans, line 177: why did you change all these types to `shared_ptr`? It is recommended C++ practice to pass references unless you really need the pointer (e.g. because the function needs to store it after the function exists). Reasons include:
- clearer interface: the user knows you don't need access afterwards
- less dependency on `shared_ptr` (I could have a `unique_ptr` for instance, which I wouldn't be able to pass).

Of course, STIR isn't quite consistent in this choice, and it is confusing that we mix conventions. But I see no good reason to change the calling conventions here (certainly as `proj_data_out` is still a reference), but I might be missing something.

If we do need to change it, it should be a `const shared_ptr<const ProjData>`.

KrisThielemans (changes requested):
Some initial comments without actual review of the "working" part of the code.

Sadly, there's white-space changes here, which seems a bit arbitrary. I'd prefer to revert these. If there are any, they should be compatible with our `clang-format`. (we will need to run this through all of STIR, but the TOF PR needs to catch up first)

NikEfth:
> Sadly, there's white-space changes here, which seems a bit arbitrary. I'd prefer to revert these. If there are any, they should be compatible with our clang-format.

Where is the ```.clang-format``` file?

KrisThielemans:
> Where is the `.clang-format` file?

in the STIR root 😄

KrisThielemans:
Sadly, #1172 will generate a lot of conflicts with this PR., and I need to merge that first (it fixes a bug). However, as #1172 simplifies the logic of the scatter umsampling, hopefully it isn't too hard to take it into account here. 

Sorry, a case of development in the same place at the same time.

NikEfth:
Ok, no worries. Let me know when to pull it or should I wait for the master?

KrisThielemans:
better to wait till it's on master I think. @markus-jehl is going to finish this soon.

markus-jehl:
Also, let me know if you have questions about the changes I made, @NikEfth.

NikEfth:
Is this PR still useful or completely superseded?

KrisThielemans:
I don't think it is completely obsolete. Certainly, there's nothing on `master` that does 3D upsampling.

It'd be great to have a 3D upsampler of course, but I haven't check the current approach nor do I know what other things are going on in Groningen, but I suppose you do.

NikEfth:
Not sure what you mean about other things in Groningen

NikEfth:
@Aparna1403J @KrisThielemans This is moving forward quite well I think. 
Fixed extending the projdata in obligue segments following @markus-jehl modifications and the first step of the 3D interpoloation which is very similar to what the non-3D does but for obligue segments, too.

Next step the michelogram interpolation and this will be finished.

NikEfth:
image

As tests are starting piling up I want to paste here a comparison of the output of ```scatter_interpolation_test_cyl_asymmetric()```. This is the 28th (out of 29) segment (1st axial position). 
On the left is the full projdata. In the middle is the 3D upsampling / interpolation and on the right going through 2D interpolation and inverse SSRB.

NikEfth:
Added support for span. Seems to be working

Issue #1167, opened 2023-02-28
Methods in ML_norm.cxx are not parallelised yet


markus-jehl:
The methods in ML_norm.cxx take much longer than they should, because they do not leverage OpenMP yet.

PR #1168, opened 2023-02-28
Parallelising fan-data to proj-data function in ML_norm.cxx 

markus-jehl:
Relates to issue: https://github.com/UCL/STIR/issues/1167

### DIFF
--- a/src/buildblock/ML_norm.cxx
+++ b/src/buildblock/ML_norm.cxx
@@ -28,6 +28,10 @@
 #include "stir/SegmentBySinogram.h"
 #include "stir/stream.h"
 
+#ifdef STIR_OPENMP
+#  include <omp.h>
+#endif
+
 #include <algorithm>
 using std::min;
 using std::max;
@@ -1087,12 +1091,15 @@ static void make_fan_data_remove_gaps_help(FanProjData& fan_data,
     const int num_physical_rings = num_rings - (num_axial_blocks-1)*num_virtual_axial_crystals_per_block;
     fan_data = FanProjData(num_physical_rings, num_physical_detectors_per_ring, new_max_delta, 2*new_half_fan_size+1);
 
-
-    shared_ptr<SegmentBySinogram<float> > segment_ptr;
-    Bin bin;
-
-    for (bin.segment_num() = proj_data.get_min_segment_num(); bin.segment_num() <= proj_data.get_max_segment_num();  ++ bin.segment_num())
+#ifdef STIR_OPENMP
+  #pragma omp parallel for schedule(dynamic)
+#endif
+    for (int segment = proj_data.get_min_segment_num(); segment <= proj_data.get_max_segment_num(); ++segment)
     {
+        shared_ptr<SegmentBySinogram<float> > segment_ptr;
+        Bin bin;
+        bin.segment_num() = segment;
+
         segment_ptr.reset(new SegmentBySinogram<float>(proj_data.get_segment_by_sinogram(bin.segment_num())));
 
         for (bin.axial_pos_num() = proj_data.get_min_axial_pos_num(bin.segment_num());
@@ -1194,11 +1201,15 @@ static void set_fan_data_add_gaps_help(ProjData& proj_data,
 
     const int num_physical_axial_crystals_per_block = num_axial_crystals_per_block - num_virtual_axial_crystals_per_block;
 
-    Bin bin;
-    shared_ptr<SegmentBySinogram<float> > segment_ptr;
-
-    for (bin.segment_num() = proj_data.get_min_segment_num(); bin.segment_num() <= proj_data.get_max_segment_num();  ++ bin.segment_num())
+#ifdef STIR_OPENMP
+  #pragma omp parallel for schedule(dynamic)
+#endif
+    for (int segment = proj_data.get_min_segment_num(); segment <= proj_data.get_max_segment_num(); ++segment)
     {
+        Bin bin;
+        shared_ptr<SegmentBySinogram<float> > segment_ptr;
+        bin.segment_num() = segment;
+
         segment_ptr.reset(new SegmentBySinogram<float>(proj_data.get_empty_segment_by_sinogram(bin.segment_num())));
 
         for (bin.axial_pos_num() = proj_data.get_min_axial_pos_num(bin.segment_num());
@@ -1348,7 +1359,9 @@ void apply_geo_norm(FanProjData& fan_data, const GeoData3D& geo_data, const bool
     FanProjData work = fan_data;
     work.fill(0);
     
-    
+#ifdef STIR_OPENMP
+  #pragma omp parallel for schedule(dynamic)
+#endif
     for (int ra = 0; ra < num_axial_crystals_per_block; ++ra)
         for (int a = 0; a < num_transaxial_crystals_per_block /2; ++a)
             // loop rb from ra to avoid double counting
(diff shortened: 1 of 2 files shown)


markus-jehl:
Before the change, the following calls were timed as:
- make_fan_data_remove_gaps: 2.8 s
- apply_efficiencies: 0.7 s
- apply_geo_norm: 2.1 s
- set_fan_data_add_gaps: 3.1 s

After the change, they were much faster:
- make_fan_data_remove_gaps: 1 s
- apply_efficiencies: 0.1 s
- apply_geo_norm: 0.4 s
- set_fan_data_add_gaps: 0.6 s

src/buildblock/ML_norm.cxx
```diff
@@ -1399,7 +1412,10 @@ void apply_geo_norm(FanProjData& fan_data, const GeoData3D& geo_data, const bool
                         
                     }
                 }
-    
+
+#ifdef STIR_OPENMP
+  #pragma omp parallel for schedule(dynamic)
```
danieldeidda, line 1417: You can probably use collapse(2) in this case it should parallelise the inner loop as well

markus-jehl: cool, will give it a go!

KrisThielemans: It will need to be guarded with OpenMP version as in other places, as Visual Studio doesn't support it.

KrisThielemans (commented):
I haven't checked but this is probably fine. The tricky (i.e. virtually impossible) one would be `iterate_efficiencies` as that updates efficiencies continuously in the loop. We tried a "one step late" scheme but it doesn't converge (see the old proceedings by Darren Hogg).

src/buildblock/ML_norm.cxx
markus-jehl: Done.

danieldeidda: Does it speed it up further?

markus-jehl: Not noticeably, to be honest. But I thought it can't hurt to use it.

KrisThielemans (changes requested):
Unfortunately, I think a lot of this is not thread-safe. Lines like
https://github.com/UCL/STIR/blob/fda27e0c8018a6fa0a531ddf3d5e2fb6755ed89c/src/buildblock/ML_norm.cxx#L1138-L1139
are not guaranteed to work. See for instance https://stackoverflow.com/a/41614045/15030207

Unfortunately, using `atomic` will likely not work for lines like that as they go via a class member, but only for vector access. Even then they will be version specific (but I'd be entirely fine only parallelising for compilers that support recent enough openmp). If `atomic` doesn't work, I think it'll need a critical section, essentially killing the speed-up. It would then need creating writable variables for every thread (which would probably break all encapsulation of `FanData` etc.

For the loops that use `set_segment`, it would be alright, as each segment is independent per thread, and `set_segment` is thread-safe.

markus-jehl:
Your comment makes sense. But the sections I've parallelised should be fine, I think. As you say, parallelising on segments should be alright. And in the places where we parallelise over ra and a, the threads only access entries at the ra and a locations. The only one that looks problematic is the first of the three collapse(2) parallelisations: there we indeed compute new indices and use them to write into the "work" variable.

KrisThielemans:
not sure if I agree. The lines I quoted are in the 1st loop and update `new_fan_data`. They are problematic. There's a loop [here](https://github.com/UCL/STIR/blob/fda27e0c8018a6fa0a531ddf3d5e2fb6755ed89c/src/buildblock/ML_norm.cxx#L1399) (3rd one?) which updated `work`. The [next one](https://github.com/UCL/STIR/blob/fda27e0c8018a6fa0a531ddf3d5e2fb6755ed89c/src/buildblock/ML_norm.cxx#L1437) updates `fan_data`. No?

markus-jehl:
Yes, the lines you originally quoted are in the first loop, but that loop only parallelises across segments. The first new link you sent is the one I also think is problematic, but the second link I think is fine again because it does update fan_data, but only in location (ra, a, ...) therefore ensuring that each thread writes a different section of fan_data.

KrisThielemans:
but updating `fan_data` is just as problematic, certainly see as we update it symmetrically (`fan_data(new_ra, new_a, new_rb, new_b) =  fan_data(new_rb, new_b, new_ra, new_a) = ...`). How do we know that another thread isn't access the "symmetric" version, or just the neghbouring one, with therefore the potential for memory corruption?

It might be alright as all the index access in the 4D array is read-only, and we only update the 1D vectors, but how do know that those 1D arrays are not adjacent for different segments?

Of course, it seems pretty unlikely that this would generate a race-condition, but I don't think we have a solid guarantee.

I don't have access to any tools to check thread-safety sadly.

markus-jehl:
Darn, I only noticed this symmetry thing now. I ran TSAN over it and even the parallelisation across segments seems to be not thread-safe.

KrisThielemans:
By the way, how reliable is TSAN (presumably with clang) for OpenMP? (I've tried it with gcc, but I had to build by own instrumented openmp library (or was it gcc?), so gave up).

markus-jehl:
I haven't used it with OpenMP before, and the more I look into it here the less I trust it. I think it might be getting confused with the various layers of bound checking and shared pointers when obtaining segments by sinogram.

KrisThielemans:
yeah.. by the way, there's no reason for this code
```c++
 shared_ptr<SegmentBySinogram<float> > segment_ptr;
 segment_ptr.reset(new SegmentBySinogram<float>(proj_data.get_segment_by_sinogram(bin.segment_num())));
```
the following is essentially equivalent but much clearer
``c++
   const auto segment(proj_data.get_segment_by_sinogram(bin.segment_num()));
```
(I'm hoping I didn't write those lines)

Note that this is going to make TSAN problems disappear.

markus-jehl:
Finally I managed to get OpenMP working with TSAN for a very simple dummy for loop just printing something to cout. Then parallelising the least problematic for loop over segments where only the segment is modified, it already throws a lot of warnings in TSAN. Furthermore, using "atomic read" on the fan_data doesn't even compile, because it expects this to work on simple operations such as "v=x", not on complex functions. Even on the "[]"-style indexing in VectorWithOffset it complained.

I'm afraid we'll have to live with the slow serial implementation for now...

KrisThielemans:
> I managed to get OpenMP working with TSAN for a very simple dummy for loop just printing something to cout.

well, that's a bit strange. Writing to `cout` is most definitely not thread-safe with `critical` section, so it should have complained! I've tried to find some doc on clang/TSAM/OpenMP but gave up. The only info i could find (but it's old) is that you need to build your own libgomp, and then `LD_PRELOAD` it.

>  "atomic read" on the fan_data doesn't even compile, because it expects this to work on simple operations such as "v=x", not on complex functions. Even on the "[]"-style indexing in VectorWithOffset it complained

I'm not surprised by the `atomic` restrictions for `fan_data`, but am disappointed it cannot handle `VectorWithOffset::operator[]`. Does it work with `std::vector::operator[]` by the way?

We don't need atomic reads anywhere. They are only necessary when someone else can write to that memory. So, in the second loop, we read `fan_data` in parallel (fine), and write to a thread-local segment (fine), and then call `set_segment`, which should have its own internal critical section., so should be fine as well. You can always check by adding your own critical section for the `set_segment`.

Updating `fan_data` though is harder.

markus-jehl:
Yes, it's all very mysterious.

> I've tried to find some doc on clang/TSAM/OpenMP but gave up. The only info i could find (but it's old) is that you need to build your own libgomp, and then `LD_PRELOAD` it.

Yes, this is how I got it to work eventually.

> I'm not surprised by the atomic restrictions for fan_data, but am disappointed it cannot handle VectorWithOffset::operator[]. Does it work with std::vector::operator[] by the way?
> We don't need atomic reads anywhere. They are only necessary when someone else can write to that memory. So, in the second loop, we read fan_data in parallel (fine), and write to a thread-local segment (fine), and then call set_segment, which should have its own internal critical section., so should be fine as well. You can always check by adding your own critical section for the set_segment.

I can try this a bit later this week!

markus-jehl:
As discussed, std::vector access works, but VectorWithOffset doesn't. There seems to be no way to use atomic, and TSAN still complains about all parallelisation (therefore TSAN can't be trusted).

I have now also tried to add an atomic_read function to FanData that has a critical section on the read, but that kills performance thoroughly.

Therefore, the best option is to parallelise only the loop that looks safe for now.

KrisThielemans (changes requested):
looks good now. can you just add something to the release notes? Thanks!

PR #1169, opened 2023-03-02
 changing SPECTUB global variables to class members

danieldeidda:


src/include/stir/recon_buildblock/ProjMatrixByBinSPECTUB.h
```diff
@@ -172,6 +171,9 @@ class ProjMatrixByBinSPECTUB :
 
   bool already_setup;
 
+  static wm_da_type wm;
```
KrisThielemans, line 174: I think this has to be
```
static SPECTUB::wm_da_type wm;
```

src/include/stir/recon_buildblock/ProjMatrixByBinSPECTUB.h
```diff
@@ -75,7 +75,7 @@ class Bin;
 End Projection Matrix By Bin SPECT UB Parameters:=
 \endverbatim
 */
-
+using namespace SPECTUB;
```
KrisThielemans, line 78: remove `using` here as we need to explicitly state it below

src/recon_buildblock/ProjMatrixByBinSPECTUB.cxx
```diff
@@ -60,14 +62,14 @@
...
-}
+//namespace SPECTUB {
+//  wm_da_type wm;
+//  wmh_type wmh;
+//  float * Rrad;
+//}
 
 START_NAMESPACE_STIR
-
+using namespace SPECTUB;
```
KrisThielemans, line 72: if this doesn't work, put it above the `START_NAMESPACE_STIR`

src/include/stir/recon_buildblock/ProjMatrixByBinSPECTUB.h
```diff
@@ -206,7 +208,10 @@ class ProjMatrixByBinSPECTUB :
   int maxszb;
 
 	
-  void compute_one_subset(const int kOS) const;
+  void compute_one_subset(const int kOS,
+                          wm_da_type& wm,
```
KrisThielemans, line 212: would need `SPECTUB::` everywhere

KrisThielemans (commented):
remove comment in https://github.com/UCL/STIR/blob/94efb22e2a2e1d6974e282a6c0d0ca6012eb667f/src/include/stir/recon_buildblock/SPECTUB_Weight3d.h#L27

remove extern declarations at https://github.com/UCL/STIR/blob/94efb22e2a2e1d6974e282a6c0d0ca6012eb667f/src/include/stir/recon_buildblock/SPECTUB_Weight3d.h#L38-L40

Will need to disable the copy-constructor which is something like
```c++
// disable copy-constructor as currently unsafe to copy due to bare pointers
ProjMatrixSPECTUB(const &ProjMatrixSPECTUB) = delete;
```

Ideally (in next step) make `Rrad` and `std::vector` such that we don't need to deallocate etc. Can then be passed to SPECTUB functions as `&Rrad[0]`, which is of course  a bit ugly... So we can leave that for later.

danieldeidda:
> Will need to disable the copy-constructor which is something like
> 
> ```c++
> // disable copy-constructor as currently unsafe to copy due to bare pointers
> ProjMatrixSPECTUB(const &ProjMatrixSPECTUB) = delete;
> ```

Where should this go I am guessing in the ProjMatrixByBinSPECTUB.h?

KrisThielemans:
notice a syntax error. that'll likely have to be
```c++
ProjMatrixSPECTUB(const ProjMatrixSPECTUB&) = delete;
```

src/include/stir/recon_buildblock/ProjMatrixByBinSPECTUB.h
danieldeidda: removing statis causes the following:
`
ProjMatrixByBinSPECTUB.cxx:879: error: binding reference of type ‘SPECTUB::wm_da_type&’ to ‘const SPECTUB::wm_da_type’ discards qualifiers
ProjMatrixByBinSPECTUB.cxx: In member function ‘virtual void stir::ProjMatrixByBinSPECTUB::calculate_proj_matrix_elems_for_one_bin(stir::ProjMatrixElemsForOneBin&) const’:
ProjMatrixByBinSPECTUB.cxx:879:30: error: binding reference of type ‘SPECTUB::wm_da_type&’ to ‘const SPECTUB::wm_da_type’ discards qualifiers
  879 |       compute_one_subset(kOS,wm,wmh,Rrad);
                                    ^~
ProjMatrixByBinSPECTUB.cxx:741:32: note:   initializing argument 2 of ‘void stir::ProjMatrixByBinSPECTUB::compute_one_subset(int, SPECTUB::wm_da_type&, SPECTUB::wmh_type&, float*) const’
  741 |                    wm_da_type& wm,
      |                    ~~~~~~~~~~~~^~ '

KrisThielemans: hmmm. probably a conflict between the member variable and the argument. It is of course a bit confusing now. I guess either rename arguments and where they are used, or remove arguments and use member variables. Maybe the latter is more logical after all. In that case, it seems `compute_one_subset` can no longer be `const`. If it has to be, then we'd have to make the members `mutable`.

KrisThielemans:
let's try to make `float * Rrad` a `const float * Rrad` where possible

KrisThielemans:
Linking errors
```
SPECTUB_Tools.cxx:(.text+0x3b0a): undefined reference to `SPECTUB::wmh
```
so some functions are still expecting the global varaibles

KrisThielemans (changes requested):
Great! It is working fine.

Can you clean-up by removing comments, and adding something to `release_5.2.htm`? (along the lines of "removed global variables in the SPECTUB code, such that we can now have multiple SPECTUB projectors (PR #1159)"

KrisThielemans:
@samdporter want to try this out? it should work now.

KrisThielemans:
@mastergari we'll have to do the same for the pinhole code presumably. can do that later.

samdporter:
> @samdporter want to try this out? it should work now.

Looks good with the tests in [here](https://github.com/samdporter/SPECTissues).

I'll do something a bit more difficult and let you know

Sam

KrisThielemans (approved):
great. @samdporter let us know when you're happy.

Ok to squash-merge?

samdporter:
> great. @samdporter let us know when you're happy.
> 
> Ok to squash-merge?

STIR happy to set up multiple projectors (so an improvement!) but unhappy when setting up multiple `OSMAPOSLReconstruction3DFloat `objects.

I'll update the repo with a failing script and copy the valgrind output here

Issue #1173, opened 2023-03-10
SingleScatterSimulation output not consistent downsample_scanner_bool's status


Robert-PrescientImaging:
I believe this is related to the status of `downsample_scanner_bool`. My testing python script:

```
def simulate_scatter_test(template_projdata: stir.ProjData, true_image: stir.FloatVoxelsOnCartesianGrid,
                          attenuation_image: stir.FloatVoxelsOnCartesianGrid,
                          downsample_scanner=False) -> stir.ProjData:

    scattered_projdata = stir.ProjDataInMemory(template_projdata.get_exam_info(),
                                               template_projdata.get_proj_data_info())
    s = stir.SingleScatterSimulation()
    s.set_template_proj_data_info(scattered_projdata.get_proj_data_info())
    s.set_exam_info(scattered_projdata.get_exam_info())

    s.set_output_proj_data_sptr(scattered_projdata)
    s.set_activity_image_sptr(true_image)
    s.set_density_image_sptr(attenuation_image)

    s.set_downsample_scanner_bool(downsample_scanner)
    s.set_up()
    s.process_data()

    print(f"Scattered projdata with downsample_scanner = {downsample_scanner}:\n"
          f"\t{scattered_projdata},\n"
          f"\tsum: {scattered_projdata.to_array().sum()}")

    print(f"Output projdata downsample_scanner = {downsample_scanner}:\n"
          f"\t{s.get_output_proj_data()},\n"
          f"\tsum: {s.get_output_proj_data().to_array().sum()}\n")


def main():
    projdata, image = construct_projdata_and_image_for_test()
    for downsample_bool in [True, False]:
        simulate_scatter_test(projdata, image, image, downsample_bool)


if __name__ == '__main__':
    main()
```

I slightly modified STIR's SingleScatterSimulation class to add `set_downsample_scanner_bool(arg)` (PR incoming). The output from the two print statements is:

```
Scattered projdata with downsample_scanner = True: 
    <Swig Object of type 'stir::shared_ptr< stir::ProjDataInMemory > *' at 0x7fae873456c0>, 
    sum: 0.0
Output projdata downsample_scanner = True: 
    <Swig Object of type 'stir::shared_ptr< stir::ProjData > *' at 0x7fadaeeedf10>, 
    sum: 737.7693481445312

Scattered projdata with downsample_scanner = False: 
    <Swig Object of type 'stir::shared_ptr< stir::ProjDataInMemory > *' at 0x7fadaeefad88>, 
[...]

PR #1174, opened 2023-03-10
Scatter simulation: extra set/get methods

Robert-PrescientImaging:
As mentioned in #1173, does not currently fix that issue.

PR adds get/set method for protected/private members:
- `downsample_scanner_bool`, 
- `num_downsample_scanne…r_rings` and 
- `num_downsample_scanner_dets`

The set methods will set the `_already_set_up` to false (if change), requiring a new `set_up()` call.

### DIFF
--- a/src/scatter_buildblock/ScatterSimulation.cxx
+++ b/src/scatter_buildblock/ScatterSimulation.cxx
@@ -828,6 +828,61 @@ set_exam_info_sptr(const shared_ptr<const ExamInfo> arg)
     this->template_exam_info_sptr = arg->create_shared_clone();
 }
 
+
+bool 
+ScatterSimulation::
+get_downsample_scanner_bool() const
+{
+    return this->downsample_scanner_bool;
+}
+
+void 
+ScatterSimulation::
+set_downsample_scanner_bool(const bool arg)
+{
+    if (arg != this->get_downsample_scanner_bool())
+    {
+        this->_already_set_up = false;
+        this->downsample_scanner_bool = arg;
+    }
+}
+
+int 
+ScatterSimulation::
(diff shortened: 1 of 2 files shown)


src/scatter_buildblock/ScatterSimulation.cxx
```diff
@@ -828,6 +828,55 @@ set_exam_info_sptr(const shared_ptr<const ExamInfo> arg)
...
+    return this->downsample_scanner_bool;
+}
+
+void 
+ScatterSimulation::
+set_downsample_scanner_bool(const bool arg)
+{
+    if (arg != this->get_downsample_scanner_bool())
+        this->_already_set_up = false;
+    this->downsample_scanner_bool = arg;
```
KrisThielemans, line 845: could in principle do
```suggestion
    if (arg != this->get_downsample_scanner_bool())
        this->_already_set_up = false;
    this->downsample_scanner_bool = arg;
```
```suggestion
    if (arg != this->get_downsample_scanner_bool())
      {
        this->_already_set_up = false;
        this->downsample_scanner_bool = arg;
      }
```
which might be a  bit cleaner, but up to you

KrisThielemans (commented):
looks good to me. any reason you still have this as draft?

Robert-PrescientImaging:
It was a draft because of #1173 and it was unknown at the time if more changes were needed. It is now ready.

src/scatter_buildblock/ScatterSimulation.cxx
Robert-PrescientImaging: I force push this for a cleaner git history

KrisThielemans (approved):

PR #1182, opened 2023-04-08
SPECT DICOM to interfile utility (rebased)

KrisThielemans:
Replaces #1122.

@danieldeidda I've rebased now on top of a commit with Ben's original check-in, such that we preserve authorship. I've done all that on top of current `master`.  I've also added release notes etc.

Maybe you can check this?


danieldeidda:
OK on it

PR #1192, opened 2023-04-27
Modified the inline error() function to also throw a runtime_error

markus-jehl:
I decided against calling the other error function, because that one is deprecated.


markus-jehl:
@KrisThielemans : these two includes in common.h introduce circular dependencies if I want to include the TextWriter in error.h and warning.h : https://github.com/UCL/STIR/blob/d23cda69ebd68ac92de4f51c0536635afbaa2a22/src/include/stir/common.h#L296

I guess this is why info.h (where the TextWriter is already used) is not included here, but instead wherever the info() function is used. I would propose to follow the same strategy for warning and error: remove them from common.h and instead include them where needed. Is that ok with you?

PR #1206, opened 2023-05-16
Adopted the same parallelisation used in back projection to avoid the…

markus-jehl:
… clang thread safety problems in FBP2D.


markus-jehl:
Still need to look into FBP3D and why the tests are failing.

KrisThielemans:
that seems to have worked! :-)

still want to do FBP3DRP before I merge?

Issue #863, opened 2021-04-23
STIR python - volume addition issues


robbietuk:
In python, I have:
```
>> type(STIR_image)
<class 'stir.FloatVoxelsOnCartesianGrid'>
```
Trying to add a scalar float results in an error:
```
>> STIR_image += 1.0
Traceback (most recent call last):
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
TypeError: unsupported operand type(s) for +=: 'stir.FloatVoxelsOnCartesianGrid' and 'float'
```
Trying to add itself onto itself results in a error:
```
>> STIR_image += STIR_image
Traceback (most recent call last):
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
TypeError: unsupported operand type(s) for +=: 'stir.FloatVoxelsOnCartesianGrid' and 'stir.FloatVoxelsOnCartesianGrid'
```

This is because the `STIR_image does not have the `__add__` method:
```
help(STIR_image)
Help on FloatVoxelsOnCartesianGrid object:
class FloatVoxelsOnCartesianGrid(Float3DDiscretisedDensityOnCartesianGrid)
 |  stir::VoxelsOnCartesianGrid< float >
 |  
 |  Method resolution order:
 |      FloatVoxelsOnCartesianGrid
 |      Float3DDiscretisedDensityOnCartesianGrid
 |      Float3DDiscretisedDensity
 |      FloatArray3D
 |      builtins.SwigPyObject
 |      builtins.object
 |  
 |  Methods defined here:
 |  
 |  __eq__(self, value, /)
 |      Return self==value.
 |  
 |  __ge__(self, value, /)
 |      Return self>=value.
 |  
 |  __gt__(self, value, /)
 |      Return self>value.
 |  
 |  __hash__(self, /)
 |      Return hash(self).
 |  
 |  __init__(self, /, *args, **kwargs)
 |      Initialize self.  See help(type(self)) for accurate signature.
 |  
 |  __le__(self, value, /)
 |      Return self<=value.
 |  
 |  __lt__(self, value, /)
 |      Return self<value.
 |  
[...]

PR #1207, opened 2023-05-23
Extend VoxelsOnCartesianGrid to include sapyb and test

Robert-PrescientImaging:
Fixes part of #863

### DIFF
--- a/src/swig/stir_voxels.i
+++ b/src/swig/stir_voxels.i
@@ -42,6 +42,13 @@
     {
       return write_to_file(filename, *$self);
     }
+
+    // add sapyb to VoxelsOnCartesianGrid
+    void sapyb(const float a,  const Array<3, float>& y, const float b)
+    { 
+      $self->sapyb(a,y,b); 
+    }
+
  }
 
 %template(Float3DDiscretisedDensity) stir::DiscretisedDensity<3,float>;
--- a/src/swig/test/python/test_buildblock.py
+++ b/src/swig/test/python/test_buildblock.py
@@ -261,3 +261,41 @@ def test_ProjData_from_to_Array3D():
     # assert every data point is equal
     assert all(a==b for a, b in zip(projdata.to_array().flat(),new_projdata.to_array().flat()))
 
+def test_xapyb_and_sapyb():
+    """
+    Test the xapyb and sapyb methods for FloatVoxelsOnCartesianGrid and ProjDataInMemory
+    """
+    test_value = 1.4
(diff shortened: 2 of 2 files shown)


Robert-PrescientImaging:
I have also tried to add support for `+=` and `-=` via the following
```
    // Define the __iadd__ method to support +=
    stir::VoxelsOnCartesianGrid<elemT>& operator+=(const stir::VoxelsOnCartesianGrid<elemT>& other) 
    {
      $self->sapyb(1.0f, other, 1.0f);
      return *$self;
    }
    
    // Define the __iadd__ method to support +=
    stir::VoxelsOnCartesianGrid<elemT>& operator-=(const stir::VoxelsOnCartesianGrid<elemT>& other) 
    {
      $self->sapyb(1.0f, other, -1.0f);
      return *$self;
    }
```
but it returns a pointer to the wrong data. I don't understand swig... 😞

Robert-PrescientImaging:
With the swig:
```
    void __iadd__(const Array<3, float>& y)
    { 
      $self->sapyb(1.0f,y,1.0f); 
    }
```
I get get the following python test working, but not the commented out `image += image` (returns `None` type).
```
    image.fill(test_value)
    # image += image
    image.__iadd__(image)
    assert image.find_max()==pytest.approx(2*test_value)
    assert image.find_min()==pytest.approx(2*test_value)
```

Robert-PrescientImaging:
I have also tried this approach, with the `%extend` scope:
```
    // Define += operator
    VoxelsOnCartesianGrid<float>& operator+=(VoxelsOnCartesianGrid<float>& y)
    { 
      self->sapyb(1.0f, y, 1.0f);
      return *self;
    }
```
but this returns a different pointer than that of `image`

KrisThielemans:
I think I'd first try to get this to work on the Arrays. Probably 
```
%extend Array {
 stir::Array& operator+=(const stir::Array& other) 
{
  $self->operator+=(other);
  return $self;
}
}

```
could do the trick. If not, doing the same but with 
%extend Array {
 stir::Array& __iadd__(const stir::Array& other) 
{
 ...
}
}
```
could work.

Question is then if it transports it to `VoxelsOnCartesianGrid`. If not, can add the same lines there. (which then I'd do via a macro of course)

Robert-PrescientImaging:
I cannot get any of the addition or in place addition to work on Arrays either, debugging is painful. I am happy to just add the already committed `sapyb` methodology to the `VoxelsOnCartesianGrid`, unless you want to try to fix the underlying issue with it not inheriting this property from `Array`?

PR #1208, opened 2023-05-24
 making ECAT8 LM input file format specific to siemens ecat8

danieldeidda:


danieldeidda:
This has been tested on the mMR data from zenodo

src/include/stir/IO/ECAT8_32bitListmodeInputFileFormat.h
```diff
@@ -46,8 +46,22 @@ public InputFileFormat<ListModeData >
     actual_can_read(const FileSignature& signature,
 		    std::istream& input) const
   {
-    // TODO need to do check that it's a siemens list file etc
-    return is_interfile_signature(signature.get_signature());
+        std::string ecat8;
+
+    for (int i=0;i<signature.size(); i++)
```
KrisThielemans, line 51: This would have to be `signature.size()-8` to avoid buffer-overrun.

src/include/stir/IO/ECAT8_32bitListmodeInputFileFormat.h
```diff
@@ -46,8 +46,22 @@ public InputFileFormat<ListModeData >
...
+    {
+
+        ecat8=std::string(1,signature.get_signature()[i])+
+                std::string(1,signature.get_signature()[i+1])+
+                std::string(1,signature.get_signature()[i+2])+
+                std::string(1,signature.get_signature()[i+3])+
+                std::string(1,signature.get_signature()[i+4])+
+                std::string(1,signature.get_signature()[i+5])+
+                std::string(1,signature.get_signature()[i+6]);
```
KrisThielemans, line 60:
```suggestion
        const std::string ecat8(signature.get_signature() + i, 8);
```
should work I believe

KrisThielemans (changes requested):
I'd prefer to do the check on `is_interfile_signature` first (i.e. if that fails, get out immediatelly).

Currently our `FileSignature` size is 1024. I looked at one of your mCT files
```
**!INTERFILE :=
!originating system :=1104
%comment:=Bin Address SubHeader for PETSYNGO_VG70B
%SMS-MI header name space :=PETLINK bin address
%SMS-MI version number :=3.7
```
So I guess it still fits in there. Can you check with that as well? (Obviously, it won't be able to read it due to header problems etc, but we should get into the parsing)

KrisThielemans:
Actually, the following might be faster
```
const std::string signature_as_string(signature.get_signature(), signature.size());
return signature_as_string.find("PETLINK") != std::string::npos;
```

danieldeidda:
> So I guess it still fits in there. Can you check with that as well? (Obviously, it won't be able to read it due to header problems etc, but we should get into the parsing)

yes it parses this too

src/include/stir/IO/ECAT8_32bitListmodeInputFileFormat.h
```diff
@@ -46,9 +46,18 @@ public InputFileFormat<ListModeData >
...
+        if(!is_interfile_signature(signature.get_signature()))
+                return false;
+        else
+        {
+            for (int i=0;i<signature.size()-8; i++)
+            {
+                const std::string signature_as_string(signature.get_signature(), signature.size());
+                return signature_as_string.find("PETLINK") != std::string::npos;
+            }
+        }
```
KrisThielemans, line 58:
```suggestion
        else
        {
            for (int i=0;i<signature.size()-8; i++)
            {
                const std::string signature_as_string(signature.get_signature(), signature.size());
                return signature_as_string.find("PETLINK") != std::string::npos;
            }
        }
```

KrisThielemans: no more need for the loop now!

danieldeidda: of course :D

KrisThielemans (approved):
seems fine to me now. Does it still work on the data?

danieldeidda:
yes it does!

PR #1209, opened 2023-05-24
attempt to make CListEventCyl...DiscreteDetectors templated for block…

danieldeidda:
…OnCylindrical

This seems to give  the same sinogram I was getting before the change. However, when using BlocksOnCylindrical the sinogram obtained is exactly the same. Would we expect this to happen?  if yes then pheraps there is no need to template thiese functions. I was expecting some difference. 

This is not most elegant way of doing this, I was trying to avoid changes in other classes but there are a few part that I am not very happy about (feedback welcome!)

### DIFF
--- a/src/include/stir/listmode/CListEventCylindricalScannerWithDiscreteDetectors.h
+++ b/src/include/stir/listmode/CListEventCylindricalScannerWithDiscreteDetectors.h
@@ -22,6 +22,7 @@
 #include "stir/Succeeded.h"
 #include "stir/DetectionPositionPair.h"
 #include "stir/ProjDataInfoCylindricalNoArcCorr.h"
+#include "stir/ProjDataInfoBlocksOnCylindricalNoArcCorr.h"
 #include "stir/listmode/CListRecord.h"
 
 START_NAMESPACE_STIR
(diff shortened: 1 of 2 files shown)


KrisThielemans:
Not sure what you mean when saying that the  sinograms are the same. If you pretend a cyl-scanner is a blocks-scanner, and unlist that way, and then fix the header, yes, you would get the same results (as we're just working on indices really).

See also #994  which was another attempt at the same problem

danieldeidda:
> Not sure what you mean when saying that the sinograms are the same. If you pretend a cyl-scanner is a blocks-scanner, and unlist that way, and then fix the header, yes, you would get the same results (as we're just working on indices really).
> 
> See also #994 which was another attempt at the same problem

what I meant was that if we try to unlist the data using different functions (according to whether we want Cylindrical or BlocksOnCylindrical) ex: find_cartesian_coordinates_given_scanner_coordinates I would expect that the events are somehow organised differently between the two geometry. But I guess as you said we are only working on indices . I suppose it will make a difference if  we get the position and not the index

KrisThielemans:
> I suppose it will make a difference if we get the position and not the index

definitely

KrisThielemans:
These changes are now obsolete after #1222

PR #1213, opened 2023-06-03
[GHA] reduce disk space requirements

KrisThielemans:
gcc-12 Ubuntu jobs are now running out of space.

try `apt clean`


KrisThielemans:
this works now after removing dotnet, android and some cached tools

PR #1221, opened 2023-07-05
Change back projectors in demos

### LINKED ISSUE #1220: demos 1, 2 and 3* cannot handle non-zero view-offset with BackProjectorByBinUsingInterpolation
Following the guide here (with correct `${EXE_LOC}`):

https://github.com/UCL/STIR/blob/d04922d0cf053f1b8fcc488365f5b5bb925ad1e2/examples/C%2B%2B/src/README.md?plain=1#L125-L148

Results in the following message:

```
rts@Roberts-MacBook-Pro src % generate_image generate_image.par

INFO: Processing next shape...

INFO: Number of voxels recomputed with finer sampling : 6267

INFO: Processing next shape...

INFO: Number of voxels recomputed with finer sampling : 1396
rts@Roberts-MacBook-Pro src % forward_project sino.hs image.hv  small.hs

WARNING: Scanner RATPET: transaxial singles_unit info is not set (probably irrelevant unless you use dead-time correction that needs this info)

INFO: Detected view-mashing factor 2 from the number of views (28) and the number of detectors per ring (112).
Adjusting the azimuthal angle offset accordingly (an extra offset of 1.60714 degrees)

WARNING: WARNING: ProjMatrixByBinUsingRayTracing used for pixel size (in x,y) that is larger than the bin size.
Backprojecting with this matrix might have artefacts at views 0 and 90 degrees.


INFO: Processing view 0 of segment 0

[CUT BACKPROJECTION INFO]
```

and error:

```
rts@Roberts-MacBook-Pro src % ${EXE_LOC}/demo1

WARNING: FactoryRegistry:: overwriting previous value of key in registry.
     key: None
Input file(default extension '.hs'):sino.hs

WARNING: ProjData::read_from_file trying to read sino.hs as Interfile

WARNING: Scanner RATPET: transaxial singles_unit info is not set (probably irrelevant unless you use dead-time correction that needs this info)

INFO: Detected view-mashing factor 2 from the number of views (28) and the number of detectors per ring (112).
Adjusting the azimuthal angle offset accordingly (an extra offset of 1.60714 degrees)
Template image file(default extension '.hv'):image.hv

ERROR: BackProjectorByBinUsingInterpolation cannot handle non-zero view-offset. Sorry
libc++abi: terminating with uncaught exception of type std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >
zsh: abort      ${EXE_LOC}/demo1
```

This is also the case for demo2. Demo3 doesn't work with the default configuration or with the `demo.par` parameter file. Providing the `demoPM.par` works fine (matrix back projector). 

The issue is the use of the `BackProjectorByBinUsingInterpolation` projector with this data.

robbietuk:
Fixes #1220

### DIFF
--- a/examples/C++/src/README.md
+++ b/examples/C++/src/README.md
@@ -146,6 +146,8 @@ To run each of the demos:
 ```
 ${EXE_LOC}/demo1
 ```
+Input a projection data filename (e.g. `sino.hs`) and image data filename (e.g. `image.hv`) when prompted.
+
 You can display the output using for instance
 ```
 manip_image output.hv
@@ -155,6 +157,10 @@ manip_image output.hv
 ```
 ${EXE_LOC}/demo2
 ```
+
+As before, input a projection data filename (e.g. `sino.hs`) and image data filename (e.g. `image.hv`) when prompted. 
+Subsequently enter `Matrix` when asked `Which type do you want? Possible values are:` (in reference to backprojector type) and `Ray Tracing` when prompted with `matrix type := `. _Other options may fail as they have deprecated._
+
 `demo2` contains a call to `stir::display`, so if you compiled STIR with `GRAPHICS=X`, you'll see the display immediately.
 
 #### `demo3`
@@ -165,7 +171,9 @@ or
 ```
 ${EXE_LOC}/demo3 demoPM.par
 ```
-The folowing command line will ask the questions interactively
+Inspect these parameter files for more details.
+
+The following command will ask the configuration questions interactively. Follow prompts as before. 
 ```
 ${EXE_LOC}/demo3 
 ```
--- a/examples/C++/src/README.md
+++ b/examples/C++/src/README.md
@@ -159,7 +159,8 @@ ${EXE_LOC}/demo2
 ```
 
 As before, input a projection data filename (e.g. `sino.hs`) and image data filename (e.g. `image.hv`) when prompted. 
-Subsequently enter `Matrix` when asked `Which type do you want? Possible values are:` (in reference to backprojector type) and `Ray Tracing` when prompted with `matrix type := `. _Other options may fail as they have deprecated._
+Subsequently, enter `Matrix` (or `parallelproj` if installed) when prompted by `Which type do you want? Possible values are:` (in reference to back projector type) and `Ray Tracing` when prompted by `matrix type := `. 
+_Other options such as `Interpolation` may fail as they have been deprecated._
 
 `demo2` contains a call to `stir::display`, so if you compiled STIR with `GRAPHICS=X`, you'll see the display immediately.
 
--- a/examples/C++/src/demo1.cxx
+++ b/examples/C++/src/demo1.cxx
@@ -23,6 +23,7 @@
     of the GNU General  Public Licence (GPL)
     See STIR/LICENSE.txt for details
 */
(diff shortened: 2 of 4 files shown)


examples/C++/src/README.md
```diff
@@ -155,6 +157,10 @@ manip_image output.hv
 ```
 ${EXE_LOC}/demo2
 ```
+
+As before, input a projection data filename (e.g. `sino.hs`) and image data filename (e.g. `image.hv`) when prompted. 
+Subsequently enter `Matrix` when asked `Which type do you want? Possible values are:` (in reference to backprojector type) and `Ray Tracing` when prompted with `matrix type := `. _Other options may fail as they have deprecated._
+
```
robbietuk, line 163: This is about the best I could do with `demo2` with minor modifications. The demo is there to demonstrate the prompting STIR has but in this instance, only a few options are valid from the provided list: Interpolation, Matrix, None and Post Smoothing.

examples/C++/src/demo3.cxx
```diff
@@ -54,7 +55,8 @@ class MyStuff: public ParsingObject
 void
 MyStuff::set_defaults()
 {
-  back_projector_sptr.reset(new BackProjectorByBinUsingInterpolation);
+  shared_ptr<ProjMatrixByBin> PM(new  ProjMatrixByBinUsingRayTracing());
+  back_projector_sptr.reset(new BackProjectorByBinUsingProjMatrixByBin(PM));
```
robbietuk, line 59: Can this be done in one line? Maybe it is best to leave as two

examples/C++/src/README.md
KrisThielemans: `parallelproj` would work as well (if it is available), just to make things complicated. not 100% sure how to phrase that. Maybe better to exclude the `Interpolation` one. (Note grammar error in "have deprecated")

examples/C++/src/demo1.cxx
```diff
@@ -55,8 +56,9 @@ int main()
 
   density_sptr->fill(0);
 
-  /////////////// back project
-  BackProjectorByBinUsingInterpolation back_projector;
+  /////////////// back project, using ray tracing
+  shared_ptr<ProjMatrixByBin> projection_matrix(new  ProjMatrixByBinUsingRayTracing());
+  BackProjectorByBinUsingProjMatrixByBin back_projector(projection_matrix);
```
KrisThielemans, line 61:
```suggestion
  auto projection_matrix_sptr = std::make_shared_ptr<ProjMatrixByBinUsingRayTracing>();
  BackProjectorByBinUsingProjMatrixByBin back_projector(projection_matrix_sptr);
```
I think this will work and is maybe somewhat cleaner?

examples/C++/src/demo3.cxx
KrisThielemans: can be done in 1 line, but could become a bit cryptic. Could use the `auto` trick from above

KrisThielemans (commented):
Thanks!

we really should add compiling/running these to our GitHub Actions, but probably leave that for later and create an issue instead.

robbietuk:
Updated the suggestions and I will make an issue with the GHA

KrisThielemans (approved):

PR #1222, opened 2023-07-08
prepare CListEventCylindricalScannerWithDiscreteDetectors for non-cylindrical

KrisThielemans:
- move all functionality to template class `CListEventScannerWithDiscreteDetectors<ProjDataInfoT>`
- keep `CListEventCylindricalScannerWithDiscreteDetectors` as derived class such that this change is backwards compatible

### DIFF
--- a/documentation/release_5.2.htm
+++ b/documentation/release_5.2.htm
@@ -107,6 +107,15 @@
     </li>
 </ul>
 
+<h3>Changes for developers</h3>
+<ul>
+  <li>moved all functionality in <code>CListEventCylindricalScannerWithDiscreteDetectors</code> to template class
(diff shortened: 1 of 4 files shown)


KrisThielemans:
@danieldeidda @VietAnhDao I prefer this over #1209 and #994. I think it's cleaner and avoids type-casting all the time. (I guess it's less flexible as it relies on compilation-time decision on what the listmode data contains, but I think that makes sense in pretty much every case anyway)

KrisThielemans:
@markus-jehl please check. It should not affect you at all, but will help later.

markus-jehl:
Looks good to me!

danieldeidda (approved):
To be honest that's what I would have liked to do anyway. So looks good to me :D

PR #1230, opened 2023-07-23
[Python] fix ProjData_read_from_file

KrisThielemans:
`stir.ProjData_read_from_file` is not available on recent SWIG (4.1) unless you use a specific switch, see https://swig.org/Doc4.1/Python.html#Python_nn20. So use `stir.ProjData.read_from_file`

### DIFF
--- a/documentation/release_5.2.htm
+++ b/documentation/release_5.2.htm
@@ -63,6 +63,8 @@
 
 <h4>Python</h4>
 <ul>
+  <li>Examples use <code>stir.ProjData.read_from_file</code> as opposed to <code>stir.ProjData_read_from_file</code>. The former is supported since SWIG 3.0, and the <a href="https://swig.org/Doc4.1/Python.html#Python_nn20">default from SWIG 4.1</a>.
+    </li>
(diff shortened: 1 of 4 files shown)


robbietuk:
Looks good to me. 

Can the swig compiler detect earlier swig versions? If so, could the `stir.ProjData_read_from_file` function wrap a `stir.ProjData.read_from_file` call? It would streamline a lot of code and make it SWIG version invariant.

robbietuk (approved):

KrisThielemans:
> Can the swig compiler detect earlier swig versions? If so, could the `stir.ProjData_read_from_file` function wrap a `stir.ProjData.read_from_file` call? It would streamline a lot of code and make it SWIG version invariant.

Not really I guess. We can do it in our own `.i` file, and add the function. However, I wouldn't bother. The `.` syntax is much more natural, and there already a long time, see e.g. in 3.0
https://swig.org/Doc3.0/Python.html#Python_nn20. I guess you might have had problems with my MATLAB swig branch, which is quite out-of-date...

Issue #1223, opened 2023-07-13
Ray tracing projector introduces axial artifacts for BlocksOnCylindrical scanner geometry


markus-jehl:
This was found in the master branch as well as in the latest TOF branch. Segment 0 is not affected, but oblique segments have cross-talk suggesting that the intrinsic_tilt and/or segment sign are assigned wrongly in certain situations.

Example from BlocksTests::run_map_orientation_test (slightly modified. left: original, right: rotated detector map):
![image](https://github.com/UCL/STIR/assets/87262701/f06060dc-0774-4c97-ae7b-874f39cc6231) ![image](https://github.com/UCL/STIR/assets/87262701/30c10de0-b818-4d95-b7f7-3b80e684a82e)

markus-jehl:
And below, a plot of a reconstruction of projdata with the same rotation for the template and the data (first) compared to a reconstruction that has a template with a different scanner intrinsic tilt than the data (the data was generated at a different tilt, then the projdata were modified to the same tilt than the one used for reconstruction):
![image](https://github.com/UCL/STIR/assets/87262701/13e338e2-bd16-410e-a0dc-f6016da18936)
![image](https://github.com/UCL/STIR/assets/87262701/d3ae565f-e8cf-4283-b1b5-1fb603c270f1)

KrisThielemans:
Can you confirm this doesn't happen with `parallelproj`?

Can you clarify what you are showing with the viewgrams please? which view, which segment? The left one looks weirder than the right one, so what is "original"?

markus-jehl:
Yes, with parallelproj it does not happen.

The viewgrams I'm showing are across all segments (0, 1, -1, 2, -2, ... going downwards), the left one is the standard SAFIRDualRingPrototype used in the test, the right one is the same scanner but with flipped tangential coordinates in the detector map. The one on the left is definitely showing some data that is axially placed wrongly, so should be a good test case to investigate the problem.

markus-jehl:
@KrisThielemans : I think the problem with the vertical line emerges in the transformation from detectors to tang_pos_num and view_num, i.e. around here https://github.com/UCL/STIR/blob/4a9391257c76e8b8b1b68c22bc81fa92033bff77/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx#L289
This is probably the same issue seen here: https://github.com/UCL/STIR/issues/1176

[...]

PR #1231, opened 2023-08-07
Fix for bug in ray tracing projection which produced axial artifacts due to flipped theta angle

markus-jehl:
Fixes https://github.com/UCL/STIR/issues/1223

### DIFF
--- a/src/include/stir/LORCoordinates.inl
+++ b/src/include/stir/LORCoordinates.inl
@@ -346,8 +346,19 @@ find_LOR_intersections_with_cylinder(LORAs2Points<coordT1>& intersection_coords,
   auto l2 = static_cast<coordT2>((-b - root) / a);
   
   // TODO won't work when coordT1!=coordT2
-  intersection_coords.p1() = d*l1 + c1;
-  intersection_coords.p2() = d*l2 + c1;
+  const CartesianCoordinate3D<coordT2> distance1 = d * l1 + c1 - coords.p1();
+  const CartesianCoordinate3D<coordT2> distance2 = d * l2 + c1 - coords.p1();
+  if (square(distance1.x()) + square(distance1.y()) + square(distance1.z())
+      < square(distance2.x()) + square(distance2.y()) + square(distance2.z()))
+  {
+    intersection_coords.p1() = d * l1 + c1;
+    intersection_coords.p2() = d * l2 + c1;
+  }
+  else
+  {
+    intersection_coords.p2() = d * l1 + c1;
+    intersection_coords.p1() = d * l2 + c1;
+  }
   assert(fabs(square(intersection_coords.p1().x())+square(intersection_coords.p1().y())
 	      - square(radius))
 	 <square(radius)*10.E-5);
--- a/src/include/stir/LORCoordinates.inl
+++ b/src/include/stir/LORCoordinates.inl
@@ -342,23 +342,11 @@ find_LOR_intersections_with_cylinder(LORAs2Points<coordT1>& intersection_coords,
   }
   const coordT2 root = static_cast<coordT2>(sqrt(argsqrt));
 
-  auto l1 = static_cast<coordT2>((-b + root) / a);
-  auto l2 = static_cast<coordT2>((-b - root) / a);
+  auto l1 = static_cast<coordT2>((-b - root) / a);
+  auto l2 = static_cast<coordT2>((-b + root) / a);
   
-  // TODO won't work when coordT1!=coordT2
-  const CartesianCoordinate3D<coordT2> distance1 = d * l1 + c1 - coords.p1();
-  const CartesianCoordinate3D<coordT2> distance2 = d * l2 + c1 - coords.p1();
-  if (square(distance1.x()) + square(distance1.y()) + square(distance1.z())
-      < square(distance2.x()) + square(distance2.y()) + square(distance2.z()))
-  {
-    intersection_coords.p1() = d * l1 + c1;
-    intersection_coords.p2() = d * l2 + c1;
-  }
-  else
-  {
-    intersection_coords.p2() = d * l1 + c1;
-    intersection_coords.p1() = d * l2 + c1;
-  }
+  intersection_coords.p1() = d * l1 + c1;
+  intersection_coords.p2() = d * l2 + c1;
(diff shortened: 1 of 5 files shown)


src/include/stir/LORCoordinates.inl
```diff
@@ -346,8 +346,19 @@ find_LOR_intersections_with_cylinder(LORAs2Points<coordT1>& intersection_coords,
   auto l2 = static_cast<coordT2>((-b - root) / a);
   
   // TODO won't work when coordT1!=coordT2
-  intersection_coords.p1() = d*l1 + c1;
-  intersection_coords.p2() = d*l2 + c1;
+  const CartesianCoordinate3D<coordT2> distance1 = d * l1 + c1 - coords.p1();
```
markus-jehl, line 349: TODO: check if this can be done just with math to improve performance.

src/include/stir/ProjDataInfoGeneric.inl
```diff
@@ -83,8 +83,22 @@ ProjDataInfoGeneric::get_tantheta(const Bin& bin) const
...
+
+  _p1.z() += z_shift.z();
+  _p2.z() += z_shift.z();
+
+  // do the below transformation to check whether the points are swapped when phi is obtained
+  // if they are swapped theta needs to be inverted, otherwise the oblique segments contain flipped streaks
+  LORAs2Points<float> lor_as_2_points(_p1, _p2);
+  const double R = sqrt(fmax(square(_p1.x()) + square(_p1.y()), square(_p2.x()) + square(_p2.y())));
+  LORInAxialAndNoArcCorrSinogramCoordinates<float> lor;
+  lor_as_2_points.change_representation(lor, R);
```
KrisThielemans, line 95: now that we have `lor`, can't we just use `lor.get_tantheta()` (or whatever it's called). I think it was a mistake to put this geometry here, as it's already in the LOR code already

src/include/stir/LORCoordinates.inl
KrisThielemans: seems to be me that due to the parametrisation `c = l*d+c1`, the smallest `l` is going to be closest to `c1`. Therefore, chosing the negative root as `l1` should do the trick, i.e. the old code *always* swapped.

KrisThielemans (changes requested):
Do we now test with segment !=0 ? You'd have hoped that this would have shown the original mistake.

src/include/stir/LORCoordinates.inl
markus-jehl: This makes sense and matches what I've seen when debugging through it. I'll test with just swapping l1 and l2.

src/include/stir/ProjDataInfoGeneric.inl
markus-jehl: I haven't found any mention of theta outside ProjDataInfo files. Can you point me to the LOR function you mean?

markus-jehl:
> Do we now test with segment !=0 ? You'd have hoped that this would have shown the original mistake.

Ah, I forgot to mention this! I tried to modify @danieldeidda 's test to include non-zero segments and look at non-trivial entries in the proj data. Unfortunately, even though the proj data look very similar, the bin values still vary in too many places to compare the proj data bin by bin (which might be why the test currently only looks at segment 0, axial_pos 0, tangential_pos 0 and then loops through the views). So I'm still unsure how to best test this without simplifying too much.

src/include/stir/ProjDataInfoGeneric.inl
KrisThielemans: right, that's because it doesn't exist. sorry.

what about then doing
```
	get_LOR(lor, bin);
```
directly without going via ` find_cartesian_coordinates_of_detection` (presumably, internally `get_LOR` does that anyway). Possibly we then have
```
       return (lor.z1() - lor.z2()) / lor.radius();
```
(I think we might have to define `radius()` in `LORInAxialAndNoArcCorrSinogramCoordinates`)

markus-jehl:
```
get_LOR(lor, bin);

return (lor.z2() - lor.z1()) / (2 * lor.radius());
```
does the trick, i.e. the inverse and using double of the radius. Is this definitely how the radius is defined on the LOR? Couldn't find the place where this conversion happens.

KrisThielemans:
> I tried to modify @danieldeidda 's test to include non-zero segments and look at non-trivial entries in the proj data. Unfortunately, even though the proj data look very similar, the bin values still vary in too many places to compare the proj data bin by bin (which might be why the test currently only looks at segment 0, axial_pos 0, tangential_pos 0 and then loops through the views).

sorry, could you elaborate a little bit? Which test was this?

Otherwise, this is good to go, as soon as you add something to the release notes of course!

markus-jehl:
> > I tried to modify @danieldeidda 's test to include non-zero segments and look at non-trivial entries in the proj data. Unfortunately, even though the proj data look very similar, the bin values still vary in too many places to compare the proj data bin by bin (which might be why the test currently only looks at segment 0, axial_pos 0, tangential_pos 0 and then loops through the views).
> 
> sorry, could you elaborate a little bit? Which test was this?
> 
> Otherwise, this is good to go, as soon as you add something to the release notes of course!

I mean the run_map_orientation_test in test_blocks_on_cylindrical_projectors.cxx. I have locally modified it to call set_blocks_projdata_info instead of set_direct_projdata_info, and then write out the projdata1 and projdata2 to compare them visually. The obvious flipped vertical lines are gone, but if I extend the bin comparison to use non-trivial positions there are some bins with large differences.

src/include/stir/ProjDataInfoGeneric.inl
markus-jehl: Perfect!

KrisThielemans:
> if I extend the bin comparison to use non-trivial positions there are some bins with large differences

hmmm. that isn't reassuring!

Anyway, let's merge this then and create an issue for that. squash-merge ok?

KrisThielemans (approved):

markus-jehl:
> > if I extend the bin comparison to use non-trivial positions there are some bins with large differences
> 
> hmmm. that isn't reassuring!
> 
> Anyway, let's merge this then and create an issue for that. squash-merge ok?

Sure! I'll create the issue right away

PR #1246, opened 2023-09-14
fix SPECTUB bug with collimator slope division by 10

samdporter:
Using set_resolution_model() with SPECTUB currently divides the slope by 10. This should not be the case as the slope is dimensionless.

Changed the code to reflect this and renamed the collimator slope variable

### DIFF
--- a/documentation/release_5.2.htm
+++ b/documentation/release_5.2.htm
@@ -26,7 +26,8 @@
 
 <h3>Bug fixes</h3>
 <ul>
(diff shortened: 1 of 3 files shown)


KrisThielemans (changes requested):
Good catch!

can you
- create an issue for it, such that it's obvious there was a bug
- add a line to release_5.2.htm in the bug fix section (if there's one) accordingly. Say that it doesn't affect results when using a .par file, but only when using `set_resolution_model` (e.g. from Python or from SIRF)

KrisThielemans (approved):

PR #1260, opened 2023-09-17
rewrite ProjDataInMemory to avoid streams

### LINKED ISSUE #1240: ProjDataInMemory::fill crashes for size > 2GB due to boost bug
Originally filed in https://github.com/NikEfth/STIR/issues/45.

This is a boost bug in `boost::interprocess::bufferstream`  https://github.com/boostorg/interprocess/issues/190. `boost::iostreams::stream<boost::iostreams::array>` is also not usable https://github.com/boostorg/iostreams/issues/164

There seems to be no `std` equivalent for these. For instance, `std::stringstream` needs to be constructed from `string` (by the way, before c++20 this means making a copy, which doubles memory size). Looks like we'll have to get rid of the `stream` unless we'd get into `streambuf` stuff, but using `ProjDataFromStream` is quite wasteful anyway.

KrisThielemans:
`ProjDataInMemory` was derived from `ProjDataFromStream`, which meant we had to write less code. However, this meant it was slower than just doing the copies from memory directly.

Also, due to a limitation in `boost::interprocess::bufferstream`, total size was limited to 2 GB.

Bits of the code are currently ugly to be compatible with both master and the TOF branch. There's some copies related to offsets from the (TOF) `ProjDataFromStream` class.

Fixes #1240

### DIFF
--- a/documentation/release_5.2.htm
+++ b/documentation/release_5.2.htm
@@ -126,6 +126,10 @@
     re-use for generic/blocksoncylindrical scanners.
     <br /><a href="https://github.com/UCL/STIR/pull/1222/">PR #1222</a>.
   </li>
+  <li>rewritten <code>ProjDataInMemory</code> to avoid streams, causing a speed-up of some
+    operations, and removing a limit of total size of 2GB.
+    <br /><a href="https://github.com/UCL/STIR/pull/1260/">PR #1260</a>.
+  </li>
 </ul>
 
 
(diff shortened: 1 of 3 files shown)


KrisThielemans:
@markus-jehl we talked about this. It's a rather drastic revision, but passes all my tests (also for TOF). Give it a go...

markus-jehl:
will try it out

markus-jehl:
The reconstruction ran through without any issues, so seems good to me!

Issue #1255, opened 2023-09-17
(potential?) data-races


KrisThielemans:
Running Intel Inspector (with the Intel compiler) gives me a few data-races. Of course, I'm not sure if these are false positives or not

<h1>when checking internal ProjDataInfo data-structures were initialised</h1>

https://github.com/UCL/STIR/blob/9e40811fac72de58d5488247788c520e447b0501/src/include/stir/ProjDataInfoGenericNoArcCorr.inl#L43-L44
conflicts with
https://github.com/UCL/STIR/blob/9e40811fac72de58d5488247788c520e447b0501/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx#L228
(and equivalent for Generic). 

This implementation of the Double-Checked-Locking(DCL) pattern was suggested at http://stackoverflow.com/questions/27975737/how-to-handle-cached-data-structures-with-multi-threading-e-g-openmp.
I see that I didn't use the `atomic write` in the .cxx [Edit: this is now fixed by #1256], but adding that doesn't seem to make Intel Inspector happy.

I can make these disappear by initialising these tables in the constructor, i.e. commenting out lines like
https://github.com/UCL/STIR/blob/9e40811fac72de58d5488247788c520e447b0501/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx#L73-L74
However, that causes a slow-down as these tables take a while to construct. For example, 
```
time ./run_test_simulate_and_recon.sh
```
before
```
real    0m23.021s
user    4m12.391s
sys     0m2.925s
```
after
```
real    0m33.966s
user    4m39.860s
sys     0m10.009s
```

<h1>detector position access in `DetectorCoordinateMap::get_coordinate_for_det_pos`</h1>

https://github.com/UCL/STIR/blob/9e40811fac72de58d5488247788c520e447b0501/src/include/stir/DetectorCoordinateMap.h#L91-L93
conflicts with some internal randoms stuff.

This will be because [std random generators are not thread-safe](https://stackoverflow.com/a/45908573).
I think we need one generator per thread (probably a vector of size `get_max_num_threads()`. @markus-jehl feel like looking at that one?

markus-jehl:
sure, I'll put the second one on my list

markus-jehl:
@KrisThielemans : can this be closed?

PR #1265, opened 2023-09-26
Create one normal distribution per omp thread to avoid data races

markus-jehl:
This addresses the second part of https://github.com/UCL/STIR/issues/1255

### DIFF
--- a/src/include/stir/DetectorCoordinateMap.h
+++ b/src/include/stir/DetectorCoordinateMap.h
@@ -93,6 +93,9 @@ public:
 	stir::CartesianCoordinate3D<float> get_coordinate_for_det_pos( const stir::DetectionPosition<>& det_pos ) const
 	{ 
 		auto coord = det_pos_to_coord.at(det_pos);
+    if (sigma == 0.0)
+      return coord;
+
 #ifdef STIR_OPENMP
 		auto thread_id = omp_get_thread_num();
 #else
@@ -123,8 +126,8 @@ public:
 
 protected:
 
- DetectorCoordinateMap(double sigma = 0.0) :
-        sigma(sigma)
+ explicit DetectorCoordinateMap(double sigma = 0.0) :
+  sigma(sigma)
   {
 #ifdef STIR_OPENMP
     generators.resize(omp_get_max_threads());
--- a/src/include/stir/DetectorCoordinateMap.h
+++ b/src/include/stir/DetectorCoordinateMap.h
@@ -27,6 +27,7 @@
 #include <map>
 #include <boost/algorithm/string.hpp>
 #include <boost/unordered_map.hpp>
+#include <omp.h>
 
 #include "stir/CartesianCoordinate3D.h"
 #include "stir/DetectionPosition.h"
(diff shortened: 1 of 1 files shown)


src/include/stir/DetectorCoordinateMap.h
```diff
@@ -27,6 +27,7 @@
 #include <map>
 #include <boost/algorithm/string.hpp>
 #include <boost/unordered_map.hpp>
+#include <omp.h>
```
KrisThielemans, line 30:
```suggestion
#ifdef STIR_OPENMP
#include <omp.h>
#endif
```

src/include/stir/DetectorCoordinateMap.h
```diff
@@ -65,14 +66,32 @@ class DetectorCoordinateMap
...
+		{ 
+      read_detectormap_from_file(filename);
+#ifdef STIR_OPENMP
+      distributions.resize(omp_get_max_threads());
+      for (auto &distribution : distributions)
+        distribution = std::normal_distribution<double>(0.0, sigma);
+#else
+      distributions.resize(1);
+      distributions[0] = std::normal_distribution<double>(0.0, sigma);
+#endif
```
KrisThielemans, line 79:
```suggestion
		DetectorCoordinateMap(sigma)
		{ 
                   read_detectormap_from_file(filename);
```
should work I believe (delegating constructor)

src/include/stir/DetectorCoordinateMap.h
```diff
@@ -88,9 +107,15 @@ class DetectorCoordinateMap
...
-		coord.z() += static_cast<float>(distribution(generator));
+#ifdef STIR_OPENMP
+    auto thread_id = omp_get_thread_num();
+#else
+    auto thread_id = 0;
+#endif
+    
+		coord.x() += static_cast<float>(distributions[thread_id](generator));
+		coord.y() += static_cast<float>(distributions[thread_id](generator));
+		coord.z() += static_cast<float>(distributions[thread_id](generator));
```
KrisThielemans, line 118:
```suggestion
                auto& distribution = distributions[thread_id];
		coord.x() += static_cast<float>(distribution(generator));
		coord.y() += static_cast<float>(distribution(generator));
		coord.z() += static_cast<float>(distribution(generator));
```

However, this raises the question if `generator` is thread-safe. I have no idea, but somehow I think it's that one that isn't, while the distribution itself is. Not sure!

src/include/stir/DetectorCoordinateMap.h
```diff
@@ -88,9 +107,15 @@ class DetectorCoordinateMap
 	stir::CartesianCoordinate3D<float> get_coordinate_for_det_pos( const stir::DetectionPosition<>& det_pos ) const
 	{ 
 		auto coord = det_pos_to_coord.at(det_pos);
-		coord.x() += static_cast<float>(distribution(generator));
-		coord.y() += static_cast<float>(distribution(generator));
-		coord.z() += static_cast<float>(distribution(generator));
+#ifdef STIR_OPENMP
+    auto thread_id = omp_get_thread_num();
+#else
+    auto thread_id = 0;
```
KrisThielemans, line 113: more white-space here apparently (unless it's GitHub).

KrisThielemans (commented):
cosmetics but one question where the thread-safety should sit: the distribution or the generator?

src/include/stir/DetectorCoordinateMap.h
markus-jehl: True, that's much cleaner!

src/include/stir/DetectorCoordinateMap.h
markus-jehl: good that you asked! I initially thought it was just the distribution (and the intel inspector complaints have gone away), but some more research showed that the generator is also not thread-safe. I'll just store a vector of instances for each.

KrisThielemans (changes requested):
I think this is fine, aside from the `explicit`.

By the way, would it be faster to have `if(sigma ==0) {return coord;}` without going throught the random generator?

src/include/stir/DetectorCoordinateMap.h
```diff
@@ -113,9 +124,19 @@ class DetectorCoordinateMap
 protected:
 
  DetectorCoordinateMap(double sigma = 0.0) :
```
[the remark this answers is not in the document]
markus-jehl: good idea! it can't hurt to handle the sigma == 0.0 case separately.

KrisThielemans (approved):

Issue #1268, opened 2023-10-02
Listmode reconstruction creates cache files with wrong numbers


markus-jehl:
The listmode reconstruction creates files numbered from very large numbers, such as e.g. 2241087299. This leads to the reconstruction failing when trying to read the cache, because then it is looking for files numbered from 0 onwards.
This can be fixed by explicitly setting the file count to 0 when creating the cache.

PR #1269, opened 2023-10-02
Expose listmode reconstruction in SWIG and fix an issue with cache naming

markus-jehl:
Fixes https://github.com/UCL/STIR/issues/1268 and makes listmode reconstruction callable from python.


KrisThielemans (approved):
looks good!

I guess no tests for the Python stuff, but none for the other reconstructions either, so that'll have to be for the future. I'm assuming you tested it and it works?

markus-jehl:
Yes, I have completed my first listmode reconstruction today and it looks good! Took a long time, though: approximately 9 minutes per set of 10 sub-iterations

KrisThielemans:
ok great. we can discuss timing later. would be good to have some idea of difference in timing when creating the cache vs re-using it. Also, we should really switch to non-view subsets, such that we can use many more, but that's another PR!

PR #1273, opened 2023-10-15
Introduce classes for ProjData indices

### LINKED ISSUE #1263: better indices for projection data
We need a more future-proof way to specify indices in projection data, e.g.
```c++
ProjData::get_viewgram(const ViewgramIndices&)
```
as opposed to what we have now
```c++
ProjData::get_viewgram(const int view_num, const int segment_num,
                               const bool make_num_tangential_poss_odd=false
#ifdef STIR_TOF
                               , const int timing_pos=0
#endif
                               )
```
(the `make_num_tangential_poss_odd` is long obsolete, but was unfortunately preserved on the TOF PR #304 so it's harder to remove now).

Having this will mean less code changes in the future once we add energy indices, layers etc.

https://github.com/NikEfth/STIR/pull/17 started this on the TOF branch, but was somewhat ambitious and got overtaken by other things.

One possible way to do this would be
```
class SegmentIndices
{ time, TOF, energy, whatever};
class ViewgramIndices: public SegmentIndices // replacing ViewsegmentNumbers
{ view_num};
class SinogramIndices: public SegmentIndices
{ axial_pos_num};
class Bin: public ViewgramIndices
{ tangential_pos_num};
```

Sadly, C++ has no "named parameters", so we shift some of the burden to constructing the relevant object, and having constructor arguments in the correct order. However, it would at least prevent loops being specific.

As noted in https://github.com/NikEfth/STIR/pull/17, loops could then be simplified by having something like this
```c++
 for (auto vg_idx: proj_data_info.get_all_viewgram_indices())
   ...
```

Anyone any suggestions for improvements?

KrisThielemans:
This PR introduces `SegmentIndices`, `ViewgramIndices` and `SinogramIndices` (and derives `Bin` from `ViewgramIndices` ) and add constructors and methods to use these, as opposed to listing indices separately. This will allow for more concise code, but also prepares for TOF (and energy/layers etc).

Using these new methods is left for the TOF branch.

`DataSymmetriesForViewSegmentNumbers` etc have not been modified yet.

Fixes #1263


KrisThielemans:
temp zenodo problems. Also, Appveyor didn't run, as last commit was only for the release notes. I'll need to push something else.

robbietuk:
I agree with the change and it looks good. I dont have time to fully review it. Are there any additional tests that should be added?

KrisThielemans:
> Are there any additional tests that should be added?

I guess so. However, the old functions are now mostly implemented in terms of the new ones, so we have the tests for free. There will be a few untested, but we will find out once we apply this to TOF.

I wanted to have this PR on 5.2, such that we can tell people to use the new interface, and it will work for the next STIR version as well (while the old functions won't work for TOF). Maybe I didn't make the latter very clear in the release notes....

KrisThielemans:
@danieldeidda this should be ready now, aside from some tiny edits to the release notes.

src/include/stir/Bin.h
```diff
@@ -91,8 +87,6 @@ class Bin
 private :
   // shared_ptr<ProjDataInfo> proj_data_info_ptr; 
   
-  int  segment;
```
danieldeidda, line 94: shouldn't we have a viewgram_indices member as well?

src/include/stir/Bin.inl
```diff
@@ -25,31 +26,24 @@
...
 Bin::Bin()
+  : time_frame(1)
 {}
 
 
 Bin::Bin(int segment_num,int view_num, int axial_pos_num,int tangential_pos_num,float bin_value)
-	 :segment(segment_num),view(view_num),
-	 axial_pos(axial_pos_num),tangential_pos(tangential_pos_num),bin_value(bin_value),time_frame(1)
-     {}
+  : ViewgramIndices(view_num, segment_num),
```
danieldeidda, line 34: oh ok I guess there is a member already and I di not see it

danieldeidda (commented):
looks good to me

src/include/stir/Bin.h
KrisThielemans: it's derived from it.

PR #1275, opened 2023-10-24
SPECTUB: write warning if image radius is larger than detector radius

### LINKED ISSUE #567: SPECTUB matrix should check rotation radius etc
Attempting to forward project SPECT data while details in the template sinogram are inconsistent with image dimensions (eg. a radius of rotation which is smaller than half of the x-dimension or the y-dimension in mm) gives the following error message:

weight3d: attpth larger than allocated
terminate called after throwing an instance of 'std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > 

Including a check for consistency and a more specific error message would be useful.

_Originally posted by @RebeccaGillen in https://github.com/UCL/STIR/issues/154#issuecomment-

KrisThielemans:
Fixes #567

### DIFF
--- a/src/recon_buildblock/ProjMatrixByBinSPECTUB.cxx
+++ b/src/recon_buildblock/ProjMatrixByBinSPECTUB.cxx
@@ -1,7 +1,7 @@
 /*
     Copyright (C) 2013, Institute for Bioengineering of Catalonia
     Copyright (C) Biomedical Image Group (GIB), Universitat de Barcelona, Barcelona, Spain.
-    Copyright (C) 2013-2014, 2019, 2020 University College London
+    Copyright (C) 2013-2014, 2019, 2020, 2023 University College London
     Copyright (C) 2023 National Physical Laboratory
     This file is part of STIR.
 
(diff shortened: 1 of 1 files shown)


KrisThielemans:
@mastergari potentially there's a similar warning that could be written for the pinhole code, but I didn't immediately see where. In any case, it is not vitally important.

mastergari:
@KrisThielemans the pinhole-SPECT code checks if a voxel is located behind or within a pinhole. These are checked when looping over the holes during weight matrix calculation.

 https://github.com/UCL/STIR/blob/ff1c20bbdad07d8b06120a0c9062016a46429960/src/recon_buildblock/PinholeSPECTUB_Weight3d.cxx#L1109
 https://github.com/UCL/STIR/blob/ff1c20bbdad07d8b06120a0c9062016a46429960/src/recon_buildblock/PinholeSPECTUB_Weight3d.cxx#L352-L361
https://github.com/UCL/STIR/blob/ff1c20bbdad07d8b06120a0c9062016a46429960/src/recon_buildblock/PinholeSPECTUB_Weight3d.cxx#L391-L400

PR #1288, opened 2023-11-14
SWIG: expose more LOR and listmode classes

KrisThielemans:
- Add missing LOR classes
- Rename `FloatLOR` to `LOR` in SWIG
- Add missing `ListRecord` etc.

Currently python3 crashes when I `import stir`, but that might be a local problem.


Robert-PrescientImaging (approved):
Looks good and great work with the test and demo

Issue #1290, opened 2023-11-23
Scatter simulation for BlocksOnCylindrical scanner geometry is relatively slow


markus-jehl:
For the NeuroLF scanner geometry, four iterations of scatter estimation take around 9 minutes of which the majority is spent doing the single scatter simulation. This is because the detectors per ring and the views are currently not downsampled (axial downsampling is the same than for cylindrical scanners). I would propose to add a downsampling in detectors and views by an integer factor, while maintaining the buckets. This would ignore gaps between blocks, but since the simulation result is quite smooth anyways that shouldn't introduce large errors.

PR #1291, opened 2023-11-24
Added downsampling of BlocksOnCylindrical scanners in scatter

markus-jehl:
Added the functionality to downsample BlocksOnCylindrical scanners in views and detectors per ring during scatter simulation, in order to speed up the computations. The default behaviour did not change.

## Testing performed
This was tested with iterative scatter estimation and simulation on NeuroLF and BPET blocks geometry and compared to non-downsampled scatter estimation. The differences were of the order of 1% when the crystals per ring and views were both downsampled by a factor of 2.

## Related issues
fixes https://github.com/UCL/STIR/issues/1290
fixes https://github.com/UCL/STIR/issues/1393
fixes https://github.com/UCL/STIR/issues/1497

  - [x] `documentation/release_XXX.md` has been updated with any functionality change (if applicable)


markus-jehl:
original scatter:
image
below are plots of the differences between downsampled scatter and original scatter, relative to abs(mean(original_scatter))=11.5
downsampling factor 1.066667 (from 32 to 30 detectors per module)
image
downsampling factor 1.28 (from 32 to 25 per module)
image
downsampling factor 1.6 (from 32 to 20 per module)
image
downsampling factor 2 (from 32 to 16 per module)
image
downsampling factor 3.2 (from 32 to 10 per module)
image

markus-jehl:
image
relative difference between original scatter and downsampled scatter with factor 2 (divided by mean intensity of areas with intensity > 0.1 => 0.5)
image

markus-jehl:
Therefore it looks like choosing integer factors for the downsampling isn't required, and the intersections between the modules remain problematic (sometimes over 10% errors in the scatter estimate). However, on the final image the differences caused by the downsampled scatter are below 3% of the average hot values.

markus-jehl:
As part of this investigation, a bug was found: https://github.com/UCL/STIR/issues/1396
This should be fixed on the ProjData side, but it raises another problem with the scatter estimation code itself: for the last iteration we first upsample from the downsampled scanner to the full-size scanner here: https://github.com/UCL/STIR/blob/8ced2d73933420457e0bc76074964b7d4ff00f0c/src/scatter_buildblock/ScatterEstimation.cxx#L937-L946
And then we run upsample_and_fit_scatter_estimate a second time here: https://github.com/UCL/STIR/blob/8ced2d73933420457e0bc76074964b7d4ff00f0c/src/scatter_buildblock/ScatterEstimation.cxx#L1017-L1026

The second call will fail once #1396 is fixed, since the input ProjData are correctly identified as not homogeneously sampled in axial direction. However, we also don't need the second interpolation step, because input and output ProjData have the same dimensions. Therefore, I would propose to replace that call with a simple inverse_SSRB followed by scatter_normalisation.undo(). This is all that is needed at this point.

src/include/stir/scatter/ScatterEstimation.h
```diff
@@ -379,6 +381,8 @@ class ScatterEstimation : public ParsingObject
   float min_scale_value;
 
   bool downsample_scanner_bool;
+  int downsampled_detectors_per_ring;
```
KrisThielemans, line 384: once we do this, it would make most sense to add the variable for the "dets per ring" as well.

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -77,6 +77,7 @@ ScatterEstimation::set_defaults()
   this->override_scanner_template = true;
   this->override_density_image = true;
   this->downsample_scanner_bool = true;
+  this->downsampled_detectors_per_ring = 0;
```
KrisThielemans, line 80: normally default variables like this should be -1, such that we know it's invalid. The code indeed checks for `<= 0`, so it's fine, but I'd prefer -1...

src/include/stir/scatter/ScatterEstimation.h
```diff
@@ -171,6 +171,8 @@ class ScatterEstimation : public ParsingObject
 
   inline void set_num_iterations(int);
 
+  inline void set_downsample_scanner(bool downsample_scanner, int downsampled_detectors_per_ring = 0);
```
KrisThielemans, line 174: once we do this, it would make most sense to add the variable for the "dets per ring" as well. 

normally default variables like this should be -1, such that we know it's invalid. The code indeed checks for <= 0, so it's fine, but I'd prefer -1...

KrisThielemans (changes requested):
I'm hoping that once this is done, we can avoid the almost repeated code for the blocks and cylindrical case, but maybe do that in stages.

markus-jehl:
> I'm hoping that once this is done, we can avoid the almost repeated code for the blocks and cylindrical case, but maybe do that in stages.

Do you mean in interpolate_projdata? This would indeed be very nice. I'll have a look at it!

src/include/stir/scatter/ScatterEstimation.h
KrisThielemans: oops sorry. yes, num rings

KrisThielemans:
> > I'm hoping that once this is done, we can avoid the almost repeated code for the blocks and cylindrical case, but maybe do that in stages.
> 
> Do you mean in interpolate_projdata? This would indeed be very nice. I'll have a look at it!

there's some in the ray-tracing projector as well, but that's on our wish-list :-)

KrisThielemans:
> The second call will fail once #1396 is fixed, since the input ProjData are correctly identified as not homogeneously sampled in axial direction. 

ok, although we could at least in principle make that check more sophisticated to see if needs to interpolate or not, but I'm fine with not doing that!

> However, we also don't need the second interpolation step, because input and output ProjData have the same dimensions. Therefore, I would propose to replace that call with a simple inverse_SSRB followed by scatter_normalisation.undo(). This is all that is needed at this point.

good point!

markus-jehl:
On a simulated cylinder, the scatter estimation has much smaller errors when we use linear interpolation instead of quadratic interpolation. I strongly suspect that this is because we quite aggressively downsample the scanner (e.g. 8 instead of 48 rings) to speed up the computation time. Quadratic interpolation simply can't cope with that level of simplification. Below is the quadratic interpolation vs. ground truth first, then the linear interpolation vs. ground truth:
image
image

markus-jehl:
Below is the difference of the downsampled (half detectors per ring) scatter estimation and the ground truth scatter followed by the non-downsampled scatter estimation. The scatter amplitude is a little over 8, so in both cases the errors are <2%.
image
And here is the same comparison with the reconstructed images (ground truth image has intensity 1, so the errors are usually below 1% inside the cylinder and only higher at the edge):
image

markus-jehl:
Plots as at the beginning of the PR: scatter for Hoffman phantom data with downsampling of 2 compared against no downsampling.
image
image

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -124,6 +127,8 @@ ScatterEstimation::initialise_keymap()
   this->parser.add_parsing_key("Scatter Simulation type", &this->scatter_simulation_sptr);
   this->parser.add_key("scatter simulation parameter filename", &this->scatter_sim_par_filename);
   this->parser.add_key("use scanner downsampling in scatter simulation", &this->downsample_scanner_bool);
+  this->parser.add_key("override number of rings", &this->downsampled_number_of_rings);
```
KrisThielemans, line 130:
```suggestion
  this->parser.add_key("override number of downsampled rings", &this->downsampled_number_of_rings);
```

src/scatter_buildblock/ScatterEstimation.cxx
```diff
@@ -834,7 +839,7 @@ ScatterEstimation::process_data()
   float local_min_scale_value = 0.5f;
   float local_max_scale_value = 0.5f;
 
-  stir::BSpline::BSplineType spline_type = stir::BSpline::quadratic;
+  stir::BSpline::BSplineType spline_type = stir::BSpline::linear;
```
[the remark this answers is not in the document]
KrisThielemans: @markus-jehl this might introduce a change in cylindrical? It could be a beneficial one, but would be surprising to this in this PR

src/buildblock/interpolate_projdata.cxx
```diff
@@ -200,96 +205,313 @@ interpolate_projdata(ProjData& proj_data_out,
...
+                  // translate the crystal position on each module from the full size scanner to the downsampled scanner
+                  const auto proj_data_in_info_ptr = dynamic_cast<const ProjDataInfoGenericNoArcCorr*>(&proj_data_in_info);
+                  const int dets_per_module_in
+                      = proj_data_in_info_ptr->get_scanner_sptr()->get_num_transaxial_crystals_per_bucket();
+                  double crystal1_out_module_idx = det1_num_out % dets_per_module_out;
+                  double crystal1_out_module_pos
+                      = static_cast<double>(
+                            std::floor(crystal1_out_module_idx
+                                       / proj_data_out_info_ptr->get_scanner_sptr()->get_num_transaxial_crystals_per_block()))
+                            * proj_data_out_info_ptr->get_scanner_sptr()->get_transaxial_block_spacing()
```
KrisThielemans, line 326: `std::floor` again not needed as `int` division. If that's not what you want, do the `double cast on one of the elements.

src/scatter_buildblock/ScatterEstimation.cxx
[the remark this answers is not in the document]
markus-jehl: Yes, this also affects cylindrical. A little bit higher up on this MR I have added difference images for both cases (8-Mar) showing that the errors are significantly smaller and more regular with linear interpolation. Would your preference be to remove this change for now and merge it as a separate MR?

src/buildblock/interpolate_projdata.cxx
```diff
@@ -200,96 +205,313 @@ interpolate_projdata(ProjData& proj_data_out,
...
+                  // find the two crystals for this bin and on which buckets (modules) they are
+                  int det1_num_out, det2_num_out;
+                  const auto proj_data_out_info_ptr = dynamic_cast<const ProjDataInfoGenericNoArcCorr*>(&proj_data_out_info);
+                  proj_data_out_info_ptr->get_det_num_pair_for_view_tangential_pos_num(det1_num_out,
+                                                                                       det2_num_out,
+                                                                                       index_out[2], /*view*/
+                                                                                       index_out[3] /* tangential pos */);
+                  const int dets_per_module_out = proj_data_out_info.get_scanner_sptr()->get_num_transaxial_crystals_per_bucket();
+                  const int det1_module = std::floor(det1_num_out / dets_per_module_out);
+                  const int det2_module = std::floor(det2_num_out / dets_per_module_out);
```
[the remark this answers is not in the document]
markus-jehl: true, this is redundant!

src/buildblock/interpolate_projdata.cxx
markus-jehl: in the first line the static_cast<double> is actually not needed, so I will remove it.

markus-jehl:
> I haven't checked the actual interpolation code, but trust you! However, some documentation what the code is doing would be good, as it's quite hard to see.

Ok, I will add an overall description.

src/scatter_buildblock/ScatterEstimation.cxx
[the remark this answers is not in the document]
KrisThielemans: Can you just create an issue for it then, pointing to the above evidence. Then we can close that issue with this PR.

Issue #1286, opened 2023-11-09
Many calls to print warnings use the deprecated function which ignores verbosity


markus-jehl:
Currently, most calls of warning() use the deprecated version of the function. This function turns out to ignore the verbosity. I would propose to replace all these calls to the deprecated version with boost::format formatted calls to the proper version.

E.g. https://github.com/UCL/STIR/blob/734a0d713df4318cea10f1a8ec5aea90ca093254/src/IO/InterfileHeader.cxx#L1397
would become:
`warning(boost::format("Interfile parsing ended up with the following scanner:\n%1%\n") % scanner_ptr_from_file->parameter_info().c_str());`

PR #1298, opened 2023-12-14
Switched some warning messages from deprecated function to boost::format

markus-jehl:
The deprecated warning function ignores verbosity. Therefore, this PR is switching selected calls over to the new warning function.

Changed some calls to deprecated warning function to new warning function using boost::format.

## Testing performed
Built and loaded proj data object that triggered some warnings. After changing verbosity these warnings were no longer printed.

## Related issues
fixes some instances of https://github.com/UCL/STIR/issues/1286

### DIFF
--- a/src/IO/InterfileHeader.cxx
+++ b/src/IO/InterfileHeader.cxx
@@ -36,6 +36,7 @@
 #include <functional>
 #include "stir/ProjDataInfoBlocksOnCylindricalNoArcCorr.h"
 #include "stir/ProjDataInfoGenericNoArcCorr.h"
+#include <boost/format.hpp>
 
 #ifndef STIR_NO_NAMESPACES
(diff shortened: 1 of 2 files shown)


KrisThielemans (commented):
It's a bit unfortunate to have to explicitly cast to `std::string`. What about adding another overload
```c++
void warning(const char * const m) { warning(std::string(m)); }
```
or something similar as work-around. This would catch all  instances that don't have extra arguments without needing to change the actual code.

markus-jehl:
unfortunately it doesn't compile because it says the call to "warning" is ambiguous: it doesn't know which deprecated function to call...
I would propose to keep this PR as is and then remove the deprecated implementation once and for all after the TOF merge.

KrisThielemans (approved):

PR #1300, opened 2023-12-14
Add CUDA build to GitHub Actions (and CMake clean-up)

KrisThielemans:
- add a GHA build with CUDA, whcih needed renaming of caches
- add missing source extensions in some CMakeLists.txt

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -41,6 +41,7 @@ jobs:
         - os: ubuntu-latest
           compiler: gcc
           compiler_version: 9
+          cuda_version: "0"
           BUILD_FLAGS: "-DSTIR_OPENMP=ON"
           BUILD_TYPE: "Release"
           parallelproj: "ON"
@@ -49,6 +50,7 @@ jobs:
         - os: ubuntu-latest
           compiler: clang
           #compiler_version:
+          cuda_version: "0"
           BUILD_FLAGS: "-DSTIR_OPENMP=ON"
           BUILD_TYPE: "Release"
           parallelproj: "ON"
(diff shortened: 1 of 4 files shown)


KrisThielemans:
@markus-jehl this works all ok, but it is the first time that we test gcc 12 with CUDA on GHA. Sadly `test_interpolate_projdata` fails
https://github.com/UCL/STIR/actions/runs/7219366591/job/19670252356#step:10:165
```
INFO: Performing symmetric interpolation test for Cylindrical scanner
INFO: Determined voxel size by dividing default_bin_size (6.52531) by zoom
WARNING: forward_project. Imaging modality unknown for either the image or the projection data or both.
Going ahead anyway.
Error : 6.3685 is larger than 0.00089545, symmetry errors larger than 0.01% of absolute values in axial direction
```
the "blocks" test works. Any idea?

(By the way, we need to set modality in these tests to avoid at least some warnings.)

KrisThielemans:
Left gcc-12, right gcc-11
image

@markus-jehl @danieldeidda any ideas?

markus-jehl:
How does the downsampled sinogram look? If that one is still symmetric, then it is definitely in the interpolation. Otherwise it must be some difference in how float numbers are rounded

KrisThielemans:
it's the downsampled data. left is gcc-12
image

markus-jehl:
This is very strange... I looked at the scanner definitions and the math checks out as far as I can tell. So it must be some weird floating point issue in how the cylinder is constructed for the downsampled scanner or similar.

KrisThielemans:
no problem in Debug build

KrisThielemans:
downsampled image in Debug mode occupies central 3 planes (out of 11). In RelWithDebug, it has 1 more non-zero plane, causing asymmetry.

KrisThielemans:
for the cylindrical scanner, the z-voxel size is 10mm, but the length of the cylinder is 40. So, this is sensitive to rounding.

KrisThielemans:
Should be fixed. @markus-jehl please have a look.

markus-jehl:
That looks clean, thank you!

PR #1319, opened 2024-01-11
TOF segment bug fix, extra members and update visualisation to add TOF

KrisThielemans:
- fixes `ProjData::get_segment_by_*(const SegmentIndices&)`
- adds some more members to `Segment` classes that use `*Indices`
- update @robbietuk 's visualisation tool to add a TOF slider etc

### DIFF
--- a/examples/python/projdata_visualisation/BackendTools/STIRInterface.py
+++ b/examples/python/projdata_visualisation/BackendTools/STIRInterface.py
@@ -1,4 +1,4 @@
-# Copyright 2022 University College London
+# Copyright 2022, 2024 University College London
 
 # Author Robert Twyman
 
@@ -23,7 +23,7 @@ class ProjDataDims(Enum):
     AXIAL_POS = auto()
     VIEW_NUMBER = auto()
     TANGENTIAL_POS = auto()
-
+    TIMING_POS = auto()
 
 class ProjDataVisualisationBackend:
     """Class used as STIR interface to the projection data for ProjDataVisualisation."""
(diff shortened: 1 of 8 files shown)


KrisThielemans:
@NicoleJurjew this might come in handy for showing your TOF data

robbietuk:
Am I right in thinking the summation over TOF bins collapses into a non-TOF view/sinogram? If so, we should add a checkbox to do this.

robbietuk:
I assume this is tested? Currently the demo uses `examples/recon_demo/smalllong.hs`, which is not TOF data.

Otherwise, merge.

robbietuk:
There is a bug, `segment_number` argument is not used. 


https://github.com/UCL/STIR/blob/c0c4c5b7aec9069f683feba22e5ce1cda43e7ca4/examples/python/projdata_visualisation/BackendTools/STIRInterface.py#L102-L117

```
return self.projdata.get_min_axial_pos_num(self.get_current_segment_num()), \ 
    self.projdata.get_max_axial_pos_num(self.get_current_segment_num()) 
```

Should read

```
return self.projdata.get_min_axial_pos_num(segment_num), \ 
    self.projdata.get_max_axial_pos_num(segment_num) 

```

KrisThielemans:
ok, but that was an existing bug then, so should have created problems in non-TOF as well. Can you check?

> I assume this is tested?

yes, on some local TOF data. It looked fine, but I probably didn't try to switch segments

robbietuk:
> ok, but that was an existing bug then, so should have created problems in non-TOF as well. Can you check?

Yes. It is actually a latent bug so no need to fix here. [c0c4c5b](https://github.com/UCL/STIR/pull/1319/commits/c0c4c5b7aec9069f683feba22e5ce1cda43e7ca4) drew my attention to it.

> yes, on some local TOF data. It looked fine, but I probably didn't try to switch segments

I think you can still merge this and I'll add an issue and PR some fixes.

Issue #1320, opened 2024-01-11
Feature suggestions for ProjData Visualization


robbietuk:
Relevant directory: https://github.com/UCL/STIR/tree/master/examples/python/projdata_visualisation

#1319 Added a TOF slider. The following are suggestions of changes to be made to ProjData Visualization

Further TOF improvements:
- [ ] Add a checkbox to sum over TOF indices, as suggested [here](https://github.com/UCL/STIR/pull/1319#issuecomment-1887589084) 
- [ ] Allow multiple ProjData matplotlib canvas widgets that can be controlled by a set of single sliders ([assumption of this suggestion](https://github.com/UCL/STIR/pull/1319#issuecomment-1887597481))
- [ ] [Add a TOF data demonstration ](https://github.com/UCL/STIR/pull/1319#issuecomment-1887624168)

General
- [x] Improve the "hideous" UI (I wrote it 😄). Example `mainLayout.addWidget(self.ProjDataVisualisationGroupBox, 1, 1)` -> `mainLayout.addWidget(self.ProjDataVisualisationGroupBox, 1, 0, 1, 2)` might help by expanding the matplotlib canvas. However window size would vary.
- [x] Add error handling for loading projdata with no binary or error when reading binary. The error should be displayed in the application window (see [here](https://github.com/UCL/STIR/issues/1320#issuecomment-1887725755))
- [ ] Colormap limit consistency and options to set values. Currently, the 2D slice determines the colormap and the limits. These should be consistent when sliders are adjusted.

robbietuk:
Suggestion: Add error handling for loading projdata with no binary or error when reading binary. The error should be displayed in the application window.

```
$ python3 examples/python/projdata_visualisation/ProjDataVisualisation.py 
ProjDataVisualisationBackend.load_data: Loading data from file: /home/rts/devel/stir/examples/PET_simulation/scatter_template.hs

WARNING: 
Interfile warning: I did not recognise the scanner neither from 
'originating_system' or 'number of detectors per ring' and 'number of rings'.


WARNING: Interfile warning: 'default_bin size (cm)' invalid.


WARNING: Interfile warning: 'axial crystals per singles unit' invalid.


WARNING: Interfile warning: 'transaxial crystals per singles unit' invalid.


WARNING: Interfile warning: 'distance between crystals in axial direction (cm)' invalid.


WARNING: Interfile warning: 'distance between crystals in transaxial direction (cm)' invalid.


[...]

PR #1321, opened 2024-01-11
ProjData Visualisation window improvement and segment num bug fix

robbietuk:
- Fixes a bug with the `segment_num` parameter not being used in the method, instead it was using an internal call to `self.get_current_segment_num()`. This limited the method to the current state of the data, rather than the abstract.
- Expand MLP canvas to full width
- Fix bug when a projdata file cannot be loaded (e.g. missing binary). Previously, the projection data would be updated and fail when loading the segment_data into memory. However, the `stir_interface.projdata` was updated. This lead to errors with slider and reading new data. The PR, uses temporary variables when projdata and segment are loaded.
- Remove `refresh_segment_data` as this was leading to issues with the above and recursive calls to `load_projdata`
- Add formatting to `print_projdata_configuration` and `print_segment_data_configuration`

## Testing performed
Yes

## Related issues
- Partially addresses #1320
- Addresses https://github.com/UCL/STIR/issues/1320#issuecomment-1887725755  regarding loading of projdata error handling
 - As mentioned: https://github.com/UCL/STIR/pull/1319#issuecomment-1887685739, addresses a bug with `segment_number` in `get_limits`

### DIFF
--- a/examples/python/projdata_visualisation/BackendTools/STIRInterface.py
+++ b/examples/python/projdata_visualisation/BackendTools/STIRInterface.py
@@ -85,8 +85,6 @@ class ProjDataVisualisationBackend:
             f"Number of non-tof sinograms:         {self.projdata.get_num_non_tof_sinograms():>10}\n\n"
         )
 
-
-
     def print_segment_data_configuration(self) -> None:
         """Prints the configuration of the segment data."""
         print(
--- a/examples/python/projdata_visualisation/BackendTools/STIRInterface.py
+++ b/examples/python/projdata_visualisation/BackendTools/STIRInterface.py
@@ -8,13 +8,12 @@
 #
 # See STIR/LICENSE.txt for details
 
-import time
+from enum import Enum, auto
 
 import numpy
+
 import stir
 import stirextra
-
-from enum import Enum, auto
 
 
 class ProjDataDims(Enum):
@@ -57,7 +56,7 @@ class ProjDataVisualisationBackend:
         self.projdata = new_projdata
         self.segment_data = new_segment_data
         print("ProjDataVisualisationBackend.load_data: Data loaded.")
-        self.print_projdata_configuration() 
+        self.print_projdata_configuration()
         return True
         
     def set_projdata(self, projdata: stir.ProjData) -> None:
--- a/examples/python/projdata_visualisation/BackendTools/STIRInterface.py
+++ b/examples/python/projdata_visualisation/BackendTools/STIRInterface.py
@@ -35,17 +35,16 @@ class ProjDataVisualisationBackend:
 
         self.segment_data = None
 
-    def load_projdata(self, filename=None) -> bool:
+    def load_projdata_from_file(self, filename: str | None = None) -> bool:
         """
         Loads STIR projection data from a file and updates the segment data in memory.
         :param filename: The filename to load the projection data from.
         :return: True if the data was loaded successfully, False otherwise.
         """
-        if filename is not None and filename != "":
-            self.projdata_filename = filename
-
-        if self.projdata_filename == "":
+        if filename is None or filename == "":
             return False
+
+        self.projdata_filename = filename
         
         print("ProjDataVisualisationBackend.load_data: Loading data from file: " + self.projdata_filename)
         try:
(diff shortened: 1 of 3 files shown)


robbietuk:
I have tested this for non-TOF sinogram data, everything seems okay. @KrisThielemans or @NicoleJurjew, could you please test for TOF data? I don't have any accessible right now.

P.S. We need some example TOF data?

KrisThielemans:
If you run recon_test_pack/run_test_listmode_recon.sh,  you'll have some (but needs ROOT)

examples/python/projdata_visualisation/BackendTools/STIRInterface.py
```diff
@@ -36,63 +36,66 @@ def __init__(self, *args, **kwargs) -> None:
...
-                self.segment_data = self.refresh_segment_data()
-                print("ProjDataVisualisationBackend.load_data: Data loaded.")
-                self.print_projdata_configuration()
-            except RuntimeError:
-                print("ProjDataVisualisationBackend.load_data: Error loading data from file: " + self.projdata_filename)
-                return False
-            return True
-        return False
+        if self.projdata_filename == "":
+            return False
```
KrisThielemans, line 48: I think this needs to be
```python
 if filename is None or filename == "":
       return False

  self.projdata_filename = filename
```

examples/python/projdata_visualisation/BackendTools/STIRInterface.py
```diff
@@ -36,63 +36,66 @@ def __init__(self, *args, **kwargs) -> None:
...
+    def get_segment_data(self, segment_number=0, timing_pos=0) -> stir.FloatSegmentByView:
+        """Returns the segment data."""
+        return self.projdata.get_segment_by_view(stir.SegmentIndices(segment_number, timing_pos))
         
     def set_projdata(self, projdata: stir.ProjData) -> None:
         """Sets the projection data stream."""
         self.projdata = projdata
         self.projdata_filename = "ProjData filename set externally, no filename"
-        self.segment_data = self.refresh_segment_data()
+        self.segment_data = self.projdata.get_segment_by_view(stir.SegmentIndices(segment_number, timing_pos))
```
KrisThielemans, line 72: looks like you can run you `get_segment_data` function

examples/python/projdata_visualisation/BackendTools/UIGroupboxProjdataDimensions.py
```diff
@@ -144,11 +143,23 @@ def update_enable_disable(self, sinogram_radio_button_state: bool):
 
         elif not sinogram_radio_button_state:
             self.disable(ProjDataDims.AXIAL_POS)
-            self.enable(ProjDataDims.VIEW_NUMBER)
+            view_num_limits = self.stir_interface.get_limits(ProjDataDims.VIEW_NUMBER,
+                                                             self.stir_interface.get_current_segment_num())
+            if (view_num_limits[0] - view_num_limits[1]) != 0:
```
KrisThielemans, line 148: fancy way if saying they're not equal... why not say `limits[1]>limits[0]`?

KrisThielemans (changes requested):
also 2 Codacy items

robbietuk:
I think there was a bug in #1319. When changing the TOF Bin slider, I am getting the same imaging (I think).
Adding a print statement to `timing_pos_refresh`:
```
    def timing_pos_refresh(self):
        """This function is called when the user changes the TOF bin value."""
        new_segment_num = self.UI_slider_spinboxes[ProjDataDims.SEGMENT_NUM].value()
        new_timing_pos = self.UI_slider_spinboxes[ProjDataDims.TIMING_POS].value()
        self.stir_interface.segment_data = self.stir_interface.get_segment_data(new_segment_num, new_timing_pos)
        print(f"TOF bin changed to {new_timing_pos}, "
              f"segment data has tof pos number: {self.stir_interface.segment_data.get_timing_pos_num()}")
        self.UI_controller_UI_change_trigger()
```
Console log:
```
TOF bin changed to -2, segment data has tof pos number: 0
```
This is calling
```
    def get_segment_data(self, segment_number=0, timing_pos=0) -> stir.FloatSegmentByView:
        """Returns the segment data."""
        return self.projdata.get_segment_by_view(stir.SegmentIndices(segment_number, timing_pos))
```
Something is going wrong on the CPP level. `get_segment_by_view` is defined here:
https://github.com/UCL/STIR/blob/151c457495721d38e1c3f00780a81055735bacb3/src/buildblock/ProjData.cxx#L409-L417

KrisThielemans:
#1319 fixed a bug in `ProjData::get_segment_by_view(SegmentIndices&)`. The relevant line is now
https://github.com/UCL/STIR/blob/151c457495721d38e1c3f00780a81055735bacb3/src/include/stir/ProjData.inl#L30
while it didn't have pass `timing_pos` through before.

Are you sure you recompiled? It does change the TOF Data for me (after that fix)

robbietuk:
> Are you sure you recompiled? It does change the TOF Data for me (after that fix)

I had recompiled but VSCode was not loading my .bash_profile with the updated install directory in the `$PYTHONPATH`. No wonder my debugging was failing.


I think that should do it. I changed more than I would like but the code probably needs a bit more with some improvements to the PyQt5 usage.

Once codacy has completed we can squash merge. You can use the details in the initial summary https://github.com/UCL/STIR/pull/1321#issue-2077383886

KrisThielemans (approved):
I haven't tried it myself yet, but favour merging ASAP. Thanks!

PR #1328, opened 2024-01-13
make coincidence_time_window arg of randoms_from_singles default

KrisThielemans:
we now have it in ProjDataInfo


KrisThielemans:
will be done in #1331

PR #1334, opened 2024-01-15
Allow even number of ToF bins

### LINKED ISSUE #1333: TOF: need to support even number of TOF bins and TOF_mashing_factor

danieldeidda:
fixing #1333


KrisThielemans:
@danieldeidda `test_DataSymmetriesForBins_PET_CartesianGrid` failed. Possibly we made a mistake in the merge, or there is something else.

danieldeidda:
something is happening with the segmentIndices::_timing_pos whcich occasionally is different. Need to go deeper in debugging though

src/include/stir/recon_buildblock/DataSymmetriesForBins_PET_CartesianGrid.inl
```diff
@@ -369,13 +369,13 @@ find_sym_op_general_bin(
...
           if ( do_symmetry_swap_s && s < 0) 
-            return new SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy(view180, axial_pos_shift, z_shift);
+            return new SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy(view180, axial_pos_shift, z_shift, proj_data_info_ptr->get_scanner_ptr()->get_max_num_timing_poss());
           else        
             return new SymmetryOperation_PET_CartesianGrid_swap_zq(view180, axial_pos_shift, z_shift, transform_z);   // s > 0
       }  
       else // segment_num = 0 
       {
-        if ( do_symmetry_swap_s && s < 0) return new SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy(view180, axial_pos_shift, z_shift);
+        if ( do_symmetry_swap_s && s < 0) return new SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy(view180, axial_pos_shift, z_shift, proj_data_info_ptr->get_scanner_ptr()->get_max_num_timing_poss());
```
danieldeidda, line 378: I guess the problem is happening  in one of these modifications

danieldeidda: Actually not sure anymore because the test is giving errors when the 2 bins are actually the same

KrisThielemans: can you paste some sample errors, for both TOF and non-TOF?

danieldeidda: WARNING: FactoryRegistry:: overwriting previous value of key in registry.
     key: None
Tests for DataSymmetriesForBins_PET_CartesianGrid
Testing span=1

INFO: Detected view-mashing factor 24 from the number of views (8) and the number of detectors per ring (384).
Adjusting the azimuthal angle offset accordingly (an extra offset of 10.7813 degrees)
	Tests with usual image size

INFO: Determined voxel size by dividing default_bin_size (3.12932) by zoom
		Testing with all symmetries

Error. Comparing symmetry bin configuration

Current bin: 	segment = -5, 	axial pos 0, 	view = 0, 	tangential_pos_num = -6, timing position index = 0
Symm bin: 		segment = -5, 	axial pos 0, 	view = 0, 	tangential_pos_num = -6, timing position index = 0
Error. Comparing symmetry bin configuration

Current bin: 	segment = -5, 	axial pos 0, 	view = 0, 	tangential_pos_num = -3, timing position index = 0
Symm bin: 		segment = -5, 	axial pos 0, 	view = 0, 	tangential_pos_num = -3, timing position index = 0
Error. Comparing symmetry bin configuration

Current bin: 	segment = -5, 	axial pos 1, 	view = 0, 	tangential_pos_num = -6, timing position index = 0
Symm bin: 		segment = -5, 	axial pos 1, 	view = 0, 	tangential_pos_num = -6, timing position index = 0
Error. Comparing symmetry bin configuration

Current bin: 	segment = -5, 	axial pos 1, 	view = 0, 	tangential_pos_num = -3, timing position index = 0
Symm bin: 		segment = -5, 	axial pos 1, 	view = 0, 	tangential_pos_num = -3, timing position index = 0
Error. Comparing symmetry bin configuration

Current bin: 	segment = -5, 	axial pos 2, 	view = 0, 	tangential_pos_num = -6, timing position index = 0
Symm bin: 		segment = -5, 	axial pos 2, 	view = 0, 	tangential_pos_num = -6, timing position index = 0
Error. Comparing symmetry bin configuration

Current bin: 	segment = -5, 	axial pos 2, 	view = 0, 	tangential_pos_num = -3, timing position index = 0
Symm bin: 		segment = -5, 	axial pos 2, 	view = 0, 	tangential_pos_num = -3, timing position index = 0
Error. Comparing symmetry bin configuration

Current bin: 	segment = -5, 	axial pos 3, 	view = 0, 	tangential_pos_num = -6, timing position index = 0
Symm bin: 		segment = -5, 	axial pos 3, 	view = 0, 	tangential_pos_num = -6, timing position index = 0
Error. Comparing symmetry bin configuration

Current bin: 	segment = -5, 	axial pos 3, 	view = 0, 	tangential_pos_num = -3, timing position index = 0
Symm bin: 		segment = -5, 	axial pos 3, 	view = 0, 	tangential_pos_num = -3, timing position index = 0
Error. Comparing symmetry bin configuration

Current bin: 	segment = -5, 	axial pos 4, 	view = 0, 	tangential_pos_num = -6, timing position index = 0
Symm bin: 		segment = -5, 	axial pos 4, 	view = 0, 	tangential_pos_num = -6, timing position index = 0
Error. Comparing symmetry bin configuration

Current bin: 	segment = -5, 	axial pos 4, 	view = 0, 	tangential_pos_num = -3, timing position index = 0
Symm bin: 		segment = -5, 	axial pos 4, 	view = 0, 	tangential_pos_num = -3, timing position index = 0

KrisThielemans:
test failure in `OSMAPOSL_parallelproj`  is weird as non-TOF, but  occurs in all Release builds (and isn't run in the Debug jobs), so not a one-off.

`test_proj_data_info` failures:
- printing of bins is also flawed (replace as in other file) so we can't see what's going on
- Debug runs have `assert` error https://github.com/UCL/STIR/actions/runs/7614648272/job/20737408998?pr=1334#step:10:921
- worrying is `Error. malloc(): corrupted top size` https://github.com/UCL/STIR/actions/runs/7614648272/job/20737408026?pr=1334#step:10:1110

Other debug asserts:
- https://github.com/UCL/STIR/actions/runs/7614648272/job/20737408998?pr=1334#step:10:174
- https://github.com/UCL/STIR/actions/runs/7614648272/job/20737408998?pr=1334#step:10:745

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -729,15 +732,15 @@ actual_compute_subset_gradient_without_penalty(TargetT& gradient,
                                  this->proj_data_sptr, 
                                  subset_num, 
                                  this->num_subsets, 
-                                 -this->max_segment_num_to_process,
+                                 0,
```
KrisThielemans, line 735: This seems the only non-TOF change. A mistake?

src/test/test_proj_data_info.cxx
```diff
@@ -147,7 +147,7 @@ ProjDataInfoTests::test_generic_proj_data_info(ProjDataInfo& proj_data_info)
         "check on min/max_tangential_pos_num being (almost) centred");
   check_if_equal(proj_data_info.get_max_tof_pos_num() - proj_data_info.get_min_tof_pos_num() + 1,
                  proj_data_info.get_num_tof_poss(), "basic check on min/max/num_tof_pos_num");
-  check_if_equal(proj_data_info.get_max_tof_pos_num() + proj_data_info.get_min_tof_pos_num(), 0,
+  check_if_equal(proj_data_info.get_max_tof_pos_num() + proj_data_info.get_min_tof_pos_num(), proj_data_info.get_max_tof_pos_num(),
                  "check on min/max_tof_pos_num being (almost) centred");
```
KrisThielemans, line 151: this check boils down to `get_min_tof_pos_num()==0`. Better to do that.

src/test/test_proj_data_info.cxx
```diff
@@ -236,7 +236,7 @@ ProjDataInfoTests::test_generic_proj_data_info(ProjDataInfo& proj_data_info)
                             || !check(diff_view_num <= 1, "round-trip get_LOR then get_bin: view")
                             || !check(diff_axial_pos_num <= 1, "round-trip get_LOR then get_bin: axial_pos")
                             || !check(diff_tangential_pos_num <= 1, "round-trip get_LOR then get_bin: tangential_pos")
-                            || !check(diff_timing_pos_num == 0, "round-trip get_LOR then get_bin: timing_pos"))
+                            || !check(diff_timing_pos_num == 0 || diff_timing_pos_num ==  proj_data_info.get_max_tof_pos_num() , "round-trip get_LOR then get_bin: timing_pos"))
```
KrisThielemans, line 239: why do we have this? If it's valid, it needs to be commented.

src/test/test_proj_data_info.cxx
```diff
@@ -293,7 +293,7 @@ ProjDataInfoTests::test_generic_proj_data_info(ProjDataInfo& proj_data_info)
                         || !check(diff_view_num <= 1, "round-trip get_LOR then get_bin (LORAs2Points): view")
                         || !check(diff_axial_pos_num <= 1, "round-trip get_LOR then get_bin (LORAs2Points): axial_pos")
                         || !check(diff_tangential_pos_num <= 1, "round-trip get_LOR then get_bin (LORAs2Points): tangential_pos")
-                        || !check(diff_timing_pos_num == 0, "round-trip get_LOR then get_bin: timing_pos"))
+                        || !check(diff_timing_pos_num == 0 || diff_timing_pos_num ==  proj_data_info.get_max_tof_pos_num(), "round-trip get_LOR then get_bin: timing_pos"))
```
KrisThielemans, line 296: as above

KrisThielemans (changes requested):

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
```diff
@@ -605,8 +605,11 @@ set_up_before_sensitivity(shared_ptr<const TargetT > const& target_sptr)
       return Succeeded::no;
     }
 
-  this->max_timing_pos_num_to_process =
-    this->proj_data_sptr->get_max_tof_pos_num();
+    this->max_num_central_timing_poss_to_process =
+      this->proj_data_sptr->get_max_tof_pos_num();
```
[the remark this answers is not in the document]
danieldeidda: mh not sure than the shift is different then zero if we want to use them all

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
danieldeidda: looks like a mistake

src/test/test_proj_data_info.cxx
danieldeidda: I have to say I do not remember but if I remove it it fails. need to think about why

src/test/test_proj_data_info.cxx
danieldeidda: this happens because the diff_timing_pos depends on view_num and when doing the roundtrip view_num goes from 0 to max and so diff_timing_pos=org_bin.timing_pos_num() + new_bin.timing_pos_num()
@KrisThielemans too long as a comment?

KrisThielemans: I think that means that https://github.com/UCL/STIR/blob/605f950f56b6536a3dbe16658d9389219ebec1f2/src/test/test_proj_data_info.cxx#L215-L216 is incorrect. The swapping should have been detected by the check on views, so after that, the TOF positions should be close.

danieldeidda: so Are you saying diff_timing_pos_num=intabs(org_bin.timing_pos_num() - new_bin.timing_pos_num()) always?

KrisThielemans: no, if view is 180 degrees, then the timing pos would have flipped sign in the previous case. Now, the timing pos would be `get_max_timing_pos_num() - org_timing_pos`, so I suppose that would mean
```
const int diff_timing_pos_num = intabs(org_bin.view_num() - new_bin.view_num()) < proj_data_info.get_num_views() - intabs(org_bin.view_num() - new_bin.view_num()) ?  
   intabs(org_bin.timing_pos_num() - new_bin.timing_pos_num()) : intabs(org_bin.timing_pos_num() - (get_max_timing_pos_num()  - new_bin.timing_pos_num())); 
```
it does make sense that this condition has to change w.r.t. the old case.

Obviously, it'd look nicer if we introduce a variable
```
const bool views_are_close = intabs(org_bin.view_num() - new_bin.view_num()) < proj_data_info.get_num_views() - intabs(org_bin.view_num() - new_bin.view_num())
```
or similar name, and use it for all checks.

KrisThielemans: fixed! thanks

src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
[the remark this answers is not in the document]
KrisThielemans: in fact, we don't support this yet, but having wrong code is a bad idea. Add a comment and throw an error if the argument isn't what we expect? 

However, it seems hard to know what the argument should be. For 2 TOF bins, the max number of central should be 1. For 3 TOF bins, it should also be 1.5... (1 could work, but how would one say then we only want to do the central one). Looks like the name is still wrong. 

Or remove...

KrisThielemans:
I've merged `master` to resolve conflict. Sadly, the MacOS build now fails because it requires C++-14 (due to upgraded boost version). https://github.com/UCL/STIR/actions/runs/7677624387/job/20926592594?pr=1334#step:9:1146
```
/usr/local/include/boost/math/tools/config.hpp:23:6: warning: #warning "The minimum language standard to use Boost.Math will be C++14 starting in July 2023 (Boost 1.82 release)" [-Wcpp]
   23 | #    warning "The minimum language standard to use Boost.Math will be C++14 starting in July 2023 (Boost 1.82 release)"
```

KrisThielemans:
@danieldeidda I'd really like to release STIR 6.0.0 soon. We can easily postpone this to 6.1. Let me know.

danieldeidda:
Hi I have been looking at it but  haven't been able to find the source of the error. Let's postpone to 6.1 then

src/include/stir/recon_buildblock/DataSymmetriesForBins_PET_CartesianGrid.inl
```diff
@@ -353,7 +353,7 @@ find_sym_op_general_bin(
     if ( !do_symmetry_swap_segment || segment_num > 0) 
     {   
       if ( do_symmetry_swap_s && s < 0) 
-        return new SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy_zq(view180, axial_pos_shift, z_shift, transform_z);
+        return new SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy_zq(view180, axial_pos_shift, z_shift, transform_z, proj_data_info_ptr->get_scanner_ptr()->get_max_num_timing_poss()-1);
```
KrisThielemans, line 356: probably has to be ` proj_data_info_ptr->get_max_tof_pos_num() - 1`

src/include/stir/recon_buildblock/DataSymmetriesForBins_PET_CartesianGrid.inl
```diff
@@ -369,13 +369,13 @@ find_sym_op_general_bin(
           return new SymmetryOperation_PET_CartesianGrid_swap_zq(view180, axial_pos_shift, z_shift, transform_z);
         else*/  
           if ( do_symmetry_swap_s && s < 0) 
-            return new SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy(view180, axial_pos_shift, z_shift);
+            return new SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy(view180, axial_pos_shift, z_shift, proj_data_info_ptr->get_scanner_ptr()->get_max_num_timing_poss()-1);
```
KrisThielemans, line 372: probably has to be ` proj_data_info_ptr->get_max_tof_pos_num() - 1`

src/include/stir/recon_buildblock/DataSymmetriesForBins_PET_CartesianGrid.inl
```diff
@@ -369,13 +369,13 @@ find_sym_op_general_bin(
...
           else        
             return new SymmetryOperation_PET_CartesianGrid_swap_zq(view180, axial_pos_shift, z_shift, transform_z);   // s > 0
       }  
       else // segment_num = 0 
       {
-        if ( do_symmetry_swap_s && s < 0) return new SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy(view180, axial_pos_shift, z_shift);
+        if ( do_symmetry_swap_s && s < 0) return new SymmetryOperation_PET_CartesianGrid_swap_xmx_ymy(view180, axial_pos_shift, z_shift, proj_data_info_ptr->get_scanner_ptr()->get_max_num_timing_poss()-1);
```
KrisThielemans, line 378: probably has to be ` proj_data_info_ptr->get_max_tof_pos_num() - 1`

danieldeidda (commented):
this seems to have fixed the problem. Let's see how the test goes

KrisThielemans:
We will still have to think about consequences for IO. In particular, Siemens TOF order needed to be handled, so this code will have to be adapted
https://github.com/UCL/STIR/blob/8ced2d73933420457e0bc76074964b7d4ff00f0c/src/IO/stir_ecat_common.cxx#L208-L212

KrisThielemans:
Should put a check in
https://github.com/UCL/STIR/blob/8ced2d73933420457e0bc76074964b7d4ff00f0c/src/buildblock/ProjDataFromStream.cxx#L152-L155
that values in the `seq` array are between `min` and `max`, to catch people using the "old" convention.

KrisThielemans:
https://github.com/UCL/STIR/pull/1389 introduced the capability to read `TOF bin order` from the Interfile header:
https://github.com/UCL/STIR/blob/8ced2d73933420457e0bc76074964b7d4ff00f0c/src/IO/InterfileHeader.cxx#L621C6-L621C19
The value of that keyword would now be different.
(Note that #1389 was after the 6.0.0 release, so we are not breaking backwards compatibility w.r.t. a released version)

src/IO/stir_ecat_common.cxx
```diff
@@ -208,7 +208,7 @@ find_timing_poss_sequence(const ProjDataInfo& pdi)
   for (int timing_pos_num = 1; timing_pos_num <= max_timing_pos_num; ++timing_pos_num)
     {
       timing_pos_sequence[2 * timing_pos_num - 1] = timing_pos_num;
```
KrisThielemans, line 210: this cannot be correct. Even the positive bins have a different number now. I think it has to be
```suggestion
      timing_pos_sequence[2 * timing_pos_num - 1] = max_timing_pos_num + timing_pos_num;
```
Note that comment line a few lines earlier will need to be adapted as well.

src/buildblock/ProjDataInfoCylindrical.cxx
```diff
@@ -534,12 +534,14 @@ ProjDataInfoCylindrical::get_LOR(LORInAxialAndNoArcCorrSinogramCoordinates<float
   const float z1 = (m_in_mm - max_a * tantheta);
   const float z2 = (m_in_mm - min_a * tantheta);
 
+  const bool swap = false; // get_tof_delta_time(bin)<0;
```
KrisThielemans, line 537: actually, this should be deleted, and the whole change w.r.t. `master` reverted

KrisThielemans (changes requested):

KrisThielemans:
we got through `ctest`, but `run_test_simulate_and_recon.sh` is failing again. 😢

KrisThielemans:
@danieldeidda although I like the approach that you've taken here (starting `timing_pos_num` from 0), we haven't been able to resolve the problems.

@NikEfth used the simpler approach of using an asymmetric range (e.g., -2,-1,0,+1), which is less intuitive, but did work and has been merged #1604.

We could still pick this up though. What do you think?

danieldeidda:
Hi,
Yes if it work I am totally on board :D

PR #1357, opened 2024-01-31
remove forward_projector from distributable_sensitivity_computation

### LINKED ISSUE #1352: sensitivity calculation with ProjData can do unnecessary forward projections
We currently set the forward projector in the call
https://github.com/UCL/STIR/blob/208fd3c3fcb8e5ff722daed12bdf42cd45f0693c/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L887
which will do
https://github.com/UCL/STIR/blob/208fd3c3fcb8e5ff722daed12bdf42cd45f0693c/src/recon_buildblock/distributable.cxx#L408
However, for Parallelproj and NiftyPET projectors, this means that the projection is done, even though it is never used.

There are 2 ways to prevent this:
- pass a 0 pointer as forward projector and adapt `distributable_computation` 
- change Parallelproj forward projector to only do the projection if at least one projection of a `related_viewgrams` (or whatever) is called.

The first options is easiest...

KrisThielemans:
It doesn't need it, and in previous versions, calling forward_projector::set_input could actually do a lot of work (e.g. in the current GPU versions)

fixes #1352

### DIFF
--- a/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
+++ b/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx
@@ -1,7 +1,7 @@
 /*
     Copyright (C) 2000 PARAPET partners
     Copyright (C) 2000-2011, Hammersmith Imanet Ltd
-    Copyright (C) 2014, 2016-2023 University College London
+    Copyright (C) 2014, 2016-2024 University College London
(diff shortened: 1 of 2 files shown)


KrisThielemans:
@markus-jehl this should speed-up the `set_up()` (i.e. sensitivity calculation) when using parallelproj.

markus-jehl:
Nice, thank you!

PR #1360, opened 2024-02-01
More c++ examples

KrisThielemans:
Added a new example to show how to use STIR as an external project. And test it in GHA. This therefore serves as a test for `STIRConfig.cmake` export etc.

Builds on top of #1358 for building, so that one needs to be merged first.


examples/C++/using_installed_STIR/demo_create_image.cxx
```diff
@@ -0,0 +1,60 @@
...
+  try
+    {
+      stir::write_to_file(output_filename, image);
+      stir::info("Image written as " + output_filename);
+    }
+  catch (...)
+    {
+      return EXIT_FAILURE;
+    }
+  return EXIT_SUCCESS;
```
robbietuk, line 59:
```suggestion
  try
    {
      stir::write_to_file(output_filename, image);
    }
  catch (...)
    {
      stir::info("Failed to write image to file " + output_filename);
      return EXIT_FAILURE;
    }
  stir::info("Image written as " + output_filename);
  return EXIT_SUCCESS;  
```

examples/C++/using_installed_STIR/demo_create_image.cxx
```diff
@@ -0,0 +1,60 @@
...
+  int output_image_size_x = 100;
+  int output_image_size_y = 104;
+  int output_image_size_z = 20;
+  float output_voxel_size_x = 5.F;
+  float output_voxel_size_y = 4.F;
+  float output_voxel_size_z = 3.F;
+  double image_duration = 100.; // secs
+  double rel_start_time = 1; //secs
+
+  auto exam_info_sptr = std::make_shared<stir::ExamInfo>(stir::ImagingModality::PT);
```
robbietuk, line 24: I don't love the use of `auto` but its fine

examples/C++/using_installed_STIR/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+
+#========================================================================
+# Author: Kris Thielemans
+# Copyright 2016 - 2020 University College London
+# Copyright 2016 - 2020 Science Technology Facilities Council
```
robbietuk, line 5: Correct dates?

robbietuk (commented):
This is great and makes using STIR as a library much easier. I was able to follow your instructions and build the `using_installed_stir` project.

examples/C++/using_installed_STIR/demo_create_image.cxx
KrisThielemans: same result really. if `write_to_file` throws, the `info` will never be  executed.

examples/C++/using_installed_STIR/demo_create_image.cxx
KrisThielemans: I start to like it more... Less verbose than
```c++
std::shared<stir::ExamInfo> exam_info_sptr(new ExamInfo(stir::ImagingModality::PT));
```

examples/C++/using_installed_STIR/CMakeLists.txt
KrisThielemans: I guess not!

examples/C++/using_installed_STIR/demo_create_image.cxx
robbietuk: True, minor point, ignore it. I am learning error handling with golang and I generally only `try` one thing at a time, especially with `catch (...)`. In this case it does not matter as it is the end of the program.

Issue #857, opened 2021-04-18
missing overrides


KrisThielemans:
https://travis-ci.com/github/UCL/STIR/jobs/499276315#L12392
```
n file included from /Users/travis/build/UCL/STIR/src/include/stir/recon_buildblock/BinNormalisationSPECT.h:22:
/Users/travis/build/UCL/STIR/src/include/stir/recon_buildblock/BinNormalisationWithCalibration.h:47:9: warning: 'get_calibration_factor' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
  float get_calibration_factor() const;
        ^
/Users/travis/build/UCL/STIR/src/include/stir/recon_buildblock/BinNormalisation.h:66:17: note: overridden virtual function is here
  virtual float get_calibration_factor() const {return -1;}
                ^
In file included from /Users/travis/build/UCL/STIR/src/recon_buildblock/recon_buildblock_registries.cxx:56:
/Users/travis/build/UCL/STIR/src/include/stir/recon_buildblock/BinNormalisationFromAttenuationImage.h:101:16: warning: 'apply' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
  virtual void apply(RelatedViewgrams<float>& viewgrams) const;
               ^
/Users/travis/build/UCL/STIR/src/include/stir/recon_buildblock/BinNormalisation.h:96:16: note: overridden virtual function is here
  virtual void apply(RelatedViewgrams<float>&) const;
               ^
```

PR #1367, opened 2024-02-05
run clang-tidy --modernize-use-override

KrisThielemans:
following prescription of #857 after getting rid of old compiler work-arounds


KrisThielemans:
I used
```
 git commit -a --author="stir_maintenance <noreply@github.com>"
```

PR #1369, opened 2024-02-05
 adding data_source for radionuclide database

danieldeidda:


KrisThielemans:
great. This seems to work!

can you add something the release notes as well (old format still supported, new format recommended). please commit that with `[ci skip]` in the commit message

Issue #1363, opened 2024-02-02
Python plot_sinogram_profiles confused by TOF


KrisThielemans:
data is now 4D, so the int sumaxis,select leads to a load of profiles.

@robbietuk can we modify this to pass tuples? Or alternatively "sumallotheraxes"

robbietuk:
Yes. Let me take a look and get back to you. Behavior control is quite hard. It might be better to merge this with https://github.com/UCL/STIR/blob/master/examples/python/projdata_visualisation/ProjDataVisualisation.py

Some notes for myself:
This utility is hard.
- The idea is to plot a 1D line profile through (multiple) ProjData.
- This means each dimension of the data should be: a selected value or summed over all rows.
- 2D data: The dimension to plot across should be given (i.e. horizontal on plot) and vertical should be a value of the other ProjData dimension or a sum of it.
- In 3D The horizontal  component dimension should be provided and the vertical should be the given values of the other dimensions or a sum.
- This extends to 4D data.

PR #1370, opened 2024-02-05
Rebuilt `plot_sinogram_profiles.py` for 4D tof data

robbietuk:
Built `plot_projdata_profiles.py` to replace `plot_sinogram_profiles.py`. This allows for TOF data to be handled. I added significant error handling and what I hope is a better user experience. It does break backwards compatabity. 

I need to add a the ability to read v6.0 STIR projection data.

## Testing performed
Include a pytest (commented out)

## Related issues
#1363


examples/python/plot_projdata_profiles.py
```diff
@@ -0,0 +1,305 @@
...
+                        type=int,
+                        default=None,
+                        help='The TOF value of the projection data to plot.'
+                             'The default is to sum over all TOF values.')
+    parser.add_argument('--axial_segment',
+                        dest="axial_segment",
+                        type=int,
+                        default=None,
+                        help='The axial segment number of the projection data to plot.'
+                             'The default is to sum over all axial segments.')
```
robbietuk, line 210: Axial and segment data are collapsed into the same dimension of the data, this means that it isn't easy to show only axial components.

examples/python/plot_projdata_profiles.py
```diff
@@ -0,0 +1,305 @@
...
+    if len(args.filenames) < 1:
+        parser.print_help()
+        exit(0)
+
+    plot_sinogram_profiles(projection_data_list=args.filenames,
+                           display_axis=args.display_axis,
+                           data_indices=[args.tof, args.axial_segment, args.view, args.tangential]
+                           )
+
+# def test_generate_1d_from_4d():
```
robbietuk, line 235: Add this to pytest within STIR?

robbietuk (commented):
A few usage observations

examples/python/plot_projdata_profiles.py
KrisThielemans: Agreed. However, many users will not be so surprised by this, as the result of `to_numpy` has this "feature" as well. It's also how GE and Siemens store data. So I think we just leave this as-is.

examples/python/plot_projdata_profiles.py
KrisThielemans: I forgot how!

KrisThielemans (changes requested):
I'm happy with this and would be ok to merge (after an update to release_6.1.htm)

src/swig/test/python/test_projdata_profiles.py
```diff
@@ -0,0 +1,79 @@
+import numpy as np
+import pytest
+
+from examples.python.plot_projdata_profiles import compress_and_extract_1d_from_nd_array
```
robbietuk, line 4: If this pytest fails, it will likely be here. This worked on my debugger but failed on my CLI, it was probably an enviroment issue.

KrisThielemans:
pytest fails with 
```
ModuleNotFoundError: No module named 'examples'
```

Could be solved by adding it to the pythonpath first (would need to be done both for GHA and AppVeyor). Alternatively, we do it better via https://github.com/UCL/STIR/issues/1392

KrisThielemans:
@robbietuk want to finish this one? I think only needs
- setting python path for pytest
- updating release notes.

I'm hoping to release 6.3 soon, as it has a major bug fix. We could leave this for later if you have no time now.

src/swig/test/python/test_projdata_profiles.py
KrisThielemans: easiest silly work-around is probably
```suggestion
import os, sys
sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))))
from examples.python.plot_projdata_profiles import compress_and_extract_1d_from_nd_array
```

KrisThielemans (approved):
This should be fine now. ok to squash merge?

KrisThielemans:
GHA is fine but AppVeyor fails
```
=========================== test session starts =============================
platform win32 -- Python 3.8.20, pytest-8.3.4, pluggy-1.5.0
rootdir: C:\projects\stir\src
collected 25 items / 1 error
=================================== ERRORS ====================================
_________ ERROR collecting swig/test/python/test_projdata_profiles.py _________
swig\test\python\test_projdata_profiles.py:8: in <module>
_________ ERROR collecting swig/test/python/test_projdata_profiles.py _________
swig\test\python\test_projdata_profiles.py:8: in <module>
    from examples.python.plot_projdata_profiles import compress_and_extract_1d_from_nd_array
..\examples\python\plot_projdata_profiles.py:35: in <module>
    def get_projdata_from_file_as_numpy(filename: str) -> np.ndarray | None:
E   TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```
I think this is because python 3.8 is too old for this type of hinting.

KrisThielemans:
I'd prefer to support python 3.8 and 3.9 for a bit longer. https://typing.readthedocs.io/en/latest/guides/modernizing.html seems to say `from __future__ import annotations` could do the trick even for python 3.7. Alternatively, https://adamj.eu/tech/2022/01/15/python-type-hints-how-to-split-types-by-python-version/ could provide some hints on how to cope with this

robbietuk:
Feel free to do anything to merge this. Thank you for getting the pytesting working. I hope my Union[] fix resolves the Appveyor issue

KrisThielemans:
`Union` didn't work, but it looks like importing `annotations` does (at least in my test). let's see

KrisThielemans:
success! (at least for 3.8)

Issue #1374, opened 2024-02-07
Scanner number of axial components is not logical for BlocksOnCylindrical


robbietuk:
Issue
------

I have have been experiencing an issue related to the number of axial crystals in a scanner when the wrong `num_rings` value is input. I had the rough configuration
```
num_axial_crystals_per_block = 1
num_axial_blocks_per_bucket = 2
num_rings = 2 * 3  // Attempting 3 axial buckets
```
The issue I found is that everything is created cleanly but when back projecting a uniform 1s to a generated volume, the back projection only occurred in 1/3 of the volume.  Note, this is only an issue with `BlocksOnCylindrical`.

CTest Example
------

Created a test to demonstrate the issue:
```
void
BlocksTests::run_my_test()

{
  //    create projadata info
  auto scannerBlocks_sptr = std::make_shared<Scanner>(Scanner::SAFIRDualRingPrototype);
  {
    scannerBlocks_sptr->set_average_depth_of_interaction(5);
    scannerBlocks_sptr->set_num_axial_crystals_per_block(1);
    scannerBlocks_sptr->set_axial_block_spacing(scannerBlocks_sptr->get_axial_crystal_spacing()
                                                * scannerBlocks_sptr->get_num_axial_crystals_per_block());
    scannerBlocks_sptr->set_transaxial_block_spacing(scannerBlocks_sptr->get_transaxial_crystal_spacing()
                                                     * scannerBlocks_sptr->get_num_transaxial_crystals_per_block());
    scannerBlocks_sptr->set_num_axial_blocks_per_bucket(2);
    scannerBlocks_sptr->set_num_rings(2 * 3);  // Attempting 3 axial buckets
    scannerBlocks_sptr->set_scanner_geometry("BlocksOnCylindrical");
    scannerBlocks_sptr->set_up();
  }

  auto proj_data_info_blocks_sptr = std::make_shared<ProjDataInfoBlocksOnCylindricalNoArcCorr>();
  proj_data_info_blocks_sptr = set_blocks_projdata_info<ProjDataInfoBlocksOnCylindricalNoArcCorr>(scannerBlocks_sptr);
  auto projdata_sptr
      = std::make_shared<ProjDataInMemory>(std::make_shared<ExamInfo>(ImagingModality::PT), proj_data_info_blocks_sptr);

  auto origin = CartesianCoordinate3D<float>(0, 0, 0);
  auto volume_dimensions = CartesianCoordinate3D<int>(-1, -1, -1);
[...]

PR #1236, opened 2023-08-24
update Array hierarchy and allocate nD arrays in a contiguous block by default

KrisThielemans:
addresses the following
- `Array` and `VectorWithOffset` did not have move constructors, so we were using the default generated ones which might not be optimal
- some copy constructors were the default generates ones, but they now needed to be added explicitly
- some assignment operators were the default generates ones, which could imply unnecessary reallocation (in 1D)
- nD `Array`s used many different separately allocated 1D Arrays. This had the consequence that memory is not contiguous (which prevents some optimisations) as well as many calls to `new[]` and `delete[]`, which turns out to be slow. The default is not to allocate one single block, and let the nD Array point into that. This happens transparent to the rest of the code. **CAVEAT**: it does mean that growing an nD array in the "first" dimensions could now be less efficient (as it will reallocate everything).

Currently `ctest` is happy on my VM and `test_Array` and `test_VectorWithOffset` work fine via `valgrind` (Ubuntu, gcc8). We'll see what GHA and AppVeyor say...

@markus-jehl this should be fine for you to test now.

Things still to do:
- [x] check if growing an nD contiguous array deallocates original block if it can
- [x] adapt other code to take advantage of this, e.g. reading/writing nD `Array`s could now read/write in 1 chunk, which would be faster, e.g. `read_data`, `convert_array`
- [ ] make nD `Array` assignment such that it doesn't create a copy first if reallocation can be avoided.
- [ ] check if we need move constructors/assignments in derived classes such as `Sinogram` etc, or are the default ones ok. (`operator=` might need implemented to benefit from previous bullet)
- [ ] check if it actually works for everything and improves practical run-times

PR #1375, opened 2024-02-08
Axial bucket projection issue

robbietuk:
In response to #1374, add an integration CTest that ensures BlocksOnCylindrical with more than 1 axial bucket back project into all z slices of a volume derived from the projection data.

Add a guard to ensure `num_rings % num_crystals_per_block == 0`.

## Testing performed
Added a new CTest

## Related issues
#1374

### DIFF
--- a/documentation/release_6.1.htm
+++ b/documentation/release_6.1.htm
@@ -84,8 +84,10 @@ from the above):</H2>
 
 <h3>Backward incompatibities</h3>
 <ul>
-<li>
-</li>
+    <li>
+        Additional checks on <code>GeometryBlocksOnCylindrical</code> scanner configuration, which may lead to an
+        error being raised, while previously the code silently proceeded.
+    </li>
 </ul>
 
 <h3>New functionality</h3>
--- a/src/buildblock/GeometryBlocksOnCylindrical.cxx
+++ b/src/buildblock/GeometryBlocksOnCylindrical.cxx
@@ -45,7 +45,8 @@ START_NAMESPACE_STIR
 
 GeometryBlocksOnCylindrical::GeometryBlocksOnCylindrical(const Scanner& scanner)
 {
-  check_scanner_configuration(scanner);
+  if (scanner.check_consistency() == Succeeded::no)
+    error("Error in GeometryBlocksOnCylindrical: scanner configuration not accepted. Please check warnings.");
   build_crystal_maps(scanner);
 }
 
@@ -57,21 +58,6 @@ GeometryBlocksOnCylindrical::get_rotation_matrix(float alpha) const
                           stir::make_1d_array(0.F, -1 * std::sin(alpha), std::cos(alpha)));
 }
 
-void
-GeometryBlocksOnCylindrical::check_scanner_configuration(const Scanner& scanner)
-{
-  { // Ensure the number of axial crystals per bucket is a multiple of the number of rings
-    int num_axial_crystals_per_bucket = scanner.get_num_axial_crystals_per_block() * scanner.get_num_axial_blocks_per_bucket();
-    if (scanner.get_num_rings() % (num_axial_crystals_per_bucket) != 0)
-      {
-        error(boost::format("Error in GeometryBlocksOnCylindrical: number of rings (%d) is not a multiple of the "
-                            "num_axial_crystals_per_bucket "
-                            "(%d) = num_axial_crystals_per_block (%d) * num_axial_blocks_per_bucket (%d)")
-              % scanner.get_num_rings() % num_axial_crystals_per_bucket % scanner.get_num_axial_crystals_per_block()
-              % scanner.get_num_axial_blocks_per_bucket());
-      }
-  }
-}
 void
 GeometryBlocksOnCylindrical::build_crystal_maps(const Scanner& scanner)
 {
--- a/src/buildblock/GeometryBlocksOnCylindrical.cxx
+++ b/src/buildblock/GeometryBlocksOnCylindrical.cxx
@@ -39,6 +39,7 @@ limitations under the License.
 #include <iostream>
 #include <fstream>
 #include <iomanip>
+#include <boost/format.hpp>
 
 START_NAMESPACE_STIR
 
@@ -63,12 +64,11 @@ GeometryBlocksOnCylindrical::check_scanner_configuration(const Scanner& scanner)
     int num_axial_crystals_per_bucket = scanner.get_num_axial_crystals_per_block() * scanner.get_num_axial_blocks_per_bucket();
     if (scanner.get_num_rings() % (num_axial_crystals_per_bucket) != 0)
       {
-        error("Error in GeometryBlocksOnCylindrical: number of rings (%d) is not a multiple of the num_axial_crystals_per_bucket "
-              "(%d) = num_axial_crystals_per_block (%d) * num_axial_blocks_per_bucket (%d)",
-              scanner.get_num_rings(),
-              num_axial_crystals_per_bucket,
-              scanner.get_num_axial_crystals_per_block(),
-              scanner.get_num_axial_blocks_per_bucket());
+        error(boost::format("Error in GeometryBlocksOnCylindrical: number of rings (%d) is not a multiple of the "
+                            "num_axial_crystals_per_bucket "
+                            "(%d) = num_axial_crystals_per_block (%d) * num_axial_blocks_per_bucket (%d)")
+              % scanner.get_num_rings() % num_axial_crystals_per_bucket % scanner.get_num_axial_crystals_per_block()
+              % scanner.get_num_axial_blocks_per_bucket());
       }
   }
 }
--- a/src/buildblock/GeometryBlocksOnCylindrical.cxx
+++ b/src/buildblock/GeometryBlocksOnCylindrical.cxx
@@ -44,6 +44,7 @@ START_NAMESPACE_STIR
 
 GeometryBlocksOnCylindrical::GeometryBlocksOnCylindrical(const Scanner& scanner)
 {
+  check_scanner_configuration(scanner);
   build_crystal_maps(scanner);
 }
 
@@ -56,16 +57,31 @@ GeometryBlocksOnCylindrical::get_rotation_matrix(float alpha) const
 }
 
 void
+GeometryBlocksOnCylindrical::check_scanner_configuration(const Scanner& scanner)
+{
+  { // Ensure the number of axial crystals per bucket is a multiple of the number of rings
+    int num_axial_crystals_per_bucket = scanner.get_num_axial_crystals_per_block() * scanner.get_num_axial_blocks_per_bucket();
+    if (scanner.get_num_rings() % (num_axial_crystals_per_bucket) != 0)
+      {
+        error("Error in GeometryBlocksOnCylindrical: number of rings (%d) is not a multiple of the num_axial_crystals_per_bucket "
+              "(%d) = num_axial_crystals_per_block (%d) * num_axial_blocks_per_bucket (%d)",
+              scanner.get_num_rings(),
+              num_axial_crystals_per_bucket,
+              scanner.get_num_axial_crystals_per_block(),
+              scanner.get_num_axial_blocks_per_bucket());
+      }
+  }
+}
+void
 GeometryBlocksOnCylindrical::build_crystal_maps(const Scanner& scanner)
 {
   // local variables to describe scanner
   int num_axial_crystals_per_block = scanner.get_num_axial_crystals_per_block();
   int num_transaxial_crystals_per_block = scanner.get_num_transaxial_crystals_per_block();
-  int num_axial_blocks = scanner.get_num_axial_blocks();
   int num_transaxial_blocks_per_bucket = scanner.get_num_transaxial_blocks_per_bucket();
   int num_axial_blocks_per_bucket = scanner.get_num_axial_blocks_per_bucket();
-  int num_transaxial_buckets = scanner.get_num_transaxial_blocks() / num_transaxial_blocks_per_bucket;
-  int num_axial_buckets = scanner.get_num_axial_blocks() / num_axial_blocks_per_bucket;
+  int num_transaxial_buckets = scanner.get_num_transaxial_buckets();
+  int num_axial_buckets = scanner.get_num_axial_buckets();
   int num_detectors_per_ring = scanner.get_num_detectors_per_ring();
   float axial_block_spacing = scanner.get_axial_block_spacing();
   float transaxial_block_spacing = scanner.get_transaxial_block_spacing();
@@ -100,7 +116,7 @@ GeometryBlocksOnCylindrical::build_crystal_maps(const Scanner& scanner)
   stir::CartesianCoordinate3D<float> start_point(start_z, start_y, start_x);
 
   for (int ax_bucket_num = 0; ax_bucket_num < num_axial_buckets; ++ax_bucket_num)
-    for (int ax_block_num = 0; ax_block_num < num_axial_blocks; ++ax_block_num)
+    for (int ax_block_num = 0; ax_block_num < num_axial_blocks_per_bucket; ++ax_block_num)
       for (int ax_crys_num = 0; ax_crys_num < num_axial_crystals_per_block; ++ax_crys_num)
         for (int trans_bucket_num = 0; trans_bucket_num < num_transaxial_buckets; ++trans_bucket_num)
           for (int trans_block_num = 0; trans_block_num < num_transaxial_blocks_per_bucket; ++trans_block_num)
--- a/src/buildblock/Scanner.cxx
+++ b/src/buildblock/Scanner.cxx
@@ -1754,20 +1754,19 @@ Scanner::check_consistency() const
     { //! Check consistency of axial and transaxial spacing for block geometry
       if (get_num_axial_buckets() != 1)
         {
-          warning(boost::format("BlocksOnCylindrical num_axial_buckets (%d) is greater than 1. This is not supported yet."
+          warning(boost::format("BlocksOnCylindrical num_axial_buckets (%d) is greater than 1. This is not supported yet. "
                                 "Consider multiplying the number of axial_blocks_per_bucket by %d.")
                   % get_num_axial_buckets() % get_num_axial_buckets());
           return Succeeded::no;
         }
-      { // Ensure the number of axial crystals per bucket is a multiple of the number of rings
-        // This asserts the number of axial
+      { // Assert that each block contains an equal number of axial crystals
         if (get_num_rings() % get_num_axial_crystals_per_bucket() != 0)
           {
-            error(boost::format("Error in GeometryBlocksOnCylindrical: number of rings (%d) is not a multiple of the "
-                                "get_num_axial_crystals_per_bucket "
-                                "(%d) = num_axial_crystals_per_block (%d) * num_axial_blocks_per_bucket (%d)")
-                  % get_num_rings() % get_num_axial_crystals_per_bucket() % get_num_axial_crystals_per_block()
-                  % get_num_axial_blocks_per_bucket());
+            warning(boost::format("Error in GeometryBlocksOnCylindrical: number of rings (%d) is not a multiple of the "
+                                  "get_num_axial_crystals_per_bucket "
+                                  "(%d) = num_axial_crystals_per_block (%d) * num_axial_blocks_per_bucket (%d)")
+                    % get_num_rings() % get_num_axial_crystals_per_bucket() % get_num_axial_crystals_per_block()
+                    % get_num_axial_blocks_per_bucket());
           }
       }
       if (get_axial_crystal_spacing() * get_num_axial_crystals_per_block() > get_axial_block_spacing())
--- a/src/buildblock/Scanner.cxx
+++ b/src/buildblock/Scanner.cxx
@@ -42,6 +42,7 @@
 #include <iostream>
 #include <algorithm>
 #include <sstream>
+#include <boost/format.hpp>
 #include "stir/warning.h"
 #include "stir/error.h"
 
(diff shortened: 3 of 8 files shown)


robbietuk:
@markus-jehl, the modification to use `num_axial_blocks_per_bucket` as the upper bound of the loop didn't work. There is another issue here.

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -871,6 +873,90 @@ BlocksTests::run_intersection_with_cylinder_test()
...
+  PM->enable_cache(false);
+  auto back_projector = std::make_shared<BackProjectorByBinUsingProjMatrixByBin>(PM);
+  back_projector->set_up(projdata_info_sptr, volume);
+  back_projector->back_project(*volume, *projdata_sptr, 0, 144);
+
+  bool test_ok = true;
+  std::ostringstream oss;
+  oss << "BlocksTests::run_back_projection_test_with_axial_buckets: Central voxel values are:" << std::endl;
+
+  auto center_axial_values = std::vector<float>(volume->get_z_size());
```
KrisThielemans, line 941: STIR uses English spelling, so please use `centre`

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -871,6 +873,90 @@ BlocksTests::run_intersection_with_cylinder_test()
...
+  //                                                                               min_ring_diff_v,
+  //                                                                               max_ring_diff_v,
+  //                                                                               scanner_sptr->get_max_num_views(),
+  //                                                                               scanner_sptr->get_max_num_non_arccorrected_bins());
+
+  auto projdata_sptr = std::make_shared<ProjDataInMemory>(std::make_shared<ExamInfo>(ImagingModality::PT), projdata_info_sptr);
+
+  auto origin = CartesianCoordinate3D<float>(0, 0, 0);
+  auto volume_dimensions = CartesianCoordinate3D<int>(-1, -1, -1);
+  auto volume = std::make_shared<VoxelsOnCartesianGrid<float>>(*projdata_info_sptr, 1, origin, volume_dimensions);
```
KrisThielemans, line 925: I'd prefer to pass the exam_info along, to avoid warnings and future compatibility, so create an `exam_info_sptr` above and use it for both `proj_data_sptr` and `volume_sptr`

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -871,6 +873,90 @@ BlocksTests::run_intersection_with_cylinder_test()
...
+
+  // Now run the test
+  volume->fill(0.0);
+  projdata_sptr->fill(1.0);
+
+  auto PM = std::make_shared<ProjMatrixByBinUsingRayTracing>();
```
KrisThielemans, line 931: I guess use `PM_sptr` and same for `back_projector`

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -871,6 +873,90 @@ BlocksTests::run_intersection_with_cylinder_test()
...
+  PM->enable_cache(false);
+  auto back_projector = std::make_shared<BackProjectorByBinUsingProjMatrixByBin>(PM);
+  back_projector->set_up(projdata_info_sptr, volume);
+  back_projector->back_project(*volume, *projdata_sptr, 0, 144);
```
KrisThielemans, line 935: don't use hard-wired number.

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -53,6 +53,7 @@
 #include "stir/recon_buildblock/BackProjectorByBinUsingProjMatrixByBin.h"
 #include "stir/IO/write_to_file.h"
 #include "stir/VoxelsOnCartesianGrid.h"
+#include "stir/PixelsOnCartesianGrid.h"
```
KrisThielemans, line 56: needed? I'd rather get rid of this clas in the future

src/buildblock/GeometryBlocksOnCylindrical.cxx
```diff
@@ -55,17 +56,32 @@ GeometryBlocksOnCylindrical::get_rotation_matrix(float alpha) const
...
+  { // Ensure the number of axial crystals per bucket is a multiple of the number of rings
+    int num_axial_crystals_per_bucket = scanner.get_num_axial_crystals_per_block() * scanner.get_num_axial_blocks_per_bucket();
+    if (scanner.get_num_rings() % (num_axial_crystals_per_bucket) != 0)
+      {
+        error("Error in GeometryBlocksOnCylindrical: number of rings (%d) is not a multiple of the num_axial_crystals_per_bucket "
+              "(%d) = num_axial_crystals_per_block (%d) * num_axial_blocks_per_bucket (%d)",
+              scanner.get_num_rings(),
+              num_axial_crystals_per_bucket,
+              scanner.get_num_axial_crystals_per_block(),
+              scanner.get_num_axial_blocks_per_bucket());
```
KrisThielemans, line 71: use the `boost::format` trick to avoid the deprecated form
```suggestion
        error(boost::format("Error in GeometryBlocksOnCylindrical: number of rings (%d) is not a multiple of the num_axial_crystals_per_bucket "
              "(%d) = num_axial_crystals_per_block (%d) * num_axial_blocks_per_bucket (%d)") %
              scanner.get_num_rings() %
              num_axial_crystals_per_bucket %
              scanner.get_num_axial_crystals_per_block() %
              scanner.get_num_axial_blocks_per_bucket());
```
(but ATM, apply this by hand and run `clang-format`)

KrisThielemans (changes requested):
house-keeping comments, not about the actual methodology (which does look fine to me)

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
robbietuk: It is needed for `centre_axial_values[z] = volume_sptr->get_plane(z).at(0).at(0);`. Is there another way to achieve this?

KrisThielemans: what about `(*volume_sptr)[z][0][0]`, or can use `at` of course. or `(*volume_sptr)[make_coordinate(z,0,0)]`.

KrisThielemans:
First of all, "buckets" consist of a number of "blocks" which have a number of "crystals". `BlocksOnCylindrical` was developed without thinking about buckets at all (corresponding to "1 block per bucket"), as that's what was needed at the time, hence the bug. Unfortunately, most scanners actually would correspond to "BucketsOnCylindrical". Furthermore, for most scanners it would make more sense to have a larger axial gap between buckets than between blocks. We currently don't have the variables for that.

#776 by @NikEfth tried to address some of this by adding a facility for extra virtual crystals between buckets. Sadly, this get stuck in changing meaning of various variables with a long-ish discussion. I cannot recall how far this got, and it will need revision now since various merges. In any case, that isn't completely general anyway as the extra spacing can only be a crystal.

The whole issue of arbitrary geometries and virtual crystals is a very difficult one. Getting the detector-map sorted is relatively easy, but things like symmetries (ok, we just switch them off), component-based normalisation, and down/upsampling for scatter and even nomenclature are hard. 

While this should be rather high on our priority list,  I suggest we don't try to resolve that in this PR, or discuss it here either. I think for now we should just fix the code such that it does a sensible thing (and document it). Luckily, the fact that the code is currently broken  for `num_axial_blocks_per_bucket > 1` means we are just bug fixing, and don't introduce extra incompatibility! 😄 

Looking at the [current definition](https://github.com/UCL/STIR/blob/ee307c721cae562ab163aee06dbab8407d861414/src/buildblock/GeometryBlocksOnCylindrical.cxx#L111-L129),the `alpha` definition is in terms of buckets. So it will put all blocks in a bucket next/parallel to eachother. That looks good to me.  So, I think that currently `BlocksOnCylindrical` ***does*** support multiple blocks per bucket in transaxial direction. However, clearly in axial direction the bucket location is completely ignored.

On to the proposed solutions. Keep in mind there are 2 different things: how do we get information (`get_num_rings()` is all over the place), and how do we define the scanner. The former is independent of the latter. The latter is where things get quite hard due to existing Interfile headers, and loads of scanner definitions. Changing the Interfile headers would need versions and all that. Not so easy therefore.

> * Add `*_bucket_spacing` variables in `Scanner` and implement usage.
>   
>   * _This is the simplest solution and uses the current methodology_

Yes, this is rather easy. In the first instance, I would suggest we do this initially with "gaps between blocks" = "gaps between buckets". So, I suggest to have only a `get_bucket_spacing()` that computes it as `num_axial_blocks_per_bucket*get_axial_block_spacing()`. i.e this is only an additional the `get`. In the next step, making the gap larger needs extra variables/parsing etc, so it's a bit more work. (not much in implementation, probably loads of work in testing).

> * Disable the use of multiple buckets per "module" and ensure each parallel plane is a single bucket (axially and transaxially).
>   
>   * _This is limiting and may break a few things?_

"module" is not STIR terminology ATM, so I have no idea what you mean @robbietuk 

> * Transition away from the use of `num_rings` to define the geometry and implement entirely from `num_crystals_per_block`, `num_blocks_per_bucket` and `num_buckets_per_(module?)` (in each dimension)
>   
>   * _I personally perfer this method as it is the simplest for the user as they no longer need to precalculate the `num_rings`, STIR does it for them._

This is a `set` change. This is a lot harder as discussed above.

So, I prefer the first option. It's a fairly easy fix,  and the 3rd option can then be done later. (Thinking how we add virtual crystals to all this gives me a headache. They are a pain, but they do enable easy use of symmetries...)

KrisThielemans (commented):
extra comments, also, don't forget to add you as `\author`, and when done `release_6.1`

robbietuk:
I agree with all of the above (https://github.com/UCL/STIR/pull/1375#issuecomment-1935082119). The third option requires reconsidering of "how do we define the scanner" which would require a "change to existing Interfile headers, and loads of scanner definitions". This is uncomfortable and I don't have time.

I agree the first option is the most useful.

> > * Add `*_bucket_spacing` variables in `Scanner` and implement usage.
> >   
> >   * _This is the simplest solution and uses the current methodology_
> 
> Yes, this is rather easy. In the first instance, I would suggest we do this initially with "gaps between blocks" = "gaps between buckets". So, I suggest to have only a `get_bucket_spacing()` that computes it as `num_axial_blocks_per_bucket*get_axial_block_spacing()`. i.e this is only an additional the `get`. In the next step, making the gap larger needs extra variables/parsing etc, so it's a bit more work. (not much in implementation, probably loads of work in testing).

For my own scanner development I need "gaps between blocks" != "gaps between buckets" but I will make this a separate PR. I suggest we hold off implementing "incorrectly" functioning `get_*_bucket_spacing()` methods until then. 

I therefore I have two options.
1. Update the `build_crystal_maps` method to calculate the pitch of multiple axial buckets with the assumption that all the axial blocks have the same pitch/spacing, regardless of bucket index. I could introduce the `get_*_bucket_spacing()` methods here but their definitions would be "incorrect".
2. Disable the use of multiple axial buckets with this geometry via an error. `num_axial_buckets > 1` does not work due to the current state of `build_crystal_maps` and computing `axial_bucket_spacing = num_axial_blocks_per_bucket*get_axial_block_spacing()` is assumptive about the geometry.

Really this is a case of trying to distinguish a line between two PRs, one a bug fix and the other an added feature. I would prefer to proceed with 2. and bug fix the current issue and then create a new PR with `bucket_spacing` interfile variables and more tests. Note, I would also have the disable the test added in this PR because it would fail.

src/buildblock/GeometryBlocksOnCylindrical.cxx
```diff
@@ -39,11 +39,14 @@ limitations under the License.
...
 #include <fstream>
 #include <iomanip>
+#include <boost/format.hpp>
 
 START_NAMESPACE_STIR
 
 GeometryBlocksOnCylindrical::GeometryBlocksOnCylindrical(const Scanner& scanner)
 {
+  if (scanner.check_consistency() == Succeeded::no)
+    error("Error in GeometryBlocksOnCylindrical: scanner configuration not accepted. Please check warnings.");
```
robbietuk, line 49: This is the primary compatibility change I have made. Some previously geometries may stop working. This was the case with the `test_interpolate_projdata.cxx` test and the change made to the down-sampled geometry.

It also assumes `check_consistency()` is correct.

src/buildblock/Scanner.cxx
```diff
@@ -1751,6 +1752,24 @@ Scanner::check_consistency() const
...
     { //! Check consistency of axial and transaxial spacing for block geometry
+      if (get_num_axial_buckets() != 1)
+        {
+          warning(boost::format("BlocksOnCylindrical num_axial_buckets (%d) is greater than 1. This is not supported yet."
+                                "Consider multiplying the number of axial_blocks_per_bucket by %d.")
+                  % get_num_axial_buckets() % get_num_axial_buckets());
+          return Succeeded::no;
+        }
+      { // Ensure the number of axial crystals per bucket is a multiple of the number of rings
+        // This asserts the number of axial
```
robbietuk, line 1763:
```suggestion
      { // Assert that each block contains an equal number of axial crystals
```

src/buildblock/Scanner.cxx
```diff
@@ -1786,22 +1805,22 @@ Scanner::check_consistency() const
...
             }
-          else
-            {
-              std::ifstream crystal_map(get_crystal_map_file_name());
-              if (!crystal_map)
-                {
-                  warning("No correct crystal map provided. Please check the file name.");
-                  return Succeeded::no;
-                }
-            }
```
robbietuk, line 1804: The git change come from fixing the scope on the `else if (get_scanner_geometry() == "Generic")` (moved the new leading `}`). The changes here are caused clang formatting.

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -905,6 +989,8 @@ BlocksTests::run_tests()
   print_time("map orientation test took: ");
   run_intersection_with_cylinder_test();
   print_time("intersection with cylinder test took: ");
+//  run_back_projection_test_with_axial_buckets(back_projector);
+//  print_time("back projection test with axial buckets took: ");
```
robbietuk, line 993: Re-enable this test but use only 1 axial bucket. A future PR to allow for multiple axial buckets.

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -871,6 +872,89 @@ BlocksTests::run_intersection_with_cylinder_test()
...
+                                                                   scanner_sptr->get_max_num_views(),
+                                                                   scanner_sptr->get_max_num_non_arccorrected_bins());
+  //  scanner_sptr->set_scanner_geometry("Cylindrical");
+  //  scanner_sptr->set_up();
+  //  auto projdata_info_sptr = std::make_shared<ProjDataInfoCylindricalNoArcCorr>(scanner_sptr,
+  //                                                                               num_axial_pos_per_segment,
+  //                                                                               min_ring_diff_v,
+  //                                                                               max_ring_diff_v,
+  //                                                                               scanner_sptr->get_max_num_views(),
+  //                                                                               scanner_sptr->get_max_num_non_arccorrected_bins());
```
robbietuk, line 918: Remove this now.

src/recon_test/test_blocks_on_cylindrical_projectors.cxx
```diff
@@ -871,6 +872,89 @@ BlocksTests::run_intersection_with_cylinder_test()
...
+    {
+      centre_axial_values[z] = (*volume_sptr)[z][0][0];
+      oss << "\tz = " << z << "/" << volume_sptr->get_max_z() << " is " << centre_axial_values[z] << std::endl;
+      if (test_ok)
+        {
+          test_ok = check(centre_axial_values[z] > 0,
+                          "BlocksTests::run_back_projection_test_with_axial_buckets: Central voxel value <= 0");
+          everything_ok = everything_ok && test_ok;
+        }
+    }
```
robbietuk, line 950: What other tests can be done to validate the geometries are acceptable? 

Could consider detector coordinate z components monotonically increase. This would require access to a sorted list of them. May require the splitting of `build_crystal_maps` into components (not a bad idea anyway).

robbietuk (commented):
Some comments on the latest changes. Running CI to ensure everything still works with `GeometryBlocksOnCylindrical` constructor changes

src/buildblock/GeometryBlocksOnCylindrical.cxx
KrisThielemans: This is fine for me.

KrisThielemans: as long as we put it in the release notes anyway!

src/buildblock/Scanner.cxx
```diff
@@ -1751,6 +1752,23 @@ Scanner::check_consistency() const
...
+        {
+          warning(boost::format("BlocksOnCylindrical num_axial_buckets (%d) is greater than 1. This is not supported yet. "
+                                "Consider multiplying the number of axial_blocks_per_bucket by %d.")
+                  % get_num_axial_buckets() % get_num_axial_buckets());
+          return Succeeded::no;
+        }
+      { // Assert that each block contains an equal number of axial crystals
+        if (get_num_rings() % get_num_axial_crystals_per_bucket() != 0)
+          {
+            error(boost::format("Error in GeometryBlocksOnCylindrical: number of rings (%d) is not a multiple of the "
```
KrisThielemans, line 1765: Looks like all the other are warnings with a return, while this calls error. I think it's best to stick to warnings here, mainly such that `test_Scanner` can proceed with other scanners (otherwise we'd need `try/catch` statements there).

src/include/stir/GeometryBlocksOnCylindrical.h
```diff
@@ -1,6 +1,7 @@
 
 /*
 Copyright 2017 ETH Zurich, Institute of Particle Physics and Astrophysics
+ Copyright (C) 2024, Prescient Imaging
```
KrisThielemans, line 4: no changes yet?

KrisThielemans (changes requested):

KrisThielemans:
> Really this is a case of trying to distinguish a line between two PRs, one a bug fix and the other an added feature. I would prefer to proceed with 2. and bug fix the current issue and then create a new PR with `bucket_spacing` interfile variables and more tests. 

This is of course up to you. I guess in the second PR, you can still do it in stages (i.e. first commit(s), assume bucket_gaps==block_gap and see if tests work).

> Note, I would also have the disable the test added in this PR because it would fail.

Which one? It seems a bit to introduce test code now that will fail soon. Or maybe you mean "disable the check on num_axial_blocks_per_bucket"? That'd of course be fine.

KrisThielemans:
I'm a bit confused by the clang-format situation. Did you run it through? I guess so, as the `pre-commit` action is fine, but I guess I'm mildly surprised at the output.

robbietuk:
> Which one? It seems a bit to introduce test code now that will fail soon. Or maybe you mean "disable the check on num_axial_blocks_per_bucket"? That'd of course be fine.

Ignore this, I modified the tests to use `num_axial_buckets=1` and allowed the tests (all pass).

> I'm a bit confused by the clang-format situation. Did you run it through? I guess so, as the pre-commit action is fine, but I guess I'm mildly surprised at the output.

I am running clang-format on save. I don't see where the problem is...

KrisThielemans:
> I am running clang-format on save. 

great

> I don't see where the problem is...

no problem. I'm just surprised by some of the formatting choices it makes, but anyway.

robbietuk:
@KrisThielemans [b758eaf](https://github.com/UCL/STIR/pull/1375/commits/b758eaf6fe19ee0f8c5ba0e4c1813f89af41bf76) CLI passed. Everything since is documentation or simple changes. I believe this is ready.

KrisThielemans:
@markus-jehl can you have a look please?

documentation/release_6.1.htm
```diff
@@ -84,8 +84,10 @@ <H2>What's new for developers (aside from what should be obvious
 
 <h3>Backward incompatibities</h3>
 <ul>
-<li>
-</li>
+    <li>
+        This required additional checks on <code>GeometryBlocksOnCylindrical</code> and
```
markus-jehl, line 88: Maybe just: "Additional checks on ... which may lead to a new error ..."

robbietuk:
> Looks good to me, apart from the release note comment that seems to refer to something that isn't there (anymore?).

The release notes comment refers to the inclusion of `Scanner::check_consistency()` in the constructor of `GeometryBlocksOnCylindrical` that may result in an error if an invalid geometry.
https://github.com/UCL/STIR/pull/1375/files#diff-cdce2d1c9cb44f5a4ab3ef4228bd84366641b8a6e8fb077b459d1e6e68bf27aaR48-R49

documentation/release_6.1.htm
```diff
@@ -84,8 +84,10 @@ <H2>What's new for developers (aside from what should be obvious
 
 <h3>Backward incompatibities</h3>
 <ul>
-<li>
-</li>
+    <li>
+        "Additional checks on <code>GeometryBlocksOnCylindrical</code> scanner configuration, which may lead to new
+        errors"
```
KrisThielemans, line 89:
```suggestion
        Additional checks on <code>GeometryBlocksOnCylindrical</code> scanner configuration, which may lead to an
        error being raised, while previously the code silently proceeded.
```

KrisThielemans (changes requested):

KrisThielemans:
thanks! Squash-merge ok?

robbietuk:
Yes. Remember there is another bug fix in there with ` else if (get_scanner_geometry() == "Generic")`

KrisThielemans:
ok. well, you could update the release notes, and/or clean-up the commits, or just let me know if you prefer squash-merge or normal-merge...

robbietuk:
I combine the past five commits into one. They were mostly documentation changes. I suggest a regular merge.

KrisThielemans:
Thanks! Will merge after #1386 as #1383 broke the gcc12 job as it runs out of disk space (first noted'/fixed on #1236, but that PR will have to wait till 6.2)

KrisThielemans (approved):

PR #1391, opened 2024-02-17
fix plot_scanner_LOR for new SWIG etc

KrisThielemans:
- use `Scanner.get_scanner_from_name` (static function convention changed in SWIG 4)
- Also plot a more interesting scanner and fix the output name for the png file.


robbietuk (approved):
Great!

PR #1432, opened 2024-05-16
small rationalising of warnings

KrisThielemans:
- remove warnings related to block variables when parsing a cylindrical scanner
- some explanation in warnings
- convert warnings on disabled symmetries to info()

### DIFF
--- a/src/IO/InterfileHeader.cxx
+++ b/src/IO/InterfileHeader.cxx
@@ -1032,7 +1032,7 @@ InterfilePDFSHeader::post_processing()
         {
           warning("Inconsistent number of TOF bins (" + std::to_string(this->num_timing_poss)
                   + ") and size of the 'TOF bin order' list (" + std::to_string(this->timing_poss_sequence.size()) + ").");
-          // return true;
+          return true;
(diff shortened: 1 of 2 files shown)


src/IO/InterfileHeader.cxx
```diff
@@ -1032,7 +1032,7 @@ InterfilePDFSHeader::post_processing()
         {
           warning("Inconsistent number of TOF bins (" + std::to_string(this->num_timing_poss)
                   + ") and size of the 'TOF bin order' list (" + std::to_string(this->timing_poss_sequence.size()) + ").");
-          // return true;
+          return true;
```
KrisThielemans, line 1035: @NicoleJurjew I've made this into a failure. (Seems weird to try and go ahead of the sizes are incorrect.). Let me know ASAP if this is not ok.

NicoleJurjew: Hey Kris, I agree, that wouldn't make sense. With the new script we're not manually editing the TOF bin order anymore, so this won't trip us up for Vision files.

PR #1430, opened 2024-05-15
Modifications for psmr2024

NikEfth:
## Checklist before requesting a review

  - [] I have performed a self-review of my code
  - [] I have added docstrings/doxygen in line with the guidance in the developer guide
  - [] I have implemented unit tests that cover any new or modified functionality (if applicable)
  - [] The code builds and runs on my machine
  - [] `documentation/release_XXX.md` has been updated with any functionality change (if applicable)

PR #1434, opened 2024-05-17
TOF mashing = TOF bins fixes

KrisThielemans:
@NikEfth As per my review. I need these changes to all tests to work. In addition, the code should still do what you want it to.

One particular change that I need is in `ProjDataInfo::get_tof_bin` where we can no longer check on `is_tof_data`, as when that is `false`, it goes into the next part of the code, but that fails as the relevant arrays are not set-up.

Sorry for the white-space changes, but this is due to precommit. If you're running Ubuntu 22.04, consider installing `clang-format` and `pre-commit` as https://github.com/UCL/STIR/blob/master/documentation/devel/git-hooks.md


KrisThielemans:
Obsolete as all fixed in #1430

Issue #1259, opened 2023-09-17
std::copy of VectorWithOffset<float> does not fall back to memmove


KrisThielemans:
for 2 `VectorWithOffset`s
```c++
std::copy(v.begin(), v.end(), v2.begin());
```
is rather slow (checked with gcc 11)

As `VectorWithOffset<float>` is essentially a wrapper around `float *`, you'd hope that it uses `memmove`, as opposed to a loop. Checking the STL code for gcc, this is decided based on (an internal version) of `is_trivially_copyable `, which turns out to be false.
This has probably consequences for lots of operations on float/int vectors/arrays.

Following the template trail is a bit hard, but it seems `is_trivially_copyable<VectorWithOffset_iter<float>>` is false. The type is defined as
https://github.com/UCL/STIR/blob/0307f8d70fc02430b95d44ae648f4bcfa1d5ee43/src/include/stir/VectorWithOffset.h#L39-L46
when I wrote this, my C++ was a lot better than it is now, and  I thought this would automatically work. Here's the [`boost::iterator_adaptor`  doc](https://www.boost.org/doc/libs/1_83_0/libs/iterator/doc/html/iterator/generic/adaptor.html)

KrisThielemans:
This is still a problem.
```c++
#include <type_traits>
#include "stir/VectorWithOffset."
int main()
{  
  static_assert(std::is_trivially_copyable_v<std::vector<float>> == true); // works ok
  static_assert(std::is_trivially_copyable_v<VectorWithOffset<float>> == true); // fails
  return 0;
}
```

KrisThielemans:
This will affect lots of places. @evgueni-ovtchinnikov could you try to replace
https://github.com/UCL/STIR/blob/a008e73a36fc4dedffe4170c698c5ed18c5972a8/src/include/stir/VectorWithOffset.inl#L517
with
```c++
    for (int i = this->get_min_index(); i <= this->get_max_index(); i++)
      this->num[i] = il.num[i];
```
or normally faster as it should fall back to `memmove`
```c++
   std::copy(il.num + il.get_min_index(), il.num + il.get_max_index() + 1, this->num + this->get_min_index());
```
Obviously, that's a non-ideal work-around, but it'd allow to check the impact.

evgueni-ovtchinnikov:
Both suggested loops take about the same time as std::copy(il.begin(), il.end(), this->begin()); (the second being a bit slower).

KrisThielemans:
[...]

PR #1438, opened 2024-05-28
Use contiguous Arrays more

KrisThielemans:
- add `copy_to`/`fill_from` specialisations for `Array` that check if they are contigous. This enhances the likelihood that the compiler will recognise that it can call `memove` as opposed to doing a loop. (related to #1259 )
- use that in `ProjDataInMemory`

### DIFF
--- a/src/buildblock/ProjDataInMemory.cxx
+++ b/src/buildblock/ProjDataInMemory.cxx
@@ -24,6 +24,7 @@
 */
 
 #include "stir/ProjDataInMemory.h"
+#include "stir/copy_fill.h"
 #include "stir/shared_ptr.h"
 #include "stir/Succeeded.h"
 #include "stir/SegmentByView.h"
@@ -94,7 +95,7 @@ copy_data_from_buffer(const Array<1, float>& buffer, Array<num_dimensions, float
 #endif
   {
     const float* ptr = buffer.get_const_data_ptr() + offset;
-    std::copy(ptr, ptr + array.size_all(), array.begin_all());
+    fill_from(array, ptr, ptr + array.size_all());
     buffer.release_const_data_ptr();
   }
 }
@@ -108,7 +109,7 @@ copy_data_to_buffer(Array<1, float>& buffer, const Array<num_dimensions, float>&
 #endif
   {
     float* ptr = buffer.get_data_ptr() + offset;
-    std::copy(array.begin_all(), array.end_all(), ptr);
+    copy_to(array, ptr);
     buffer.release_data_ptr();
   }
 }
--- a/src/buildblock/ProjDataInMemory.cxx
+++ b/src/buildblock/ProjDataInMemory.cxx
@@ -14,7 +14,7 @@
     Copyright (C) 2002 - 2011-02-23, Hammersmith Imanet Ltd
     Copyright (C) 2011, Kris Thielemans
     Copyright (C) 2016, University of Hull
-    Copyright (C) 2016, 2019, 2020, 2023, UCL
+    Copyright (C) 2016, 2019, 2020, 2023, 2024, UCL
     Copyright (C) 2020,  Rutherford Appleton Laboratory STFC
     This file is part of STIR.
 
@@ -117,30 +117,11 @@ copy_data_to_buffer(Array<1, float>& buffer, const Array<num_dimensions, float>&
 Viewgram<float>
 ProjDataInMemory::get_viewgram(const int view_num,
                                const int segment_num,
-                               const bool make_num_tangential_poss_odd
-#ifdef STIR_TOF
-                               ,
-                               const int timing_pos
-#endif
-) const
+                               const bool make_num_tangential_poss_odd,
+                               const int timing_pos) const
 {
-  Viewgram<float> viewgram(proj_data_info_sptr,
-                           view_num,
-                           segment_num
-#ifdef STIR_TOF
-                           ,
-                           timing_pos
-#endif
-  );
-  Bin bin(segment_num,
-          view_num,
-          this->get_min_axial_pos_num(segment_num),
-          this->get_min_tangential_pos_num()
-#ifdef STIR_TOF
-              ,
-          timing_pos
-#endif
-  );
+  Bin bin(segment_num, view_num, this->get_min_axial_pos_num(segment_num), this->get_min_tangential_pos_num(), timing_pos);
+  Viewgram<float> viewgram(proj_data_info_sptr, bin);
 
   for (bin.axial_pos_num() = get_min_axial_pos_num(segment_num); bin.axial_pos_num() <= get_max_axial_pos_num(segment_num);
        bin.axial_pos_num()++)
@@ -176,19 +157,9 @@ ProjDataInMemory::set_viewgram(const Viewgram<float>& v)
     }
   const int segment_num = v.get_segment_num();
   const int view_num = v.get_view_num();
-#ifdef STIR_TOF
   const int timing_pos = v.get_timing_pos_num();
-#endif
 
-  Bin bin(segment_num,
-          view_num,
-          this->get_min_axial_pos_num(segment_num),
-          this->get_min_tangential_pos_num()
-#ifdef STIR_TOF
-              ,
-          timing_pos
-#endif
-  );
+  Bin bin(segment_num, view_num, this->get_min_axial_pos_num(segment_num), this->get_min_tangential_pos_num(), timing_pos);
 
   for (bin.axial_pos_num() = get_min_axial_pos_num(segment_num); bin.axial_pos_num() <= get_max_axial_pos_num(segment_num);
        bin.axial_pos_num()++)
@@ -208,10 +179,8 @@ ProjDataInMemory::get_index(const Bin& this_bin) const
   if (!(this_bin.axial_pos_num() >= get_min_axial_pos_num(this_bin.segment_num())
         && this_bin.axial_pos_num() <= get_max_axial_pos_num(this_bin.segment_num())))
     error("ProjDataInMemory::this->get_index: axial_pos_num out of range : %d", this_bin.axial_pos_num());
-#ifdef STIR_TOF
   if (!(this_bin.timing_pos_num() >= get_min_tof_pos_num() && this_bin.timing_pos_num() <= get_max_tof_pos_num()))
     error("ProjDataInMemory::this->get_index: timing_pos_num out of range : %d", this_bin.timing_pos_num());
-#endif
 
   const int index = static_cast<int>(std::find(segment_sequence.begin(), segment_sequence.end(), this_bin.segment_num())
                                      - segment_sequence.begin());
@@ -225,7 +194,6 @@ ProjDataInMemory::get_index(const Bin& this_bin) const
 
   // Now we are just in front of  the correct segment
   {
-#ifdef STIR_TOF
     if (proj_data_info_sptr->get_num_tof_poss() > 1)
       {
         const int timing_index
@@ -235,7 +203,6 @@ ProjDataInMemory::get_index(const Bin& this_bin) const
         assert(offset_3d_data > 0);
         segment_offset += static_cast<streamoff>(timing_index) * offset_3d_data;
       }
-#endif
     // skip axial positions
     const streamoff ax_pos_offset = (this_bin.axial_pos_num() - get_min_axial_pos_num(this_bin.segment_num())) * get_num_views()
                                     * get_num_tangential_poss();
@@ -255,30 +222,11 @@ ProjDataInMemory::get_index(const Bin& this_bin) const
 Sinogram<float>
 ProjDataInMemory::get_sinogram(const int ax_pos_num,
                                const int segment_num,
-                               const bool make_num_tangential_poss_odd
-#ifdef STIR_TOF
-                               ,
-                               const int timing_pos
-#endif
-) const
+                               const bool make_num_tangential_poss_odd,
+                               const int timing_pos) const
 {
-  Sinogram<float> sinogram(proj_data_info_sptr,
-                           ax_pos_num,
-                           segment_num
-#ifdef STIR_TOF
-                           ,
-                           timing_pos
-#endif
-  );
-  Bin bin(segment_num,
-          this->get_min_view_num(),
-          ax_pos_num,
-          this->get_min_tangential_pos_num()
-#ifdef STIR_TOF
-              ,
-          timing_pos
-#endif
-  );
+  Sinogram<float> sinogram(proj_data_info_sptr, ax_pos_num, segment_num, timing_pos);
+  Bin bin(segment_num, this->get_min_view_num(), ax_pos_num, this->get_min_tangential_pos_num(), timing_pos);
 
   detail::copy_data_from_buffer(this->buffer, sinogram, this->get_index(bin));
 
@@ -307,66 +255,31 @@ ProjDataInMemory::set_sinogram(const Sinogram<float>& s)
     }
   int segment_num = s.get_segment_num();
   int ax_pos_num = s.get_axial_pos_num();
-#ifdef STIR_TOF
   int timing_pos = s.get_timing_pos_num();
-#endif
-  Bin bin(segment_num,
-          this->get_min_view_num(),
-          ax_pos_num,
-          this->get_min_tangential_pos_num()
-#ifdef STIR_TOF
-              ,
-          timing_pos
-#endif
-  );
+  Bin bin(segment_num, this->get_min_view_num(), ax_pos_num, this->get_min_tangential_pos_num(), timing_pos);
 
   detail::copy_data_to_buffer(this->buffer, s, this->get_index(bin));
   return Succeeded::yes;
 }
 
 SegmentBySinogram<float>
-ProjDataInMemory::get_segment_by_sinogram(const int segment_num
-#ifdef STIR_TOF
-                                          ,
-                                          const int timing_pos_num
-#endif
-) const
+ProjDataInMemory::get_segment_by_sinogram(const int segment_num, const int timing_pos_num) const
 {
-  SegmentBySinogram<float> segment(proj_data_info_sptr,
-                                   segment_num
-#ifdef STIR_TOF
-                                   ,
-                                   timing_pos_num
-#endif
-  );
   const Bin bin(segment_num,
                 this->get_min_view_num(),
                 this->get_min_axial_pos_num(segment_num),
-                this->get_min_tangential_pos_num()
-#ifdef STIR_TOF
-                    ,
-                timing_pos_num
-#endif
-  );
+                this->get_min_tangential_pos_num(),
+                timing_pos_num);
+  SegmentBySinogram<float> segment(proj_data_info_sptr, bin);
   detail::copy_data_from_buffer(this->buffer, segment, this->get_index(bin));
   return segment;
 }
 
 SegmentByView<float>
-ProjDataInMemory::get_segment_by_view(const int segment_num
-#ifdef STIR_TOF
-                                      ,
-                                      const int timing_pos
-#endif
-) const
+ProjDataInMemory::get_segment_by_view(const int segment_num, const int timing_pos) const
 {
   // TODO rewrite in terms of get_sinogram as this doubles memory temporarily
-  return SegmentByView<float>(get_segment_by_sinogram(segment_num
-#ifdef STIR_TOF
-                                                      ,
-                                                      timing_pos
-#endif
-                                                      ));
+  return SegmentByView<float>(get_segment_by_sinogram(segment_num, timing_pos));
 }
 
 Succeeded
@@ -387,12 +300,8 @@ ProjDataInMemory::set_segment(const SegmentBySinogram<float>& segmentbysinogram_
   const Bin bin(segment_num,
                 this->get_min_view_num(),
                 this->get_min_axial_pos_num(segment_num),
-                this->get_min_tangential_pos_num()
-#ifdef STIR_TOF
-                    ,
-                segmentbysinogram_v.get_timing_pos_num()
-#endif
-  );
+                this->get_min_tangential_pos_num(),
+                segmentbysinogram_v.get_timing_pos_num());
 
   detail::copy_data_to_buffer(this->buffer, segmentbysinogram_v, this->get_index(bin));
   return Succeeded::yes;
--- a/src/include/stir/Array.h
+++ b/src/include/stir/Array.h
@@ -229,13 +229,15 @@ public:
   //! return minimum of all the elements
   inline elemT find_min() const;
 
-  //! Fill elements with value n (overrides VectorWithOffset::fill)
+  //! Fill elements with value \c n
+  /*!
+    hides VectorWithOffset::fill
+   */
   inline void fill(const elemT& n);
-
-  //! Sets elements below value to the value (overrides VectorWithOffset::fill)
+  //! Sets elements below value to the value
   inline void apply_lower_threshold(const elemT& l);
 
-  //! Sets elements above value to the value (overrides VectorWithOffset::fill)
+  //! Sets elements above value to the value
   inline void apply_upper_threshold(const elemT& u);
 
   //! checks if the index range is 'regular'
--- a/src/include/stir/copy_fill.h
+++ b/src/include/stir/copy_fill.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (C) 2020 University College London
+    Copyright (C) 2020, 2024 University College London
     This file is part of STIR.
 
     SPDX-License-Identifier: Apache-2.0
@@ -16,6 +16,7 @@
   \author Kris Thielemans
 */
 
+#include "stir/Array.h"
 #include "stir/ProjDataInMemory.h"
 
 START_NAMESPACE_STIR
@@ -42,6 +43,49 @@ struct CopyFill
   static void fill_from(T& stir_object, iterT iter, iterT iter_end)
   {
     std::copy(iter, iter_end, stir_object.begin_all());
+  }
+};
+
+//! Helper class for stir::copy_to and stir::fill_from
+/*! Specialisation that uses underlying 1D iterators for contiguous Array objects
+ */
+template <int num_dimensions, class elemT>
+struct CopyFill<Array<num_dimensions, elemT>>
+{
+  typedef Array<num_dimensions, elemT> Array_type;
+
+  template <typename iterT>
+  static iterT copy_to(const Array_type& stir_array, iterT iter)
+  {
+    if (stir_array.is_contiguous())
+      {
+        // std::cerr << "Using 1D std::copy for copy_to\n";
+        auto beg = stir_array.get_const_full_data_ptr();
+        auto ret = std::copy(beg, beg + stir_array.size_all(), iter);
+        stir_array.release_const_full_data_ptr();
+        return ret;
+      }
+    else
+      {
+        // std::cerr<<"Using normal std::copy for copy_to\n";
+        return std::copy(stir_array.begin_all(), stir_array.end_all(), iter);
+      }
+  }
+
+  template <typename iterT>
+  static void fill_from(Array_type& stir_array, iterT iter, iterT iter_end)
+  {
+    if (stir_array.is_contiguous())
+      {
+        // std::cerr << "Using 1D std::copy for fill_from\n";
+        std::copy(iter, iter_end, stir_array.get_full_data_ptr());
+        stir_array.release_full_data_ptr();
+      }
+    else
+      {
+        // std::cerr<<"Using normal std::copy for fill_from\n";
+        std::copy(iter, iter_end, stir_array.begin_all());
+      }
   }
 };
 
--- a/src/test/test_Array.cxx
+++ b/src/test/test_Array.cxx
@@ -43,7 +43,9 @@
 
 #include "stir/ArrayFunction.h"
 #include "stir/array_index_functions.h"
+#include "stir/copy_fill.h"
 #include <functional>
+#include <algorithm>
 
 // for open_read/write_binary
 #include "stir/utilities.h"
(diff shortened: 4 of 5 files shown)


KrisThielemans:
@markus-jehl I'm still hunting for an elusive speed-up. Maybe this one does help a bit...

markus-jehl:
This merge broke something in the reconstruction. Projections still work, but when I do the initial reconstruction it comes out full of NaN‘s in the FoV. Haven‘t had time to look into the changes yet, but maybe this is also what impacts you on the psmr branch (since I also got a test failure on the MacOS build on my fork after pulling this merge).

KrisThielemans:
ah. sorry.

This PR is quite innocent really. Essentially the only change is in `ProjDataInMemory::copy_data_from_buffer` (and the `to` version), where it uses a new specialisation of `fill_from` and `copy_to` for `Array`. That checks that if it's a contiguous array, we'll fall back to a `std::copy` using `float *` as opposed to full iterators, as the former should be faster (any good compiler will call `memmove` or similar).

There are some other formatting changes as I got rid of `#ifdef STIR_TOF`, which gives a lot of "noise" due to reformatting sadly.

I've added tests for all this, but I suppose none of our tests do a recon with `ProjDataInMemory` though.

KrisThielemans:
@markus-jehl I cannot find a problem, nor reproduce it. Are you sure it is this PR?

Suggested debugging steps:

1. disable the `CopyFill<Array<>>` specialisation (just comment out [this](https://github.com/UCL/STIR/blob/master/src/include/stir/copy_fill.h#L52-L90))
2. if step 1 works, then re-enable and replace the `if (stir_array.is_contiguous())` statements in those lines with `if(false)`.
3. if step 1 didn't work, revert the `ProjDataInMemory::copy_data_from_buffer` (and `to`) to use `std::copy` directly (really shouldn't do anything as `CopyFill` should then fall-back to the same `std::copy` anyway)
4. if step 3 didn't work, then revert `ProjDataInMemory` completely (it'd have to be something in the `STIR_TOF` edits I guess)

thanks for your help!

markus-jehl:
I'm fairly certain it is this one: I went back to each PR in my fork and this is the one that first displayed the issue.

Will go through your suggested debug steps today!

markus-jehl:
The problem is the first `if(stir_array.is_contiguous())`. I noticed that we're not actually checking if some other process has already obtained full access: https://github.com/UCL/STIR/blob/31e730eeb4dd5d2fb49c3d2f404ed7bc02e45115/src/include/stir/Array.inl#L283

So it's probably running into some race condition. The only thread-safe way to handling this would be mutexes and locks, but this is likely detrimental to performance. Do you have an idea where there could be multiple processes trying to access the same array? Then this could be solved logically rather than by brute force.

KrisThielemans:
> The problem is the first `if(stir_array.is_contiguous())`. 

which one is the first? L60?

> I noticed that we're not actually checking if some other process has already obtained full access

that's true.

> What is the purpose of _full_pointer_access, actually? It seems to be never checked and since it's only used to read data, not modify them it shouldn't be required.

For `get_data_ptr`, we do check if access is on/off
https://github.com/UCL/STIR/blob/31e730eeb4dd5d2fb49c3d2f404ed7bc02e45115/src/include/stir/VectorWithOffset.inl#L611-L613

I put those in as a safety check really. `std::vector::data()` doesn't do that. It just says that the "behaviour is undefined" if you modify the underlying vector (e.g. by a `resize()`, which would affect read access as well). I felt it was safer to check, at the expense of more elaborate code (so maybe it wasn't a great decision).

> Still, I don't understand why `std::copy(beg, beg + stir_array.size_all(), iter);` gives issues while `std::copy(stir_array.begin_all(), stir_array.end_all(), iter);` is fine...

I have no clue.

Note that this whole stuff is not thread-safe: if multiple processes modify an `Array` at the same time, stuff is going to happen. That' s not documented, but should be rather obvious. So, I don't think we do that.

I'm not so sure how to proceed. Can you find out where the NaNs are created? Is the sensitivity image ok for instance?

Can you please create an issue, just referring to this discussion here?

markus-jehl:
Yes, L60 is the problematic one. I'm currently running with as much debug output as possible to narrow down where it happens.

Sure, will create an issue.

markus-jehl:
This is the subsensitivity image, so the NaN's come from there:
image

KrisThielemans:
ok. Narrowing it down. Relevant lines are
https://github.com/UCL/STIR/blob/31e730eeb4dd5d2fb49c3d2f404ed7bc02e45115/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L830-L832
https://github.com/UCL/STIR/blob/31e730eeb4dd5d2fb49c3d2f404ed7bc02e45115/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L871

By the way, do your `ctest` and `recon_test_pack` work?

markus-jehl:
All tests with ctest pass, but recon_test_pack I first need to look up how to run. Will also look into the distributable sensitivity computation, since that already sounds like it might be a good candidate :-)

KrisThielemans:
> All tests with ctest pass

interesting, as that is using OSMAPOSL etc as well

`recon_test_pack`, see e.g.: https://github.com/conda-forge/stir-feedstock/blob/6d4fd1aaf32bfe034bbe7440988cedac3f2d12da/recipe/meta.yaml#L66-L79

the `distributable_computation` is the hardest part in STIR (sits in distributable.cxx). Essentially, it is a multi-threaded loop where it's going to call a callback. That makes me think you could try and set threads=1, or compile without openmp.

thanks for debugging!

markus-jehl:
Interestingly it looks exactly the same in repeat runs and also suing just 1 thread. So my suspicion is that the `is_contiguous` function sometimes gives false positives. But I'm still looking for the flaw in the logic...

KrisThielemans:
it certainly is easier if it's an ordinary bug as opposed to a race-condition! I guess I'd recommend compiling without OpenMP then and running `valgrind`.

Other (more desperate) thing to try: use a more conservative approach for `is_contiguous`:
- add a `bool Array::_is_contiguous` member
- set it to `true` in `Array::init()`
- set it to false in the 2 constructors that don't call `init`, i.e. [here](https://github.com/UCL/STIR/blob/31e730eeb4dd5d2fb49c3d2f404ed7bc02e45115/src/include/stir/Array.inl#L115) and [here](https://github.com/UCL/STIR/blob/31e730eeb4dd5d2fb49c3d2f404ed7bc02e45115/src/include/stir/Array.inl#L125), and in `resize()`
- let `is_contiguous` return `_is_contiguous`.

Vital to run `ctest` after that!

If that works, we could try to check consistency between the old and new `is_contiguous` (although the new one is overly pessimistic at the moment).

markus-jehl:
I was actually thinking about this pessimistic approach! Because `is_contiguous` is called many times, I was worried if it might actually be computationally heavy in its own right. The member variable would solve this problem. Will give it a try.

I agree, running valgrind or a sanitiser should highlight where the issue is. For this I first need to create a minimal working example of the issue (currently I just test it in python with an iterative reconstruction that would take forever in serial with a sanitiser attached to it...). The recon_test_pack passes as well, btw.

markus-jehl:
Just an update: I've now ported the test across to C++ and I get this weird looking sensitivity map even when disabling the new copy. So the more I dig the less I understand... Will update you when I have more clarity.

KrisThielemans:
Ok. Any chance you can do this on vanilla master as opposed to your branch?

markus-jehl:
Yes, I'm currently doing it on the official master. It looks like in C++ some additional problems are caused by reading ProjDataInMemory from file and using them - as if some of that memory is either not persistent for long enough or gives issues with copying.

KrisThielemans:
By the way, if you have a small test example, potentially we could add it as a test, as it's escaping all other tests somehow. Can also do a `git bisect` I guess. thanks, and sorry for the trouble.

markus-jehl:
The address sanitiser actually already throws an error in the first line of my test, where I'm reading in ProjDataInMemory as such:
`auto inputProjData = std::make_shared<stir::ProjDataInMemory>(*(stir::ProjDataInMemory::read_from_file("filename.hs")));`

This is the ASAN output:
```
=================================================================
==1174870==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6170000301d0 at pc 0x55eb98f589ad bp 0x7fff7c0a8730 sp 0x7fff7c0a7ef8
READ of size 92160 at 0x6170000301d0 thread T0
    #0 0x55eb98f589ac in __interceptor_memmove (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x1839ac) (BuildId: 8775ca2bc7777fd8de9cf6411d87eded039071d8)
    #1 0x55eb9951fff8 in float* stir::CopyFill<stir::Array<3, float> >::copy_to<float*>(stir::Array<3, float> const&, float*) (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x74aff8) (BuildId: 8775ca2bc7777fd8de9cf6411d87eded039071d8)
    #2 0x55eb9951b93d in stir::ProjDataInMemory::set_segment(stir::SegmentBySinogram<float> const&) (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x74693d) (BuildId: 8775ca2bc7777fd8de9cf6411d87eded039071d8)
    #3 0x55eb9951b9d1 in stir::ProjDataInMemory::set_segment(stir::SegmentByView<float> const&) (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x7469d1) (BuildId: 8775ca2bc7777fd8de9cf6411d87eded039071d8)
    #4 0x55eb994e2093 in stir::ProjData::fill(stir::ProjData const&) (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x70d093) (BuildId: 8775ca2bc7777fd8de9cf6411d87eded039071d8)
    #5 0x55eb9951dfa3 in stir::ProjDataInMemory::ProjDataInMemory(stir::ProjData const&) (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x748fa3) (BuildId: 8775ca2bc7777fd8de9cf6411d87eded039071d8)
    #6 0x55eb991c7e94 in decltype(new ((void*)(0))stir::ProjDataInMemory(std::declval<stir::ProjData&>())) std::construct_at<stir::ProjDataInMemory, stir::ProjData&>(stir::ProjDataInMemory*, stir::ProjData&) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_construct.h:97:39
    #7 0x55eb991c7e94 in void std::allocator_traits<std::allocator<stir::ProjDataInMemory> >::construct<stir::ProjDataInMemory, stir::ProjData&>(std::allocator<stir::ProjDataInMemory>&, stir::ProjDataInMemory*, stir::ProjData&) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/alloc_traits.h:518:4
    #8 0x55eb991c7e94 in std::_Sp_counted_ptr_inplace<stir::ProjDataInMemory, std::allocator<stir::ProjDataInMemory>, (__gnu_cxx::_Lock_policy)2>::_Sp_counted_ptr_inplace<stir::ProjData&>(std::allocator<stir::ProjDataInMemory>, stir::ProjData&) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr_base.h:519:4
    #9 0x55eb991c7e94 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::__shared_count<stir::ProjDataInMemory, std::allocator<stir::ProjDataInMemory>, stir::ProjData&>(stir::ProjDataInMemory*&, std::_Sp_alloc_shared_tag<std::allocator<stir::ProjDataInMemory> >, stir::ProjData&) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr_base.h:651:6
    #10 0x55eb991c7e94 in std::__shared_ptr<stir::ProjDataInMemory, (__gnu_cxx::_Lock_policy)2>::__shared_ptr<std::allocator<stir::ProjDataInMemory>, stir::ProjData&>(std::_Sp_alloc_shared_tag<std::allocator<stir::ProjDataInMemory> >, stir::ProjData&) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr_base.h:1342:14
    #11 0x55eb991c7e94 in std::shared_ptr<stir::ProjDataInMemory>::shared_ptr<std::allocator<stir::ProjDataInMemory>, stir::ProjData&>(std::_Sp_alloc_shared_tag<std::allocator<stir::ProjDataInMemory> >, stir::ProjData&) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:409:4
    #12 0x55eb991c7e94 in std::shared_ptr<stir::ProjDataInMemory> std::allocate_shared<stir::ProjDataInMemory, std::allocator<stir::ProjDataInMemory>, stir::ProjData&>(std::allocator<stir::ProjDataInMemory> const&, stir::ProjData&) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:862:14
    #13 0x55eb991c7e94 in std::shared_ptr<stir::ProjDataInMemory> std::make_shared<stir::ProjDataInMemory, stir::ProjData&>(stir::ProjData&) /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:878:14
    #14 0x55eb991c7e94 in TestClientReconComponent::TestArrayBug() /workspace/neurolf/reconstruction/test/unit/ReconstructionComponentTests.cpp:492:24
    #15 0x55eb9920907e in ____C_A_T_C_H____T_E_S_T____18() /workspace/neurolf/reconstruction/test/unit/ReconstructionComponentTests.cpp:987:24
    #16 0x55eb9902b759 in Catch::TestInvokerAsFunction::invoke() const /workspace/neurolf/reconstruction/libs/install/include/catch2/catch.hpp:14317:9
    #17 0x55eb99096d36 in Catch::TestCase::invoke() const /workspace/neurolf/reconstruction/libs/install/include/catch2/catch.hpp:14156:15
    #18 0x55eb99096d36 in Catch::RunContext::invokeActiveTestCase() /workspace/neurolf/reconstruction/libs/install/include/catch2/catch.hpp:13016:27
    #19 0x55eb990943a9 in Catch::RunContext::runCurrentTest(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) /workspace/neurolf/reconstruction/libs/install/include/catch2/catch.hpp:12989:17
    #20 0x55eb9909129f in Catch::RunContext::runTest(Catch::TestCase const&) /workspace/neurolf/reconstruction/libs/install/include/catch2/catch.hpp:12750:13
    #21 0x55eb9909f32f in Catch::(anonymous namespace)::TestGroup::execute() /workspace/neurolf/reconstruction/libs/install/include/catch2/catch.hpp:13343:45
    #22 0x55eb9909f32f in Catch::Session::runInternal() /workspace/neurolf/reconstruction/libs/install/include/catch2/catch.hpp:13549:39
    #23 0x55eb9909c4eb in Catch::Session::run() /workspace/neurolf/reconstruction/libs/install/include/catch2/catch.hpp:13505:24
    #24 0x55eb990d56d2 in int Catch::Session::run<char>(int, char const* const*) /workspace/neurolf/reconstruction/libs/install/include/catch2/catch.hpp:13227:30
    #25 0x55eb990d56d2 in main /workspace/neurolf/reconstruction/libs/install/include/catch2/catch.hpp:17504:29
    #26 0x7f7c09d04d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #27 0x7f7c09d04e3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #28 0x55eb98f3eb64 in _start (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x169b64) (BuildId: 8775ca2bc7777fd8de9cf6411d87eded039071d8)

0x6170000301d0 is located 0 bytes to the right of 720-byte region [0x61700002ff00,0x6170000301d0)
allocated by thread T0 here:
    #0 0x55eb98ffc88d in operator new[](unsigned long) (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x22788d) (BuildId: 8775ca2bc7777fd8de9cf6411d87eded039071d8)
    #1 0x55eb99212e3a in stir::VectorWithOffset<float>::reserve(int, int) /workspace/neurolf/reconstruction/libs/install/include/STIR-6.2/stir/VectorWithOffset.inl:395:45
    #2 0x55eb99337963 in stir::detail::VectorWithOffset_iter<stir::Array<1, float> > std::__copy_move<false, false, std::random_access_iterator_tag>::__copy_m<stir::detail::VectorWithOffset_iter<stir::Array<1, float> const>, stir::detail::VectorWithOffset_iter<stir::Array<1, float> > >(stir::detail::VectorWithOffset_iter<stir::Array<1, float> const>, stir::detail::VectorWithOffset_iter<stir::Array<1, float> const>, stir::detail::VectorWithOffset_iter<stir::Array<1, float> >) (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x562963) (BuildId: 8775ca2bc7777fd8de9cf6411d87eded039071d8)

SUMMARY: AddressSanitizer: heap-buffer-overflow (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x1839ac) (BuildId: 8775ca2bc7777fd8de9cf6411d87eded039071d8) in __interceptor_memmove
Shadow bytes around the buggy address:
  0x0c2e7fffdfe0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2e7fffdff0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2e7fffe000: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2e7fffe010: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2e7fffe020: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x0c2e7fffe030: 00 00 00 00 00 00 00 00 00 00[fa]fa fa fa fa fa
  0x0c2e7fffe040: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c2e7fffe050: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2e7fffe060: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2e7fffe070: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c2e7fffe080: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==1174870==ABORTING
```

markus-jehl:
Ok, progress! Managed to debug into STIR to see where it's failing. All that is needed is to handle the case where an array has only one index. Currently it is not entering the for loop and therefore not checking whether the sub-array is contiguous (plus the last sub-array was never checked). Changing the for loop from `i < this->get_max_index()` to `i <= this->get_max_index()` and handling the one case where we index `i+1` fixes this:
```
template <int num_dimensions, typename elemT>
bool
Array<num_dimensions, elemT>::is_contiguous() const
{
  auto mem = &(*this->begin_all());
  for (auto i = this->get_min_index(); i <= this->get_max_index(); ++i)
    {
      if (!(*this)[i].is_contiguous())
        return false;
      mem += (*this)[i].size_all();
      if (i < this->get_max_index() && mem != &(*(*this)[i + 1].begin_all()))
        return false;
    }
  return true;
}
```

KrisThielemans:
great! That was a bug. Can you submit a PR?

KrisThielemans:
I suggest
```C++
     if (i == this->get_max_index() 
       return true;
     mem += (*this)[i].size_all();
      if (mem != &(*(*this)[i + 1].begin_all()))
        return false;
```
and remove the final `return false`. Might be slightly easier to understand the logic.

KrisThielemans:
By the way, while
```c++
auto inputProjData = std::make_shared<stir::ProjDataInMemory>(*(stir::ProjDataInMemory::read_from_file("filename.hs")));
```
is fine, it might be clearer to use
```c++
std::shared<stir::ProjDataInMemory> inputProjData(stir::ProjDataInMemory::read_from_file("filename.hs"));
```

markus-jehl:
Yes, I'll issue a PR. But first I also want to understand the memory leak that still exists.

KrisThielemans:
Ok. Doesn't TSAN tell you where it came from? Valgrind could.

markus-jehl:
This is all it says:
```
==1342194==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1584 byte(s) in 3 object(s) allocated from:
    #0 0x55da4942f77d in operator new(unsigned long) (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x22777d) (BuildId: ee84e9178c2086875508efe40ec877644497b58a)
    #1 0x55da49917699 in stir::ProjData::read_from_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::_Ios_Openmode) (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x70f699) (BuildId: ee84e9178c2086875508efe40ec877644497b58a)

Indirect leak of 24576 byte(s) in 3 object(s) allocated from:
    #0 0x55da4942f88d in operator new[](unsigned long) (/workspace/neurolf/reconstruction/build/test/nlf.reconstruction.tests.asan+0x22788d) (BuildId: ee84e9178c2086875508efe40ec877644497b58a)
    #1 0x7f8958273023 in std::basic_filebuf<char, std::char_traits<char> >::_M_allocate_internal_buffer() (/lib/x86_64-linux-gnu/libstdc++.so.6+0x115023) (BuildId: e37fe1a879783838de78cbc8c80621fa685d58a2)

SUMMARY: AddressSanitizer: 26160 byte(s) leaked in 6 allocation(s).
```
I suspect it is when the object goes out of scope at the end of the test that the memory is not correctly freed. Has something changed there recently?

KrisThielemans:
Seems unrelated to the rest. Nothing changed there.

It's not so useful that it doesn't say what object. Clearly, it cannot be the `ProjDataFromMemory` data itself, as the leak  would then be a lot larger.

Just for sanity, please try the above constructor instead.

markus-jehl:
I actually did (must have also tried before and then given up), but it gives me the following error:
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:150:7: note: candidate constructor not viable: no known conversion from 'shared_ptr<stir::ProjData>' to 'const shared_ptr<stir::ProjDataInMemory>' for 1st argument
      shared_ptr(const shared_ptr&) noexcept = default; ///< Copy constructor
      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:304:7: note: candidate constructor not viable: no known conversion from 'shared_ptr<stir::ProjData>' to 'shared_ptr<stir::ProjDataInMemory>' for 1st argument
      shared_ptr(shared_ptr&& __r) noexcept
      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:357:17: note: candidate constructor not viable: no known conversion from 'shared_ptr<stir::ProjData>' to 'std::nullptr_t' for 1st argument
      constexpr shared_ptr(nullptr_t) noexcept : shared_ptr() { }
                ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:160:2: note: candidate template ignored: could not match '_Yp *' against 'shared_ptr<stir::ProjData>'
        shared_ptr(_Yp* __p) : __shared_ptr<_Tp>(__p) { }
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:296:2: note: candidate template ignored: requirement 'is_constructible<std::__shared_ptr<stir::ProjDataInMemory, __gnu_cxx::_S_atomic>, const std::shared_ptr<stir::ProjData> &>::value' was not satisfied [with _Yp = stir::ProjData]
        shared_ptr(const shared_ptr<_Yp>& __r) noexcept
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:313:2: note: candidate template ignored: requirement 'is_constructible<std::__shared_ptr<stir::ProjDataInMemory, __gnu_cxx::_S_atomic>, std::shared_ptr<stir::ProjData>>::value' was not satisfied [with _Yp = stir::ProjData]
        shared_ptr(shared_ptr<_Yp>&& __r) noexcept
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:325:11: note: candidate template ignored: could not match 'weak_ptr' against 'shared_ptr'
        explicit shared_ptr(const weak_ptr<_Yp>& __r)
                 ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:332:2: note: candidate template ignored: could not match 'auto_ptr' against 'shared_ptr'
        shared_ptr(auto_ptr<_Yp>&& __r);
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:340:2: note: candidate template ignored: could not match 'unique_ptr' against 'shared_ptr'
        shared_ptr(unique_ptr<_Yp, _Del>&& __r)
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:408:2: note: candidate template ignored: could not match '_Sp_alloc_shared_tag' against 'shared_ptr'
        shared_ptr(_Sp_alloc_shared_tag<_Alloc> __tag, _Args&&... __args)
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:148:17: note: candidate constructor not viable: requires 0 arguments, but 1 was provided
      constexpr shared_ptr() noexcept : __shared_ptr<_Tp>() { }
                ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:177:2: note: candidate constructor template not viable: requires 2 arguments, but 1 was provided
        shared_ptr(_Yp* __p, _Deleter __d)
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:194:2: note: candidate constructor template not viable: requires 2 arguments, but 1 was provided
        shared_ptr(nullptr_t __p, _Deleter __d)
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:257:2: note: candidate constructor template not viable: requires 2 arguments, but 1 was provided
        shared_ptr(const shared_ptr<_Yp>& __r, element_type* __p) noexcept
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:284:2: note: candidate constructor template not viable: requires 2 arguments, but 1 was provided
        shared_ptr(shared_ptr<_Yp>&& __r, element_type* __p) noexcept
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:417:7: note: candidate constructor not viable: requires 2 arguments, but 1 was provided
      shared_ptr(const weak_ptr<_Tp>& __r, std::nothrow_t) noexcept
      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:214:2: note: candidate constructor template not viable: requires 3 arguments, but 1 was provided
        shared_ptr(_Yp* __p, _Deleter __d, _Alloc __a)
        ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/shared_ptr.h:233:2: note: candidate constructor template not viable: requires 3 arguments, but 1 was provided
        shared_ptr(nullptr_t __p, _Deleter __d, _Alloc __a)
```
It's as if it can't convert pointers to convertable objects, only the objects themselves. With my old approach I initialise a new pointer with the converted object, so it worked.

My suspicion is that the leak might be in the interfile parsing somewhere, but I haven't yet found an obvious allocation somewhere that might cause this.

markus-jehl:
Update:
When I use `std::shared_ptr<stir::ProjDataInMemory> inputProjData(std::dynamic_pointer_cast<stir::ProjDataInMemory>(stir::ProjDataInMemory::read_from_file("filename.hs")));` then it compiles, but gives a segmentatoin fault later when I call sapyb on inputProjData to subtract randoms from it. Is this because the underlying object is still some file stream that is closed prematurely?

KrisThielemans:
> It's as if it can't convert pointers to convertable objects, only the objects themselves. With my old approach I initialise a new pointer with the converted object, so it worked.

oops, sorry. it needs a `std::move`
```
std::shared_ptr<stir::ProjDataInMemory> inputProjData(std::move(stir::ProjDataInMemory::read_from_file("filename.hs"));
```

The `cast` will fail, as the object disappears. Not "prematurely" in a C++ sense, as the `unique_ptr` gets destructed immediately, so the object is gone.

KrisThielemans:
> My suspicion is that the leak might be in the interfile parsing somewhere, but I haven't yet found an obvious allocation somewhere that might cause this.

oh right. There might even be an issue on that. I'll keep ignoring that for a bit longer. sorry...

markus-jehl:
> > It's as if it can't convert pointers to convertable objects, only the objects themselves. With my old approach I initialise a new pointer with the converted object, so it worked.
> 
> oops, sorry. it needs a `std::move`
> 
> ```
> std::shared_ptr<stir::ProjDataInMemory> inputProjData(std::move(stir::ProjDataInMemory::read_from_file("filename.hs"));
> ```
> 
> The `cast` will fail, as the object disappears. Not "prematurely" in a C++ sense, as the `unique_ptr` gets destructed immediately, so the object is gone.

Hmm... The move still gives me the same error: `note: candidate constructor not viable: no known conversion from 'shared_ptr<stir::ProjData>' to 'const shared_ptr<stir::ProjDataInMemory>' for 1st argument`

KrisThielemans:
ok. Of course, `read_from_file` doesn't return a `ProjDataInMemory` object. ignore me!

markus-jehl:
> > My suspicion is that the leak might be in the interfile parsing somewhere, but I haven't yet found an obvious allocation somewhere that might cause this.
> 
> oh right. There might even be an issue on that. I'll keep ignoring that for a bit longer. sorry...

I couldn't find an issue related to this. Should I create one?

KrisThielemans:
Neither can I. Please do so. Thanks a lot!

PR #1449, opened 2024-06-09
More openMP: find_min/max() and sum()

KrisThielemans:
Originally in #1430, but as that needs more work, creating a separate PR. Sadly, in most cases (except LAFOV), the functions that I parallelised were pretty fast already.

There's more work to do `+=` etc, as it might be a bad idea to parallelise this for `VectorWithOffset`, as then the nD-Array functions could create a lot of threads. I guess it needs specialisation for `Array<1,elemT>`.

Will likely be rebased on top of #1442 once that is merged, although I could merge as well as it's independent.

@markus-jehl I haven't parallelised `is_contiguous`, as that's also a bit more work, see e.g.
https://stackoverflow.com/questions/9793791/parallel-openmp-loop-with-break-statement. Alternatively, we introduce an extra member (if it turns out to be worth it, as that'd occur a lot in current nD Arrays).

### DIFF
--- a/documentation/release_6.2.htm
+++ b/documentation/release_6.2.htm
@@ -100,7 +100,7 @@ from the above):</H2>
     but non-TOF instead. This did not happen in our normal reconstruction code, and would have thrown an error
     if it occured.
     <br>
-    Fixed in <a href=https://github.com/UCL/STIR/pull/1427>issue #1427</a>.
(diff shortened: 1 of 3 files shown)


KrisThielemans:
Same linking problem for `clang` as in https://github.com/UCL/STIR/pull/1430#issuecomment-2143564633

markus-jehl:
all good!

PR #1458, opened 2024-06-14
Auto use_tofsens

### LINKED ISSUE #1456: automatically set use-time-of-flight-sensitivities in Poisson...ProjData
At present, we default `use_tofsens=false`, as this is the best for most scanners. However, for Siemens Vision 600 (and others), the e7tools norm is TOF. At present, this will result in a run-time error
https://github.com/UCL/STIR/blob/4cb2a5d885fea5d2aa4d2f8c53ce8e14561df117/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L561-L564

We could try and detect this though, e.g. insert some lines like this [here](https://github.com/UCL/STIR/blob/4cb2a5d885fea5d2aa4d2f8c53ce8e14561df117/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L640)
```c++
if (auto norm_pd_ptr = dynamic_cast<BinNormalisationFromProjData const *>(this->normalisation_sptr.get()))
{
  if (!this->use_tofsens && norm_pd_ptr->get_norm_proj_data_sptr()->get_num_timing_poss() > 1)
   {
      info("Detected TOF normalisation data, so using time-of-flight sensitivities");
      this->use_tofsens = true;
    }
}
```
@NicoleJurjew what do you think?

KrisThielemans:
Fixes #1456

@NicoleJurjew could you try this PR? I've attempted to generalise it a bit.

### DIFF
--- a/documentation/release_6.2.htm
+++ b/documentation/release_6.2.htm
@@ -66,7 +66,7 @@
     This should mean that the <tt>use time-of-flight sensitivities</tt> keyword should now
     rarely (if ever) be necessary.
     <br>
-    <a href=https://github.com/UCL/STIR/pull/1456>PR #1456</a>
+    <a href=https://github.com/UCL/STIR/pull/1458>PR #1458</a>
     </li>
 </ul>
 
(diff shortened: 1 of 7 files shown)


NicoleJurjew:
Hi Kris, I checked and these changes work for my use-case!

Thank you!

Issue #1280, opened 2023-10-26
Siemens mMR single scatter has unexpected scale


KrisThielemans:
As observed in https://github.com/UCL/STIR/issues/1163, the scatter scale factors for the NEMA phantom had to be around .2, but normally STIR single scatter is scaled to have the correct absolute values. #1279 fixed the thresholds for the scale factor, such that the example now works, but it is still unclear why the scatter scales are so much lower than expected.

KrisThielemans:
This is still a problem for SIRF as it uses the default settings for the scatter estimator
https://github.com/UCL/STIR/blob/bdf04aca863ac6cffdfd635efdb7ec3db16533b3/src/scatter_buildblock/ScatterEstimation.cxx#L88-L89

markus-jehl:
@KrisThielemans : can this be closed?

KrisThielemans:
Sadly not. We still don't know why the single scatter estimate is wrongly scaled for the mMR.

KrisThielemans:
This was indeed fixed by #1531. For the NEMA IQ data used in the examples, I now get the following scale factors (before filtering) when disabling all thresholding:
![Image](https://github.com/user-attachments/assets/8c2e2db6-56c0-4ef4-a07e-0392fda400f5)
Corresponding axial profile through the reconstruction image (averaged over a central square of size 40x40 voxels):
![Image](https://github.com/user-attachments/assets/71138c1c-c565-455a-9765-d86a58adc7c5)
where the bump around 70 is due to the spheres.

Note that there is no bed in the current mu-map.

Plots generated with
```python
import numpy
import matplotlib.pyplot as plt
import os
os.system("sed -n '/upsample_and_fit_scatter_estimate.*before/{;n;p;}' output/scatter.log|tr -d '{}' > scalefactors_before_threshold.csv")

b=numpy.loadtxt('scalefactors_before_threshold.csv',delimiter=',')
plt.figure()
plt.plot(b.transpose());
plt.legend(['1','2','3','4','5'])

import stir
im=stir.FloatVoxelsOnCartesianGrid.read_from_file('final_activity_image_42.hv')
import stirextra
arr=stirextra.to_numpy(im)
p=numpy.mean(arr[:,120:160,120:160],axis=(1,2))
plt.figure()
plt.plot(p)
plt.show()

```

PR #1463, opened 2024-06-23
lower default min-scale-factor for scatter tail-fitting

KrisThielemans:
Set default to 0.05 (was .4). This prevents problems with the mMR (and others?).

Addresses https://github.com/UCL/STIR/issues/1280

@markus-jehl does this affect you?


markus-jehl:
should be ok!

PR #1464, opened 2024-06-24
Allow TOF mashing in SSRB

NicoleJurjew:

  - [x] `documentation/release_XXX.md` has been updated with any functionality change (if applicable)


NicoleJurjew:
Checks have been performed by mashing uniform 1 sinograms:

image

KrisThielemans:
I had to undo the normalisation for "num_mashed_tof_bins". The projectors are set-up such that the total number of counts remains the same independent of number of TOF bins/width etc (i.e. the same as non-TOF). Therefore, when mashing TOF bins and setting `do_norm=1`, we do **not** need to divide by the the number of contributing TOF bins.

I've now added a test to check all this via reconstruction.

This should be good to go. Let's see how the tests are doing.

PR #1486, opened 2024-07-31
Prevent `std::normal_distribution` being called with `sigma=0.0`

### LINKED ISSUE #1485: Debug assert error caused by normal distribution sigma equals 0.0 - MSVC 14.40
Building a cmake project ontop of STIR with CMake (because of #1483). 

I am using the example https://github.com/UCL/STIR/blob/master/examples/C%2B%2B/using_installed_STIR/demo_create_image.cxx but building in Debug.

Modifying the program to:

``` cpp
#include <iostream>
#include "stir/Scanner.h"
using namespace std;

int main()
{
	cout << "Here are all the scanner names:\n" << stir::Scanner::list_all_names() << endl;
	return 0;
}
```

I get an error in `C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.40.33807\include\random`, specifically l:3371 
```
        void _Init(_Ty _Mean0, _Ty _Sigma0) noexcept { // set internal state
            _STL_ASSERT(0.0 < _Sigma0, "invalid sigma argument for normal_distribution");
            _Mean  = _Mean0;
            _Sigma = _Sigma0;
        }
```
Where the `_STL_ASSERT` is throwing an exception because `0.0 == _Sigma0`. This stems from 
```cpp
distributions[0] = std::normal_distribution<double>(0.0, sigma);
``` 
in 
https://github.com/UCL/STIR/blob/8c57bc93894e9245b6c9b90e4ce755f958d656ac/src/include/stir/DetectorCoordinateMap.h#L122-L136

My guess is this hasn't been caught because CI only tests Release on MSVC. This is the only issue I have found so far.

______

<h2>What is the fix?</h2>

I am not sure what the logic behind the use of `sigma=0.0` to define `distributions[0]` as a normal distribution with `mean=0` and `sigma=0` is not a well-defined probability distribution. It essentially collapses to a single point at 0. ~Likely, the plan was to set `distributions[0]=0`. Does this seem correct?~

robbietuk:

Add a guard to `DetectorCoordinateMap` to prevent `std::normal_distribution<double>(0.0, 0.0)` being called.

## Testing performed

Built STIR in debug on windows machine and build project using STIR in debug. Fixes assert `0.0 < sigma` issue observed in #1485.

## Related issues

Fixes #1485

### DIFF
--- a/src/include/stir/DetectorCoordinateMap.h
+++ b/src/include/stir/DetectorCoordinateMap.h
@@ -123,6 +123,8 @@ protected:
   explicit DetectorCoordinateMap(double sigma = 0.0)
       : sigma(sigma)
   {
+    if (sigma == 0.0)
+      return;
 #  ifdef STIR_OPENMP
     generators.resize(omp_get_max_threads());
     distributions.resize(omp_get_max_threads());


KrisThielemans (commented):
I cannot check from here if the actual calls check first if sigma is zero. 

@markus-jehl can you please review?

robbietuk:
The [`get_coordinate_for_det_pos()` method](https://github.com/UCL/STIR/blob/8c57bc93894e9245b6c9b90e4ce755f958d656ac/src/include/stir/DetectorCoordinateMap.h#L92-L108) does but this assert error/fix is in the constructor.

markus-jehl:
This looks good to me! Makes no sense creating the distributions and generators if they are not needed.

Issue #1200, opened 2023-05-06
listmode recon with caching fails on GitHub Actions MacOS with clang 16


KrisThielemans:
Images are reconstructed as zero. Example log file
[OSMAPOSL_test_lm_frame_2.log](https://github.com/UCL/STIR/files/11413135/OSMAPOSL_test_lm_frame_2.log)
from https://github.com/UCL/STIR/actions/runs/4902794721/jobs/8754848390

As I have no access to a Mac, I cannot investigate this.

KrisThielemans:
we no longer see failures, so presumably #1566 fixed this problem. I will close this.

PR #1489, opened 2024-08-19
GHA MacOS jobs tweaks

KrisThielemans:
- Use latest (?) gcc
- Re-enable clang test, to see if https://github.com/UCL/STIR/issues/1200 is resolved with clang 18


KrisThielemans:
Reverting the clang test due to #1491 and #1490

PR #1495, opened 2024-08-23
Add support for get_axial_length for cylindrical scanners.

### LINKED ISSUE #1493: Scanner get_axial_length() gives negative values with Cylindrical geometries
Relevant code

https://github.com/UCL/STIR/blob/e03fbd03976abdd543724aa2e1c2315fbd71e115/src/include/stir/Scanner.inl#L298-L303

The issue appears to be with some of the old preconfigured scanner geometries. 
This snippet:
```
auto scanner = Scanner::get_scanner_from_name("ECAT EXACT");
std::cout << scanner.get_axial_length() << std::endl;
```
will log `-9` to the console. The problem is with `get_axial_block_spacing()` and `get_axial_crystal_spacing()` returning `-1` for these cylindrical scanners.

I guess these are in the `functions to get block geometry info` section:
https://github.com/UCL/STIR/blob/e03fbd03976abdd543724aa2e1c2315fbd71e115/src/include/stir/Scanner.h#L414-L431
but these tools would be useful for cylindrical geometries too.

_____

The solution could be to add an if statement, based upon scanner geometry, to return the current implementation for Blocks on Cylindrical or a new implementation for cylindrical.

robbietuk:

Adds support for cylindrical scanners to call `get_axial_length()` and return a sensible value.

I believe
```
get_num_rings() * get_ring_spacing();
```
is the correct implementation over
```
get_ring_spacing() * (get_num_rings() - 1);
```
because a scanner with 1 ring will have an axial length of that ring. That is, unless axial bin size should be incorporated into the calculation.

## Related issues

Fixes #1493

### DIFF
--- a/src/include/stir/Scanner.inl
+++ b/src/include/stir/Scanner.inl
@@ -298,6 +298,10 @@ Scanner::get_axial_block_spacing() const
 float
 Scanner::get_axial_length() const
 {
+  if (get_scanner_geometry() == "Cylindrical")
+  {
+    return get_num_rings() * get_ring_spacing();
+  }
   return get_num_axial_buckets() * get_num_axial_blocks_per_bucket() * get_axial_block_spacing()
          - (get_axial_block_spacing() - (get_num_axial_crystals_per_block() - 1) * get_axial_crystal_spacing());
 }
--- a/src/include/stir/Scanner.inl
+++ b/src/include/stir/Scanner.inl
@@ -299,9 +299,13 @@ float
 Scanner::get_axial_length() const
 {
   if (get_scanner_geometry() == "Cylindrical")
-  {
-    return get_num_rings() * get_ring_spacing();
-  }
+    {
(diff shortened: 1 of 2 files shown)


src/include/stir/Scanner.inl
```diff
@@ -298,6 +298,10 @@ Scanner::get_axial_block_spacing() const
 float
 Scanner::get_axial_length() const
 {
+  if (get_scanner_geometry() == "Cylindrical")
+  {
+    return get_num_rings() * get_ring_spacing();
+  }
```
KrisThielemans, line 304:
```suggestion
  if (get_scanner_geometry() == "Cylindrical")
    {
      return get_num_rings() * get_ring_spacing();
    }
  else if (get_scanner_geometry() == "Generic")
    {
      error("get_axial_length not yet implemented for Generic scanner geometry");
    }
```
fixing white space as well as safe-guard for Generic

PR #1496, opened 2024-08-25
fix OSX clang problems

### LINKED ISSUE #1490: Clang 18 on MacOS tests fail due to uncaught (but intentional) exceptions
We have a few tests that check if errors are thrown, e.g.
https://github.com/UCL/STIR/blob/209cbb634b841300886d431a37d4181188f20743/src/test/test_DateTime.cxx#L105-L114
However,  with clang18 on MacOS these tests all fail
https://github.com/UCL/STIR/actions/runs/10450627407/job/28935385772?pr=1489#step:12:751
```
Testing DICOM DateTime to epoch functionality

The next test should throw an error

ERROR: Time_Zone info '+020' does not fit DICOM standard
libc++abi: terminating due to uncaught exception of type std::runtime_error: 
ERROR: Time_Zone info '+020' does not fit DICOM standard
```
This is pretty weird, at `catch (...)` should catch all exceptions AFAIK, so this seems like a clang bug to me.

However, we know what we're throwing
https://github.com/UCL/STIR/blob/209cbb634b841300886d431a37d4181188f20743/src/buildblock/error.cxx#L62
So we could just as well (or even better) catch `std::runtime_error`.

KrisThielemans:
fixes #1490

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -93,15 +93,15 @@ jobs:
           BUILD_TYPE: "Debug"
           ROOT: "OFF"
           ITK: "OFF"
-        # Currently disabled due to problems with listmode recon, see https://github.com/UCL/STIR/issues/1200
-        #- os: macOS-latest
-        #  compiler: clang
-        #  compiler_version: 16
-        #  BUILD_FLAGS: "-DSTIR_OPENMP=OFF"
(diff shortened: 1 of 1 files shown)


KrisThielemans:
This approach didn't work. We fixed the issue in #1490 by adapting clang flags.

PR #1499, opened 2024-08-29
MLEstimateComponentBasedNormalisation refactor

### LINKED ISSUE #1498: `ML_estimate_component_based_normalisation` and `apply_norm_factors3D` cannot be used without saving to disk
These are executable files for computing normalization and a brief description of how they are implmented:
- [`find_ML_normfactors3D`](https://github.com/UCL/STIR/blob/master/src/utilities/find_ML_normfactors3D.cxx): calls [`ML_estimate_component_based_normalisation`](https://github.com/UCL/STIR/blob/master/src/recon_buildblock/ML_estimate_component_based_normalisation.cxx) and the single function there. Additionally, the documentation in [`ML_estimate_component_based_normalisation.h`](https://github.com/UCL/STIR/blob/master/src/include/stir/recon_buildblock/ML_estimate_component_based_normalisation.h) is lacking.
- [`find_ML_normfactors`](https://github.com/UCL/STIR/blob/master/src/utilities/find_ML_normfactors.cxx): implements much of the functionality in the main function.
- [`apply_normfactors3D`](https://github.com/UCL/STIR/blob/master/src/utilities/apply_normfactors3D.cxx): implements things in the main function, at least is very short.
- [`apply_normfactors`](https://github.com/UCL/STIR/blob/master/src/utilities/apply_normfactors.cxx): does the same as `apply_normfactors3D` _(code duplication?)_

Most of the following pertains to `find_ML_normfactors3D` and `find_ML_normfactors`. The above functionality saves the computed data to be saved to disk, rather than returning/utilizing a shared_ptr.

__________

The core functionality of these executables should be split into various components.
1. The main function should handle only the arguments handling and saving data.
1. Mostly, functions/classes that take the core arguments and compute/process the data.
1. Can handle saving of the data, if specified by the arguments.
1. Function/ object returns a pointer to the computed data, that can be saved by the main exe.

_____

To implement this:
- The exe's need to maintain backwards compatibility.
- Replace the `ML_estimate_component_based_normalisation` function with a class/function.. 
- Allow member variables/argument to determine if data is saved to disk.
- Store/return computed data as members that can be saved/handled by the exe main functions.

robbietuk:

WIP

Refactor of `ML_estimate_component_based_normalisation`.
- Created a class.
- Simplified the system logic
- Add the ability option to save the data or not

## Remaining tasks

- [x] Add methods to `get_efficiencies()`, `get_norm_geo_data()` and `get_norm_block_data()`
- [x] Add methods to get and set the boolean values of `do_geo`, `do_block`, `do_symmetry_per_block`, `do_KL`, `do_display`, `do_save_to_file`
- [x] Improve documentation defining the `efficiencies`, `norm_geo_data`, `norm_block_data` variables (even after all this work I am still slightly confused).
- [x] Add unit tests

## Testing performed

Added a new unit test for basic functionality.

## Related issues

Fixes #1498


robbietuk:
Thus far, I have not (intentionally) changed any of the functionality of the `find_ML_normfactors3D` utility. I have only used refactoring to move the logic into a class, split functionality into component methods, created members.

The only new introduction is `do_save_to_file`.

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
```diff
@@ -84,225 +127,274 @@ ML_estimate_component_based_normalisation(const std::string& out_filename_prefix
...
+                                                  num_physical_transaxial_crystals_per_basic_unit / 2,
+                                                  num_physical_rings,
+                                                  num_physical_detectors_per_ring); // inputes have to be modified
+
+  measured_block_data = BlockData3D(num_axial_blocks, num_transaxial_blocks, num_axial_blocks - 1, num_transaxial_blocks - 1);
+  norm_block_data_ptr
+      = std::make_shared<BlockData3D>(num_axial_blocks, num_transaxial_blocks, num_axial_blocks - 1, num_transaxial_blocks - 1);
+
+  make_fan_data_remove_gaps(model_fan_data, *model_projdata_sptr);
+  make_fan_data_remove_gaps(measured_fan_data, *measured_projdata_sptr);
```
robbietuk, line 148: My cpp-foo might not be up to snuff here. The target of the shared_ptr is being passed by reference into the function. Is this okay?

robbietuk (commented):
A few comments. I am not 100% sure about updating to using `shared_ptr`s but the idea is that the data can be accessed without copying.

Additionally, I am not convinced there is any testing of this code.

src/buildblock/ML_norm.cxx
```diff
@@ -1141,23 +1145,19 @@ make_fan_data_remove_gaps(FanProjData& fan_data, const ProjData& proj_data)
...
-  if (proj_data.get_proj_data_info_sptr()->is_tof_data())
-    error("make_fan_data: Incompatible with TOF data. Abort.");
-
   const ProjDataInfo& proj_data_info = *proj_data.get_proj_data_info_sptr();
   get_fan_info(num_rings, num_detectors_per_ring, max_delta, fan_size, proj_data_info);
 
   if (proj_data.get_proj_data_info_sptr()->get_scanner_ptr()->get_scanner_geometry() == "Cylindrical")
     {
-      auto proj_data_info_ptr = dynamic_cast<const ProjDataInfoCylindricalNoArcCorr* const>(&proj_data_info);
+      const auto proj_data_info_ptr = dynamic_cast<const ProjDataInfoCylindricalNoArcCorr* const>(&proj_data_info);
```
KrisThielemans, line 1153: `const` not needed, as already in the `dynamic_cast`. It doesn't do any harm of course.

src/include/stir/recon_buildblock/ML_estimate_component_based_normalisation.h
```diff
@@ -14,25 +15,171 @@
...
+        \param do_symmetry_per_block Whether to perform block normalization calculations with symmetry
+        \param do_KL Whether to perform Kullback-Leibler divergence calculations and display the KL value. This can be slow.
+        \param do_display Whether to display the progress of the algorithm.
+        \param do_save_to_file Whether to save the each iteration of the efficiencies, geo data and block data to file.
 */
 void ML_estimate_component_based_normalisation(const std::string& out_filename_prefix,
-                                               const ProjData& measured_data,
-                                               const ProjData& model_data,
+                                               const std::shared_ptr<ProjData>& measured_data,
+                                               const std::shared_ptr<ProjData>& model_data,
```
KrisThielemans, line 40: if we use `shared_ptr`, please rename to `measured_data_sptr`. A question though is if we need to (i.e. does the object need access to these after the constructor?)

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
```diff
@@ -3,6 +3,7 @@
     Copyright (C) 2019-2020, 2022, University College London
     Copyright (C) 2016-2017, PETsys Electronics
     Copyright (C) 2021, Gefei Chen
+    Copyright (C) 2024
```
KrisThielemans, line 6: who owns the copyright if your current work?

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
```diff
@@ -84,225 +127,274 @@ ML_estimate_component_based_normalisation(const std::string& out_filename_prefix
...
+  efficiencies_ptr = std::make_shared<DetectorEfficiencies>(IndexRange2D(num_physical_rings, num_physical_detectors_per_ring));
+
+  measured_geo_data = GeoData3D(num_physical_axial_crystals_per_basic_unit,
+                                num_physical_transaxial_crystals_per_basic_unit / 2,
+                                num_physical_rings,
+                                num_physical_detectors_per_ring); // inputes have to be modified
+  norm_geo_data_ptr = std::make_shared<GeoData3D>(num_physical_axial_crystals_per_basic_unit,
+                                                  num_physical_transaxial_crystals_per_basic_unit / 2,
+                                                  num_physical_rings,
+                                                  num_physical_detectors_per_ring); // inputes have to be modified
```
KrisThielemans, line 141: 2 `GeoData3D` objects? not great.

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
KrisThielemans: as long as `make_fan_data_remove_gaps` doesn't store references, pointers or whatever to the content of `*measured_projdata_stpr`, then it is fine. (and it doesn't, as we'd have had compilation warnings previously anyway)

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
```diff
@@ -84,225 +127,274 @@ ML_estimate_component_based_normalisation(const std::string& out_filename_prefix
...
+  make_fan_sum_data(data_fan_sums, measured_fan_data);
+  make_geo_data(measured_geo_data, measured_fan_data);
+  make_block_data(measured_block_data, measured_fan_data);
+  if (do_display)
     {
-      make_fan_data_remove_gaps(measured_fan_data, measured_data);
+      display(measured_block_data, "raw block data from measurements");
+    }
 
-      /* TEMP FIX */
```
KrisThielemans, line 113: I can't remember where this came from and would have to review, but are you sure we can cut it?

src/include/stir/recon_buildblock/ML_estimate_component_based_normalisation.h
```diff
@@ -14,25 +15,171 @@
...
+   */
+  float compute_threshold_for_KL();
+
+  // Constructor parameters
+  //! The prefix for the output files
+  std::string out_filename_prefix;
+  //! The measured projection data
+  const std::shared_ptr<ProjData>& measured_projdata;
+  //! The model projection data
+  const std::shared_ptr<ProjData>& model_projdata;
```
KrisThielemans, line 146: definitely not use references here.

KrisThielemans (changes requested):
some initial comments. I didn't review everything.

I'm not sure if we should provide access to the `efficiencies_sptr` etc. `shared_ptr` are nice, but to be avoided when we can. If we have a `const auto get_efficiencies()` member, there seems tobe no point in making `efficiencies` etc into a `shared_ptr`.

KrisThielemans:
>Additionally, I am not convinced there is any testing of this code.

there are tests in the `recon_test_pack` I believe.

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
robbietuk: I had to check, its actually me. Do you want my name here or blank?

KrisThielemans: yours! And we'll need another letter then :-)

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
robbietuk: They are not new https://github.com/UCL/STIR/blob/33a0106df6c796dbe718eea9c860565a41ff7c47/src/recon_buildblock/ML_estimate_component_based_normalisation.cxx#L92-L99

They are both used in `iterate_geo_norm(*norm_geo_data_ptr, measured_geo_data, fan_data);` Not sure how to simplify this

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
robbietuk: `make_fan_data_remove_gaps` is purely functional.

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
robbietuk: This `TEMP FIX` does 
```
if (model_fan_data(ra, a, rb, b) == 0)
    measured_fan_data(ra, a, rb, b) = 0;
```
and then sets the `threshold_for_KL = measured_fan_data.find_max() / 100000.F;`

This isn't removed, refactored into `compute_threshold_for_KL`. GitHub is not displaying these changes very well.

src/buildblock/ML_norm.cxx
robbietuk: The compiler is probably optimizing this anyway and its such a minor change. Lets leave it in but I see your point.

src/include/stir/recon_buildblock/ML_estimate_component_based_normalisation.h
robbietuk: Correct, the member is not used outside of the constructor. I have removed it.

robbietuk:
At this point, I want to add an interface with `BinNormalisationPETFromComponents`. Is the any specific reason for no direct setters of the eff, geo, block factors into `BinNormalisationPETFromComponents`?
https://github.com/UCL/STIR/blob/33a0106df6c796dbe718eea9c860565a41ff7c47/src/include/stir/recon_buildblock/BinNormalisationPETFromComponents.h#L60-L62
`apply_normfactors3D` uses streams to set these values
https://github.com/UCL/STIR/blob/33a0106df6c796dbe718eea9c860565a41ff7c47/src/utilities/apply_normfactors3D.cxx#L71-L83. I would prefer not to do this.

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
```diff
@@ -18,291 +19,371 @@
...
+                                                  do_save_to_file);
+  estimator.process();
+}
 
-  const int num_transaxial_blocks = measured_data.get_proj_data_info_sptr()->get_scanner_sptr()->get_num_transaxial_blocks();
-  const int num_axial_blocks = measured_data.get_proj_data_info_sptr()->get_scanner_sptr()->get_num_axial_blocks();
+MLEstimateComponentBasedNormalisation::MLEstimateComponentBasedNormalisation(
+    std::string out_filename_prefix_v,
+    const ProjData& measured_projdata_v,
+    const ProjData& model_projdata_v,
```
robbietuk, line 66: Should there be a check that these ProjData are the same "shape". Are there any tests for that?

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
KrisThielemans: ok fine. Sorry. Just correct "inputes" (although I don't know what those comments mean)

KrisThielemans:
> Is the any specific reason for no direct setters of the eff, geo, block factors into `BinNormalisationPETFromComponents`?

I was just running out of time...

Note that `BinNormalisationPETFromComponents` was intended to form a parent class for other component-based norms, including Siemens and GE. Not 100% sure how feasible that'll be, but it'd remove a lot of obscure/copied code. I'm not saying this because you need to do it, but just to be aware of the longer term plan. Clearly, adding setters/getters is not a problem.

On another note, the "block" stuff really doesn't work and shouldn't be used. It was intended to cope with timing-alignment factors (for many scanners these are per block) but it introduces too much freedom (which I didn't know when I wrote that code). Timing alignment should be parametrised by 1 factor per block, not a norm-factor for every block-pair. Just saying that such that you don't spend too much time on that (i.e. maybe don't expose it)

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
KrisThielemans: I haven't checked, but this fix will not just be important for KL, but probably also for the iterative update. (it'll detect 0/0, as oposed 1/0).

src/recon_buildblock/BinNormalisationPETFromComponents.cxx
```diff
@@ -173,6 +173,20 @@ BinNormalisationPETFromComponents::allocate(
...
+    {
+      error("BinNormalisationPETFromComponents: internal error: allocate called on MLEstimateComponentBasedNormalisation "
+            "without it having processed the data");
+    }
+  this->proj_data_info_sptr = pdi_sptr;
+  this->efficiencies = *normalization_estimator.get_efficiencies();
+  this->geo_data = *normalization_estimator.get_geo_data();
+  this->block_data = *normalization_estimator.get_block_data();
+  this->_already_allocated = true;
+}
```
robbietuk, line 189: I believe this is the cleanest way to implement all these changes. Here is a use case from my cpp program.

``` cpp
    MLEstimateComponentBasedNormalisation ml_estimator(out_filename_tmp, *measured_projdata, *model_activity_projdata,
                                                       num_eff_iterations, num_iterations, do_geo, do_block,
                                                       do_symmetry_per_block, do_KL, do_display, do_save_to_file);
    ml_estimator.process();

    // Apply the normalization factors
    BinNormalisationPETFromComponents bin_normalisation;
    bin_normalisation.allocate(measured_projdata->get_proj_data_info_sptr(), ml_estimator);
    bin_normalisation.set_up(measured_projdata->get_exam_info_sptr(), measured_projdata->get_proj_data_info_sptr());

    // Compute the projdata
    ProjDataInMemory normalization_projdata(*measured_projdata);
    normalization_projdata.fill(1.F);
    bin_normalisation.undo(normalization_projdata);
```

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
robbietuk: It was probably a WIP note from before. I will remove them.

src/recon_buildblock/BinNormalisationPETFromComponents.cxx
robbietuk: I acknowledge https://github.com/UCL/STIR/issues/1498#issuecomment-2322294430, however this does work and basically provides an alternative setter of the fields.

robbietuk:
Currently seeing the oddest error. The minimum tangential position is filled with inf.

```cpp
TEST(TestNormalizationCalculator, TBD)
{
    auto scanner = PETScanner();
    auto measured_projdata = get_empty_projdata(scanner, 1);
    measured_projdata->fill(1.F);

    auto model_projdata = std::make_shared<ProjDataInMemory>(*measured_projdata);
    model_projdata->fill(1.F);


    int num_eff_iterations = 1;
    int num_iterations = 1;
    bool do_geo = false;
    bool do_block = false;
    bool do_symmetry_per_block = false;
    bool do_KL = false;
    bool do_display = false;
    bool do_save_to_file = false;
    stir::MLEstimateComponentBasedNormalisation ml_estimator("", *measured_projdata, *model_projdata,
                                                             num_eff_iterations, num_iterations, do_geo, do_block,
                                                             do_symmetry_per_block, do_KL, do_display, do_save_to_file);
    ml_estimator.process();

    // Apply the normalization factors
    BinNormalisationPETFromComponents bin_normalisation;
    bin_normalisation.allocate(measured_projdata->get_proj_data_info_sptr(), ml_estimator);
    bin_normalisation.set_up(measured_projdata->get_exam_info_sptr(), measured_projdata->get_proj_data_info_sptr());

    // Compute the projdata
    ProjDataInMemory normalization_projdata(*measured_projdata);
    normalization_projdata.fill(1.F);
    bin_normalisation.apply(normalization_projdata);
    std::cout << "Complete calculating normalization factors" << std::endl;
    std::cout << "Max: " << normalization_projdata.find_max() << std::endl;
    std::cout << "Min: " << normalization_projdata.find_min() << std::endl;
    auto filename = "test_norm";
    normalization_projdata.write_to_file(filename);
```

It outputs 
```
Complete calculating normalization factors
Max: inf
Min: 1
```

_____________

Visualization

image
All voxels are 1 for every slice, except the `tangential_position = 0`. 

____________

I don't think this is a preexisting issue but I will check the following later
```Powershell
find_ML_normfactors3D.exe test_exe .\ones.hs .\twos.hs 1 1
apply_normfactors3D.exe test_apply_exe test_exe .\ones.hs 0 1 1
```

src/recon_buildblock/ML_estimate_component_based_normalisation.cxx
KrisThielemans: They should be. Not sure if it's checked anywhere.

src/recon_buildblock/BinNormalisationPETFromComponents.cxx
KrisThielemans: I'd rather do it the other way around I'm afraid. Your way is very different from anything else in STIR.

I very much prefer to let `MLEstimateComponentBasedNormalisation` have a `shared_ptr<BinNormalisationPETFromComponents> normalisation_sptr` member, which we can get via `get_output()` (returning a `shared_ptr<const BinNorm...>`). The `MLEstimateComponentBasedNormalisation::get_efficiencies()` would then just forward to `normalisation_sptr->get_efficiencies()` (note that I wouldn't let those return `shared_ptr` then, but references). Any reason not to do it this way?

KrisThielemans:
Infinities tend to come from 1/0. A candidate would be the `TEMP FIX` stuff.

KrisThielemans:
Note that you'll have to rename class and file to `MLEstimateComponentBasedNormalisation` (no `_` in class names according to STIR conventions)

KrisThielemans:
not sure what you did there, but the diff shows lots of things that shouldn't be there (from recent updates on `master`). Maybe you rebased as opposed to merged with master (although it doesn't look like that)? History is likely messed up unfortunately.

robbietuk:
I rebased onto master but GitHub doesn't handle that well it seems. I have reverted the rebase.

robbietuk:
This is failing because of a divide by 0. This because the projdata is 192 tangential positions but the fan size is calculated as 191 by
```cpp
   const int half_fan_size = min(proj_data_info.get_max_tangential_pos_num(), -proj_data_info.get_min_tangential_pos_num()); 
   fan_size = 2 * half_fan_size + 1; 
```
in [here](https://github.com/UCL/STIR/blob/e168b074b3f7d21b4ebf8872680b4bb115022719/src/buildblock/ML_norm.cxx#L973-L995). There seems to be some requirement for the fan size to be odd?
______

Regardless, during `BindNormalisationPETFromComponents::set_up` is called it constructs the projdata, which eventually reaches 

https://github.com/UCL/STIR/blob/e168b074b3f7d21b4ebf8872680b4bb115022719/src/buildblock/ML_norm.cxx#L1167-L1244

The issue is in the last for loop and the `bin.tangential_pos_num()` being initialized at `-half_fan_size = -95` whereas the `proj_data.get_min_tangential_pos_num() returns `-96`. Therefore, the first row `tangential_pos` is treated as a gap and set to 0. This zero is later used in the denominator of a division that leads to the infinity values in the test.

KrisThielemans:
> There seems to be some requirement for the fan size to be odd?

yes, that is the case at the moment. Code could be enhanced, but I was too lazy (having 1% extra data won't make or break the efficiency estimation)

KrisThielemans:
> Therefore, the first row `tangential_pos` is treated as a gap and set to `0`. This `0` is later used in the denominator of a division that leads to the infinity values in the test.

yes, so the TEMP FIX could/should handle this (the model has to be zero in that bin). Or you could trim the proj-data first (with SSRB)

robbietuk:
> > Therefore, the first row `tangential_pos` is treated as a gap and set to `0`. This `0` is later used in the denominator of a division that leads to the infinity values in the test.
> 
> yes, so the TEMP FIX could/should handle this (the model has to be zero in that bin). Or you could trim the proj-data first (with SSRB)

The `TEMP FIX` code sits in the `MLEstimateComponentBasedNormalisation` class. This class correctly populates the `DetectorEfficiencies` with uniform `0.707*` values in the test. It also has the shape (`24`,`384`) = (`num_rings`, `num_det_per_ring`). I'm not sure that is the problem. 

The issue is when the `DetectorEfficiencies` are applied to the `invnorm_proj_data_sptr` in `BindNormalisationPETFromComponents::set_up`. Because of the `half_fan_size` calculation, the min tangential position is not accounted for in the efficiencies.

I am not a huge fan of trimming the projdata with SSRB. I am trying to understand the code to figure out why the `min_tangential_pos_num` is not able to be populated

KrisThielemans:
It's not the efficiencies themselves that are the problem, but the `FanData`
https://github.com/UCL/STIR/blob/e168b074b3f7d21b4ebf8872680b4bb115022719/src/buildblock/ML_norm.cxx#L745-L746
Moreover, there's plenty of loops doing
``
for (bin.tangential_pos_num() = -half_fan_size; bin.tangential_pos_num() <= half_fan_size; ++bin.tangential_pos_num())
```
This isn't going to be so easy to fix.

**Comment by KrisThielemans** [comment]
Note that #1430 updates the `ML_norm` files as well. We didn't manage to finish that one...

**Comment by robbietuk** [comment]
I am not up to date with the theory in the papers. Do you expect any issue if I fix the fan data by removing the `half_fan_size` and `fan_size` calculations and replace them with methods that better derive from `get_num_tangential_poss()`. I feel this is the problem now.

**Comment by robbietuk** [comment]
I changed the scanner to "Discovery 690", with an odd number of tangential positions (`381` I believe). This allows the test to pass. 

The previous issue with `0` and `+inf` values in the resulting projection data is pushed to issue #1511

File: src/include/stir/recon_buildblock/MLEstimateComponentBasedNormalisation.h [hunk 10]
```diff
@@ -0,0 +1,200 @@
...
+
+  //! The threshold for the Kullback-Leibler divergence calculation
+  float threshold_for_KL;
+  FanProjData model_fan_data;
+  FanProjData fan_data;
+  DetectorEfficiencies data_fan_sums;
+  BlockData3D measured_block_data;
+  GeoData3D measured_geo_data;
+
+  bool data_processed = false;
```
robbietuk, line 191: Document

src/include/stir/recon_buildblock/MLEstimateComponentBasedNormalisation.h
```diff
@@ -0,0 +1,200 @@
...
+                                        bool do_display_v,
+                                        bool do_save_to_file_v);
+
+  /*!
+  \brief Run the normalisation estimation algorithm using the parameters provided in the constructor.
+  */
+  void process();
+
+  //! Check if the data has been processed
+  bool has_processed_data() const;
```
robbietuk, line 92:
```suggestion
  //! Check if the data has been processed
  bool data_is_processed() const;
```
Also make all setter methods change the returned value to false.

robbietuk (commented):
A few check before I believe this is done.

robbietuk:
Pretty sure this PR is ready for review. I can't find any more outstanding issues. #1511 covers the other issue with the fan data and even numbers.  Ill add to the release notes.

robbietuk:
This should be merged into v6.3. I have been using a branch derived from here for a while with no major issues.

src/include/stir/recon_buildblock/MLEstimateComponentBasedNormalisation.h
```diff
@@ -0,0 +1,316 @@
...
+                                        bool do_block_v,
+                                        bool do_symmetry_per_block_v,
+                                        bool do_KL_v,
+                                        bool do_display_v,
+                                        bool do_save_to_file_v);
+
+  /*!
+  \brief Run the normalisation estimation algorithm using the parameters provided in the constructor.
+  */
+  void process();
```
KrisThielemans, line 89: Many STIR classes have `process_data()` instead, so use this here as well.

Also, we need a ` Succeeded set_up() `, even if just returns `Succeeded::yes`.

src/utilities/find_ML_normfactors3D.cxx
```diff
@@ -114,7 +114,8 @@ main(int argc, char** argv)
                                             do_block,
                                             do_symmetry_per_block,
                                             do_KL,
-                                            do_display);
+                                            do_display,
```
KrisThielemans, line 117: This call needs to be rewritten in terms of the class, such that we don't use the deprecated function anymore.

src/recon_test/test_MLEstimateComponentBasedNormalisation.cxx
```diff
@@ -0,0 +1,120 @@
+/*
+Copyright (C) 2024, Robert Twyman skelly
+    This file is part of STIR.
+    SPDX-License-Identifier: Apache-2.0
+    See STIR/LICENSE.txt for details
+*/
+/*!
+  \file
+  \ingroup recon_test
+  \brief Test program to ensure MLEstimateComponentBasedNormalisation works as expected and
```
KrisThielemans, line 10:
```suggestion
  \brief Test program to ensure stir::MLEstimateComponentBasedNormalisation works as expected and
```

src/recon_test/test_MLEstimateComponentBasedNormalisation.cxx
```diff
@@ -0,0 +1,120 @@
...
+         produces the correct normalization factors from components.
+         This is a very basic test that ensure the basic functionality.
```
KrisThielemans, line 12:
```suggestion
         This is a test on basic functionality only.
```

src/recon_test/test_MLEstimateComponentBasedNormalisation.cxx
```diff
@@ -0,0 +1,120 @@
...
+
+  /*! Runs a test to check the normalization calculation with efficiencies only.
+   * Uses two uniform projdata sets, one with 1s and the other with 2s to estimate the normalization factors.
+   * The efficiencies are then checked to ensure they are within the expected range.
+   * The normalization factors are then computed from PET components and applied to a uniform projdata set.
+   */
+  void test_normalization_calculation_with_efficiencies()
+  {
+    const auto scanner = std::make_shared<Scanner>(*Scanner::get_scanner_from_name("Discovery 690"));
+    const auto exam_info = std::make_shared<ExamInfo>(ImagingModality::PT);
```
KrisThielemans, line 47: rename to `scanner_sptr` and `exam_info_sptr`

src/recon_test/test_MLEstimateComponentBasedNormalisation.cxx
```diff
@@ -0,0 +1,120 @@
...
+    exam_info->patient_position = PatientPosition(PatientPosition::HFS);
+
+    const shared_ptr<ProjDataInfo> projdata_info(ProjDataInfo::ProjDataInfoCTI(
```
KrisThielemans, line 50: use `ProjDataInfo::construct_proj_data_info` (same signature)

src/recon_test/test_MLEstimateComponentBasedNormalisation.cxx
```diff
@@ -0,0 +1,120 @@
...
+        /*scanner_ptr*/ scanner,
+        /*span*/ 1,
+        /*max_delta*/ 23,
+        /*views*/ scanner->get_num_detectors_per_ring() / 2,
+        /*tang_pos*/ scanner->get_default_num_arccorrected_bins(),
+        /*arc_corrected*/ false,
+        /*tof_mash_factor*/ -1));
```
KrisThielemans, line 57: why -1? Set to 0 for non-TOF.

src/recon_test/test_MLEstimateComponentBasedNormalisation.cxx
```diff
@@ -0,0 +1,120 @@
...
+
+    const auto measured_projdata = std::make_shared<ProjDataInMemory>(ProjDataInMemory(exam_info, projdata_info, false));
+    measured_projdata->fill(1.F);
+
+    const auto model_projdata = std::make_shared<ProjDataInMemory>(*measured_projdata);
+    model_projdata->fill(2.F);
+
+    constexpr int num_eff_iterations = 6;
```
KrisThielemans, line 65: Can we make this smaller (to run faster)?

KrisThielemans (changes requested):
Small changes required, but unfortunately quite a lot of them.

We cannot rename functions/files in 6.3, so I've flagged those up. Deprecations need to be listed in the release notes.

KrisThielemans:
Also, you'll need to run `pre-commit run --all` (see the devel doc). Also check the Codacy report.

KrisThielemans:
@robbietuk where are we with this? I think it might be best to postpone for after 6.3, as that needs to be completed soon.

robbietuk:
Postpone for now unless you can handle the review items. Sorry, I don't have much free time for this

Issue #1507, opened 2024-09-10
tangential position range inconsistency


SomayehSaghamanesh:
There are two different definitions (as below), which result in inconsistency in tangential position range. This is problematic at least in cases handling full tangential positions.
In ProjDataInfo.cxx:
  min_tangential_pos_num = -(num_tang_poss / 2);
  max_tangential_pos_num = min_tangential_pos_num + num_tang_poss - 1;

In ProjDataInfoCylindricalNoArcCorr.cxx and ProjDataInfoGenericNoArcCorr.cxx :
  const int min_tang_pos_num = -(num_detectors / 2) + 1;
  const int max_tang_pos_num = -(num_detectors / 2) + num_detectors;

robbietuk:
The class inheritance is as follows
![image](https://github.com/user-attachments/assets/bc2a34a7-78fc-43e0-9c56-523c3fe7cf4d)

I suggest using this https://github.com/UCL/STIR/blob/e168b074b3f7d21b4ebf8872680b4bb115022719/src/buildblock/ProjDataInfo.cxx#L123-L129
in the child classes for consistency?

Maybe worth noting, the definition of the max/min tang poss in `ProjDataInfoCylindricalNoArcCorr` is very old (circa 2003). Somethings may break because they were built upon this methodology.

SomayehSaghamanesh:
[...]

PR #1508, opened 2024-09-10
Tangential position range consistency

SomayehSaghamanesh:
Changed min/max_tang_pos_num definitions in two classes to be consistent with ProjDataInfo.cxx. 

## Testing performed
It is built successfully but fails in 8 tests (in make test). Probably needs more modification?
I could handle projection files with tang180 (with lm_to_projdata and convert_projdata_types) without error. Before implementing above changes, I got below error:
"ERROR: The tangential_pos range (-90 to 89) for this projection data is too large. Maximum supported range is from -89 to 90"

## Related issues
#1507

### DIFF
--- a/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx
+++ b/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx
@@ -178,8 +178,8 @@ ProjDataInfoCylindricalNoArcCorr::initialise_uncompressed_view_tangpos_to_det1de
   assert(fabs(get_phi(Bin(0, 0, 0, 0)) - v_offset) < 1.E-4);
   assert(fabs(get_phi(Bin(0, get_num_views(), 0, 0)) - v_offset - _PI) < 1.E-4);
 #endif
-  const int min_tang_pos_num = -(num_detectors / 2) + 1;
-  const int max_tang_pos_num = -(num_detectors / 2) + num_detectors;
+  const int min_tang_pos_num = -(num_detectors / 2);
+  const int max_tang_pos_num = -(num_detectors / 2) + num_detectors - 1;
 
   if (this->get_min_tangential_pos_num() < min_tang_pos_num || this->get_max_tangential_pos_num() > max_tang_pos_num)
     {
--- a/.gitignore
+++ b/.gitignore
@@ -71,6 +71,7 @@ install_manifest.txt
 build/
 install/
 .vscode/
+build_safir_updates/
 
 # MATLAB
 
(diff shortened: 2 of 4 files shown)


src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx
```diff
@@ -178,8 +178,8 @@ ProjDataInfoCylindricalNoArcCorr::initialise_uncompressed_view_tangpos_to_det1de
   assert(fabs(get_phi(Bin(0, 0, 0, 0)) - v_offset) < 1.E-4);
   assert(fabs(get_phi(Bin(0, get_num_views(), 0, 0)) - v_offset - _PI) < 1.E-4);
 #endif
-  const int min_tang_pos_num = -(num_detectors / 2) + 1;
-  const int max_tang_pos_num = -(num_detectors / 2) + num_detectors;
+  const int min_tang_pos_num = -(num_detectors / 2);
+  const int max_tang_pos_num = -(num_detectors / 2) + num_detectors - 1;
```
KrisThielemans, line 182: I'm not sure about this.  We have
https://github.com/UCL/STIR/blob/0e667f713967f45f76cbb906805e3058e9e09f20/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx#L206-L208
Subtracting
```
diff =  ((tp_num >> 1) + ((tp_num + 1) >> 1) + num_detectors/2 ) % num_detectors
```
Checking some examples (and relying on https://github.com/UCL/STIR/blob/0e667f713967f45f76cbb906805e3058e9e09f20/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx#L168-L169 , it transpires that this is equal to 
```
diff =  (tp_num + num_detectors/2 ) % num_detectors
```
Therefore, setting `tp_num = -(num_detectors/2)` gives `diff==0`. On the other hand, using the current  `max_tang_pos_num` also gives us `diff==0`. Both are "self-coincidences", which don't make a lot of sense.

It is probably ok to have a too large look-up table (although see [this comment](https://github.com/UCL/STIR/blob/0e667f713967f45f76cbb906805e3058e9e09f20/src/buildblock/ProjDataInfoCylindricalNoArcCorr.cxx#L270-L271), but I think that if you're having this problem, your `num_tangential_poss` is too large. The max is `num_detectors -1` really (which arguably we should check).

KrisThielemans (commented):
I'm reluctant to merge this. If anything, we should keep the current max. However, this goes rather deep, and as you say 8 tests failed, I cannot see how we can merge as-is to cope with your case where `num_tangential_poss=num_detectors_per_ring` (which should mean that the first bin is always zero)

PR #1510, opened 2024-09-10
Adds ProjDataInMemory::read_from_file to return a ProjDataInMemory object

robbietuk:

Previously, 
``` cpp
auto pd1 = ProjData::read_from_file(fname);
auto pd2 = ProjDataInMemory::read_from_file(fname);
```
would both return a `ProjData` object.

This change, overwrites the `ProjData::read_from_file()` method for `ProjDataInMemory`. This allows `ProjDataInMemory::read_from_file()` to return a `ProjDataInMemory` object.

## Testing performed
CI

### DIFF
--- a/src/buildblock/ProjDataInMemory.cxx
+++ b/src/buildblock/ProjDataInMemory.cxx
@@ -353,13 +353,7 @@ ProjDataInMemory::ProjDataInMemory(const ProjDataInMemory& proj_data)
 shared_ptr<ProjDataInMemory>
 ProjDataInMemory::read_from_file(const std::string& filename, const std::ios::openmode openmode)
 {
-#if 1
-  const auto proj_data = ProjData::read_from_file(filename, openmode);
-  return std::make_shared<ProjDataInMemory>(*proj_data);
-#else
-  // Alternative implementation
   return std::make_shared<ProjDataInMemory>(*ProjData::read_from_file(filename, openmode));
-#endif
 }
 
 float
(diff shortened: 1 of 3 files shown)


src/buildblock/ProjDataInMemory.cxx
```diff
@@ -350,6 +350,18 @@ ProjDataInMemory::ProjDataInMemory(const ProjDataInMemory& proj_data)
...
+shared_ptr<ProjDataInMemory>
+ProjDataInMemory::read_from_file(const std::string& filename, const std::ios::openmode openmode)
+{
+#if 1
+  const auto proj_data = ProjData::read_from_file(filename, openmode);
+  return std::make_shared<ProjDataInMemory>(*proj_data);
+#else
+  // Alternative implementation
+  return std::make_shared<ProjDataInMemory>(*ProjData::read_from_file(filename, openmode));
+#endif
```
robbietuk, line 362: Not sure which implementation is best. I think the second

```cpp
return std::make_shared<ProjDataInMemory>(*ProjData::read_from_file(filename, openmode));
```
might prevent an additional copy. I am just not sure about the data being correctly loaded.

KrisThielemans: they're identical really. it needs to make a copy of the data anyway as `PojData::read_from_file` will create a "read from disk" object.

robbietuk:
@KrisThielemans Are you happy with this implementation? It can be squash merged.

KrisThielemans (changes requested):
actually, the `open_mode` should be removed (or at least ignored). `ProjDataInMemory` will always be read/write, and we're making a copy anyway. The initial `ProjData::read_from_file` should always be read-only (more chance that it works), which is the default.

Also, add something to `release_6.3.htm`

KrisThielemans (approved):

PR #1549, opened 2024-12-10
ProjDataInMemory: avoid allocating with zero unless required

evgueni-ovtchinnikov:

Unnecessary 0-fill of newly created acquisition data reduced.

## Testing performed

The following SIRF Python script (employing STIR via SIRF Python interface) was used for comparing STIR's `ProjDataInMemory` algebra performance with `numpy`:
```
import sirf.STIR as pet
pet.AcquisitionData.set_storage_scheme('memory')

acq = pet.AcquisitionData('prompts.hs')
x = acq.clone()
y = x*2

start = timeit.default_timer()
z = x + y
elapsed = timeit.default_timer() - start
print(f'z = x + y time {elapsed}')

start = timeit.default_timer()
x_arr = x.as_array()
y_arr = y.as_array()
elapsed = timeit.default_timer() - start
print(f'as_array time {elapsed}')
numpy_time = elapsed

start = timeit.default_timer()
z_arr = x_arr + y_arr
elapsed = timeit.default_timer() - start
print(f'numpy z = x + y time {elapsed}')
numpy_time += elapsed

start = timeit.default_timer()
z.fill(z_arr)
elapsed = timeit.default_timer() - start
print(f'fill time {elapsed}')
numpy_time += elapsed
print(f'total numpy z = x + y time {numpy_time}')
```
With a recent STIR, the output looks like
```
z = x + y time 0.5256878180002786
as_array time 0.23524911599997722
numpy z = x + y time 0.1330225659999087
fill time 0.07107986500022889
total numpy z = x + y time 0.4393515470001148
```
demonstrating that STIR's `ProjDataInMemory` algebra is much less efficient than `numpy` algebra applied to `ProjDataInMemory` data copied into `numpy` array. The investigation of this issue revealed that a considerable chunk of the computation time was wasted on zero-filling the newly created `ProjDataInMemory` object where `x + y` data is to end up.

This PR succeeded in reducing such waste, so that the timings have become
```
z = x + y time 0.24671401699015405
as_array time 0.23049703100696206
numpy z = x + y time 0.13044680299935862
fill time 0.0659959859913215
total numpy z = x + y time 0.4269398199976422
```


KrisThielemans:
Ubuntu jobs fail, see #1550 

MacOS is unaffected by that, but has
```
ERROR: NumericVectorWithOffset::xapyb: index ranges don't match
```
(note that this will be seen in Debug mode only. Otherwise it will likely lead to a segfault)

src/include/stir/Array.h
```diff
@@ -579,6 +588,8 @@ class Array<1, elemT> : public NumericVectorWithOffset<elemT, elemT>
     \arg data_ptr should start to a contiguous block of correct size
   */
   inline void init(const IndexRange<1>& range, elemT* const data_ptr, bool copy_data);
+
+  bool init_with_zeros_= 1;
```
KrisThielemans, line 592:
```suggestion
  bool init_with_zeros_= true;
```

src/buildblock/ProjDataInMemory.cxx
```diff
@@ -73,12 +73,18 @@ ProjDataInMemory::ProjDataInMemory(shared_ptr<const ExamInfo> const& exam_info_s
...
+  }
 #if 0  
   float *b = new float[this->size_all()];
   if (initialise_with_0)
     memset(b, 0, this->size_all()*sizeof(float));
   return b;
 #else
+  this->buffer.set_initialise_with_zeros(true);
   this->buffer = Array<1, float>(static_cast<int>(this->size_all()));
 #endif
```
KrisThielemans, line 89:
```suggestion
    this->buffer.set_initialise_with_zeros(initialise_with_0);
    this->buffer.resize(0, this->size_all()-1);
```
Note the `-1` (STIR gives min/max indices, not like Python)

KrisThielemans (changes requested):
While this would work, it could be a bit surprising to the user, as the effect of `set_initialise_with_zeros()` would be permanent. In the use-case that you have in mind, that'd be fine, but it'd have to come with serious warnings in the doxygen.

I'm rather in favour of having an explicit `bool` argument to `grow` and `resize` (defaulting to `true`). We could then have that for the constructor as well.

KrisThielemans:
Please also use `pre-commit`, see https://github.com/UCL/STIR/blob/master/documentation/devel/git-hooks.md

KrisThielemans:
No point in pushing until I've sorted out https://github.com/UCL/STIR/pull/1552, which you'll have to merge here.

KrisThielemans:
#1552 is now merged, so merge `master` back on this branch.

src/buildblock/ProjDataInMemory.cxx
evgueni-ovtchinnikov: done

KrisThielemans:
> I'm rather in favour of having an explicit `bool` argument to `grow` and `resize` (defaulting to `true`). We could then have that for the constructor as well.

I still think this would be a better option. @danieldeidda @markus-jehl what do you think?

markus-jehl:
I tend to agree with @KrisThielemans, that explicitly stating it may be cleaner rather than relying on an internal setting on individual instances of the Array class. Just because I can imagine running into weird issues and struggling for a while before noticing what is going on 😅

KrisThielemans (changes requested):
I have some small comments, but realise now a conceptual difficulty: you had to add an arg to the virtual `VectorWithOffset::grow`. However, there it is unused (and would necessarily make sense, as "0" might not work for an arbitrary type). This is therefore still surprising to the user.

All this is more complicated due to the fact that adding default values to virtual functions is not possible and changing default values for virtual functions isn't recommended, see e.g. https://softwareengineering.stackexchange.com/questions/355074/is-overriding-a-pure-virtual-function-with-default-arguments-good-or-bad.

Could we do something like this
```
// make a "no_op" function object
virtual VectorWithOffset::grow(min_idx, max_idx, func& f); // apply `f` on new elements
virtual VectorWithOffset::grow(min_idx, max_idx); // no `f`, i.e. current version

virtual Array::grow(min_idx, max_idx); // assign with 0 (i.e. current version)
using virtual Array::grow(min_idx, max_idx, f); // use base-class version
```
if we don't want to initialise, then we could pass the no_op  function.

Difficulties:
- it's a bit ugly to have to pass `no_op` (caused by the current Array::grow default being the wrong thing)
- to avoid overhead, we'd have to detect if `f == no_op` and avoids loops.

Not sure how to resolve this, unfortunately. Any ideas?

KrisThielemans:
We need to discuss this further. The `grow` virtual function situation is currently a mess, which probably causes confusion here. Also, to be able to use CUDA managed pointers, we will likely need to pass a `std::allocator` (as in `std::vector`). That might then open up an opportunity to use a default allocator that doesn't initialise anything, see https://stackoverflow.com/a/21028912 for a solution.

Tough stuff....

src/include/stir/Array.h
```diff
@@ -451,13 +451,13 @@ class Array<1, elemT> : public NumericVectorWithOffset<elemT, elemT>
   inline virtual void grow(const IndexRange<1>& range);
 
   // Array::grow initialises new elements to 0
-  inline void grow(const int min_index, const int max_index) override;
+  inline virtual void grow(const int min_index, const int max_index, bool initialise_with_0 = true);
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: suggested change implemented

src/include/stir/Array.h
```diff
@@ -451,13 +451,13 @@ class Array<1, elemT> : public NumericVectorWithOffset<elemT, elemT>
...
 
   //! Array::resize initialises new elements to 0
   inline virtual void resize(const IndexRange<1>& range);
 
   // Array::resize initialises new elements to 0
-  inline void resize(const int min_index, const int max_index) override;
+  inline virtual void resize(const int min_index, const int max_index, bool initialise_with_0 = true);
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: suggested change implemented (commit 241c04180133615814cf158e040305592784f6c7)

src/include/stir/Array.h
```diff
@@ -569,6 +569,15 @@ class Array<1, elemT> : public NumericVectorWithOffset<elemT, elemT>
...
   //@}
 
+  void set_initialise_with_zeros(bool iwz)
+  {
+    init_with_zeros_ = iwz;
+  }
+  bool get_initialise_with_zeros() const
+  {
+    return init_with_zeros_;
+  }
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: removed

src/include/stir/Array.h
```diff
@@ -579,6 +588,8 @@ class Array<1, elemT> : public NumericVectorWithOffset<elemT, elemT>
     \arg data_ptr should start to a contiguous block of correct size
   */
   inline void init(const IndexRange<1>& range, elemT* const data_ptr, bool copy_data);
+
+  bool init_with_zeros_ = true;
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: removed

src/include/stir/VectorWithOffset.h
```diff
@@ -194,7 +194,7 @@ class VectorWithOffset
       resize() in a derived class, it is probably safest to overload
       grow() as well.
   */
-  inline virtual void grow(const int min_index, const int max_index);
+  inline virtual void grow(const int min_index, const int max_index, bool initialise_with_0 = true);
```
KrisThielemans, line 197: we'll need the same strategy for `grow` as for `resize`

src/include/stir/VectorWithOffset.h
```diff
@@ -207,6 +207,8 @@ class VectorWithOffset
       \todo in principle reallocation could be avoided when the new range would fit in the
       old one by shifting.
     */
+  // inline virtual void resize(const int min_index, const int max_index, bool initialise_with_0 = true);
+  inline virtual void resize(const int min_index, const int max_index, bool initialise_with_0);
   inline virtual void resize(const int min_index, const int max_index);
```
KrisThielemans, line 212: will need a brief doxygen comment, saying it effectively set `initialise_with_0=true`

KrisThielemans (changes requested):
I think we now don't need the
```
virtual void VectorWithOffset::grow(const int min_index, const int max_index, bool initialise_with_0 = true);
```
at all (and same for `resize` of course), i.e. the point was that `initialise_with_0 ` doesn't necessarily make sense for a `VectorWithOffset<elemT>` (depends on `elemT`), so we avoid having that new functionality there, but only add it for `Array`.

(If this is correct, then obviously the `Array` version will no longer have `override`).

I think we're nearly good to go, but please clean up commented out code.

KrisThielemans (approved):

src/include/stir/Array.inl
```diff
@@ -609,7 +623,7 @@ template <class elemT>
 void
 Array<1, elemT>::grow(const int min_index, const int max_index)
 {
-  resize(min_index, max_index);
+  resize(min_index, max_index, true);
```
KrisThielemans, line 626:
```suggestion
  resize(min_index, max_index);
```

PR #1556, opened 2024-12-20
Minor clean-up of ProjDataInfoBlocksOnCylindricalNoArcCorr

KrisThielemans:
- remove unused file
- fix some includes
- make 2 deprecated members private (on the long path to resolving #105). Note that this in principle breaks backwards compatibility, but they are unused, and I'd prefer to merge this in 6.3 to make some progress...


KrisThielemans:
I removed some dynamic checks on `ProjDataInfoBlocksOnCylindrical` and use the fall-back to generic. Note however that this code was never used, as currently the `ProjDataInfoCylindrical` case will always be used (as `ProjDataInfoGeneric` is derived from it).

Issue #1491, opened 2024-08-19
listmode Hessian test fails on MacOS CLang 18


KrisThielemans:
https://github.com/UCL/STIR/actions/runs/10450627407/job/28935385772?pr=1489#step:12:207
```
----- testing concavity via Hessian-vector product (accumulate_Hessian_times_input)

INFO: Loaded 8019 prompts from list-mode file

INFO: Caching Additive corrections for : 8019 events.

INFO: Listmode gradient calculation: starting loop with 1 thread

INFO: Computation times for distributable_computation, CPU 0.01s, wall-clock 0.001722s

INFO: Loaded 8019 prompts from list-mode file

INFO: Caching Additive corrections for : 8019 events.

INFO: Listmode gradient calculation: starting loop with 1 thread

INFO: Computation times for distributable_computation, CPU 0s, wall-clock 0.001746s
Error : 0 is larger than 0, 
FAIL: PoissonLLListModeData: Computation of x^T H x = 0.000000 > 0 (Hessian) and is therefore NOT concave
 >target image max=0.999849
 >target image min=1.16359e-05
 >output image max=0
 >output image min=0
Writing diagnostic files to PoissonLLListModeData_concavity_out.hv, *target.hv
```

Other tests seem to be fine

KrisThielemans:
Possibly related to #1200

KrisThielemans:
Identical error seen on AppVeyor build for a private version of STIR, for both Visual Studio 17 2022 and Visual Studio 16 2019. We don't have this on the public version of STIR appveyor builds. As far as I can see, the difference between those is that the public version uses Appveyor Build worker image Visual Studio 2015, while the private one uses Build worker image Visual Studio 2019. (I'm not 100% sure what the consequences of this are, but presumably it means that underlying software is more recent on the 2019 version.)

KrisThielemans:
#1623 re-enabled tests, and they're working fine. I don't know yet if there's an underlying problem or not.

KrisThielemans:
we no longer see failures, so presumably #1566 fixed this problem. I will close this.

PR #1566, opened 2025-02-20
FIX Without OPENMP, LM_distributable_computation doesn't write output.

### LINKED ISSUE #1422: MacOS job fails on the list-mode Hessian-concavity test
#1418 introduced Hessian calculation for the LM objective function. However, the MacOS job fails on the Hessian-concavity test (and was disabled). Also, `run_test_listmode_recon.sh` is not used on GHA.

The output of H*dx is 0 on MacOS, but not on other systems. I have no idea why. Most likely it doesn't read the events properly, but it does read the correct number of events, so that's hard to understand. Here's the failure diagnostics
```
----- testing concavity via Hessian-vector product (accumulate_Hessian_times_input)
INFO: Loaded 8019 prompts from list-mode file
INFO: Caching Additive corrections for : 8019 events.
INFO: Listmode gradient calculation: starting loop with 1 thread
INFO: Computation times for distributable_computation, CPU 0.03s, wall-clock 0.037424s
INFO: Loaded 8019 prompts from list-mode file
INFO: Caching Additive corrections for : 8019 events.
INFO: Listmode gradient calculation: starting loop with 1 thread
INFO: Computation times for distributable_computation, CPU 0.04s, wall-clock 0.037748s
Error : 0 is larger than 0, 
FAIL: PoissonLLListModeData: Computation of x^T H x = 0.000000 > 0 (Hessian) and is therefore NOT concave
 >target image max=0.999849
 >target image min=1.16359e-05
 >output image max=0
 >output image min=0
```

_Originally posted by @KrisThielemans in https://github.com/UCL/STIR/issues/1418#issuecomment-2109518132_

ashgillman:
Should fix #1422 #1491, maybe #1200. 

G'day! Long time no see.

I'm having a little explore into the gantry shift PR to see what is required, but first step is to get master building on my machine (OSX).

I'm not familiar with this code, but it looks like the idea was to store the outputs per thread in `local_ooutput_image_sptrs` and sum them up at the end. Without OpenMP (required if Clang), it is treated as one thread, but the writing back out never happend.

## Testing performed
I've tested with `ctest`, and tests now passing.

  - [-] I have added docstrings/doxygen in line with the guidance in the developer guide
  - [X] The code builds and runs on my machine

### DIFF
--- a/src/include/stir/recon_buildblock/distributable.txx
+++ b/src/include/stir/recon_buildblock/distributable.txx
@@ -1,6 +1,7 @@
 /*
     Copyright (C) 2024 University College London
     Copyright (C) 2020, 2022, Univeristy of Pennsylvania
+    Copyright (C) 2025, Commonweath Scientific and Industrial Research Organisation
     This file is part of STIR.
 
     SPDX-License-Identifier: Apache-2.0 AND License-ref-PARAPET-license
@@ -16,6 +17,7 @@
 
   \author Nikos Efthimiou
   \author Kris Thielemans
+  \author Ashley Gillman
 */
 #include "stir/shared_ptr.h"
 #include "stir/recon_buildblock/distributable.h"
@@ -139,8 +141,7 @@ LM_distributable_computation(const shared_ptr<ProjMatrixByBin> PM_sptr,
                   local_double_out_ptrs[thread_num]);
       }
   }
-#ifdef STIR_OPENMP
-  // flatten data constructed by threads
+  // flatten data constructed by threads (or collapse unitary dim if no threading)
   {
     if (double_out_ptr != NULL)
       {
@@ -157,7 +158,6 @@ LM_distributable_computation(const shared_ptr<ProjMatrixByBin> PM_sptr,
             *output_image_ptr += *(local_output_image_sptrs[i]);
       }
   }
-#endif
   CPU_timer.stop();
   wall_clock_timer.stop();
   info(boost::format("Computation times for distributable_computation, CPU %1%s, wall-clock %2%s") % CPU_timer.value()
--- a/src/include/stir/recon_buildblock/distributable.txx
+++ b/src/include/stir/recon_buildblock/distributable.txx
@@ -1,7 +1,7 @@
 /*
-    Copyright (C) 2024 University College London
-    Copyright (C) 2020, 2022, Univeristy of Pennsylvania
-    Copyright (C) 2025, Commonweath Scientific and Industrial Research Organisation
+    Copyright (C) 2024, 2025 University College London
+    Copyright (C) 2020, 2022, University of Pennsylvania
+    Copyright (C) 2025, Commonwealth Scientific and Industrial Research Organisation
     This file is part of STIR.
 
     SPDX-License-Identifier: Apache-2.0 AND License-ref-PARAPET-license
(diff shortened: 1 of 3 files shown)


ashgillman:
See also: https://github.com/KrisThielemans/STIR-1/commit/cab1c8e207b9834e60a12865d1b7e79ba75b6469

But the re-enabling the build fails with:
```
/Users/runner/work/_temp/9e40082b-aa77-4c9f-b976-fcf58e7dc9a9.sh: line 6: /opt/homebrew/opt/llvm/bin/clang: No such file or directory
```

```
 + brew install llvm@16
==> Downloading https://ghcr.io/v2/homebrew/core/llvm/16/manifests/16.0.6_1
==> Fetching llvm@16
==> Downloading https://ghcr.io/v2/homebrew/core/llvm/16/blobs/sha256:e799fe5056da9a07d79474f7a7290e0a465474eb004cfc614775f290197e10e5
==> Pouring llvm@16--16.0.6_1.arm64_sonoma.bottle.tar.gz
==> Caveats
To use the bundled libc++ please add the following LDFLAGS:
  LDFLAGS="-L/opt/homebrew/opt/llvm@16/lib/c++ -Wl,-rpath,/opt/homebrew/opt/llvm@16/lib/c++"

llvm@16 is keg-only, which means it was not symlinked into /opt/homebrew,
because this is an alternate version of another formula.

If you need to have llvm@16 first in your PATH, run:
  echo 'export PATH="/opt/homebrew/opt/llvm@16/bin:$PATH"' >> /Users/runner/.bash_profile

For compilers to find llvm@16 you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/llvm@16/lib"
  export CPPFLAGS="-I/opt/homebrew/opt/llvm@16/include"
```

So it uses a different path because 16 is non-default.

https://formulae.brew.sh/formula/llvm

Default should be 19

ashgillman:
Okay, so failing a lot of tests with Clang.

Seems like it fails whenever a test deliberately throws an exception, not being caught. Possibly related: https://github.com/llvm/llvm-project/issues/94292

I'm using clang 16 locally with no error. Probably the best way is to revert the 16->19 change and fix the CC CXX paths, but I'll await what you think, Kris, rather than using up all the project's GHA credits.

KrisThielemans:
I'm amazed. This seems a bug that is not specific to MacOS at all. I see that we don't test with OpenMP=OFF at all before this PR! I think it'd be good to add an Ubuntu gcc test without OpenMP as well.

I think there's a mistake though.
https://github.com/UCL/STIR/blob/729527cfc31b6e5f6d54d4f4722ffcd484f31685/src/include/stir/recon_buildblock/distributable.txx#L96 initialises `local_double_out_ptrs[0]` with `double_out_ptr`. So, I think we'll be double counting the `double_out_ptr` after this fix. This will currently not be used/tested, but please check.

I think for the image, originally it was also intended to avoid the final copy, i.e. process directly into `output_image_ptr`. However, I'm not sure if that'd be correct (without spending more time than I have) due to accumulation vs overwriting.

Possibly the best/clearest solution might be to always use the code in https://github.com/UCL/STIR/blob/729527cfc31b6e5f6d54d4f4722ffcd484f31685/src/include/stir/recon_buildblock/distributable.txx#L77-L88, using `num_threads` and `max_num_threads` variables initialised appropriately.

KrisThielemans:
Regarding the exception problem, this is #1490. I have no energy for this, so using clang 16 on GHA seems best.

The fact that the bug is not MacOS specific implies you could do those fixes in a separate PR if you prefer.

danieldeidda:
BTW I installed master with openMP off and run "run_tests_listmode_recon.sh" i get no issue

KrisThielemans:
> BTW I installed master with openMP off and run "run_tests_listmode_recon.sh" i get no issue

Interesting. Not sure if I understand that, but I've now added a GHA test without OpenMP for Ubuntu as well, so at least we should be safe now.

src/include/stir/recon_buildblock/distributable.txx
```diff
@@ -1,6 +1,7 @@
 /*
     Copyright (C) 2024 University College London
     Copyright (C) 2020, 2022, Univeristy of Pennsylvania
+    Copyright (C) 2025, Commonweath Scientific and Industrial Research Organisation
```
KrisThielemans, line 4:
```suggestion
    Copyright (C) 2025, Commonwealth Scientific and Industrial Research Organisation
```

KrisThielemans:
@ashgillman I've now simplified the code. If it all works, I suggest a squash-merge for this PR, as we went back and forth a bit.

Thanks for finding the bug!

KrisThielemans:
I've had to fix my code again, so decided to clean-up history, keeping your and my updates, so a merge seems best.

fingers crossed...

KrisThielemans (approved):

KrisThielemans:
@ashgillman please check final code. Merge (my) tomorrow!

KrisThielemans:
AppVeyor fails due to unrelated problem, see #1570.

KrisThielemans:
@danieldeidda would you be able to check this PR both with and without OpenMP? (Please run `run_tests_listmode_recon.sh` and `ctest`).

ashgillman:
Apologies, I haven't looked back over this thread since Friday.

If you can hold off, I will run it tomorrow. If you need to merge ASAP, I see no issue given the build runs.

KrisThielemans:
no hurry. Let me know when done.

ashgillman:
Yes! Both ctest and recon_test_pack/run_tests.sh run without reporting any error!

PR #1573, opened 2025-03-08
enable branching ratio in BinNormalisationWithCalibration

KrisThielemans:
We were not using the radionuclide in the exam_info used for set_up, but an (uninitialised) class radionuclide class member, therefore always defaulting to 1 (afer writing a warning).

However, this fix means that images will now have a different scale.

Also, remove the get/set_branching_ratio members, as they shouldn't really be part of the normalisation object.

This therefore breaks backwards-compatibility!

(There are currently still some typos here.)

***This branch will undergo forced pushes, so be careful where you merge it!***


danieldeidda:
I also just realised that if there is a calibration factor in the header this is not applied (SPECT)

PR #1577, opened 2025-03-16
SPECTUB clean-up

KrisThielemans:
- make `wmh` and some other arguments to functions `const`
- remove some functions that are not used in STIR
- replace `boost::constants::pi` with STIR's `_PI`
- remove unused include files
- replace math.h with cmath (needing some extra care, as gcc seems to miss `std::cosf` etc)
- remove some commented-out code

There should be no functionality change here.

Note though that as the SPECTUB code contains structs with pointers, having a `const` struct isn't very definite (as whatever is being pointed to can still be changed).


KrisThielemans:
@mastergari @danieldeidda There are a bit too many changes here, but if you ignore all the "eliminate dead code" stuff, it should be harmless....

KrisThielemans:
@jchan192, feel like doing a recon with this PR and see if it changes anything? (it shouldn't!)

KrisThielemans:
I'm going to merge this as nobody seems to have time for further testing.

Issue #1089, opened 2022-10-27
Reading SPECT DICOM raw data


lukepolson:
Hi all,

I'm trying to open projection data stored in a DICOM file. I've established a conda environment using:

https://stir.sourceforge.net/wiki/index.php/Installing_STIR_with_conda

When I try to open the projection data from the DICOM file using STIR in python, I get an error:
`projdata = stir.ProjData.read_from_file('test.dcm')`

`
ERROR: 
ProjData::read_from_file could not read projection data test.dcm.
Unsupported file format? Aborting.
`

I see in the user manual that if one compiles STIR so that it can "see" the ITK library, then one is able to open DICOM files. Is this conda environment compiled so that it can see ITK, and if not, is there another image which is?

In summary: how can I open a DICOM file of projection data using the python STIR package?

EDIT: Running 

`stir.FloatVoxelsOnCartesianGrid_read_from_file('test.dcm')`

works, but the data is not on a cartesian grid, since it represents projection data. Perhaps there's some fields missing, or a way to convert from `FloatVoxelsOnCartesianGrid` to `ProjData`?

KrisThielemans:
Hi Luke

at present, ITK is only used to read images. The conda version of STIR should therefore be able to read DICOM images, but sadly not SPECT DICOM acquisition data.

@danieldeidda you have this functionality somewhere. Is this something that we can share?

danieldeidda:
Hi, 
I do have this functionality but as part of  a private repo. I can however, make a new STIR utility and a PR

danieldeidda:
The functionality is now here #1182 feel free to try it on

KrisThielemans:
#1282 is now merged. Feedback welcome

smanwell:
I am currently trying to make use of the SPECT_dicom_to_interfile utilitiy with a projection dataset where there are two detector heads that are separated by 180 degrees with respect to one another and I have noticed a few (what I think are...) issues.

he SPECT_dicom_to_interfile tool looks at the DICOM Attribute "Number of Frames in Rotation" (0x0054, 0x0053) to determine the number of projections contains in the dataset. For this particular dataset that value is 60. For this dataset, however, there are two items within its Detector Information Sequence corresponding the projection images acquired by its two detector heads. Should the total number of projections should not be 120 (i.e. 60 per detector head)?

This dataset also contains the projection data for lower and upper scatter windows. Running this utlility as is, triggers the following warning message

[...]

PR #1579, opened 2025-03-26
Enhanced support multi head spect projection data

### LINKED ISSUE #1184: SPECT projection DICOM: support multiple heads
Currently we're getting
```
WARNING: num_projections (60) * num_energy_windows (2) != num_frames (240)
```
The header written is essentially only for the first head, i.e.
```
!number of projections := 60
!extent of rotation := 180
!process status := acquired
!SPECT STUDY (acquired data) :=
!direction of rotation := CCW
start angle := 180.041
orbit := non-circular
radii := {60 numbers here}
```
At present, we have to manually adjust the header
- double number of projections and extent of rotation
- append radii from relevant dicom field

Resolving this is relatively complicated, also because our Interfile support is incomplete #1102

smanwell:
This changes is related to #1184: SPECT projection DICOM: support multiple heads.
Changes have been introduced in the SPECT_dicom_to_interfile utility to accommodate such datasets.
Relevant details have been described [here](https://github.com/UCL/STIR/issues/1089#issuecomment-1898882582).

Additional changes include: 
- support for decompressing pixel data, if required (using gdcm API, no new dependencies introduced).
- removal of unused variable and commented code

## Testing performed
The code was executed on several multi-head SPECT projection datasets, including those with and without compressed pixel data.
White-box testing was performed to confirm that exception handling was working as expected in the function: _SPECTDICOMData::get_proj_data(const std::string& output_file_).

## Related issues
This change fixes #1184.

  - [X] The code builds and runs on my machine (built on Visual Studio 2022, Windows 11 OS)

### DIFF
--- a/src/utilities/SPECT_dicom_to_interfile.cxx
+++ b/src/utilities/SPECT_dicom_to_interfile.cxx
@@ -32,6 +32,7 @@
 #include "stir/error.h"
 #include "stir/warning.h"
 #include "stir/Succeeded.h"
+#include "stir/IO/Interfileheader.h"
 
 enum class EnergyWindowInfo
 {
@@ -81,7 +82,7 @@ private:
   int actual_frame_duration = 0; // frame duration in msec
   int num_of_rotations = 0;
   std::string direction_of_rotation, isotope_name;
-  int extent_of_rotation;
+  float extent_of_rotation;
   float calibration_factor;
   std::string rotation_radius;
 
@@ -382,7 +383,7 @@ SPECTDICOMData::open_dicom_file(bool is_planar)
   std::string no_of_det_as_str;
   std::string start_angle_as_string;
   std::string angular_step_as_string;
-  std::string extent_of_rotation_as_string;
+  std::string scan_arc_as_string;
   std::string radius_as_string;
   std::string actual_frame_duration_as_string;
   std::string calib_factor_as_string;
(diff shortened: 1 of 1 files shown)


KrisThielemans (changes requested):
some code clean-up. I haven't checked actual logic...

danieldeidda:
BTW @smanwell  I notice recently with some siemens data that was acquired with energy windows for scatter estimation, that every other sinogram is saved as if there was a  change of rotation orientation. Did you have the same issue?

src/utilities/SPECT_dicom_to_interfile.cxx
```diff
@@ -306,9 +388,10 @@ SPECTDICOMData::open_dicom_file()
       std::cout << "Patient name: " << patient_name << std::endl;
     }
   std::string no_of_proj_as_str;
+  std::string no_of_det_as_str;
   std::string start_angle_as_string;
   std::string angular_step_as_string;
-  std::string extent_of_rotation_as_string;
+  std::string scan_arc_as_string;
```
danieldeidda, line 394: why the change of name? isn't the old one the same in the DICOM tag?

smanwell:
@KrisThielemans Thanks for your feedback, I'll incorporate your suggestions as soon as I can return my focus to this project.

smanwell:
@danieldeidda I'm not quite sure what you mean. Is it possible to provide a screen capture of some of the DICOM fields that are involved?

src/utilities/SPECT_dicom_to_interfile.cxx
```diff
@@ -379,59 +460,96 @@ SPECTDICOMData::open_dicom_file()
...
+
+      // Get the radial positions of the detector head(s) for all heads.
+      if (GetDetectorInfo(file, DetectorInfo::RadialPosition, radius_as_string) == stir::Succeeded::yes)
+        {
+          rotation_radius = radius_as_string;
+          std::cout << "Radial Positions: " << rotation_radius << std::endl;
+        }
+
+      if (GetDICOMTagInfo(file, gdcm::Tag(0x0018, 0x1143), extent_of_rotation_as_string) == stir::Succeeded::yes)
+        {
```
danieldeidda, line 537: is this not a repetition then, you have it already in scan arc

src/utilities/SPECT_dicom_to_interfile.cxx
```diff
@@ -379,59 +460,96 @@ SPECTDICOMData::open_dicom_file()
...
+      if (GetDICOMTagInfo(file, gdcm::Tag(0x0018, 0x1143), scan_arc_as_string) == stir::Succeeded::yes)
+        {
+          scan_arc = std::stof(scan_arc_as_string);
+          std::cout << "Scan Arc: " << std::fixed << std::setprecision(6) << scan_arc << std::endl;
+        }
+
+      // Assert that the factors were valid before calculating the extent of rotation.
+      if (num_of_detectors > 0 && scan_arc > 0)
+        {
+          extent_of_rotation = num_of_detectors * scan_arc;
```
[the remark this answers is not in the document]
smanwell: Before I made changes to this code, the extent of rotation was equated to the DICOM scan arc. For the dual-head datasets that I had used for testing, the scan arc (and thus the extent of rotation) was simply 180 degrees. As a result, all of the projection images were interpreted as belonging to erroneous angular positions and the reconstructed had severe artifacts.

Take a look at a dataset with multiple (dual) heads. If, for example, the detectors are separated by 180 deg wrt each other, each will have a scan arc of 180 deg. But the full extent of the rotation is given as the product of the number of detectors with the scan arc. For a single headed system, then yes I'd expect the scan arc to be 360 deg and equal to the extent of rotation.

If there were three heads and each was positioned at equal angles wrt each other - 120 deg. the full extent of rotation, once again, would be given as scan arc * number of detectors.

I've not come across cases with more complex detector arrangements, so I haven't seen if my change holds for all cases.

PR #1580, opened 2025-03-28
[CMake] fix HDF5 version number in STIRConfig.cmake

### LINKED ISSUE #1557: exported STIRConfig.cmake can have problems with "too much" version info
On @categascoigne 's Ubuntu 22.04, we followed https://github.com/SyneRBI/SIRF/wiki/Building-SIRF-and-CIL-with-conda. This gave an error when configuring SIRF  where CMake complained about the line generated from
https://github.com/UCL/STIR/blob/7508babaf5e8147a3563f1642a47301af4c0834d/src/cmake/STIRConfig.cmake.in#L102
```
CMake Error at /home/cate/devel/buildConda/INSTALL/lib/cmake/STIR-6.2/STIRConfig.cmake:128 (find_package): 
  find_package called with invalid argument "1.14.4-3"
```
CMake was happy once we hand-edited the (exported) `STIRConfig.cmake` to use version 1.14.4.

We will therefore have to trim the version keyword(s).

KrisThielemans:
We now only require version major.minor for HDF5, as `HDF5_VERSION` could otherwise contain extra patch info, causing problems with `find_package(STIR)`.

Fixes #1557

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -290,6 +290,8 @@ endforeach()
 set(STIR_REGISTRIES ${INSTALLED_STIR_REGISTRIES})
 
 # New create STIRConfig.cmake
+# First strip patch info from HDF5_VERSION, as on some systems it might be set as x.y.z-something, which causes problems for find_package.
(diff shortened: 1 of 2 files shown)


KrisThielemans:
One job fails due to failure to download ROOT.

PR #1588, opened 2025-04-11
Trying to provide access to STIR data via a Python array

evgueni-ovtchinnikov:

### DIFF
--- a/src/include/stir/Array.inl
+++ b/src/include/stir/Array.inl
@@ -269,8 +269,8 @@ elemT*
 Array<num_dimensions, elemT>::get_full_data_ptr()
 {
   this->_full_pointer_access = true;
-//  if (!this->is_contiguous())
-//    error("Array::get_full_data_ptr() called for non-contiguous array.");
+  //  if (!this->is_contiguous())
+  //    error("Array::get_full_data_ptr() called for non-contiguous array.");
   return &(*this->begin_all());
 };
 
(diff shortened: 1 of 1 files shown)


KrisThielemans:
No, this cannot work. The check is there to make sure that access to the pointer is safe. If the data is not contiguous, accessing data through the pointer will cause segfaults.

The question is why the data is not contiguous in the first place. There is #1468. Could you add your use-case in that issue?

evgueni-ovtchinnikov:
@KrisThielemans @casperdcl I applied my _temporary_ quick-fix so that Casper could try to use in Python the `C++` address returned by `ImageData.address`.

@casperdcl this is a simple test I did myself (having applied my quick-fix):

in `stir_data_containers.h` I replaced
```
		size_t address() {
		    return reinterpret_cast<size_t>(_data->get_full_data_ptr());
		}
```
with
```
		size_t address() {
		    float* ptr = _data->get_full_data_ptr();
		    std::cout << (size_t)ptr << '\n';
		    std::cout << *ptr << '\n';
		    std::cout << *(ptr + 1) << '\n';
		    return reinterpret_cast<size_t>(ptr);
		}
```
and in `PET/acquisition_data.py` I added
```
    img_data = acq_data.create_uniform_image(2.0)
    ptr = img_data.address
    print(ptr)
    img_arr = img_data.as_array()
    print(img_arr[0,0,0])
    print(img_arr[0,0,1])
    exit()
```
after line 83. The output from C++ and Python coincided (@casperdcl better fill `img_data` with 1,2,3,... and check the contiguity of the data directly in `address()` - I was too tired to do this on Friday evening, sorry).

Issue #1590, opened 2025-04-14
Replace boost::format with fmt::format/std::format


markus-jehl:
When building STIR with CUDA support, I got obscure build errors linked to boost functions. Removing the boost/format.h dependency in the FactoryRegistry fixed the build. This lead to an investigation for potential replacements for the boost::format calls.

fmt is a very lean and fast alternative that has made it into the standard library for C++20 and has a very similar function definition for fmt::format/std::format, so this is the obvious candidate.

Implementation questions:
-  create a stir/print.h that chooses between fmt and std versions, like we do/did for shared_ptr, such that we don’t have to replace all those fmt’s to std in a couple of years?

- not so sure about adding another dependency though, especially as it’s temporary. Could use either a submodule and put it in include/ or directly commit files (but the latter might get into trouble with clang-format reformatting things).

-  do this for boost::format first, and then run clang-tidy as a separate commit (such that we can add it to .git- blame-ignore-revs).

The build error message was:

[ 49%] Building CUDA object src/recon_buildblock/CMakeFiles/recon_buildblock.dir/CUDA/CudaRelativeDifferencePrior.cu.o
cd /workspace/stir-public/build/src/recon_buildblock && /usr/local/cuda/bin/nvcc -forward-unknown-to-host-compiler -ccbin=/usr/bin/clang++ -DITK_IO_FACTORY_REGISTER_MANAGER --options-file CMakeFiles/recon_buildblock.dir/includes_CUDA.rsp -O3 -DNDEBUG -std=c++17 -arch=all -Xcompiler=-fPIC   -fopenmp=libomp -MD -MT src/recon_buildblock/CMakeFiles/recon_buildblock.dir/CUDA/CudaRelativeDifferencePrior.cu.o -MF CMakeFiles/recon_buildblock.dir/CUDA/CudaRelativeDifferencePrior.cu.o.d -x cu -c /workspace/stir-public/STIR/src/recon_buildblock/CUDA/CudaRelativeDifferencePrior.cu -o CMakeFiles/recon_buildblock.dir/CUDA/CudaRelativeDifferencePrior.cu.o
/usr/include/boost/type_traits/is_base_and_derived.hpp(142): error: incomplete type is not allowed
          detected during:
            instantiation of class "boost::detail::is_base_and_derived_impl2<B, D> [with B=boost::in_place_factory_base, D=std::locale]"
(217): here
[...]

PR #1591, opened 2025-04-15
Add {fmt} library as a git submodule and replace calls to boost::format

markus-jehl:

- added {fmt} library as a git submodule
- added stir/format.h to decide between using fmt::format or std::format, depending on C++ version
- TODO: removed all calls to boost::format
- TODO: used clang-tidy to replace sprintf with fmt::format

## Related issues
closes https://github.com/UCL/STIR/issues/1590


KrisThielemans:
Seems we'll need a `find_package(fmt)` in the `STIRConfig.cmake.in`, but this is of course rather undesirable, as now we have an extra external dependency anyway for anyone who wants to use STIR. Far better would be to make `fmt` a private dependency. I think this would work if we don't use it in src/include/stir/FactoryRegistry.inl.

src/include/stir/format.h
```diff
@@ -0,0 +1,49 @@
...
+
+  \author Markus Jehl
+  \author Kris Thielemans
+*/
+
+#include <string>
+#include <utility> // for std::forward
+#include "stir/common.h"
+
+#if defined(__cpp_lib_format) && (__cpp_lib_format >= 201907L)
```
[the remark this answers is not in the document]
markus-jehl: I don't think it is clang specific: https://en.cppreference.com/w/cpp/utility/format (see Notes section)

src/include/stir/FactoryRegistry.inl
```diff
@@ -58,9 +57,9 @@ FactoryRegistry<Key, Factory, Compare>::add_to_registry(const Key& key, Factory
   typename FactoryMap::iterator iter = m.find(key);
   if (iter != m.end())
     {
-      warning(boost::format("FactoryRegistry:: overwriting previous value of key in registry.\n"
-                            "     key: %1%")
-              % key);
+      std::ostringstream s;
```
KrisThielemans, line 60: will need to include `sstream`

KrisThielemans:
CMake targets are useful, but a pain...

I have no idea why still fails after including it in `STIRConfig.cmake`. I see that the log doesn't say it found fmt (while it does say it found boost). A bug in fmt?

Potentially we could
- use fmt's `FMT_HEADER_ONLY` and use the work-around from https://github.com/UCL/STIR/blob/4e57186eec3887c181d00034dd9ea4d71a401e96/src/IO/CMakeLists.txt#L121-L129 (although this could be somewhat outdated according to the CMake issue). This work-around is ugly of course, and as we'd need to repeat it many times, maybe it needs a macro.
- somehow make the target part of the STIRConfig.cmake.

One more of these very simple ideas that turns into a nightmare. Sigh.

CMakeLists.txt
```diff
@@ -104,6 +104,26 @@ option(DISABLE_Parallelproj_PROJECTOR "disable use of Parallelproj projector" OF
...
+                        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                        RESULT_VARIABLE GIT_SUBMOD_RESULT)
+        if(NOT GIT_SUBMOD_RESULT EQUAL "0")
+            message(FATAL_ERROR "git submodule update --init --recursive failed with ${GIT_SUBMOD_RESULT}, please checkout submodules")
+        endif()
+    endif()
+endif()
+
+if(NOT EXISTS "${PROJECT_SOURCE_DIR}/external_helpers/CMakeLists.txt")
+    message(FATAL_ERROR "The submodules were not downloaded! GIT_SUBMODULE was turned off or failed. Please update submodules and try again.")
```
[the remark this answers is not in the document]
markus-jehl: good spot! I'll just point it to fmt for now, as that is the only submodule currently.

markus-jehl:
Calling find_package(fmt) earlier on in STIRConfig.cmake.in fixed the issue! Now an installed STIR can be used without any changes.

Not sure what the Codacy issue is about - do you have any experience with this @KrisThielemans ?

markus-jehl:
Also, I seem to have gotten appveyor into a weird state :-/

markus-jehl:
The clang-tidy printf and fprintf replacement is a bit underwhelming so far - only picked up a few instances and skipped many others (maybe because I don't build some code). But it also turns out that it doesn't replace sprintf.

KrisThielemans:
> Not sure what the Codacy issue is about

it complains about
```
return internal_format::format(fmt, std::forward<Args>(args)...);
```
with "Comparison of a boolean expression with an integer." Doesn't have anything to do with us, so let's ignore it.

documentation/release_6.3.htm
```diff
@@ -132,6 +132,13 @@ <h3>Changed functionality</h3>
     a <code>shared_ptr&lt;const ProjDataInfo&gt;</code>.<br>
     <a href=https://github.com/UCL/STIR/pull/1315>PR #1315</a>
   </li>
+  <li>
+    As <code>stir::Array</code> template arguments might change in the future,
+    it is now recommended to include <tt>stir/ArrayFwd.h</tt> when using forward
+    declaration and use the <code>ArrayType</code> template-alias in places where
+    a rectangular array that might live on a GPU is intended.<br>
+    <a href=https://github.com/UCL/STIR/pull/1589>PR #1589</a>
+  </li>
```
KrisThielemans, line 141: why does this appear as a change in this PR? something wrong with the rebasing?

src/IO/GIPL_ImageFormat.cxx
```diff
@@ -129,7 +130,7 @@ Image::~Image()
 void
 Image::GiplRead(char* filename)
 {
-  printf("Read %s\n", filename);
+  fmt::print("Read {}", filename);
```
KrisThielemans, line 133: would we have to use `internal_format::print` for these types of statements?

KrisThielemans:
Also, ideally we keep the clang-tidy changes in a separate commit, which we then add to `.git-blame-ignore-revs`. sorry

src/IO/GIPL_ImageFormat.cxx
markus-jehl: good point, yes we should

documentation/release_6.3.htm
markus-jehl: hmm... weird. Let me look into it.

recon_test_pack/run_tests_modelling.sh
```diff
@@ -126,7 +126,7 @@ generate_image ${INPUTDIR}generate_p5.par
 fi
 assemble_images p0005-p5.${imgext} p0005.hv p5.hv
 # test
-extract_single_images_from_parametric_image p0005-p5_%d.hv p0005-p5.${imgext}
+extract_single_images_from_parametric_image p0005-p5_{}.hv p0005-p5.${imgext}
```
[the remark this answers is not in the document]
markus-jehl: hmm... I assumed it works since the tests passed. But when I run it locally I get an error. I'll also have a look at this.

markus-jehl:
> > Also, I seem to have gotten appveyor into a weird state :-/
> 
> It reports
> 
> ```
> CMake Error at CMakeLists.txt:123 (message):
>   The {fmt} submodule was not downloaded! GIT_SUBMODULE was turned off or
>   failed.  Please update submodules and try again.
> ```
> 
> this seems to say that your attempt to auto-update the module fails. Not sure why... I'm not sure if we should do this though. I think I'd be surprised if building my software suddenly updates submodules. I think I'd prefer to give explicit instructions how to clone, as well as an error message that tells them what to do (i.e. either use C++-20, or use the explicit git command). In that case, we'll have to modify `appveyor.yml`. https://www.appveyor.com/docs/how-to/private-git-sub-modules/ suggests to add
> 
> ```
> install:
>   - git submodule update --init --recursive
> ```
> 
> [appveyor/ci#899](https://github.com/appveyor/ci/issues/899) claims this doesn't work well with caching and gives a far more complicated solution. Maybe we can ignore that for a while.

Unfortunately, this does not work because we currently do a shallow clone of the repository:
https://www.appveyor.com/docs/how-to/repository-shallow-clone/

Do you think we could switch to specifying a clone depth?

KrisThielemans:
> Do you think we could switch to specifying a clone depth?

sure

markus-jehl:
Even with STIR master I get the following error when I run "./run_tests_modelling.sh" from within STIR/recon_test_pack. Am I doing something wrong or is this test currently broken?

KrisThielemans:
weird. it works for me. Indeed, it works in the Actions. see for instance
https://github.com/UCL/STIR/actions/runs/14836231076/job/41648053623?pr=1591#step:17:1945

KrisThielemans:
> Bad news: fmt definitely doesn't support the "%d" syntax, so for input files we'll have to revert to boost if we want to be backward compatible.
> 

sigh. what about the following then: if the string contains a %, write a warning recommending to use the new syntax and fall-back to `boost::format`, otherwise use `fmt`. That will allow us to deprecate it later. If you have linking problems with `boost::format`, we could do an "replace %d with {} and hope for the best".

> Also, I managed to debug into the problematic line in run_tests_modelling.sh and the problem is the "if(*p)". I notice that in my case the character array for the first line shows as "3655\r", so *p interprets to "\r". Turned out to be caused by line endings: my file had CRLF line endings, and when changing the line endings for plasma.if to LF, then reading the file works fine.

That's pretty weird. and `git status` shows no change in the file? what does the following say?
```
git check-attr -a -- test_modelling_input/plasma.if
```
(in my case: nothing).

I checked that on Ubuntu, my file has Unix EOL (i.e. \n), while on Windows it has DOS EOL (i.e. \r \n). Somehow you have a mixture. Maybe this is affected by a git switch for preference?

Can you check another text file, e.g.
```
$ od -c test_modelling_input/ECAT_931_projdata_template.hs|head
0000000   !   I   N   T   E   R   F   I   L   E           :   =  \n   n
0000020   a   m   e       o   f       d   a   t   a       f   i   l   e
...
```
Of course, ideally we'd rewrite the code to cope with "DOS EOL on Unix" (`KeyParser` does this), but I doubt you feel like doing that!


> Afterwards I also had to change the "for fr in `count 23 28`"-style loops to "for fr in $(seq 23 28)", since count does not seem to be supported in ubuntu. Now it runs through fine.

`count` is a STIR script that should have been installed and in your path. Of course, your mod is better (as the script uses `bash` already), but there's bound to be some other occurences of `count` elsewhere, so I wouldn't commit that change.

markus-jehl:
That's a good idea! I'll do that.

Strangely, all files have CRLF line endings in my case, but none show up as changed in git. It must be something strange with how I cloned the repo - I'll look into it.

Ah ok, I would have never guessed that count is a STIR script! Where is the source code for it? The usage in the shell script gives me a ": invalid option" error.

KrisThielemans:
> Strangely, all files have CRLF line endings in my case, but none show up as changed in git. It must be something strange with how I cloned the repo - I'll look into it.
> 

`git config --global core.autocrlf input` and all that, I guess

> Ah ok, I would have never guessed that count is a STIR script! Where is the source code for it? The usage in the shell script gives me a ": invalid option" error.

https://github.com/UCL/STIR/blob/master/scripts/count

My guess this is also a CRLF thing. I think we need to add it in .gitattributes (`*.sh` files are already taken care of)

.gitattributes
```diff
@@ -2,6 +2,7 @@
 [attr]cppfiles
 [attr]cfiles
 [attr]pyfiles
+* text eol=lf
```
KrisThielemans: hmmm. that's potentially a lot of them. any way to list what files this affects?

KrisThielemans: I'd prefer something like `scripts/*` if that's possible, or a `.gitattributes` inside the `scripts` folder.

KrisThielemans: @markus-jehl any chance you can look into this? It'd be good to merge this soon.

KrisThielemans (changes requested):

.gitattributes
markus-jehl: sure, let's start with scripts/* and add others if the issue appears elsewhere

.gitattributes
```diff
@@ -16,7 +17,6 @@
 *.root  -text -diff
 *.gz     -text -diff
 *.nii     -text -diff
-*.sh    eol=lf
```
KrisThielemans: let's keep this. There's some scripts in examples, SimSET etc as well.

Please commit with `[ci skip]`

KrisThielemans:
sorry, could you update `release-6.3.htm` as well? Or even add minimal info to the STIR developer's guide (if appropriate). Commit again with `[ci skip]'. Thanks!

PR #1596, opened 2025-05-21
Create SPECT subsets from ProjData

samdporter:
Add ability to create SPECT subsets with partitioner (as separate ProjData)

## Testing performed
None, as yet. Tests require template data (I'll sort this soon)

## Related issues
Fixes segfaults arising from using partitioner on SPECT ProjData


examples/samples/SPECT_Interfile_header.hs
```diff
@@ -5,7 +5,7 @@
 !imaging modality := nucmed
 
 ; name of file with binary data
-name of data file := somefile.s
+name of data file := SPECT_Interfile_header.hs
```
KrisThielemans, line 11:
```suggestion
; Note: we are "self-referring" to the header here such that this file can be read by STIR as a "template".
; (This is used in one of our tests.)
; Obviously, this would normally not be the case.
name of data file := SPECT_Interfile_header.hs
```

src/recon_buildblock/SPECTUB_Tools.cxx
```diff
@@ -30,6 +36,216 @@ namespace SPECTUB
 #define abs(a) ((a) >= 0 ? (a) : (-a))
 #define SIGN(a) (a < -EPSILON ? -1 : (a > EPSILON ? 1 : 0))
 
+#define DELIMITER1 '#' // delimiter character in input parameter text file
```
KrisThielemans: All of this was removed in 9a85f5471524051a3463698823bdc1a1cacfe485. Either you need to merge master on here, or you did some copy-paste.

KrisThielemans (changes requested):
3 small changes required.

Also, how different is src/test/test_proj_data_info_subsets_spect.cxx from PET? I didn't check, but it's likely most of it is the same. Easy to avoid overlap? (could derive both from one class)

samdporter:
Hmm... this worked on my local machine - segfaults here.

src/recon_buildblock/SPECTUB_Tools.cxx
```diff
@@ -13,6 +13,12 @@
 // system libraries
 #include <stdlib.h>
 #include <math.h>
+#include <boost/math/special_functions/fpclassify.hpp>
+#include "stir/error.h"
+#include "stir/stream.h"
+#include <boost/format.hpp>
+#include <boost/math/constants/constants.hpp>
+
```
KrisThielemans: can't see why we'd need these now. Probably a left-over.

src/test/test_proj_data_info_subsets_spect.cxx
```diff
@@ -0,0 +1,648 @@
...
+
+  // First we make a full forward projection with a reduced segment range
+  shared_ptr<ProjDataInfo> reduced_seg_range_pdi_sptr(template_sino_sptr->get_proj_data_info_sptr()->clone());
+
+  // Print the initial segment range
+  std::cerr << "Initial segment range: min_segment_num = " << reduced_seg_range_pdi_sptr->get_min_segment_num()
+            << ", max_segment_num = " << reduced_seg_range_pdi_sptr->get_max_segment_num() << std::endl;
+
+  // Attempt to reduce the segment range
+  reduced_seg_range_pdi_sptr->reduce_segment_range(-1, 1);
```
KrisThielemans: this will be where the assert is raised. Of course,
1) it should never have got here, which means that L453 is wrong. I've created #1627
2) this test doesn't make sense for SPECT

KrisThielemans (changes requested):

examples/samples/SPECT_Interfile_header.hs
samdporter, line 1: Looks like these comments are now duplicated

KrisThielemans: sorry. feel free to undo

PR #1610, opened 2025-08-22
Fixed stringstream overwriting rather than appending

bwynneHEP:
Fixed a bug in use of ostringstream. No functional changes, only output error messages made more legible.

## Testing performed
Deliberately created error condition, saw incorrect message before change and correct message after.


KrisThielemans (approved):
Looks good. Let's see what CI says. thanks a lot!

PR #1617, opened 2025-09-20
When computing sensitivity, avoid a superfluous ProjDataInMemory in most cases

### LINKED ISSUE #1614: unnecessary creation of ProjDataInMemory for sensitivity calculation
We currently do
https://github.com/UCL/STIR/blob/24f18bcd9ba541d64e8fccc2e36c32fa039b25d7/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L836-L838
However, we shouldn't really need to have this (potentially very large) object.

Checking `RPC_process_related_viewgrams_sensitivity_computation`
https://github.com/UCL/STIR/blob/24f18bcd9ba541d64e8fccc2e36c32fa039b25d7/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L1520-L1527
it seems to me that we can replace the `else` case with
```c++
measured_viewgrams_ptr->fill(1.F);
back_projector_sptr->back_project(*measured_viewgrams_ptr);
```
in which case we ignore the content of the "measured" projection data (which actually is the `sens_proj`).

So, in `distributable_sensitivity_computation`
https://github.com/UCL/STIR/blob/24f18bcd9ba541d64e8fccc2e36c32fa039b25d7/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L1392-L1393
we can set the "read proj data" argument to `false`, and therefore we can pass any projection data, including the measured one, as an argument.

Final check: this projection data needs to have the correct dimensions, as we do
https://github.com/UCL/STIR/blob/24f18bcd9ba541d64e8fccc2e36c32fa039b25d7/src/recon_buildblock/distributable.cxx#L184-L197

KrisThielemans:
Optimise code such that we only use a ProjDataInMemory when using TOF projectors and non-TOF projector for the sensitivity.

Fixes #1614

### DIFF
--- a/documentation/release_6.3.htm
+++ b/documentation/release_6.3.htm
@@ -87,6 +87,14 @@
     <li>
       Default ECAT scanner configurations updated to use a negative intrinsic tilt.
     </li>
+    <li>
+      When computing the sensitivity images in <code>PoissonLogLikelihoodWithLinearModelForMeanAndProjData</code>,
+      we now avoid creating an extra <code>ProjDataInMemory</code> in most cases (there are still some corner
+      cases for TOF data when using non-TOF projector for the sensitivity, but then the memory overhead is small).
+      This enables LAFOV PET reconstructions with the "ray tracing" matrix (we use less memory, and we avoid running into a current
+      <code>ProjDataInMemory</code> <a href=https://github.com/UCL/STIR/issues/1505> limitation on the number of bins</a>).<br>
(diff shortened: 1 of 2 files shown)


KrisThielemans:
Sorry @DANAJK ignore the  review request

NikEfth:
I have more on this. I will push in a week or so.

PR #1623, opened 2025-10-05
[GHA] test clang 18 and 21 on OSX

### LINKED ISSUE #1490: Clang 18 on MacOS tests fail due to uncaught (but intentional) exceptions

KrisThielemans:
Remove linking flags pointing to homebrew's libc++ as this is no longer recommended on the homebrew page https://formulae.brew.sh/formula/llvm. In addition, https://github.com/llvm/llvm-project/issues/94292#issuecomment-2149342607 says it causes problem.

Hopefully fixes #1490


KrisThielemans:
clang-18 compiled! Needed to upgrade {fmt} to fix #1624

PR #1625, opened 2025-10-06
fix clang -latomic problem for Linux AND MacOS

### LINKED ISSUE #1494: OSX Clang OpenMP fails on -latomic
To prevent problems on Linux clang, I inserted
https://github.com/UCL/STIR/blob/e03fbd03976abdd543724aa2e1c2315fbd71e115/src/CMakeLists.txt#L139-L142
However, this gives linking problems on OSX clang on conda in https://github.com/conda-forge/stir-feedstock/pull/121
```
ld: library not found for -latomic
```
Ideally we'd convert this to a link test.

See also https://github.com/llvm/llvm-project/issues/73361 for the upstream problem.

KrisThielemans:
Fixes #1494


KrisThielemans:
test commit first to see failure

Issue #1426, opened 2024-05-15
priors are actually penalties (naming problem)


KrisThielemans:
Our priors increase for "worse" images, i.e. the sign choice is different from what the name implies.

KrisThielemans:
https://github.com/UCL/STIR/pull/1629 introduced "penalty" in the new names

PR #1629, opened 2025-10-08
GPU gibbs priors refactoring

ColomboMatte0:
## Main changes:
See presentation:
[https://www.ccpsynerbi.ac.uk/wp-content/uploads/2025/09/GibbsPrior_presentation.pdf]

[GibbsPrior.pptx](https://github.com/user-attachments/files/22780798/GibbsPrior.pptx)

##Minor changes

- added SWIG bindings for the original version  CudaRelativedifferencePrior
- moved GeneralizedPrior.cxx into GeneralizedPrior.inl
- modified src/utilities/stir_timings.cxx to include timings with the new priors (i' ve created  also a python notebook used to generate the results of the presentation not included in this pull request, let me know if i should add it (where?) )

Please tick the following: 

 - [ x] The content of this Pull Request (the Contribution) is intentionally submitted for inclusion in STIR (the Work) under the terms and conditions of the [Apache-2.0](https://www.apache.org/licenses/LICENSE-2.0) License.
 - [ x] I (or my institution) have signed the STIR Contribution License Agreement (not required for small changes).


src/include/stir/cuda_utilities.cuh
```diff
@@ -64,6 +64,54 @@ array_to_host(Array<num_dimensions, elemT>& stir_array, const elemT* dev_data)
...
+// Generic, architecture-compatible atomicAdd for double and convertible types
+template <typename elemT>
+__device__ inline double
+atomicAddGeneric(double* address, elemT val)
+{
+  double dval = static_cast<double>(val);
+#if __CUDA_ARCH__ >= 600
+  return atomicAdd(address, dval);
+#else
+  // Use atomicCAS for double-precision atomicAdd on older architectures
```
KrisThielemans:
```suggestion
#  error Either ugrade your GPU to compute capability 6 or check the code at __FILE__, __LINE__
  // Use atomicCAS for double-precision atomicAdd on older architectures
```

src/utilities/stir_timings.cxx
```diff
@@ -281,17 +300,41 @@ class Timings : public TimedObject
...
-    delete im;
+
+    this->prior_sptr->compute_gradient(*this->output_sptr, *this->image_sptr);
+
   }
 
+  void prior_grad_times_input()
+  {
+    double result =this->prior_sptr->compute_gradient_times_input(*this->input_sptr, *this->image_sptr);
+    result += 2; // to avoid compiler warning about unused variable
```
KrisThielemans:
```suggestion
    this->prior_sptr->compute_gradient_times_input(*this->input_sptr, *this->image_sptr);
```
try this to avoid Codacy warning

KrisThielemans: or `static_cast<void>(this->...)`

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
```diff
@@ -32,6 +32,15 @@
...
 
+#ifndef __CUDACC__
+  #ifndef __host__
+    #define __host__
+  #endif
+  #ifndef __device__
+    #define __device__
+  #endif
+#endif
+
```
KrisThielemans: should this be somewhere generic? if we have no place, can leave it here.

ColomboMatte0: Can we do something about it direclty from  CmakeLists.txt? not sure exactly how to avoid putting this in each  **.h that needs to be compiled by both cpu and gpu.

src/utilities/stir_timings.cxx
ColomboMatte0: Yes it works `this->prior_sptr->compute_gradient_times_input(*this->input_sptr, *this->image_sptr);` like this i modify the method also for the compute value and commit.

src/include/stir/cuda_utilities.cuh
ColomboMatte0: Done.

ColomboMatte0: No, it’s not working like this. Since we’re trying to compile with -arch=all, it will throw an error at compile time no matter what. I think the easiest solution would be to just remove -arch=all from the compile flags, but I honestly have no idea where to change it.

src/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ option(BUILD_SHARED_LIBS
        "Use shared libraries" OFF)
 
 ### Settings for external libraries
+# Locate and include the CUDA Toolkit headers
```
KrisThielemans:
```suggestion
```

src/include/stir/cuda_utilities.cuh
KrisThielemans: I don't understand that, but also I feel this is a problem not worth spending time on. Let's just `#include "stir/error.h"` and replace this with `error("Either upgrade...");`, which will compile.

Also, use doxygen docstring, like this
```c++
   //! Generic, architecture-compatible atomicAdd for double and convertible types
   /*! 
     \ingroup CUDA
     \warning Currently calls error() if CUDA ....
  */
```
More worrying is if that means if the optimal code is never used, but I guess you'll find out quickly!

src/include/stir/cuda_utilities.h
```diff
@@ -64,6 +64,55 @@ array_to_host(Array<num_dimensions, elemT>& stir_array, const elemT* dev_data)
...
+      assumed = old;
+      double updated = __longlong_as_double(assumed) + dval;
+      old = atomicCAS(address_as_ull, assumed, __double_as_longlong(updated));
+  } while (assumed != old);
+
+  return __longlong_as_double(old);
+#endif
+}
+
+inline void
```
KrisThielemans, line 140: Ideally add 1 line doxygen (I know the others don't have one)

src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.cuh
```diff
@@ -0,0 +1,818 @@
...
+
+
+
+START_NAMESPACE_STIR
+
+// CUDA kernels for CudaGibbsPrior
+
+template <class elemT, typename PotentialT>
+__global__ static void
+Value_Kernel(double* output,
```
KrisThielemans: Best to call this `CudaGibbsPrior_value_kernel` (note STIR naming conventions for functions are `some_func`). Or we put this in a `CudaGibbsPriorKernels` namespace or so.

Ideally add a tiny doxygen comment.

src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.cuh
```diff
@@ -0,0 +1,818 @@
...
+  __syncthreads();
+
+  //block pair-wise reduction in shared memory
+  int block_threads = blockDim.x * blockDim.y * blockDim.z;
+  blockReduction(block_sum_shared, thread_in_block, block_threads);
+
+  // One thread per block performs final atomic add in double
+  if (thread_in_block == 0)
+      //TODO atomicAddGeneric is defined in include/stircuda_utilities.cuh, we need it since for CUDA_ARCH__ < 600 
+      //we cannot perform atomicAdd on double. AtomicAddGeneric has not been tested for CUDA_ARCH__ < 600.
```
KrisThielemans: I don't see why we need this comment here (it belongs in the doxygen of the function)

src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.cuh
```diff
@@ -0,0 +1,818 @@
...
+  //Local sync on the threads in the block
+  __syncthreads();
+
+  //block pair-wise reduction in shared memory
+  int block_threads = blockDim.x * blockDim.y * blockDim.z;
+  blockReduction(block_sum_shared, thread_in_block, block_threads);
+
+  if (thread_in_block == 0)
+      //TODO atomicAddGeneric is defined in include/stircuda_utilities.cuh, we need it since for CUDA_ARCH__ < 600 
+      //we cannot perform atomicAdd on double. AtomicAddGeneric has not been tested for CUDA_ARCH__ < 600.
```
KrisThielemans: remove here as well

src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.cuh
```diff
@@ -0,0 +1,818 @@
...
+                current *= kappa[inputIndex] * kappa[neighbourIndex];
+              
+              sum += current;
+            }
+  gradient[inputIndex] = sum * 2.0 * penalisation_factor;
+}
+
+template <class elemT, typename PotentialT>
+__global__ static void
+Gradient_dot_input_Kernel(double* output,
```
KrisThielemans: `CudaGibbsPrior_gradient_dot_input_kernel`

src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.cuh
```diff
@@ -0,0 +1,818 @@
...
+  if (thread_in_block == 0)
+      //TODO atomicAddGeneric is defined in include/stircuda_utilities.cuh, we need it since for CUDA_ARCH__ < 600 
+      //we cannot perform atomicAdd on double. AtomicAddGeneric has not been tested for CUDA_ARCH__ < 600.
+      atomicAddGeneric(output, block_sum_shared[0]);
+
+}
+
+template <class elemT, typename PotentialT>
+__global__ static void
+Hessian_diagonal_Kernel(elemT* Hessian_diag,
```
KrisThielemans: CudaGibbsPrior_Hessian_diagonal_kernel`

src/include/stir/recon_buildblock/CUDA/CudaRelativeDifferencePrior.h
```diff
@@ -21,7 +21,11 @@
 
 #include "stir/recon_buildblock/RelativeDifferencePrior.h"
 
+#ifdef SWIG
+namespace stir {
+#else
 START_NAMESPACE_STIR
+#endif
```
KrisThielemans: unlikely this is needed. We pass this define to the call to swig in the CMakeLists.txt. We'd need it everywhere else as well.

src/include/stir/recon_buildblock/GeneralisedPrior.h
```diff
@@ -63,6 +64,8 @@ class GeneralisedPrior : public RegisteredObject<GeneralisedPrior<DataT>>
       Note that a row corresponds to an object of `DataT`.
       The method (as implemented in derived classes) should store the result in \c prior_Hessian_for_single_densel.
    */
+  virtual double compute_gradient_times_input(const DataT& input, const DataT& current_estimate);
```
KrisThielemans: move before doxygen comment, and add one

src/include/stir/recon_buildblock/GeneralisedPrior.h
```diff
@@ -74,6 +77,8 @@ class GeneralisedPrior : public RegisteredObject<GeneralisedPrior<DataT>>
       Instead, accumulate_Hessian_times_input() should be used. This method remains for backwards comparability.
        \warning The derived class should accumulate in \a output.
   */
+  virtual void compute_Hessian_diagonal(DataT& Hessian_diagonal, const DataT& current_estimate) const;
```
KrisThielemans: doxygen comment

src/include/stir/recon_buildblock/GibbsPrior.inl
```diff
@@ -0,0 +1,594 @@
...
+    : only_2D(only_2D_v)
+{
+  this->penalisation_factor = penalisation_factor_v;
+}
+
+template <typename elemT, typename PotentialT>
+bool
+GibbsPrior<elemT,PotentialT>::is_convex() const
+{
+  return true;
```
KrisThielemans: hmmm. shouldn't we forward this to the patential function?

src/include/stir/recon_buildblock/GibbsPrior.inl
```diff
@@ -0,0 +1,594 @@
...
+    error("GibbsPrior: set_up has not been called");
+  if (this->penalisation_factor == 0)
+    {
+      prior_gradient.fill(0);
+      return;
+    }
+    
+  const bool do_kappa = !is_null_ptr(kappa_ptr);
+
+  #pragma omp parallel for collapse(3)
```
KrisThielemans: more than likely, we'll need to check openMP version here. MSVC (used on Appveyor and conda) is stuck with OpenMP 2... if you see errors, do a search for collapse to copy some code

src/include/stir/recon_buildblock/GibbsPrior.h
```diff
@@ -0,0 +1,167 @@
...
+  virtual void set_kappa_sptr(const shared_ptr<const DiscretisedDensity<3, elemT>>&);
+
+  //! Set up the prior for a target image. Must be called before use.
+  Succeeded set_up(shared_ptr<const DiscretisedDensity<3, elemT>> const& target_sptr) override;
+
+protected:
+  //! Image and weight boundary indices for efficient and safe neighborhood looping.
+  CartesianCoordinate3D<int> Image_dim;
+  CartesianCoordinate3D<int> Image_max_indices;
+  CartesianCoordinate3D<int> Image_min_indices;
```
KrisThielemans: no capitals (`image_dim` etc). Follow STIR naming conventions in the developer's guide.

src/include/stir/recon_buildblock/GibbsPrior.h
```diff
@@ -0,0 +1,167 @@
...
+  shared_ptr<const DiscretisedDensity<3, elemT>> get_kappa_sptr() const;
+
+  //! Set the kappa image (spatially-varying penalty factors).
+  virtual void set_kappa_sptr(const shared_ptr<const DiscretisedDensity<3, elemT>>&);
+
+  //! Set up the prior for a target image. Must be called before use.
+  Succeeded set_up(shared_ptr<const DiscretisedDensity<3, elemT>> const& target_sptr) override;
+
+protected:
+  //! Image and weight boundary indices for efficient and safe neighborhood looping.
```
KrisThielemans: sadly needs a bit more for doxygen grouping, see https://www.doxygen.nl/manual/grouping.html
/*! \name indices 
    Image and weight boundary indices for efficient and safe neighborhood looping.
*/
//@{
Cartesian...
Cartesian ...
...
//@}
```

File: src/include/stir/recon_buildblock/GibbsPrior.inl [hunk 18]
```diff
@@ -0,0 +1,594 @@
...
+#include "stir/BasicCoordinate.h"
+#include "stir/IO/read_from_file.h"
+#include "stir/is_null_ptr.h"
+#include "stir/info.h"
+#include "stir/warning.h"
+#include "stir/error.h"
+#include <algorithm>
+#include <omp.h>
+using std::min;
+using std::max;
```
KrisThielemans: no `using` in .h or .inl, as then everyone has them

src/include/stir/recon_buildblock/GibbsPrior.inl
```diff
@@ -0,0 +1,594 @@
...
+#include "stir/IndexRange3D.h"
+#include "stir/IO/write_to_file.h"
+#include "stir/BasicCoordinate.h"
+#include "stir/IO/read_from_file.h"
+#include "stir/is_null_ptr.h"
+#include "stir/info.h"
+#include "stir/warning.h"
+#include "stir/error.h"
+#include <algorithm>
+#include <omp.h>
```
KrisThielemans: needs to be between `#ifdef STIR_OPENMP`

src/include/stir/recon_buildblock/GibbsPrior.inl
```diff
@@ -0,0 +1,594 @@
...
+  this->check(current_image_estimate);
+  if (this->_already_set_up == false)
+    error("GibbsPrior: set_up has not been called");
+  if (this->penalisation_factor == 0)
+    return 0.;
+    
+  const bool do_kappa = !is_null_ptr(kappa_ptr);
+
+  double result = 0.0;
+  #pragma omp parallel for collapse(3) reduction(+:result)
```
KrisThielemans: needs to be between `#ifdef STIR_OPENMP` and some version checks (see other comment)

src/include/stir/recon_buildblock/GibbsQuadraticPrior.h
```diff
@@ -0,0 +1,187 @@
...
+    SPDX-License-Identifier: Apache-2.0
+
+    See STIR/LICENSE.txt for details
+*/
+/*!
+  \file
+  \ingroup priors
+  \ingroup CUDA
+  \brief Declaration of class stir::GibbsQuadraticPrior, stir::CudaGibbsQuadraticPrior
+  * and the potential function stir::QuadraticPotential
```
KrisThielemans:
```suggestion
  and the potential function stir::QuadraticPotential
```

src/include/stir/recon_buildblock/GibbsQuadraticPrior.h
```diff
@@ -0,0 +1,187 @@
...
+  #endif
+  #ifndef __device__
+    #define __device__
+  #endif
+#endif
+
+START_NAMESPACE_STIR
+
+/*!
+  \file GibbsQuadraticPrior.h
```
KrisThielemans: wrong doxygen. Note that it applies to the first C++ line that follows. In any case, I'd keep GibbsQuadraticPrior doc as short as possible.

src/include/stir/recon_buildblock/GibbsQuadraticPrior.h
```diff
@@ -0,0 +1,187 @@
...
+
+  \author Matteo Neel Colombo
+  \author Kris Thielemans
+*/
+
+#ifndef __stir_recon_buildblock_GibbsQuadraticPrior_H__
+#define __stir_recon_buildblock_GibbsQuadraticPrior_H__
+
+
+#include "stir/recon_buildblock/GibbsPrior.h"
```
KrisThielemans:
```suggestion
#include "stir/recon_buildblock/GibbsPrior.h"
#ifdef STIR_WITH_CUDA
#include "stir/recon_buildblock/CudaGibbsPrior.h"
#endif 
```

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
KrisThielemans: maybe we could put it in `cudautilities.h`? and put all current stuff within `#ifdef __CUDACC__`. It makes some sense to me, as anyone who uses these macros will need to think about cuda anyway.

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
```diff
@@ -0,0 +1,239 @@
...
+START_NAMESPACE_STIR
+ 
+ /*!
+  \file GibbsRelativeDifferencePrior.h
```
KrisThielemans: as for the QuadraticPrior. This doxygen will be associated to the potential at the moment. But make it shorter.

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
```diff
@@ -0,0 +1,239 @@
...
+*/
+/*!
+  \file
+  \ingroup priors
+  \ingroup CUDA
+  \brief Declaration of class stir::CudaRelativeDifferencePrior, stir::CudaGibbsRelativeDifferencePrior and 
+  * the potential function stir::RelativeDifferencePotential
+
+  \author Matteo Neel Colombo
+  \author Kris Thielemans
```
KrisThielemans, line 20: as this copies a lot of code of the original RelativeDifferencePrior, we need to add the original authors and copyright here.

src/include/stir/recon_buildblock/GibbsPenalty.h
```diff
@@ -0,0 +1,167 @@
+/*
+    Copyright (C) 2025, University College London
```
KrisThielemans, line 2: as this copies a lot of code of the original QuadraticPrior, we need to add the original authors and copyright here.

src/recon_buildblock/GeneralisedPrior.cxx
```diff
@@ -26,72 +26,6 @@
...
-GeneralisedPrior<TargetT>::check(TargetT const& current_estimate) const
-{
-  if (!_already_set_up)
-    error("The prior should already be set-up, but it's not.");
-}
-
-#ifdef _MSC_VER
-// prevent warning message on instantiation of abstract class
-#  pragma warning(disable : 4661)
-#endif
```
KrisThielemans, line 93: these need to remain in the .cxx, remove from .h

src/recon_buildblock/recon_buildblock_registries.cxx
```diff
@@ -98,6 +100,8 @@ static PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixBy
 
 static FilterRootPrior<DiscretisedDensity<3, float>>::RegisterIt dummy4;
 static QuadraticPrior<float>::RegisterIt dummy5;
+static GibbsQuadraticPrior<float>::RegisterIt dummyGibbsQuadratic;
```
KrisThielemans: don't we need `#ifdef STIR_WTIH_CUDA` and the relevant statics here?

src/include/stir/cuda_utilities.cuh
ColomboMatte0: what i meant is that since we are compiling with `CMAKE_CUDA_ARCHITECTURES "all"` i cannot put a `# error ` if the `__CUDA_ARCH__ < 600` otherwise i get the error at compile time regardless of my current GPU. By the way i found that in STIR/CMakeLists.txt at line 220 we have  `set(CMAKE_CUDA_ARCHITECTURES "all")` shall we consider changing it?

The solution you proposed for me is not working since i cannot use stir/error.h inside a __device__ method, i need to modify stir/error.h to have __host__ __device__ methods.

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
ColomboMatte0: yes, it's a good idea. i renamed  `cuda_utilities.cuh `-> `cuda_utilities.h` so that now it can be included by .cxx file as well

src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.cuh
ColomboMatte0: i moved it in the doxygen comments for each kernel

src/include/stir/recon_buildblock/GibbsPrior.inl
ColomboMatte0: absolutely!

src/include/stir/recon_buildblock/GibbsPrior.inl
ColomboMatte0: for now i modified it like this:

```
  #ifdef STIR_OPENMP
    #if _OPENMP >= 201107 // OpenMP 3.1 or newer supports collapse(3)
      #pragma omp parallel for collapse(3)
    #else // just parallelize the outermost loop
      #pragma omp parallel for 
    #endif
  #endif

  for (int z = Image_min_indices.z(); z <= Image_max_indices.z(); ++z)
    for (int y = Image_min_indices.y(); y <= Image_max_indices.y(); ++y)
      for (int x = Image_min_indices.x(); x <= Image_max_indices.x(); ++x)
        {...}
```
This means that if collapse(3) is not available, only the z-loop will be parallelized.
Let me know if you find this appropriate.
Alternatively — although more complex — to parallelize the entire process without using collapse(3), we would need to linearize the indices and use a single loop.

ColomboMatte0: From a performance standpoint, in my experience there is no difference between parallelizing only the outer loop and parallelizing all of them.

src/recon_buildblock/recon_buildblock_registries.cxx
ColomboMatte0: GibbsQuadraticPrior is the CPU version so i don't think we need guards there.

KrisThielemans: oops. sorry. But then we'll need CUDA versions (within the guards)

src/include/stir/recon_buildblock/GibbsPrior.inl
KrisThielemans: for sure. no need to make things too complicated for older compilers. we just want it to compile.

src/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ endif()
 
 #### Swig related
 option(BUILD_SWIG_PYTHON
-       "Use SWIG to generate python bindings for STIR (experimental)" OFF)
+       "Use SWIG to generate python bindings for STIR (experimental)" ON)
```
KrisThielemans: undo this change. it doesn't belong in the PR.

KrisThielemans:
Looks like you're getting very close. Let me know when I need to approve running of the Actions. (Appveyor gives some information already_

src/recon_buildblock/recon_buildblock_registries.cxx
ColomboMatte0: yes i ve added them as well with the guards.

src/recon_buildblock/recon_buildblock_registries.cxx
```diff
@@ -149,6 +149,8 @@ static ProjectorByBinPairUsingNiftyPET::RegisterIt gpu_pair;
 
 #ifdef STIR_WITH_CUDA
 static CudaRelativeDifferencePrior<float>::RegisterIt gpu_rdp;
+static CudaGibbsQuadraticPrior<float>::RegisterIt gpu_gibbs_quadratic;
```
KrisThielemans: more than likely you're missing the corresponding `#include` statements

src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.h
```diff
@@ -22,7 +22,9 @@
 #include "stir/recon_buildblock/GibbsPrior.h"
 #include "stir/Array.h"
 #include "stir/DiscretisedDensity.h"
+#ifdef STIR_WITH_CUDA
```
KrisThielemans: is CudaGibbsPrior.h included in any file `ifndef STIR_WITH_CUDA`? If not, the guard should not be necessary.

src/swig/stir.i
```diff
@@ -141,8 +141,14 @@
 #include "stir/recon_buildblock/QuadraticPrior.h"
 #include "stir/recon_buildblock/PLSPrior.h"
 #include "stir/recon_buildblock/RelativeDifferencePrior.h"
+#include "stir/recon_buildblock/CUDA/CudaRelativeDifferencePrior.h"
```
KrisThielemans: guard on all CUDA files

KrisThielemans (changes requested):
all executables compiled I think, but SWIG compilation fails. I think it needs another guard.

KrisThielemans:
some warnings for the CUDA job
```
/home/runner/work/STIR/STIR/src/include/stir/cuda_utilities.h(96): warning #177-D: variable "dval" was declared but never referenced
      double dval = static_cast<double>(val);
             ^
          detected during:
            instantiation of "void stir::CudaGibbsPrior_value_kernel(double *, const elemT *, const float *, const elemT *, __nv_bool, int3, int3, int3, int3, int3, PotentialT) [with elemT=float, PotentialT=stir::QuadraticPotential<float>]" at line 541 of /home/runner/work/STIR/STIR/src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.cuh
            instantiation of "double stir::CudaGibbsPrior<elemT, PotentialT>::compute_value(const stir::DiscretisedDensity<3, elemT> &) [with elemT=float, PotentialT=stir::QuadraticPotential<float>]" 

/home/runner/work/STIR/STIR/src/include/stir/cuda_utilities.h(114): warning #940-D: missing return statement at end of non-void function "stir::atomicAddGeneric(double *, elemT) [with elemT=float]"
```
error in the CUDA job
```
/home/runner/work/STIR/STIR/src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.h:26:10: fatal error: cuda_runtime.h: No such file or directory
   26 | #include <cuda_runtime.h>
```
I guess we need to install something else in
https://github.com/UCL/STIR/blob/4210550bfaa19ce8942b52b397ee8bf2a6134a0c/.github/workflows/build-test.yml#L220-L225
Here's the doc: https://github.com/marketplace/actions/cuda-toolkit.
Or @casperdcl might know.

src/recon_buildblock/recon_buildblock_registries.cxx
ColomboMatte0: CudaGibbsQuadraticPrior is defined in GibbsQuadraticPrior.h which contains the CPU and GPU versions + the potential function, so i need just one #include to define the prior.

src/swig/stir.i
ColomboMatte0: sure, i forgot about swig guards.

src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.h
ColomboMatte0: the swig ones were including it without cuda guards. now i tested it on my machine and i can compile  STIR without CUDA.

src/include/stir/cuda_utilities.cuh
ColomboMatte0:
```
/home/runner/work/STIR/STIR/src/include/stir/cuda_utilities.h(96): warning #177-D: variable "dval" was declared but never referenced
      double dval = static_cast<double>(val);
             ^
          detected during:
            instantiation of "void stir::CudaGibbsPrior_value_kernel(double *, const elemT *, const float *, const elemT *, __nv_bool, int3, int3, int3, int3, int3, PotentialT) [with elemT=float, PotentialT=stir::QuadraticPotential<float>]" at line 541 of /home/runner/work/STIR/STIR/src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.cuh
            instantiation of "double stir::CudaGibbsPrior<elemT, PotentialT>::compute_value(const stir::DiscretisedDensity<3, elemT> &) [with elemT=float, PotentialT=stir::QuadraticPotential<float>]" 

/home/runner/work/STIR/STIR/src/include/stir/cuda_utilities.h(114): warning #940-D: missing return statement at end of non-void function "stir::atomicAddGeneric(double *, elemT) [with elemT=float]"
```

This warning is still related to this issue, currently i just commented out the code for `__CUDA_ARCH__ < 600` so it's complaing that i m not using` dval`.

KrisThielemans:
> ```
> In file included from /home/runner/work/STIR/STIR/src/include/stir/recon_buildblock/GibbsQuadraticPrior.h:30,
>                  from /home/runner/work/STIR/STIR/src/recon_buildblock/recon_buildblock_registries.cxx:24:
> /home/runner/work/STIR/STIR/src/include/stir/recon_buildblock/CUDA/CudaGibbsPrior.h:26:10: fatal error: cuda_runtime.h: No such file or directory
>    26 | #include <cuda_runtime.h>
> ```

> I guess we need to install something else...

that's unlikely to be it, as we have it already in https://github.com/UCL/STIR/blob/19ef64e9669425f0d2a7cb0d0fef665eea8cbc32/src/recon_buildblock/CUDA/CudaRelativeDifferencePrior.cu#L30

Instead, I guess this happens because if CMake dependency on CUDA. The registries shouldn't depend on the cuda_runtime. Another guard wiht `CUDA_CC` might fix this.

src/include/stir/recon_buildblock/GibbsPrior.inl
```diff
@@ -0,0 +1,623 @@
+/*
+    Copyright (C) 2025, University College London
+    Copyright (C) 2025, University of Milano-Bicocca
```
KrisThielemans: a lot of stuff copied from the old QuadraticPrior presumably, so need to preserve its copyright

src/include/stir/recon_buildblock/GibbsQuadraticPrior.h
```diff
@@ -0,0 +1,189 @@
...
+  ; next defaults to 0, set to 1 for 2D inverse Euclidean weights, 0 for 3D
+  only 2D:= 0
+  ; next can be used to set weights explicitly. Needs to be a 3D array (of floats).
+  ' value of only_2D is ignored
+  ; following example uses 2D 'nearest neighbour' penalty
+  ; weights:={{{0,1,0},{1,0,1},{0,1,0}}}
+  ; use next parameter to specify an image with penalisation factors (a la Fessler)
+  ; kappa filename:=
+  ; use next parameter to get gradient images at every subiteration
+  gradient filename prefix:=
```
KrisThielemans: These are the ones "in addition to"? Doesn't look like it. I think this should be
```suggestion
  ; keywords from GibbsPrior
```

src/include/stir/recon_buildblock/GibbsPrior.h
```diff
@@ -0,0 +1,170 @@
...
+  \author Kris Thielemans
+  \author Matteo Neel Colombo
+  \author Sanida Mustafovic
+
+*/
+
+#ifndef __stir_recon_buildblock_GibbsPrior_H__
+/**
+ * @file GibbsPrior.h
+ * \ingroup priors
```
KrisThielemans: This doc should be moved down as class doc (remove the @file obviously, as you have one above already). That way, it will be transported to python etc (file doc isn't).

KrisThielemans: Add the parsing parameters here, such that we don't need them in the derived classes

src/include/stir/recon_buildblock/GibbsQuadraticPrior.h
```diff
@@ -0,0 +1,189 @@
...
+    Cuda Gibbs Quadratic Prior Parameters:=
+    ; next defaults to 0, set to 1 for 2D inverse Euclidean weights, 0 for 3D
+    only 2D:= 0
+    ; next can be used to set weights explicitly. Needs to be a 3D array (of floats).
+    ' value of only_2D is ignored
+    ; following example uses 2D 'nearest neighbour' penalty
+    ; weights:={{{0,1,0},{1,0,1},{0,1,0}}}
+    ; use next parameter to specify an image with penalisation factors (a la Fessler)
+    ; kappa filename:=
+    ; use next parameter to get gradient images at every subiteration
```
KrisThielemans: remove here

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
```diff
@@ -0,0 +1,240 @@
...
+  ; next defaults to 0, set to 1 for 2D inverse Euclidean weights, 0 for 3D
+  only 2D:= 0
+  ; next can be used to set weights explicitly. Needs to be a 3D array (of floats).
+  ; weights:={{{0,1,0},{1,0,1},{0,1,0}}}
+  ; gamma value :=
+  ; epsilon value :=
+  ; use next parameter to specify an image with penalisation factors (a la Fessler)
+  ; kappa filename:=
+  ; use next parameter to get gradient images at every subiteration
+  gradient filename prefix:=
```
KrisThielemans: refer to GibbsPrior parameters, only keep gamma and epsilon, indicate their default values

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
```diff
@@ -0,0 +1,240 @@
...
```
KrisThielemans: as above

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
```diff
@@ -0,0 +1,240 @@
+//
+//
+/*
+    Copyright (C) 2000- 2011, Hammersmith Imanet Ltd
```
KrisThielemans: remove this, nothing here anymore from the good old days

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
```diff
@@ -0,0 +1,240 @@
...
+  \file
+  \ingroup priors
+  \ingroup CUDA
+  \brief Declaration of class stir::CudaRelativeDifferencePrior, stir::CudaGibbsRelativeDifferencePrior and 
+  the potential function stir::RelativeDifferencePotential
+
+  \author Kris Thielemans
+  \author Matteo Neel Colombo
+  \author Imraj Singh
+  \author Sanida Mustafovic
```
KrisThielemans: remove Sanida, as that was all the basic GibssPrior stuff. Put yourself first!

src/recon_buildblock/CUDA/CudaGibbsRelativeDifferencePrior.cu
```diff
@@ -0,0 +1,68 @@
...
+CudaGibbsRelativeDifferencePrior<elemT>::CudaGibbsRelativeDifferencePrior(const bool only_2D, float penalisation_factor,float gamma_v, float epsilon_v)
+  : base_type(only_2D, penalisation_factor) {this->potential.gamma = gamma_v; this->potential.epsilon = epsilon_v;}
+
+template <typename elemT>
+void
+CudaGibbsRelativeDifferencePrior<elemT>::set_defaults()
+{
+  base_type::set_defaults();
+  this->potential.gamma = 2;
+  this->potential.epsilon = 1e-7F;
```
KrisThielemans: are these default values anywhere in the doxygen?

src/recon_buildblock/GeneralisedPrior.cxx
```diff
@@ -26,73 +26,14 @@
 
 START_NAMESPACE_STIR
 
-template <typename TargetT>
```
KrisThielemans, line 29: as these were all moved now, the copyright header from this .cxx needs to be merged with the .inl (maybe nothing to do, I didn't check).

In fact, I wonder if we still need this .cxx

src/include/stir/recon_buildblock/GibbsPrior.h
KrisThielemans: Ideally, write something brief on what members the `PotentialT` should have. (In C++20, this can be encoded in Concepts I think, but let's not go there, as STIR is still C++17)

KrisThielemans (changes requested):
A lot of copyright/doxygen comments in this review, for clean-up.

Code reduction question: couldn't we add the `is_convex` to the Potential function, such that `GibbsPrior::is_convex()` just calls `potential.is_convex()`. Then we don't need it in all the derived classes.

possibly we could do the same for the parsing stuff to avoid having to repeat it for the CPU and CUDA versions, something like 
```c++
GibbsPrior::initialise_keymap()
{
  this->parser.add_start_key(this->get_parsing_name());
  base_type::initialise_keymap();
  this->potential.initialise_keymap(this->parser);
  this->parser.add_stop_key("END " + this->get_parsing_name());
}

RelativeDifferencePotential::initialise_keymap(KeyParser& parser)
{
  parser.add_key("gamma value", &this->gamma);
  parser.add_key("epsilon value", &this->epsilon);
}

std::string GibbsRelativeDifferencePrior::get_parsing_name() const
{
  return "Gibbs Relative Difference Prior";
}
```
No more need for any `initialise_keymap` in `GibbsRelativeDifferencePrior` and `CudaGibbsRelativeDifferencePrior`. But feel free to leave this for later...

Did you do LogCosh already? Could be quite easy, although I realise you've run out of time. The advantage though would be that we can deprecate all the old priors, and say they will be removed in 7.0. (That will be a PR reducing STIR code base dramatically :-)).

KrisThielemans:
> I think I found the problem in CudaGibbsPrior.h. It’s related to the #include <cuda_runtime.h> directive. I can’t simply guard it with #ifdef CUDA because it’s required for CUDA-specific variable types (dim3 and int3). In the last commit, I tried guarding those variables as well.
> 
> In CudaRelativeDifferencePrior.h, they solved this issue by defining their own struct ccpdim3, allowing the CPU compiler to work with it, and then only in the **.cu they copied from ccpdim3 to dim3 which is the one to use for CUDA.

I'd prefer to use this latter work-around if possible, as it removes a huge "include dependency" (which is why I suggested it at the time). I'm a bit afraid of class definitions that change depending on `#ifdef`. Surely that will be ill-defined.

KrisThielemans:
I wonder if we should take the opportunity to address #1426, i.e. call all your new classes `....Penalty`. I couldn't face this with the old priors, but those are going to be removed anyway. What do you think?

src/include/stir/recon_buildblock/GibbsPrior.inl
ColomboMatte0: so should i add here Sanida as author and  Hammersmith Imanet Ltd copyright here?

KrisThielemans: Checking some history, we need to add authors
- Robert Twyman 
- Sanida Mustafovic

copyright: 
- UCL copyright 2022
- Hammersmith Imanet Copyright 2001 - 2005

thanks!

src/recon_buildblock/GeneralisedPrior.cxx
ColomboMatte0: Not sure to understand the problem.

KrisThielemans:
> I’ve added the two main changes you suggested (parsing and the is_convex part). 

cool! 

> As for the log-cosh prior, I could add it easily, but I want to point out that my code is still missing the parabolic surrogate implementation. Since that’s required for the log-cosh prior, it might be better to wait and add it later — once I create an intermediate class inheriting from GibbsPrior that implements the parabolic surrogate. After that, I’ll add the LogCosh class and update the GibbsQuadratic prior accordingly.
> I’ll have time to do this after the MIC; it shouldn’t take too long.

ok. no problem. I'll release 6.3 before MIC, but it won't be the last release ever :-)

src/recon_buildblock/GeneralisedPrior.cxx
KrisThielemans: This file is now essentially empty, aside from the instantiations. We don't really need those, so I think we can remove it.

src/include/stir/recon_buildblock/GibbsQuadraticPenalty.h
```diff
@@ -0,0 +1,173 @@
...
+  void initialise_keymap(KeyParser& parser)
+  {
+    // No parameters needed for quadratic potential
+  }
+};
+#endif // SWIG
+
+template <typename elemT>
+class QuadraticPotential;
+
```
KrisThielemans, line 94: for clarity, ideally put this like this
```
#ifdef IGNORESWIG
// current stuff above

#else // IGNORESWIG
template <typename elemT>
class QuadraticPotential;
#endif
```

src/include/stir/recon_buildblock/GibbsQuadraticPrior.h
```diff
@@ -0,0 +1,173 @@
...
+  ; next defaults to 0, set to 1 for 2D inverse Euclidean weights, 0 for 3D
+  only 2D:= 0
+  ; next can be used to set weights explicitly. Needs to be a 3D array (of floats).
+  ' value of only_2D is ignored
+  ; following example uses 2D 'nearest neighbour' penalty
+  ; weights:={{{0,1,0},{1,0,1},{0,1,0}}}
+  ; use next parameter to specify an image with penalisation factors (a la Fessler)
+  ; kappa filename:=
+  ; use next parameter to get gradient images at every subiteration
+  gradient filename prefix:=
```
KrisThielemans: cut these as above

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
KrisThielemans: resolve, but 2000-2011 is still there?

drop Sanida, add Robert Twyman. (sorry)

src/include/stir/recon_buildblock/GibbsRelativeDifferencePrior.h
KrisThielemans: resolved, but still there?

ColomboMatte0:
> I wonder if we should take the opportunity to address #1426, i.e. call all your new classes `....Penalty`. I couldn't face this with the old priors, but those are going to be removed anyway. What do you think?

Yes, that makes more sense — I guess it’s better not to call them priors. Should I rename everything to something like GibbsPenalty, GibbsQuadraticPenalty, etc.?

ColomboMatte0:
i don t think we can put set_defaults in GibbsPrior under const, the method is modifying internal attributes of the class. Am i wrong?

KrisThielemans:
> i don t think we can put set_defaults in GibbsPrior under const, the method is modifying internal attributes of the class. Am i wrong?

no, you're right. Not `const`. But can still be a (non-const) member of the `potential` which we can call.

ColomboMatte0:
> RegisteredParsingObject::get_registered_name() exists, so get_parsing_name() can just return get_registered_name() i think.

Yes, that’s what I was thinking, but I have to do it in the derived class, not in GibbsPrior.

KrisThielemans:
why? (both of these should be `virtual`)

KrisThielemans:
> I thought you wanted to remove the implementation from the derived class and move it to GibbsPrior.

yes, so `GibbsPrior::get_parsing_name()` returns `GibbsPrior::get_registered_name()`. Virtual functions will sort it out. I've tested this locally and it works fine. I'll push this to your branch (sorry) to make things simpler.

KrisThielemans:
Want me to fix `set_defaults()` as well? Let me know.

KrisThielemans:
Nice and simple solution :-) should work.

I now vote for moving everything in the `*Penalty.cxx` to the `.inl/cuh`, but as you prefer. In any case, delete `GeneralisedPrior.cxx`.

I suppose other things remaining:
- handle the old CUDA architectures, or some other alternative
- switching to new priors in some of the recon_test_pack
- check discrepancies
- release notes.

Anything else?

ColomboMatte0:
I removed the .cxx and .cu files for the specific priors, so now everything is contained in the headers. I encountered a very annoying problem though:
Everything works fine for the CPU classes, but the GPU ones are no longer able to work properly. I get a compilation error :

`recon_buildblock/recon_buildblock_registries.cxx.o: undefined reference to `stir::CudaGibbsPenalty<float, stir::QuadraticPotential<float> `

So there is something strange going on with the instantiations to my understanding, because the error is fixed once i instantiate somewhere:

`template class CudaGibbsPenalty<float, QuadraticPotential<float>>;`

So for now, since i don t have other ideas i created a file called CudaGibbsPriorInstantiations.cu which take care of the instantiations of cuda priors base class.

recon_test_pack/OSMAPOSL_test_PM_QP.par
```diff
@@ -22,12 +22,12 @@ projector pair type := Matrix
   End Ray tracing matrix parameters :=
   End Projector Pair Using Matrix Parameters :=
 
-prior type := quadratic
-  Quadratic Prior Parameters:=
+prior type := Cuda Gibbs Quadratic
+  Cuda Gibbs Quadratic Penalty Parameters:=
```
KrisThielemans: CI will fail on this as no CUDA run-time. I guess we just remove `Cuda`. In future, someone could detect in `run_tests.sh` is the run-time is present and repeat the tests with the CUA prior, but let's do this later! (After merging the PR, feel free to create an issue such that we don't forget)

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
```diff
@@ -0,0 +1,797 @@
...
+#include <algorithm>
+
+START_NAMESPACE_STIR
+
+// CUDA kernels for CudaGibbsPenalty
+
+//! CUDA kernel for Gibbs penalty value calculation using the potential function value and block reduction
+/*!
+ * Each thread processes one voxel, computes local sum, then performs parallel reduction in shared memory.
+ * Result from each block is then atomically added to output. Requires CUDA compute capability ≥6.0 for double atomics.
```
KrisThielemans, line 52: add 
```
 \ingroup priors
  \ingroup CUDA
```

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.h
```diff
@@ -0,0 +1,139 @@
...
+  \brief Declaration of the stir::CudaGibbsPenalty class
+
+  \author Kris Thielemans
+  \author Matteo Neel Colombo
+*/
+
+#ifndef __stir_recon_buildblock_CUDA_CudaGibbsPenalty_H__
+#define __stir_recon_buildblock_CUDA_CudaGibbsPenalty_H__
+
+#include "stir/Array.h"
```
KrisThielemans, line 22: replace all includes with
```c++
#include "stir/recon_buildblock/GibbsPenalty.h"
```
we need this one!

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.h
```diff
@@ -0,0 +1,139 @@
...
+struct int3
+{
+  int x = 0, y = 0, z = 0;
+};
+#else
+#  include <cuda_runtime.h>
+#endif
+
+#include "stir/shared_ptr.h"
+#include <string>
```
KrisThielemans: What do you think about moving this to `cuda_utilities.h`, and including that instead? Might be useful for others later. I think it will prevent conflicts if we need to define `dim3` etc elsewhere as well, but as you prefer.

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.h
```diff
@@ -0,0 +1,139 @@
...
+
+protected:
+  // GPU block and grid dimensions
+
+  dim3 block_dim;
+  dim3 grid_dim;
+
+  // Variable used for shared memory operations
+  int threads_per_block = block_dim.x * block_dim.y * block_dim.z;
+  size_t shared_mem_bytes = threads_per_block * sizeof(elemT);
```
KrisThielemans: Do these initialisations make sense here? Are `block_dim` already initialised by nvcc or so? I don't think so, as you have the initialisation elsehwere anyway)

src/recon_buildblock/CUDA/CudaGibbsPenaltiesInstantiations.cu
```diff
@@ -0,0 +1,17 @@
...
+
+#include "stir/recon_buildblock/GibbsQuadraticPenalty.h"
+#include "stir/recon_buildblock/GibbsRelativeDifferencePenalty.h"
+#include "stir/recon_buildblock/CUDA/CudaGibbsPenalty.h"
+
+START_NAMESPACE_STIR
+
+// Ensure the CUDA base template is instantiated for the potentials we use
+template class CudaGibbsPenalty<float, QuadraticPotential<float>>;
+template class CudaGibbsPenalty<float, RelativeDifferencePotential<float>>;
```
KrisThielemans: I don't think we need this really. The compiler will instantiate them because we use them. These days I'm in favour of avoiding it, as it creates extra code to maintain.

KrisThielemans (changes requested):
final clean-up. I need to think about the instantiation question...

What about the following idea...
```c++
#ifndef __CUDA__
struct cuda_dim3 {blabla}
#else
typedef dim3 cuda_dim3;
#endif
```
(and same for `int3`) and using `cuda_dim3` in the class def. That way we avoid name clashes (and make things clearer, I think).

recon_test_pack/OSMAPOSL_test_PM_QP.par
ColomboMatte0: i forgot to remove Cuda, i was just testing it locally to see if everything was ok.

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
ColomboMatte0: But that's the documentation for just one kernel. The class have already the`  \ingroup priors  \ingroup CUDA ` in the CudaGibbsPenalty.h

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.h
ColomboMatte0: I think is  a good idea!

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.h
ColomboMatte0: yes probably better to move them to set_up

ColomboMatte0: actually they are useless there i realized i was already doing this in the set_up():
```
 threads_per_block = block_dim.x * block_dim.y * block_dim.z;
  shared_mem_bytes = threads_per_block * sizeof(elemT);
```

ColomboMatte0:
> final clean-up. I need to think about the instantiation question...
> 
> What about the following idea...
> 
> ```c++
> #ifndef __CUDA__
> struct cuda_dim3 {blabla}
> #else
> typedef dim3 cuda_dim3;
> #endif
> ```
> 
> (and same for `int3`) and using `cuda_dim3` in the class def. That way we avoid name clashes (and make things clearer, I think).

I think it's cleaner.

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
KrisThielemans: Doxygen has different lists for files and classes. The `\file` doc applies to the file, not to any of the classes that it contains (a bit weird, I know).

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
```diff
@@ -0,0 +1,796 @@
...
+#include <cuda_runtime.h>
+#include <numeric>
+#include <algorithm>
+
+START_NAMESPACE_STIR
+
+// CUDA kernels for CudaGibbsPenalty
+
+//! CUDA kernel for Gibbs penalty value calculation using the potential function value and block reduction
+/*!
```
KrisThielemans, line 48:
```suggestion
/*!
 * \ingroup CUDA
 * \ingroup priors
```

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
```diff
@@ -0,0 +1,796 @@
...
+            current *= kappa[inputIndex] * kappa[neighbourIndex];
+
+          sum += current;
+        }
+  gradient[inputIndex] = sum * 2.0 * penalisation_factor;
+}
+
+//! CUDA kernel for Gibbs penalty, computes the dot product of the gradient with a given input image using the potential function
+//! first derivative.
+/*!
```
KrisThielemans, line 210:
```suggestion
/*!
 * \ingroup CUDA
 * \ingroup priors
```

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
```diff
@@ -0,0 +1,796 @@
...
+  // block pair-wise reduction in shared memory
+  int block_threads = blockDim.x * blockDim.y * blockDim.z;
+  blockReduction(block_sum_shared, thread_in_block, block_threads);
+
+  if (thread_in_block == 0)
+    atomicAddGeneric(output, block_sum_shared[0]);
+}
+
+//! CUDA kernel for Gibbs penalty, computes the Hessian diagonal using the potential function second derivative.
+/*!
```
KrisThielemans, line 300:
```suggestion
/*!
 * \ingroup CUDA
 * \ingroup priors
```

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
```diff
@@ -0,0 +1,796 @@
...
+            {
+              current *= kappa[inputIndex] * kappa[neighbourIndex];
+            }
+          sum += current;
+        }
+  Hessian_diag[inputIndex] = sum * 2.0 * penalisation_factor;
+}
+
+//! CUDA kernel for Gibbs penalty, computes the Hessian times a given input image using the potential function second derivatives.
+/*!
```
KrisThielemans, line 372:
```suggestion
/*!
 * \ingroup CUDA
 * \ingroup priors
```

KrisThielemans (changes requested):

KrisThielemans (approved):

KrisThielemans:
@casperdcl can you help with the following? @ColomboMatte0  commited sometimes as `root@DESKTOP-U7AVDMQ`.
This doesn't feel good, and  would likely create difficult with our mailmap. What is the best way to correct this? It'd be nice to preserve history in this PR, but we're going to squash-merge this. First commit was by the `root` user, so I guess it will the squash-merge will be assigned to `root`.

I suppose I could do the squash locally, correct the author, then push that to `master`, then close this PR. Not so great, as github will claim it wasn't merged.

ColomboMatte0:
i guess there are some instantiation problem for the cpu classes now? amazing!

KrisThielemans:
> I don't know why this only appears now though.

Looks like we might not have run the Actions workflow recentl (as I need to approve it)

ColomboMatte0:
> Ubuntu jobs fail, and are running SWIG 4.2.0. MacOS and AppVeyor jobs are fine, and are running SWIG 4.4.0.

I guess it's related to having removed the .cxx and .cu. The constructors where working fine back there.

KrisThielemans:
MacOS clang 21 failure due to #1420, all the rest got through! So no, you didn't do anything wrong :-)

KrisThielemans:
@ColomboMatte0 I've created #1647. Let's see...

PR #1632, opened 2025-10-14
Python: addition of as_array() and fill(numpy.ndarray)

KrisThielemans:
This adds `np_arr = stir_object.as_array()` and `stir_object.fill(np_arr)` directly via SWIG. These should be faster than the `stirextra` functions, as they are implemented in C++.

Usage is the same as in SIRF, which makes it easier to convert SIRF to STIR Python.

### DIFF
--- a/src/swig/stirextra.py
+++ b/src/swig/stirextra.py
@@ -62,13 +62,11 @@ def to_numpy(stirdata):
     """
     return the data in a STIR image or other Array as a numpy array
     """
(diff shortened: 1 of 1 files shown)


KrisThielemans:
This should be ready aside from:
- test `ProjData.as_array()` (currently we test only `ProjDataInMemory.as_array()`)
- edit `stirextra.to_numpy` to use `as_array()` presumably
- add release notes

Wou

PR #1643, opened 2025-10-22
Enable MSVC from ninja and adjust AppVeyor

KrisThielemans:
- make sure STIR compiles when using ninja. (it doesn't set the "exceptions" compiler flag, nor does it define `WIN32`, which we use)
- use ninja in Appveyor
- try to use [ccache](https://ccache.dev/)


KrisThielemans:
Current status is that it works. However, the ccache is reset at every build.

KrisThielemans:
I'll merge this even though the ccache doesn't work yet, as the ninja output is more useful, and we need the MSVC fixes in any case.

Issue #1537, opened 2024-11-15
Wrong TOF bin assigned for oblique projections


NicoleJurjew:
I was comparing projections of a uniform "plate", covering the entire axial FOV with parallelproj and STIR.

![Image](https://github.com/user-attachments/assets/45d2844d-d3d9-4b47-bdb1-84b8eac9ad26)

I plotted line profiles as follows:

Central slice of the direct segment, orthogonal view, central tangential position, over all TOF bins:
![Image](https://github.com/user-attachments/assets/b7f10b44-9dcc-4cee-96eb-6a9583bda62b)

Projection in an oblique segment, orthogonal view, central tangential position, over all TOF bins:
![Image](https://github.com/user-attachments/assets/5ee5d892-df24-4752-9971-63c6f8891f63)

For the given phantom, the activity is still in the centre of the LOR, i.e. the peak should be reached in the central TOF bin.

PR #1645, opened 2025-10-24
TOF bin fix for RayTracingMatrix

z-k-li:
<strike>Project LOR to mean z-plane for maintaining the concentric-circle model for TOF bins.</strike> Edit: fix mismatch in image/projdata coordinate system affecting TOF calculation

## Testing performed
Tested the same as issue 1537 and got consistent results with paralleproj.

## Related issues

Fixes issue #1537 

## Contribution Notes

Please tick the following: 

 - [x] The content of this Pull Request (the Contribution) is intentionally submitted for inclusion in STIR (the Work) under the terms and conditions of the [Apache-2.0](https://www.apache.org/licenses/LICENSE-2.0) License.
 - [ ] I (or my institution) have signed the STIR Contribution License Agreement (not required for small changes).

### DIFF
--- a/src/include/stir/recon_buildblock/ProjMatrixByBin.inl
+++ b/src/include/stir/recon_buildblock/ProjMatrixByBin.inl
@@ -121,7 +121,7 @@ ProjMatrixByBin::apply_tof_kernel(ProjMatrixElemsForOneBin& probabilities) const
   const CartesianCoordinate3D<float> point1 = lor2.p1();
   const CartesianCoordinate3D<float> point2 = lor2.p2();
 
-  // Coordinate system correction:
+  // Coordinate system correction: TODO remove in future with ORIGIN shift PR
   // LOR coordinates have origin at scanner center (z=0 at center of all rings)
   // Image coordinates have origin at first ring (z=0 at ring 0)
   // Calculate the offset: distance from first ring to scanner center


z-k-li:
image

KrisThielemans:
@z-k-li @NikEfth it seems to work! However, can you explain the geometric reasoning? I don't understand why the z-axis should be treated different from the others (aside from the difference in z-origin between coordinate systems flagged up in #1537).

This PR effectively sets `diff.z=0` I believe. (it doesn't change `middle`, but once `diff.z==0`, `middle.z` no longer matters). The (unchanged) formula for computing the TOF location is 
https://github.com/UCL/STIR/blob/1f2556b1722b5070e1a83fb8b87ba56262f52157/src/include/stir/recon_buildblock/ProjMatrixByBin.inl#L132

so this now becomes a distance of the projection in the transverse plane, as opposed to a projection along the line.

So, in my opinion, this PR gives "circular TOF bins" in the transverse plane, not circular along the LOR. This could be how the Quadra works, and seems to be how parallelproj works (@gschramm might confirm), but it makes no sense to me...

Maybe my geometrical insight is wrong.

KrisThielemans:
Right, that's what this PR does. However, it makes no sense to me that a scanner would do that. It measures a time difference. It would surprise me if they then do a TOF bin assignment that depends on the ring difference. (With this PR, the TOF bin will "stretch" for higher ring diff). Of course, the actual time difference would become larger for a longer LOR, and a vendor could try and compensate for that.

(I don't think the test in #1537 actually tests this stretching. It tests if the TOF bin is in the centre. So we cannot infer from it what parallelproj does regarding stretching)

z-k-li:
> sure, but the question is what a scanner does for TOF.
> 
> 
> 
> @z-k-li easy to do a sim of a line source close to the edge of the transaxial FOV? forward project with parallelproj and this PR, ideally even do a GATE sim. Look at oblique segments. The question is if the line source appears in the same TOF bin for all segments (I think it should with this PR), or not.

Yes, I will do this!

z-k-li:
image
A new update, as suggested by Kris, I tested the line source from edge of FOV. The ray_new is the new solution which is overlapped fully with the parallelproj. However, as expected, the circular TOF bins in the transverse plane solution suffers from strenching so it is not matching the parallelproj.

src/include/stir/recon_buildblock/ProjMatrixByBin.inl
```diff
@@ -114,22 +115,35 @@ ProjMatrixByBin::get_proj_matrix_elems_for_one_bin(ProjMatrixElemsForOneBin& pro
...
 {
-
   LORInAxialAndNoArcCorrSinogramCoordinates<float> lor;
   proj_data_info_sptr->get_LOR(lor, probabilities.get_bin());
   const LORAs2Points<float> lor2(lor);
   const CartesianCoordinate3D<float> point1 = lor2.p1();
   const CartesianCoordinate3D<float> point2 = lor2.p2();
 
-  // The direction can be from 1 -> 2 depending on the bin sign.
+  // Coordinate system correction:
```
KrisThielemans:
```suggestion
  // Coordinate system correction: TODO remove in future with ORIGIN shift PR
```

KrisThielemans:
I have some tests for this. And they work.

I'm tempted to squash-merge this, and then do a new PR for the test. Ok?

Ideally, you add somethng in release_6.4.htm (might have to merge `master` first)

PR #1651, opened 2025-10-27
fix remaining bug in ObjectiveFunction::accumulate_Hessian_times_input with non-zero prior

### LINKED ISSUE #1648: GeneralisedObjectiveFunction::accumulate_*Hessian* bugs when prior is non-zero
In both lines below, the last argument should be `input`, not `output`
https://github.com/UCL/STIR/blob/4c2785605362f1e83ee4999a8391b7b285fa4c03/src/recon_buildblock/GeneralisedObjectiveFunction.cxx#L295
https://github.com/UCL/STIR/blob/4c2785605362f1e83ee4999a8391b7b285fa4c03/src/recon_buildblock/GeneralisedObjectiveFunction.cxx#L397
The effect of this is that the Hessian of the prior is computed in `output` (after already accumulating the log-likelihood part into that)

KrisThielemans:
#1649 still didn't fix everything. Tests didn't catch it, so add some more tests.

Fixes #1648

### DIFF
--- a/src/swig/test/python/test_PoissonProjDataObjectiveFunction.py
+++ b/src/swig/test/python/test_PoissonProjDataObjectiveFunction.py
@@ -1,4 +1,4 @@
-# Test file for STIR buildblock py.test
+# Test file for STIR objective functions
 # Use as follows:
 # on command line
 #     pytest test_PoissonProjDataObjectiveFunction.py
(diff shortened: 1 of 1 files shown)


KrisThielemans:
One job failed due to Zenodo time-out

Issue #1652, opened 2025-10-28
extra SWIG warnings after Gibbs prior merge


KrisThielemans:
I see
```
.../src/swig/stir_priors.i:134: Warning 330: Template forward class 'stir::QuadraticPotential< float >' cannot be used to instantiate a full template class with name 'QuadraticPotential3DFloat'.
.../src/swig/stir_priors.i:146: Warning 330: Template forward class 'stir::RelativeDifferencePotential< float >' cannot be used to instantiate a full template class with name 'RelativeDifference3DFloat'.
```
@ColomboMatte0 that makes sense, as we have the `#ifdef SWIG` there. Remind me why we did? (possibly the `__host__` stuff).

Not really urgent...

PR #1653, opened 2025-10-28
fixed swig warning about potential function

ColomboMatte0:
__host__ and __device__ flags declared as empty for SWIG

#1652


KrisThielemans:
I'm optimistic...

KrisThielemans:
The usual failure in SRT2D, otherwise good!

Issue #1666, opened 2025-11-17
Create function to compute crystal histogram from projdata


markus-jehl:
For the singles prompt method of estimating randoms from singles (https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0162096), additional information on the prompts is required. From the prompts projdata, a histogram of crystal contributions to the prompts needs to be created, for which there is currently no function in STIR.

PR #1667, opened 2025-11-17
Prompt histogram calculation

markus-jehl:
Added a function to translate a projdata input into a histogram of events each crystal contributed to.

## Related issues
fixes https://github.com/UCL/STIR/issues/1666

## Contribution Notes

Please tick the following: 

 - [ ] The content of this Pull Request (the Contribution) is intentionally submitted for inclusion in STIR (the Work) under the terms and conditions of the [Apache-2.0](https://www.apache.org/licenses/LICENSE-2.0) License.
 - [ ] I (or my institution) have signed the STIR Contribution License Agreement (not required for small changes).

### DIFF
--- a/documentation/release_6.4.htm
+++ b/documentation/release_6.4.htm
@@ -51,6 +51,12 @@
 
   <h3>Changed functionality</h3>
   <ul>
+    <li>
+      OpenMP parallelisation was added to the function to create a 2d histogram of hits per crystal 
+      from a proj data object, <code>make_fan_sum_data</code>.
+      <br>
+      <a href=https://github.com/UCL/STIR/pull/1667>PR #1667</a>
+    </li>
   </ul>
   <h4>Changes to examples</h4>
   <ul>


src/buildblock/ProjDataInMemory.cxx
```diff
@@ -350,7 +350,7 @@ ProjDataInMemory::read_from_file(const std::string& filename)
 }
 
 float
-ProjDataInMemory::get_bin_value(Bin& bin)
+ProjDataInMemory::get_bin_value(Bin& bin) const
```
KrisThielemans: good catch. the `bin` argument should be `const&` as well I guess

src/buildblock/create_crystal_histogram.cxx
```diff
@@ -0,0 +1,129 @@
+/*!
+
+  \file
+  \ingroup projdata
+
+  \brief Implementation of stir::create_crystal_histogram
+
+  \author Kris Thielemans
```
KrisThielemans: not really!

markus-jehl: haha, copying done poorly 😅

KrisThielemans:
I think you're right. I can only see numerical issues. What happens if you use `Array<2,double>` for `data_fan_sums` (or your equivalent)

markus-jehl:
We're essentially dealing with integers in my example, so floating point issues don't come into it. The reason for the difference was that I'm summing from min_tangential_pos to max_tangential_pos, and data_fan_sums goes from -half_fan_size (which is min_tangential_pos + 1) to half_fan_size. This is also why in my output the histogram is not completely homogeneous:
image

Note, the asymmetry is just due to my test, where I fill the projdata with homogeneous values. In realistic cases the min_tangential_pos bin would be essentially zero anyways, so then the data will be symmetric. Therefore I think that my approach is actually more accurate, no?

src/buildblock/ML_norm.cxx
```diff
@@ -1447,42 +1447,70 @@ make_fan_sum_data_help(Array<2, float>& data_fan_sums,
                        const TProjDataInfo& proj_data_info,
                        const ProjData& proj_data)
 {
-  if (proj_data.get_proj_data_info_sptr()->is_tof_data())
-    error("make_fan_data: Incompatible with TOF data. Abort.");
```
KrisThielemans, line 1451: we're still not looping over TOF, so either we put that in (which might be tricky to do with compatibility with blocksoncylindrical), or we keep these lines.

src/buildblock/ML_norm.cxx
```diff
@@ -1447,42 +1447,70 @@ make_fan_sum_data_help(Array<2, float>& data_fan_sums,
...
-                const float value = (*segment_ptr)[bin.axial_pos_num()][bin.view_num()][bin.tangential_pos_num()];
-                data_fan_sums[ra][a] += value;
-                data_fan_sums[rb][b] += value;
-              }
-      }
+      for (bin.axial_pos_num() = proj_data.get_min_axial_pos_num(bin.segment_num());
+           bin.axial_pos_num() <= proj_data.get_max_axial_pos_num(bin.segment_num());
+           ++bin.axial_pos_num())
+        {
+          auto sinogram = proj_data.get_sinogram(bin.axial_pos_num(), bin.segment_num());
```
KrisThielemans:
```suggestion
          const auto sinogram = proj_data.get_sinogram(bin);
```
could work, potentially needs `SinogramIndices(bin)`

src/buildblock/ML_norm.cxx
```diff
@@ -1447,42 +1447,70 @@ make_fan_sum_data_help(Array<2, float>& data_fan_sums,
...
+                  // Construct bin with appropriate values
+                  // Sadly cannot be done in the loops above for OpenMP 2.0 compatibility
+                  Bin parallel_bin(bin);
+                  parallel_bin.view_num() = view_num;
+                  parallel_bin.tangential_pos_num() = tangential_pos_num;
+
+                  std::vector<DetectionPositionPair<>> det_pos_pairs;
+                  non_tof_proj_data_info_sptr->get_all_det_pos_pairs_for_bin(
+                      det_pos_pairs, parallel_bin); // using the default argument to ignore TOF here
+                  for (unsigned int i = 0; i < det_pos_pairs.size(); ++i)
```
KrisThielemans: could use new-style loop
```c++
  for (const auto& dp_pair : det_pos_pairs)
```
clearer?

KrisThielemans (changes requested):
some minor changes. Can you also add something to release_6.4.htm?

src/buildblock/ML_norm.cxx
markus-jehl: true! let's keep these lines for now

KrisThielemans (approved):

PR #1668, opened 2025-11-23
[CMake] depend on both HDF5::HDF5 and hdf5::hdf5_cpp

KrisThielemans:
We need to work-around a CMake (up to a least 4.1) bug that the hdf5::hdf5_cpp target is not fully populated. See https://gitlab.kitware.com/cmake/cmake/-/issues/27379

Hopefully, this will resolve SIRF linking errors with HDF5, see https://github.com/SyneRBI/SIRF-SuperBuild/pull/973

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -29,6 +29,11 @@ option(BUILD_EXECUTABLES
 option(BUILD_SHARED_LIBS 
        "Use shared libraries" OFF)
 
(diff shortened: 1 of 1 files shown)


KrisThielemans:
Merging this, as it is a logical work-around, and it resolves a SIRF linking problem with missing `H5::*` (i.e. C++ functions) when using the SIRF-SuperBuild and `USE_SYSTEM_HDF5=OFF`.

PR #1674, opened 2026-01-19
TOF improvements

KrisThielemans:
- add test for symmetric kernel of TOF (see #1645)
- remove zeroes from the cached projection matrix. Before, for every TOF bin, we still had row with the same size as the row of the non-TOF matrix. This meant a lot more memory was used than needed, and a potentially lot of multiplications with zero.

### DIFF
--- a/documentation/release_6.4.htm
+++ b/documentation/release_6.4.htm
@@ -53,7 +53,14 @@
   <ul>
     <li>When using a (ray tracing) matrix as projector, previously the matrix stored elements for the whole LOR,
       while many elements would be zero. We now only store the non-zero ones, resulting in a speed-up for the Siemens Vision Quadra of about a factor 1.7. Memory usage is also reduced.<br>
-      <a href=https://github.com/UCL/STIR/pull/1674># 1674</a>
+      <a href=https://github.com/UCL/STIR/pull/1674># 1674</a>.
+    </li>
+    <li>
+      OpenMP parallelisation was added to the function to create a 2d histogram of hits per crystal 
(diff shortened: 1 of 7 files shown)


KrisThielemans:
@z-k-li maybe you could have a look at this

KrisThielemans:
@z-k-li any news? Note that I suspect that this will make the first (full) iteration slower but later ones faster. It'd be interesting to check the timings reported.

z-k-li:
> @z-k-li any news? Note that I suspect that this will make the first (full) iteration slower but later ones faster. It'd be interesting to check the timings reported.

Hello Kris, Sorry for the late reply, I only just had time to test this.

The speed improvement seems big. 

I found that the new PR takes about 150 s per subiteration (wall clock time), whereas the old version takes about 280 s per sub-iteration. I did not observe substantial difference between different full iterations.

Overall, for a 4i5s reconstruction of a 220 × 220 × 645 MRD85 dataset using 16 CPU cores, the new PR completes in ~70 minutes, compared with ~124 minutes for the old version.

KrisThielemans:
That's great news! Maybe you could report memory usage here as well, but I'll merge.

ChristianHinge:
@KrisThielemans I ran this for our quadra data (zoom 1,  4i5s). It took 41 minutes, 55GB max ram.  For zoom 0.5 it is 16 minutes, 19GB max ram. Awesome improvements!

KrisThielemans:
@ChristianHinge that's great! Can you give some details on your workstation's processor?

Issue #1679, opened 2026-02-20
enable CUDA managed pointers


KrisThielemans:
To avoid copying data to/from GPU all the time, CUA managed pointers might come to the rescue. This would become more attractive once supported by parallelproj (WIP by @gschramm ). A possible strategy would then be
1. change these to CUDA managed pointers to avoid them being copied every time.
    https://github.com/UCL/STIR/blob/b67c9f4a84ba4ba2c8fabb937514c2c26b575673/src/include/stir/recon_buildblock/Parallelproj_projector/ParallelprojHelper.h#L50-L51
2. somehow add optional argument to `Array` etc to allocate with CUDA managed pointer (could probably be done by templating the allocator/destructor)
3. make images by default CUDA-managed
4. add CUDA numerical operations (either as member or global)
5. replace some iterator-based loops for numerical operations of images with numerical operations
6. investigate what happens with projection data.

KrisThielemans:
@Dimitra-Kyriakopoulou proposes some small helpers to wrap `CuVec` to `stir::Array` in https://github.com/Dimitra-Kyriakopoulou/STIR/commit/3b1ea8f6c2bc534af3b607207f9af14e00fb8d0d. It uses the following constructor
https://github.com/UCL/STIR/blob/46da4ef2b8b38487b08f5aa48c6e38b568a584cc/src/include/stir/Array.h#L151
This approach seems quite interesting.

See also https://github.com/UCL/STIR/pull/1689#issuecomment-4085525010

@casperdcl any comments?

casperdcl:
> @Dimitra-Kyriakopoulou proposes some small helpers to wrap `CuVec` to `stir::Array`

lgtm

PR #1689, opened 2026-03-07
Upgrade to parallelproj 2.0 and use cuvec

KrisThielemans:
See https://github.com/KUL-recon-lab/libparallelproj

<strike>Currently this PR is on top of #1676, while at least initially there is no good reason for this. Look therefore only at the last commit(s) and ignore the `test_Array` failure. Sorry</strike>

WARNING: Commits here will be rebased/squashed etc. The PR will probably also be split in 2 or 3 other PRs.

@gschramm @markus-jehl feel free to comment :-)

### DIFF
--- a/documentation/release_6.4.htm
+++ b/documentation/release_6.4.htm
@@ -12,15 +12,14 @@
   <h2>Overall summary</h2>
 
     <p>
-      This version is 100% backwards compatible with STIR 6.3, aside from three important bug fixes,
-      which will change results (see below).
+      This version is 100% backwards compatible with STIR 6.3.
     </p>
     <p>
       This is a release with many new features, as listed below. Highlights are
       <ul>
 
       </ul>
-      But there are many other changes from other contibutors as well, see below.
+      But there are many other changes from other contributors as well, see below.
       Of course, there is also the usual code-cleanup and some improvements to the documentation.
       Overall overview and release management was by Kris Thielemans (UCL) with help
       from Daniel Deidda (NPL).
@@ -29,8 +28,8 @@
     <h2>Patch release info</h2>
     <ul>
       <li>
-        6.4.0 released x/2025<br>
-        <a href="https://github.com/UCL/STIR/milestone/12">GitHub Milestone 6.3</a>
+        6.4.0 released XXX/2026<br>
+        <a href="https://github.com/UCL/STIR/milestone/13">GitHub Milestone 6.4</a>
       </li>
     </ul>
 
@@ -61,7 +60,25 @@
       <br>
       <a href=https://github.com/UCL/STIR/pull/1667>PR #1667</a>.
     </li>
+    <li>
+      Several optimisations were contributed as part of the <a href="https://www.ccpsynerbi.ac.uk/events/airbi-hackathon/">SyneRBI AI-RBI hackathon</a>:
+      <ul>
+        <li>
+          Compatibility with <a href="https://github.com/KUL-recon-lab/libparallelproj/">libparallelproj 2.0</a> and usage of CUDA managed pointers via
+          <a href="https://github.com/AMYPAD/cuvec">CuVec</a> for internal variables in our parallelproj interface and <code>CudaGibbsPenalty</code>.
+          This results in a ~20% speed-up, but also code simplification.
+          <br>
+          <a href=https://github.com/UCL/STIR/pull/1689>PR #1689</a>.
+        </li>
+        <li>
+          Extra constructors for array, image and projdata classes that allow <code>std::move</code> for input arrays.
+          <br>
+          <a href=https://github.com/UCL/STIR/pull/1693>PR #1693</a> and <a href=https://github.com/UCL/STIR/pull/1694>PR #1694</a>.
+        </li>
+      </ul>
+    </li>
   </ul>
+
   <h4>Changes to examples</h4>
   <ul>
   </ul>


KrisThielemans:
Current status:
- run-time problem with CUDA after insert CuVeC for parallelproj is on a single computer, but not on 2 others (including one with the same docker set-up). @markus-jehl can you put the error message here for future reference?
- `stir_timings` with NeuroLF data gives about 20% speed-up of PP projections, likely because `xstart`, `xend` remain on the GPU. @denproc, could you add the run-time comparison here?
- CudaGibbsPenalty hierarchy finished by @denproc 

Things still to do:
- [x] re-enable chunking (@markus-jehl ?)
- [ ] possibly look at doing `TOF_transpose` on GPU (could be different PR)
- [x] move cuvec submodule into `external_helpers` and insert optional `find_package(cuvec)` similar to `{fmt}` (Kris)
- [x] merge `master` to remove conflicts (Kris)
- [x] ideally re-instate parallelproj 1.* compatibility to be able to include this in STIR 6.*
- [ ] potentially pull this PR apart and/or clean up git history (mostly the back and forth on handling cuvec in CMake) (Kris)

ColomboMatte0:
> @ColomboMatte0 could you have a quick look at the CudaGibbPenalty changes? Main changes is that `cuvec` handles memory management for Cuda managed pointers. @denproc found some missing "sync" statements as well.

Which sync are we talking about?
Some synchronizations were intentionally omitted because an explicit memory copy from the GPU (e.g., cudaMemcpy) already triggers a synchronization. Adding an additional synchronization after the copy would therefore be redundant and only introduce unnecessary overhead.

denproc:
> > @ColomboMatte0 could you have a quick look at the CudaGibbPenalty changes? Main changes is that `cuvec` handles memory management for Cuda managed pointers. @denproc found some missing "sync" statements as well.
> 
> Which sync are we talking about? Some synchronizations were intentionally omitted because an explicit memory copy from the GPU (e.g., cudaMemcpy) already triggers a synchronization. Adding an additional synchronization after the copy would therefore be redundant and only introduce unnecessary overhead.

3 instances of `cudaDeviceSynchronize()` have been added to follow the existing synchronisation use. 
One instance is in the context of `cudaMemcpy` in line [493](https://github.com/KrisThielemans/STIR-1/blob/f834cafe961a6464380e9b5fe2e2f559a39dc5cf/src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh#L493) and was added to make it consistent with the existing use of synchrosiation in line [582](https://github.com/KrisThielemans/STIR-1/blob/f834cafe961a6464380e9b5fe2e2f559a39dc5cf/src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh#L582).
Two others in line [625](https://github.com/KrisThielemans/STIR-1/blob/f834cafe961a6464380e9b5fe2e2f559a39dc5cf/src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh#L625) and [669](https://github.com/KrisThielemans/STIR-1/blob/f834cafe961a6464380e9b5fe2e2f559a39dc5cf/src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh#L669) were added to match existing use in line [537](https://github.com/KrisThielemans/STIR-1/blob/f834cafe961a6464380e9b5fe2e2f559a39dc5cf/src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh#L537).

If it is not needed, could we also potentially omit the existing syncs?

src/include/stir/cuda_utilities.h
```diff
@@ -22,6 +22,10 @@
 #include "stir/Array.h"
 #include "stir/info.h"
 #include "stir/error.h"
+#ifdef __CUDACC__
+#  include <cuda_runtime.h>
+#  include "cuvec.cuh"
+#endif
```
casperdcl, line 28: Since non-cuda is handled by `CUVEC_DISABLE_CUDA`
```suggestion
#endif
#include "cuvec.cuh"
```

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "external_helpers/fmt"]
 	path = external_helpers/fmt
 	url = https://github.com/fmtlib/fmt.git
+[submodule "cuvec"]
+	path = cuvec
+	url = https://github.com/AMYPAD/cuvec
```
casperdcl:
```suggestion
[submodule "CuVec"]
	path = CuVec
	url = https://github.com/AMYPAD/CuVec
```

src/CMakeLists.txt
```diff
@@ -262,13 +262,24 @@ if (STIR_ENABLE_EXPERIMENTAL)
   ADD_SUBDIRECTORY(${CMAKE_SOURCE_DIR}/src/experimental)
 endif()
 
+set(CUVEC_INCLUDE_DIR "${CMAKE_SOURCE_DIR}/cuvec/cuvec/include")
```
casperdcl:
```suggestion
set(CUVEC_INCLUDE_DIR "${CMAKE_SOURCE_DIR}/CuVec/cuvec/include")
```

src/include/stir/cuda_utilities.h
```diff
@@ -71,6 +74,13 @@ array_to_device(elemT* dev_data, const Array<num_dimensions, elemT>& stir_array)
     }
 }
 
+template <int num_dimensions, typename elemT>
+inline void
+array_to_device(CuVec<elemT>& dev_data, const Array<num_dimensions, elemT>& stir_array)
```
casperdcl, line 88: Should we be checking `size()` on these?

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
```diff
@@ -491,17 +470,17 @@ CudaGibbsPenalty<elemT, PotentialT>::compute_value(const DiscretisedDensity<3, e
   if (this->_already_set_up == false)
     error("CudaGibbsPenalty: set_up has not been called");
 
-  cudaMemset(d_scalar, 0, sizeof(double));
+  d_scalar[0] = 0.0;
```
casperdcl, line 473: Weird interface. Can't you pass a `double` or `float` (rather than a (`cuda`)`malloc`ed `double *`)?

src/recon_buildblock/Parallelproj_projector/BackProjectorByBinParallelproj.cxx
```diff
@@ -248,7 +250,7 @@ BackProjectorByBinParallelproj::get_output(DiscretisedDensity<3, float>& density
 
       TOF_transpose(mem_for_PP_back, STIR_mem, _helper, 0);
 
-      joseph3d_back_tof_sino(_helper->xend.data(),
+      joseph3d_tof_sino_back(_helper->xend.data(),
```
casperdcl: Probably revert and add an `#ifndef` since @gschramm said this was a typo?

casperdcl:
I'd also recommend in future enabling gpu-only, i.e.

- only copy some things between host & device on first & last iterations
- no explicit sync in intermediate iterations

See e.g. all the `sync=False` statements in https://github.com/NiftyPET/NIPET/blob/master/examples/demo.ipynb

src/recon_buildblock/Parallelproj_projector/BackProjectorByBinParallelproj.cxx
KrisThielemans: no, @gschramm made a change in naming (and apologised for it)

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
KrisThielemans: sure we could. Volunteers welcome.

.gitmodules
KrisThielemans: ok. I should have done this in the `external_helpers` folder. I guess best to do this together with the move to avoid contributors to this PR to have trouble twice.

src/CMakeLists.txt
KrisThielemans: see above

KrisThielemans:
I guess I was a bit naive and hoped we wouldn't need any sync statements at all with the CUDA managed memory... TBH, that's a bit confusing to me. Needing syncs seems contrary to the CUDA managing if it's on the CPU or GPU transparently, as it seems to imply we need a sync whenever we want to do something with the CUDA managed pointer that could involve CPU. Is that correct?

ColomboMatte0:
From what I understood about the CUDA managed memory, the transparency is about data location (automatic CPU↔GPU), not about execution ordering. The sync is needed because GPU kernels are asynchronous — the host continues executing immediately after a kernel launch, so without a sync you risk reading managed memory before the kernel has finished writing to it.

The old cudaMemcpy(DeviceToHost) didn't need an explicit sync because it's implicitly synchronizing — it blocks until all previous GPU operations complete. std::copy on managed memory has no such guarantee.

So to answer directly: yes, you need a sync after an async kernel launch before accessing managed memory from the CPU — but not in other cases (e.g. CPU writing to managed memory before a kernel, or kernel→kernel on the same stream).

gschramm:
libparallelproj has `cudaDeviceSynchronize()` after every kernel launch - see e.g. [here](https://github.com/KUL-recon-lab/libparallelproj/blob/3071da97250138b82e13375eeb00db26c93cd190/src/joseph3d_fwd.cu#L107)

If you want to do host read/write after the kernel launch, you indeed need a `cudaDeviceSynchronize()`.

I think the only case where you can omit this is when you want to do "more GPU" work - e.g. projecting the same image again along different LORs into a separate array.

markus-jehl:
@KrisThielemans Finishing up the chunking now. Do you have any TOF data I could use for testing that implementation as well?

I'll also replace the d_scalar with a double, and place the error message here once I'm done with the chunking.

KrisThielemans:
https://petric.tomography.stfc.ac.uk/data/ has plenty of data. All GE data and Siemens vision 600 are TOF.

Note that the SIRF mult_factors are 1/stir_multfactors.

markus-jehl:
> run-time problem in `stir_timings` with CUDA after insert CuVeC for parallelproj is on a single computer, but not on 2 others (including one with the same docker set-up)

`ERROR: CudaGibbsPrior: CUDA error in CudaGibbsPenalty: cudaMalloc for d_input_data: invalid device ordinal` is thrown after the first cuda call in `CudaGibbsPenalty<elemT, PotentialT>::set_up`. Therefore it is not directly linked to d_input_data but points more towards parallelprojHelper with CuVec leaving cuda in a weird state. This is reproducible behaviour on my machine with a Docker image and local builds of libparallelproj v2.0.3 and STIR. On a comparable machine with the same Docker and local builds, the issue did not occur.

KrisThielemans:
could it be related to device syncing?

markus-jehl:
Rebuilding the Docker image from scratch without cache resolved the issue, so it must have been an out-of-date dependency!

KrisThielemans:
> From what I understood about the CUDA managed memory, the transparency is about data location (automatic CPU↔GPU), not about execution ordering. The sync is needed because GPU kernels are asynchronous — the host continues executing immediately after a kernel launch, so without a sync you risk reading managed memory before the kernel has finished writing to it.

ok, that is clear.

> 
> The old cudaMemcpy(DeviceToHost) didn't need an explicit sync because it's implicitly synchronizing — it blocks until all previous GPU operations complete. std::copy on managed memory has no such guarantee.
> 

ok

What about the following then: change `array_t_host` to
```c++
void array_to_host(Array<num_dimensions, elemT>& stir_array, const CuVec<elemT>& dev_data, bool sync=true)
```
and drop the explicit syncs that @denproc inserted. This way `array_to_host` for a `CuVec` acts the same as for a CUDA (ordinary) ptr, and we have less code again.

Good idea?

Dimitra-Kyriakopoulou:
Dear Professor @KrisThielemans,
 - I tested your `array_to_host(..., bool sync = true)` suggestion and pushed it here:
  `Dimitra-Kyriakopoulou/STIR:project08_pr1689_sync_cleanup`
  commit `354f20950`

   It passed the focused STIR `#1689` tests (`test_blocks_on_cylindrical_projectors`, `test_OSMAPOSL_parallelproj`, `test_priors`),   rebuilt through reduced SIRF, and still passed the Project 8 Stage 1/2/3 checks on the GPU VM:
    - Stage 1 RMSE: `0.0`
    - Stage 2 RMSE: `0.0`
    - Stage 3 RMSE: `5.241103662179114e-10`

- I also checked the other two follow-up points:
  - `d_scalar` does not currently look like a good standalone cleanup, because it is already a size-1 `CuVec<double>` reduction buffer, so it is not a raw device pointer that still needs converting.
  - I ran a separate TOF/chunking validation on the exact `#1689` stack. Forward projection matched exactly between chunked and unchunked runs; backprojection differences were very small and appear consistent with floating-point accumulation-order effects rather than a functional issue. For a reduced TOF geometry with `8` subset views:
    - chunk count `7`: backward RMSE `2.0786981487323249e-05`, max abs `0.00030517578125`
    - chunk count `3`: backward RMSE `2.0888311642741219e-05`, max abs `0.000335693359375`

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
markus-jehl: Actually, this seems like it is not possible. The kernel will set d_scalar to some value and it is being read out on the host again later. So memory copies seem required. Or am I missing something?

ColomboMatte0: d_scalar is where the atomicAdd operations point to, so it needs to live in device memory — that's how all thread blocks accumulate into the same location without race conditions. A plain double won't work here: scalars passed to CUDA kernels are copied by value, so any writes inside the kernel stay local and never make it back to the host. atomicAdd also explicitly requires a device pointer.

KrisThielemans:
I've put the parallelproj 1.* compatibility in and tested it on my Linux system without CUDA. It'd be great if someone could test it with CUDA.

markus-jehl:
> I've put the parallelproj 1.* compatibility in and tested it on my Linux system without CUDA. It'd be great if someone could test it with CUDA.

Just tested it - all fine.

Dimitra-Kyriakopoulou:
Dear Professor @KrisThielemans ,
 I prepared four small follow-up commits on my fork:

  - `project08_pr1689_array_to_host_sync_helper`
    `120807e47`
    Moves the CuVec host synchronization into `array_to_host(...)`.

  - `project08_pr1689_pair_chunks`
    `57b9d975a`
    Exposes `num_gpu_chunks` at the
  `ProjectorByBinPairUsingParallelproj` parameter-file level.

  - `project08_pr1689_projdata_external_buffer`
    `de12a0032`
    Adds a minimal `ProjDataInMemory` constructor so externally
  allocated contiguous `Array` storage can be adopted.

  - `project08_pr1689_array_cuvec_bridge`
    `cabf5d3df`
    Adds a small optional helper to let `Array` share ownership of
  `CuVec`-backed storage.

I am not entirely sure whether the preferred route here is to open separate follow-up PRs against `parallelproj2.0`. In case not, I am drawing your attention to these commits in case any of them are useful and you would like to pick them. 

THANK YOU!!
Dimitra

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
casperdcl: Oh it's an output var. ok.

P.S. haven't reviewed but hope as few atomics are used as possible.[^atom]

[^atom]: [random example](https://forums.developer.nvidia.com/t/can-i-avoid-using-atomicadd-with-this-kernel/36499)

KrisThielemans:
> Dear Professor @KrisThielemans , I prepared four small follow-up commits on my fork:
> 
> * `project08_pr1689_array_to_host_sync_helper`
>   `120807e47`
>   Moves the CuVec host synchronization into `array_to_host(...)`.

I'll do this myself, as we need some documentation etc

> * `project08_pr1689_pair_chunks`
>   `57b9d975a`
>   Exposes `num_gpu_chunks` at the
>   `ProjectorByBinPairUsingParallelproj` parameter-file level.

I've created https://github.com/KrisThielemans/STIR-1/pull/11

> * `project08_pr1689_projdata_external_buffer`
>   `de12a0032`
>   Adds a minimal `ProjDataInMemory` constructor so externally
>   allocated contiguous `Array` storage can be adopted.

I think this is already merged on master? https://github.com/UCL/STIR/pull/1694/

> * `project08_pr1689_array_cuvec_bridge`
>   `cabf5d3df`
>   Adds a small optional helper to let `Array` share ownership of
>   `CuVec`-backed storage.
> 

Are you referring to https://github.com/Dimitra-Kyriakopoulou/STIR/commit/3b1ea8f6c2bc534af3b607207f9af14e00fb8d0d? This is an interesting approach. I'd like to get @casperdcl's opinion on that. However, let's not do that here, but in https://github.com/UCL/STIR/issues/1679. See below.

> I am not entirely sure whether the preferred route here is to open separate follow-up PRs against `parallelproj2.0`. In case not, I am drawing your attention to these commits in case any of them are useful and you would like to pick them.
> 

Thanks! I generally prefer PRs with a smal aim. Some of the above have nothing to do with this PR, really, but only with the overall project, so I'd keep those for separate PRs. It also keeps the discussion focused.

src/include/stir/cuda_utilities.h
KrisThielemans: what about calling `resize()` always? No penalty if it's the correct size I guess.

KrisThielemans:
anyone can check, please.

Dimitra-Kyriakopoulou:
Dear Professor @KrisThielemans,
I checked `cbc8e4f` on the GPU VM, and the focused CUDA/STIR tests passed:
  - `test_blocks_on_cylindrical_projectors`
  - `test_OSMAPOSL_parallelproj`
  - `test_priors`

However for safety, I think it would still be good if someone else could also check, because I already made errors today ... 
In particular, 
  
  > I think this is already merged on master?

  Indeed! I am really sorry about that ...

  > Are you referring to
  https://github.com/Dimitra-Kyriakopoulou/STIR/commit/3b1ea8f6c2bc534af3b607207f9af14e00fb8d0d?

  Yes. `cabf5d3df` was the prior version; `3b1ea8f` is the current one.

  **I am really sorry** for all the carelessness, and **thank you so much** for your reply!!
  Dimitra

KrisThielemans:
@casperdcl I'm afraid I don't have the time to fix the CMake for using `find_package(cuvec)`. It's a bit tricky as we then need to add it to our `STIRConfig.cmake` (but only when using an external package). I think it's fine as-is.

src/include/stir/cuda_utilities.h
KrisThielemans: I'm going to leave this for now.. At the moment, `CUVEC_DISABLE_CUDA` isn't necessarily added for all STIR targets, so it feels a bit unsafe to rely on it.

KrisThielemans:
This should be fine now, aside from the release notes. I'd appreciate a full check and review from a few of you :-)

src/include/stir/recon_buildblock/CUDA/CudaGibbsPenalty.cuh
ColomboMatte0: atomicAdd is performed on the partial sums produced at the block level rather than per thread. As a result, the number of atomic operations scales with the number of blocks, leading to minimal impact on performance.

markus-jehl:
> This should be fine now, aside from the release notes. I'd appreciate a full check and review from a few of you :-)

Tested it and reviewed the code. All good from my side.

KrisThielemans:
I'm not going to change history on this and merge.

Thanks a lot all for your contributions!

src/include/stir/cuda_utilities.h
casperdcl: Moved to #1695

Issue #1307, opened 2024-01-06
ProjDataInfoGeneric et al need TOF changes


KrisThielemans:
`ProjDataInfoGenericNoArcCorr` currently contains a lot of copies of the `ProjDataInfocylindricalNoArcCorr` functions. this means the TOF changes need to be ported to the generic version. We also have to check the Block functions.

Ideally we do this by cleaning up the hierarchy, as opposed to creating even more duplication.

KrisThielemans:
Historical note: this was done by @pkhateri and @roemicha (on my recommendation) to be able to make the changes necessary for the blocks/generic implementation without disturbing the rest of STIR.

KrisThielemans:
The practical issue is that `ProjDataInfoGeneric` is currently derived from `ProjDataInfoCylindrical`, and therefore `ProjDataInfoGenericNoArcCorr` cannot inherit from `ProjDataInfoCylindricalNoArcCorr`, unless we'd use virtual class derivation (leading to "diamond" class hierarchies), which would likely create a lot of trouble.

The more conceptual difficulty is that `ProjDataInfoCylindricalNoArcCorr` really handles 2 types of data:
- data with certain views and `sin(s/R)` type of sampling
- interleaved data as corresponding to a scanner with discrete detectors (see [doc here](https://github.com/UCL/STIR/blob/7508babaf5e8147a3563f1642a47301af4c0834d/src/include/stir/ProjDataInfoCylindricalNoArcCorr.h#L45-L59)).

The "logical" thing to do seems to be to remove the "interleaved" functionality from `ProjDataInfoCylindricalNoArcCorr` (as that can be represented geometrically by blocks-on-cylindrical with a single crystal per block).

This is however hard, as currently the "generic"/"blocks" types are only handling span=1 and no mashing. Indeed, when there are (sizeable) axial gaps between blocks, strictly speaking "span" probably doesn't make a lot of sense, but in practice, many scanners use it anyway (as either the gaps are small, or they are "filled" with virtual crystals).

Other notes:
- current ProjDataInfoGeneric` is almost empty (mainly just `delete`s members in `ProjDataInfoCylindrical`). It should be merged with `ProjDataInfoGenericNoArcCorr`. A more appropriate name would be `ProjDataInfoDiscretisedScanner` or something.
[...]

PR #1696, opened 2026-03-20
simplify ProjDataInfoGeneric hierarchy

KrisThielemans:
This PR implements the steps discussed in https://github.com/UCL/STIR/issues/1307#issuecomment-2555972620


KrisThielemans:
Step 1 and 2 implemented.

Warning: I will squash some commits here once it works.

KrisThielemans:
Steps 1 & 2 are done, although I suddenly see segfaults in the MacOSDebug build in CI. I hope they will disappear with the later updates as that will be hard to debug (I have no access to a Mac).

Note that step 3 is for v7, as at the moment the `ProjDataInfoBlocksOnCylindricalNoArcCorr` still has a useful function that it checks it is indeed a "blocks" scanner.

We could rename `ProjDataInfoGenericNoArcCorr` in this PR as the final step (with a `using` to preserve compatibility), as it's current name really makes no sense. Maybe we should leave that for v7 though.

I think we only need 2 small changes to enable TOF (currently I still call `error()` when it's TOF data in the generic case), but to do that, we would need new test cases. In the interest of merging, we should probably do that in a different PR.

KrisThielemans:
@markus-jehl could you try this out, please?

KrisThielemans:
The code is a bit safer now. (I started testing TOF and ran into some unexpected errors because of the "deleted" functions, so now we call `error` instead). I've committed that in this PR as it doesn't have anything to do with TOF really.

KrisThielemans:
Enabling a block scanner in `test_proj_data_info` runs into trouble because `ProjDataInfoGenericNoArcCorr::get_bin(lor)` only handles `LORAs2Points` where the end points are directly in the detector coordinate map. This turns out not to be the case in `test_generic_proj_data_info` (where they are end points on a cylinder). However, coping with this would be very slow (it seems to need an expensive loop to see which of all the detector coordinates the LOR goes through...).

Of course, this problem is identical to what is currently on `master`, so it shouldn't stop us from merging. (It's a pity though, as it'd give us a bunch of TOF tests for free)

KrisThielemans:
@markus-jehl @NikEfth @danieldeidda I will merge this soon. It is working exactly as before, to the best of my knowledge.

Do test, even after the merge...

markus-jehl:
@KrisThielemans sorry, only saw this now. Managed to get around testing it now: works fine for me.

documentation/release_6.4.htm
```diff
@@ -141,6 +141,11 @@ <h3>New functionality</h3>
 
   <h3>Changed functionality</h3>
   <ul>
+    <li>
+      Simplified the <code>ProjDataInfoGenericNoArcCorr</code> class hierarchy by merging `<code>ProjDataInfoGeneric</code> into it
+      and deriving it from <code>ProjDataInfoCylindricalNoArcCorr</code>. See <a href=https://github.com/UCL/STIR/pull/1307>Issue #1307</a> for rationale.<br>
```
KrisThielemans, line 146:
```suggestion
      and deriving it from <code>ProjDataInfoCylindricalNoArcCorr</code>. See <a href=https://github.com/UCL/STIR/pull/1307>Issue #1307</a> for rationale.<br>
      <strong>WARNING:</strong> this change might have unexpected consequences for developers:
      when checking for the type of <code>ProjDataInfo</code> with <code>dynamic_cast</code> with
      a sequence of <code>if<code> statements (or equivalent), make sure you test for <code>ProjDataInfoGeneric</code>
      before testing for <code>ProjDataInfoCylindricalNoArcCorr</code>.(or even <code>ProjDataInfoCylindrical</code>). It is therefore strongly **recommended** to test via<br>
      <code>
      proj_data.get_proj_data_info_sptr()->get_scanner_ptr()->get_scanner_geometry() == "Cylindrical")
      </code><br>
      etc
      <br>
```

PR #1712, opened 2026-05-22
Fixes making the Python interface better (and some code reduction)

### LINKED ISSUE #284: SWIG: add downcasting

### LINKED ISSUE #662: Unable to set/get prior kappa image from STIR-python (matlab)

### LINKED ISSUE #707: STIR-python priors are GeneralisedPrior objects rather than child class priors

### LINKED ISSUE #1709: SWIG does not generate a constructor for ProjDataInfoCylindricalNoArcCorr
STIR Python can create `ProjDataInfoGenericNoArcCorr` objects, but not `ProjDataInfoCylindricalNoArcCorr` and it claims the class is abstract. This is due to https://github.com/swig/swig/issues/2333
and the line https://github.com/UCL/STIR/blob/1e15f4cd67f51421ae6d94654888c226413dbd1a/src/include/stir/ProjDataInfoCylindricalNoArcCorr.h#L251
adding the default value solves the problem, as in
https://github.com/UCL/STIR/blob/1e15f4cd67f51421ae6d94654888c226413dbd1a/src/include/stir/ProjDataInfoGenericNoArcCorr.h#L208

KrisThielemans:
This PR accumulates quite a few things that reduce C++ code a bit, but especially improve the SWIG-generated Python interface.

# C++ 
- introduce `DataWithProjDataInfo` class, containing a `shared_ptr<ProjDataInfo const>` and methods to get it and `get_min/max*` forwarding functions. This reduces duplicate C++ code for many classes related to projection data and also allows fixing things on the SWIG side only once.
- add run-time checks on consistency of sizes of an `Array` input and the `proj_data_info` in `Segment*`, `Viewgram`, `Sinogram`. Previously, we had a few `assert()` statements, so checks where only run in `Debug` mode. Also, we had no checks on TOF indices. 
- The previous change uncovered a bug in `ArcCorrection`, which actually didn't work for TOF data.
- Use covariant return types of `ProjDataInfo*::clone()`. e.g. `ProjDataInfoCylindricalNoArcCorr::clone()` now returns a    `ProjDataInfoCylindricalNoArcCorr*` as opposed to `ProjDataInfo*`. This is mostly useful for SWIG (see below)
- remove obsolete code for old compilers (`#ifdef STIR_NO_COVARIANT_RETURN_TYPES`)

# Python
- add downcasting for most hierarchies (see #284). This makes the returned objects much more useful. Unfortunately, implementation needs an explicit lists of the "leaves" (and a few work-arounds for SWIG limitations). Luckily, if a leaf is missing, it will just fall back to the default.<br>
This means the following examples now return objects of the "leaf" type (exposing all their methods) as opposed to a wrapped object only giving methods of the base-class:
   ```python
   pdi = proj_data.get_proj_data_info()
   prior = obj_fun.get_prior()
   im_copy = image.clone()
   ```
   Fixes #707<br>
Fixes #662

- add Python `__repr__` pointing the user to `parameter_info()` for a few more  hierarchies.
- enable @danieldeidda's  `get_lor()` etc functions for all `ProjDataInfo*NoArcCorr`
- add `DetectionPositionPairVector` (useful for `get_all_bins_for_det_pos_pair`)
- avoid some more SWIG warnings due to shadowing
- instantiate more `Coordinate2D` with `%template`
- allow creation of `ProjDataInfoCylindricalNoArcCorr` objects by working around SWIG bug<br>
Fixes #1709

# Documentation
- update Python examples for LORs (minor clean-up of `plot_scanner_LORs.py`, add `coordinates_vs_bins.py`)
- update STIR-developers-overview on TOF conventions etc


KrisThielemans:
I've added a Python example to translate between bins and detectors, show LORs and compare with backprojection: `coordinates_vs_bins.py`. Plots:
- based on detector coordinates. 
image
- (TOF) back-projection 
image

KrisThielemans:
I've now updated the figures  shown above to let the "y" axis point downwards, as per the STIR conventions.

PR #1731, opened 2026-07-21
Simple test for get_m() and scanner z_shift()

NikEfth:
Simple quick test for get_m

### DIFF
--- a/src/include/stir/Scanner.h
+++ b/src/include/stir/Scanner.h
@@ -107,6 +107,7 @@ class Succeeded;
 class Scanner
 {
   friend class BlocksTests;
+  friend class ProjDataInfoCylindricalNoArcCorrTests;
 
 public:
   /************* static members*****************************/
--- a/src/include/stir/Scanner.h
+++ b/src/include/stir/Scanner.h
@@ -3,9 +3,10 @@
     Copyright (C) 2000-2010, Hammersmith Imanet Ltd
     Copyright (C) 2011-2013, King's College London
     Copyright (C) 2016, University of Hull
-    Copyright (C) 2016, 2019, 2021, 2023 UCL
+    Copyright (C) 2016, 2019, 2021, 2023, 2024, 2026 UCL
     Copyright 2017 ETH Zurich, Institute of Particle Physics and Astrophysics
     Copyright (C 2017-2018, University of Leeds
+    Copyright (C) 2026, UMCG
     This file is part of STIR.
 
     SPDX-License-Identifier: Apache-2.0 AND License-ref-PARAPET-license
@@ -37,6 +38,7 @@
 #include "stir/CartesianCoordinate3D.h"
 #include "stir/DetectorCoordinateMap.h"
 #include "stir/shared_ptr.h"
+#include "stir/deprecated.h"
 #include <string>
 #include <list>
 #include <vector>
@@ -574,7 +576,9 @@ public:
   inline Succeeded find_detection_position_given_cartesian_coordinate(DetectionPosition<>& det_pos,
                                                                       const CartesianCoordinate3D<float>& cart_coord) const;
 
-  const shared_ptr<const DetectorCoordinateMap> get_detector_coordinate_map_sptr() const { return detector_map_sptr; }
+  shared_ptr<const DetectorCoordinateMap> STIR_DEPRECATED get_detector_map_sptr() const { return detector_map_sptr; }
+  shared_ptr<const DetectorCoordinateMap> get_detector_coordinate_map_sptr() const { return detector_map_sptr; }
+  const DetectorCoordinateMap& get_detector_coordinate_map() const { return *detector_map_sptr; }
 
 private:
   bool _already_setup;
@@ -617,26 +621,22 @@ private:
   //! This number corresponds the the least significant clock digit.
   float size_timing_pos;
 
-  //!
-  //! \brief scanner info needed for block geometry
+  //! \name scanner info needed for block geometry
   //! \author Parisa Khateri
-  //! A negative value indicates unknown.
+  //@{ A negative value indicates unknown.
   std::string scanner_geometry;     /*! scanner geometry */
   float axial_crystal_spacing;      /*! crystal pitch in axial direction in mm*/
   float transaxial_crystal_spacing; /*! crystal pitch in transaxial direction in mm*/
   float axial_block_spacing;        /*! block pitch in axial direction in mm*/
   float transaxial_block_spacing;   /*! block pitch in transaxial direction in mm*/
-
+  //@}
   std::string crystal_map_file_name;
-  shared_ptr<DetectorCoordinateMap> detector_map_sptr; /*! effective detection positions including average DOI */
+  shared_ptr<const DetectorCoordinateMap> detector_map_sptr; /*! effective detection positions including average DOI */
 
-  //! Since the two setters either create a new DetectorCoordinateMap or make a deep-copy
-  //! here the function justs moves from the the temporal shared_ptr to the local.
-  void install_detector_coordinate_map_sptr(shared_ptr<DetectorCoordinateMap> new_map_sptr);
-  //! This function creates a new DetectorCoordinateMap
-  void set_detector_map(const DetectorCoordinateMap::det_pos_to_coord_type& coord_map);
+  //! Directly set the \c shared_ptr (use with caution)
+  void set_detector_coordinate_map_sptr(shared_ptr<const DetectorCoordinateMap> new_map_sptr);
   //! This function performs a deep copy of the object
-  void set_detector_coordinate_map_sptr(const shared_ptr<const DetectorCoordinateMap>& coord_map_sptr);
+  void set_detector_coordinate_map(const DetectorCoordinateMap& coord_map);
   void initialise_max_FOV_radius();
 
   // function to create the maps
--- a/src/include/stir/Scanner.h
+++ b/src/include/stir/Scanner.h
@@ -107,6 +107,7 @@ class Succeeded;
 class Scanner
 {
   friend class BlocksTests;
+  friend class ProjDataInfoCylindricalNoArcCorrTests;
 
 public:
   /************* static members*****************************/
@@ -573,7 +574,7 @@ public:
   inline Succeeded find_detection_position_given_cartesian_coordinate(DetectionPosition<>& det_pos,
                                                                       const CartesianCoordinate3D<float>& cart_coord) const;
 
-  shared_ptr<const DetectorCoordinateMap> get_detector_map_sptr() const { return detector_map_sptr; }
+  const shared_ptr<const DetectorCoordinateMap> get_detector_coordinate_map_sptr() const { return detector_map_sptr; }
 
 private:
   bool _already_setup;
@@ -629,7 +630,13 @@ private:
   std::string crystal_map_file_name;
   shared_ptr<DetectorCoordinateMap> detector_map_sptr; /*! effective detection positions including average DOI */
 
+  //! Since the two setters either create a new DetectorCoordinateMap or make a deep-copy
+  //! here the function justs moves from the the temporal shared_ptr to the local.
+  void install_detector_coordinate_map_sptr(shared_ptr<DetectorCoordinateMap> new_map_sptr);
+  //! This function creates a new DetectorCoordinateMap
   void set_detector_map(const DetectorCoordinateMap::det_pos_to_coord_type& coord_map);
+  //! This function performs a deep copy of the object
+  void set_detector_coordinate_map_sptr(const shared_ptr<const DetectorCoordinateMap>& coord_map_sptr);
   void initialise_max_FOV_radius();
 
   // function to create the maps
--- a/src/test/test_proj_data_info.cxx
+++ b/src/test/test_proj_data_info.cxx
@@ -44,6 +44,7 @@
 #include <algorithm>
 #include <math.h>
 #include "stir/CPUTimer.h"
+#include "stir/format.h"
 
 using std::cerr;
 using std::setw;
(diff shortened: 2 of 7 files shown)


src/test/test_proj_data_info.cxx
```diff
@@ -128,12 +129,13 @@ ProjDataInfoTests::set_blocks_projdata_info(shared_ptr<Scanner> scanner_sptr)
...
-                                                                    max_ring_diff_v,
-                                                                    scanner_sptr->get_max_num_views(),
-                                                                    scanner_sptr->get_max_num_non_arccorrected_bins());
+  auto proj_data_info_blocks_sptr
+      = std::make_shared<TProjDataInfo>(scanner_sptr,
+                                        num_axial_pos_per_segment,
+                                        min_ring_diff_v,
+                                        max_ring_diff_v,
+                                        scanner_sptr->get_max_num_views(),
+                                        scanner_sptr->get_max_num_non_arccorrected_bins() / bin_fraction);
```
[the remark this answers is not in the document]
NikEfth: I don't think we need to guard this, as it is very limited.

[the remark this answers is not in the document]
KrisThielemans: sure

src/test/test_proj_data_info.cxx
```diff
@@ -102,7 +103,7 @@ class ProjDataInfoTests : public RunTests
   void test_generic_proj_data_info(ProjDataInfo& proj_data_info);
 
   template <class TProjDataInfo>
-  shared_ptr<TProjDataInfo> set_blocks_projdata_info(shared_ptr<Scanner> scanner_sptr);
+  shared_ptr<TProjDataInfo> set_blocks_projdata_info(shared_ptr<Scanner> scanner_sptr, int bin_fraction = 1);
```
KrisThielemans: probably should make this a `static` member

src/test/test_proj_data_info.cxx
```diff
@@ -1242,6 +1247,150 @@ ProjDataInfoCylindricalNoArcCorrTests::run_tests()
...
+ProjDataInfoCylindricalNoArcCorrTests::run_get_m_test()
+{
+  // ExamInfo not required for get_m() consistency checks.
+  //-- create projdata info Blocks on Cylindrical
+  auto scannerCyl_sptr = std::make_shared<Scanner>(Scanner::SAFIRDualRingPrototype);
+  scannerCyl_sptr->set_scanner_geometry("Cylindrical");
+  scannerCyl_sptr->set_up();
+
+  auto proj_data_info_cyl_sptr = std::make_shared<ProjDataInfoCylindricalNoArcCorr>();
+  proj_data_info_cyl_sptr = set_blocks_projdata_info<ProjDataInfoCylindricalNoArcCorr>(scannerCyl_sptr, 2);
```
KrisThielemans:
```suggestion
  auto proj_data_info_cyl_sptr = set_blocks_projdata_info<ProjDataInfoCylindricalNoArcCorr>(scannerCyl_sptr, 2);
```

src/test/test_proj_data_info.cxx
```diff
@@ -1242,6 +1247,150 @@ ProjDataInfoCylindricalNoArcCorrTests::run_tests()
...
+  scannerBlocks_sptr->set_num_transaxial_blocks_per_bucket(1);
+
+  scannerBlocks_sptr->set_axial_block_spacing(scannerBlocks_sptr->get_axial_crystal_spacing()
+                                              * scannerBlocks_sptr->get_num_axial_crystals_per_block());
+  scannerBlocks_sptr->set_transaxial_block_spacing(scannerBlocks_sptr->get_transaxial_crystal_spacing()
+                                                   * scannerBlocks_sptr->get_num_transaxial_crystals_per_block());
+  scannerBlocks_sptr->set_up();
+
+  auto proj_data_info_blocks_sptr = std::make_shared<ProjDataInfoBlocksOnCylindricalNoArcCorr>();
+  proj_data_info_blocks_sptr = set_blocks_projdata_info<ProjDataInfoBlocksOnCylindricalNoArcCorr>(scannerBlocks_sptr, 2);
```
KrisThielemans:
```suggestion
  auto proj_data_info_blocks_sptr = set_blocks_projdata_info<ProjDataInfoBlocksOnCylindricalNoArcCorr>(scannerBlocks_sptr, 2);
```

src/test/test_proj_data_info.cxx
```diff
@@ -1242,6 +1247,150 @@ ProjDataInfoCylindricalNoArcCorrTests::run_tests()
...
+
+  //-- create projdata info generic
+  shared_ptr<const DetectorCoordinateMap> map_sptr = scannerBlocks_sptr->get_detector_map_sptr();
+
+  //-- Copy the detector locations
+  DetectionPosition<> det_pos, det_pos_ord;
+  DetectorCoordinateMap::det_pos_to_coord_type coord_map_reordered;
```
KrisThielemans: I guess in this case it isn't really `reordered`. Do we really need this step?

src/test/test_proj_data_info.cxx
```diff
@@ -1242,6 +1247,150 @@ ProjDataInfoCylindricalNoArcCorrTests::run_tests()
...
+  scannerGeneric_sptr->set_up();
+
+  auto proj_data_info_generic_sptr = std::make_shared<ProjDataInfoGenericNoArcCorr>();
+  proj_data_info_generic_sptr = set_blocks_projdata_info<ProjDataInfoGenericNoArcCorr>(scannerGeneric_sptr, 2);
+
+  Bin bin;
+  CartesianCoordinate3D<float> det_cyl_1, det_cyl_2, det_gen_1, det_gen_2, det_blk_1, det_blk_2;
+  float z_shift_cyl = (scannerCyl_sptr->get_axial_length() - scannerCyl_sptr->get_ring_spacing()) / 2.F;
+  float z_shift_gen = scannerGeneric_sptr->get_coordinate_for_det_pos(DetectionPosition<>(0, 0, 0)).z();
+  float z_shift_blk = scannerBlocks_sptr->get_coordinate_for_det_pos(DetectionPosition<>(0, 0, 0)).z();
```
KrisThielemans: This choice means we've strange `+` and `-` choices below. What about flipping the sign for `z_shift_gen` and `z_shift_blk`, and using `-z_shift*` below?

KrisThielemans (changes requested):

KrisThielemans:
@nikefth could we get this across the line? We're piling up PRs :-)

src/test/test_proj_data_info.cxx
NikEfth: I guess for the time will be fine

NikEfth:
Done :)

src/test/test_proj_data_info.cxx
KrisThielemans: could you clarify in a comment in the code why we need this (if we do), rename the variable to something else (as I see no reordering). Sorry. I should know...

NikEfth: OK, I can rename ```coord_map_reorder``` but I don't have a way to get the ```DetectorCoordinateMap::det_pos_to_coord_type``` from ```shared_ptr<const DetectorCoordinateMap> map_sptr``` I need to go through the loops. 
Is your comment abou the naming? Or the there another way to set it?

NikEfth: I can make a quick get const pointer function to avoid duplication. But that will expose the private member.

NikEfth:
I think this is done

NikEfth:
I am sorry for the typos and mistakes this afternoon. Now it should be fine.

src/include/stir/DetectorCoordinateMap.h
```diff
@@ -84,6 +84,8 @@ class DetectorCoordinateMap
   /*! applies sorting to standard STIR order */
   void set_detector_map(const det_pos_to_coord_type& coord_map);
 
+  const det_pos_to_coord_type& get_det_pos_to_coord_type() const { return det_pos_to_coord; }
```
KrisThielemans: This function has the wrong name. it should be either `get_det_pos_to_coord()` or `get_detector_map()` (for consistency with `set_...`. I suppose the latter is then better.

KrisThielemans (changes requested):
I guess we can merge once we rename `get_det_pos_to_coord_type()`. However, having a closer look at this, I think the difficulty really sits in `Scanner`. This has
https://github.com/UCL/STIR/blob/45c403c7f67350dbb2b60d1f2921b72e1efa0985/src/include/stir/Scanner.h#L631-L633

I think the cleanest way to resolve this is to add `get_detector_coordinate_map() const` and `set_detector_coordinate_map()` (where the current `set_detector_map()` would then call `set_detector_coordinate_map()`) to `Scanner`. I'd even deprecate the current `Scanner::set_detector_map`, as it uses some internal stuff.

If we do that, we wouldn't need the mod to `DetectorCoordinateMap`.

Up to you...

src/include/stir/DetectorCoordinateMap.h
NikEfth: removed

NikEfth:
I was not 100% if the setters should be deep-copy of the DetectorCoordinateMap. But in the previous implementation a new object was created ... so I kept that.

src/include/stir/Scanner.h
```diff
@@ -573,7 +574,7 @@ class Scanner
   inline Succeeded find_detection_position_given_cartesian_coordinate(DetectionPosition<>& det_pos,
                                                                       const CartesianCoordinate3D<float>& cart_coord) const;
 
-  shared_ptr<const DetectorCoordinateMap> get_detector_map_sptr() const { return detector_map_sptr; }
+  const shared_ptr<const DetectorCoordinateMap> get_detector_coordinate_map_sptr() const { return detector_map_sptr; }
```
KrisThielemans: can't delete the old version for backwards compatibility. I suggest to prefix with `STIR_DEPRECATED`(after including the header).

I'm really not sure why we'd return a `const shared_ptr`. See e.g. https://stackoverflow.com/questions/8716330/purpose-of-returning-by-const-value

src/include/stir/Scanner.h
```diff
@@ -629,7 +630,13 @@ class Scanner
   std::string crystal_map_file_name;
   shared_ptr<DetectorCoordinateMap> detector_map_sptr; /*! effective detection positions including average DOI */
 
+  //! Since the two setters either create a new DetectorCoordinateMap or make a deep-copy
+  //! here the function justs moves from the the temporal shared_ptr to the local.
+  void install_detector_coordinate_map_sptr(shared_ptr<DetectorCoordinateMap> new_map_sptr);
+  //! This function creates a new DetectorCoordinateMap
   void set_detector_map(const DetectorCoordinateMap::det_pos_to_coord_type& coord_map);
+  //! This function performs a deep copy of the object
+  void set_detector_coordinate_map_sptr(const shared_ptr<const DetectorCoordinateMap>& coord_map_sptr);
```
KrisThielemans: Usnig `shared_ptr` makes things needlessly restrictive/confusing(as it's deep-copy anyway). I prefer
```suggestion
  void set_detector_coordinate_map(const DetectorCoordinateMap& coord_map);
```
That way, it doesn't really even need documentation that it's a deep-copy.

src/include/stir/Scanner.h
```diff
@@ -629,7 +630,13 @@ class Scanner
   std::string crystal_map_file_name;
   shared_ptr<DetectorCoordinateMap> detector_map_sptr; /*! effective detection positions including average DOI */
 
+  //! Since the two setters either create a new DetectorCoordinateMap or make a deep-copy
+  //! here the function justs moves from the the temporal shared_ptr to the local.
+  void install_detector_coordinate_map_sptr(shared_ptr<DetectorCoordinateMap> new_map_sptr);
```
KrisThielemans: we don't have the `install` name anywhere. Given my suggestion below to rename `set_detector_coordinate_map_sptr` to `set_detector_coordinate()`, I think it would make perfect sense to call this one `set_detector_coordinate_map_sptr`.

(Do we need it though? I'd be happy enough to make it `private`, but as you prefer)

KrisThielemans (changes requested):

src/include/stir/Scanner.h
NikEfth: I know, but the getter returns a shared_ptr ... I looked more symmetric this way

src/include/stir/Scanner.h
KrisThielemans: oops. it's already `private`. sorry.

I'll rename it anyway.

KrisThielemans (approved):
@NikEfth I've made some changes and a tiny clean-up (also of some unrelated things that I spotted).

KrisThielemans:
segfault in CI for MacOS-conda-latest-compiler
```
35/74 Test #35: test_interpolate_projdata .......................................................***Exception: SegFault  1.32 sec

INFO: Performing symmetric interpolation test for BlocksOnCylindrical scanner

INFO: Determined voxel size by dividing default_bin_size (2) by zoom

INFO: Disabling all symmetries for the projector except for symmetry_z since they are not implemented in block geometry yet.

WARNING: ProjMatrixByBinUsingRayTracing used for pixel size (x,y)=(2,2) that is smaller than the central bin size (7.747817) divided by num_tangential_LORs (1).
This matrix will completely miss some voxels for some (or all) views. It is therefore to best to increase 'number of rays in tangential direction to trace for each bin'.
```

Second run crashed [later](https://github.com/UCL/STIR/actions/runs/33985461246/job/101391305160#step:16:145) in "asymmetric interpolation test for BlocksOnCylindrical scanner".

This seems unrelated to this PR, but it didn't occur in other Actions.

KrisThielemans:
I'm running `test_interpolate_projdata` through valgrind on my Ubuntu 24.04. No problems. The above segfault could therefore be race condition, hopefully only appearing in MacOS clang.

KrisThielemans:
@NikEfth this is ready now as far as I'm concerned. I intend to squash most commits before merging (today).

KrisThielemans:
I've re-run CI 3 times now without any trouble. I therefore hope this was due to some temporary problems on the runner.

I'll go ahead and merge.

NikEfth:
TBH, I think there is a race codition somewhere. I have seen similar crashes from time to time, especially in my mac. I don't think that it appears on this branch

PR #1732, opened 2026-07-23
PETSIRD: fix delayeds and some corner cases

KrisThielemans:
- properly read delayed_events_are_stored
- handle corner cases where there are no prompts/delayeds in a time block. Before, we were aborting.


KrisThielemans:
One job failure due to zenodo time-out. re-running...

@NikEfth this has been confirmed to be working ok on the NEMA mMR by @alanosys. Without it, STIR didn't return any delayeds.

NikEfth:
ah, OK, I agree with the changes.

PR #1746, opened 2026-08-10
Better generic/blocks-on-cylindrical support: ROOT support and enabled TOF

### LINKED ISSUE #1307: ProjDataInfoGeneric et al need TOF changes

### LINKED ISSUE #1520: Core dump while trying to access Scanner object in C++ project using STIR as a dependancy
Found this issue in an code base. I can replicate the issue with a modification of `using_installed_STIR` C++ example https://github.com/UCL/STIR/tree/master/examples/C%2B%2B/using_installed_STIR

## Setup

Fresh STIR clone, build and install, most options like ROOT, Python, OpenMP are disabled. 
```
$cd ${STIR_REPO}/examples/C++/using_installed_STIR 
$cmake . -DSTIR_DIR=${STIR_REPO}/install/lib/cmake/STIR-6.2 # No change in output between `CMAKE_BUILD_TYPE` $configurations
$make
$./demo_create_image
```

This all works fine, create the expected output and test interfiles are created.

----------

Modify the `demo_create_image.cxx` file to 
```cxx
#include "stir/Scanner.h"

int
main()
{
  std::cout << "Running demo_create_image" << std::endl;
  auto stir_scanner = stir::Scanner::get_scanner_from_name("D690");
  std::cout << "Constructed Scanner" << std::endl;
  // Basic test to ensure accessible scanner properties
  std::cout << stir_scanner->get_num_axial_blocks() << std::endl;
  std::cout << "Done" << std::endl;
  return EXIT_SUCCESS;
}
```
```
$ make && ./demo_create_image
Consolidate compiler generated dependencies of target demo_create_image
[100%] Built target demo_create_image
Running demo_create_image
Constructed Scanner
Floating point exception (core dumped)
```

It produces the `Floating point exception (core dumped)` during `stir_scanner->get_num_axial_blocks()`

---------

I have a setup with a windows install and I don't get this issue. I am not sure what I am missing...

### LINKED ISSUE #1752: Interfile parsing of BlocksOnCylindrical uses wrong default for axial crystal spacing
https://github.com/UCL/STIR/blob/6f3c5e6231bfc3aa9fab0057135e0dc1f4c06ebd/src/IO/InterfileHeader.cxx#L1098-L1099
note use of transaxial spacing to set the axial spacing.

This bug occurs when the interfile header has a known scanner model AND it doesn't have the `axial_distance_between_crystals (cm)` keyword.

KrisThielemans:
Add support for blocks on cylindrical scanners for reading GATE ROOT list-mode files.

Currently no tests for it yet. Help appreciated. @NikEfth @VietAnhDao


KrisThielemans:
All ctests work on my machine. However, the headers of the test files on Zenodo need updating, so CI will currently still fail.

We could fix the headers during CI, but I suppose it's better to upload a new version on Zenodo anyway.

IMPORTANT: to get all the tests to work, I had to enable TOF for blocks-on-cylindrical (and fix a few bugs elsewhere).

NikEfth:
I can work on this after Thursday

KrisThielemans:
CI failures:
- zenodo download failed (but CI would have failed anyway)
- `test_OSMAPOSL_parallelproj` fails. I've seen that on my machine as well, although I don't understand why. (It seems to be reconstructing arc-corrected data).

src/buildblock/Scanner.cxx
```diff
@@ -1550,8 +1561,6 @@ Scanner::set_params(Type type_v,
     set_scanner_geometry("Cylindrical");
   else
     set_scanner_geometry(scanner_geometry_v);
-
-  set_up();
```
KrisThielemans, line 1554: Probably best to re-instate this, but I had some trouble during testing.

KrisThielemans:
> * `test_OSMAPOSL_parallelproj` fails. I've seen that on my machine as well, although I don't understand why. (It seems to be reconstructing arc-corrected data).

reconstructed image with parallelproj seems to use very different dimensions. Still no idea why.
image

KrisThielemans:
I had to call `Scanner:set_up()` in more cases. Tests should work now (aside from the Zenodo problem)

KrisThielemans:
`test_OSMAPOSL_parallelproj` still fails on GHA, but in a different way (reconstructed images are apparently zero), and not on my laptop. grrr

KrisThielemans:
I was a bit surprised that `test_consistency_with_GATE` is working fine with the old (i.e. cylindrical) .hroot files, but of course it should, as otherwise we would have broken something.

I still think it's better to test with "blocks" .hroot. (I suppose we could test with both, but I'm not sure that's really a good idea).

KrisThielemans:
Appveyor build for VS 2019 fails while compiling `test_IO_ITKMulticomponent.cxx` with

https://ci.appveyor.com/project/KrisThielemans/stir/builds/54710682/job/jucuo2u66rcx3u2d#L2838
```
C:\projects\stir\src\include\stir/LORCoordinates.inl(297): warning C4244: '=': conversion from 'double' to 'coordT', possible loss of data
        with
        [
            coordT=size_t
        ]
C:\projects\stir\src\include\stir/modulo.h(140): error C2668: 'stir::modulo': ambiguous call to overloaded function
C:\projects\stir\src\include\stir/modulo.h(87): note: could be 'int stir::modulo(const int,const int)'
C:\projects\stir\src\include\stir/modulo.h(70): note: or       'float stir::modulo(const float,const float)'
C:\projects\stir\src\include\stir/modulo.h(52): note: or       'double stir::modulo(const double,const double)'
C:\projects\stir\src\include\stir/modulo.h(140): note: while trying to match the argument list '(const FloatOrDouble, FloatOrDouble)'
        with
        [
            FloatOrDouble=size_t
        ]
C:\projects\stir\src\include\stir/LORCoordinates.inl(134): note: see reference to function template instantiation 'FloatOrDouble stir::to_0_2pi<coordT>(const FloatOrDouble)' being compiled
        with
        [
            FloatOrDouble=size_t,
            coordT=size_t
        ]
C:\projects\stir\src\include\stir/LORCoordinates.inl(130): note: while compiling class template member function 'stir::LORInCylinderCoordinates<coordT>::LORInCylinderCoordinates(const stir::LORInAxialAndNoArcCorrSinogramCoordinates<coordT> &)'
        with
        [
            coordT=size_t
        ]
C:\projects\stir\src\include\stir/LORCoordinates.inl(492): note: see reference to function template instantiation 'stir::LORInCylinderCoordinates<coordT>::LORInCylinderCoordinates(const stir::LORInAxialAndNoArcCorrSinogramCoordinates<coordT> &)' being compiled
        with
        [
            coordT=size_t
        ]
C:\projects\stir\src\include\stir/LORCoordinates.inl(367): warning C4244: 'argument': conversion from 'float' to 'coordT', possible loss of data
        with
        [
            coordT=size_t
        ]
C:\projects\stir\src\include\stir/LORCoordinates.inl(407): note: see reference to function template instantiation 'stir::Succeeded stir::find_LOR_intersections_with_cylinder<coordT,coordT>(stir::LORInCylinderCoordinates<coordT> &,const stir::LORAs2Points<coordT> &,const double)' being compiled
        with
        [
            coordT=size_t
        ]
C:\projects\stir\src\include\stir/LORCoordinates.inl(406): note: while compiling class template member function 'stir::Succeeded stir::LORAs2Points<coordT>::change_representation(stir::LORInCylinderCoordinates<coordT> &,const double) const'
        with
        [
            coordT=size_t
        ]
C:\projects\stir\src\include\stir/LORCoordinates.inl(492): note: see reference to class template instantiation 'stir::LORAs2Points<coordT>' being compiled
        with
        [
            coordT=size_t
        ]
C:\projects\stir\src\include\stir/LORCoordinates.inl(369): warning C4146: unary minus operator applied to unsigned type, result still unsigned
C:\projects\stir\src\include\stir/LORCoordinates.inl(370): warning C4146: unary minus operator applied to unsigned type, result still unsigned
```

All other compilers including VS 2022 are fine.

It's trying to instantiate `LORInCylinderCoordinates<size_t> which should indeed fail. (Other warnings are related to `LORInAxialAndNoArcCorrSinogramCoordinates<size_t>` ).

It looks like this is caused by
https://github.com/UCL/STIR/blob/3da6880a5b87bb3de8043ccc039848cb1061adb2/src/test/IO/test_IO_ITKMulticomponent.cxx#L62
and the `RunTests` addition in this PR
```c++
template <class T>
bool
RunTests::check_if_equal(const LORInAxialAndNoArcCorrSinogramCoordinates<T>& a,
                         const LORInAxialAndNoArcCorrSinogramCoordinates<T>& b,
                         const std::string& str)
```
Looks like a compiler bug to me.

KrisThielemans:
Aside from problems with downloading files from Zenodo, <strike>this is all fine now</strike> (edited after more tests).

I'll revert the change that removed the call to `set_up()` from `set_params()` to avoid backwards compatibility problems, and make sure it is only enabled in 7.0.

@VietAnhDao @emikhay this PR that should help you with GATE simulations for blocks and hence NeuroLF. As long as you edit the .hroot file.

Summary of all changes is in the release_6.5.htm. Commit is 0eedd70e12d051bdf01d55b23aeef3df80eda76a.

KrisThielemans:
`test_consistency_with_GATE` now works locally with both blocks and cylindrical (including TOF), and both `ProjDataInfoGenericNoArcCorr::get_LOR` and the `ProjDataInfoCylindricalNoArcCorr::get_LOR` are consistent with `CListEventROOT::get_LOR`. I think all tests will work.

However... the swaps in the code make no logical sense to me. There are some strange choices, so I wonder if there's 2 bugs somewhere in `master` (and now here) that compensate each other...

KrisThielemans:
Unfortunately, due to the `Scanner::set_up()` changes, this PR has become unwieldy. It might be better to split it into 2.

src/listmode_buildblock/CListRecordROOT.cxx
```diff
@@ -38,20 +39,53 @@ CListEventROOT::CListEventROOT(const shared_ptr<const ProjDataInfo>& proj_data_i
...
+  if (_det_pos.timing_pos() >= 0)
+    {
+      _det_pos.pos1() = det1;
+      _det_pos.pos2() = det2;
+    }
+  else
+    {
+      _det_pos.pos1() = det2;
+      _det_pos.pos2() = det1;
+    }
```
KrisThielemans, line 59: This is really rather worrying, but i had to do this to get tests to work.

NikEfth: I my mind I don't really see a bug (but I should put this in paper first). It is messy -yes. But, the problem starts from the fact that gate simulations have only positive time differences. Then we need figure out when to make them negative and swap them accordingly. BUT scanners come with time differences already centered to 0 and we need to make them swap according to STIR conventions. 

So, there are two paths. The scanner path centered in 0 has to have the simple swapping according to STIR detector locations and how the Eulers formula returns p1 and p2 after conversions to radial coordintates. 

The ROOT needs 1. Center to 0 with swaps according to time 2. swaps according to detector position (to match the scanner path).

KrisThielemans: >  gate simulations have only positive time differences

that isn't the case as far as I can see (I suppose that we wouldn't have uncovered problems in the test if it was).

I don't think I understand what you write below that, but indeed, the `get_LOR()` choices are determined by
- how we convert to radial coordinates from 2 points
- conventions used in the projectors (currently only checked the RT-matrix) as far as order along the LOR goes and then applying the TOF kernel

The current situation is a dance to get all of this work together, but it is messy. I would also say it is surprising to the user that the order of the points in `get_LOR()`  changes depending on the TOF bin.

NikEfth: They _must_ start as positive. 
Having that said, I do remember that some rogue simulations at the very first nanoseconds can have negative differences. But we should ignore these, or do nothing they are very few events. 

I will have a look in the particular root file. 

image

NikEfth: > The current situation is a dance to get all of this work together

ABSOLUTELY my feeling since I first worked on the TOF :D :D

KrisThielemans:
@NikEfth As this PR follows (and is consistent with) `ProjDataInfoCylindricalNoArcCorr`, maybe the way forward is to

1. document the choices made
2. possibly introduce a helper for doing the swaps (not sure where it should sit, but I guess we could have 2: one in `ProjDataInfo` and one in `DetectionPositionPair`. The problem for the latter is that it doesn't know if the number of TOF bins is even or not, so it will need an extra argument.
3. merge
4. create an issue pointing to some weird choices.

Suggestions for step 2? (or we leave that for another PR, e.g. #1755 )

NikEfth:
Sure, let's merge. because I want to finish the PR on GATE 10 and then the even TOF bins. In the even bins we can discuss more.

PR #1755, opened 2026-09-11
* Fixed mashing for even number of TOF bins.

### LINKED ISSUE #1751: potential problems with even TOF bins and TOF mashing
#1604 introduced changes for even number of TOF bins.
https://github.com/UCL/STIR/blob/6f3c5e6231bfc3aa9fab0057135e0dc1f4c06ebd/src/buildblock/ProjDataInfo.cxx#L221-L223
I think this doesn't work with our TOF_mashing calculations, such as
https://github.com/UCL/STIR/blob/6f3c5e6231bfc3aa9fab0057135e0dc1f4c06ebd/src/include/stir/ProjDataInfoCylindricalNoArcCorr.inl#L185
(For odd number of TOF bins, 0->0, but for even number (either in the original or in the mashed), this doesn't seem to preserve the centre.

@NikEfth @danieldeidda what do you think?

NikEfth:
This PR fixes #1751. With even number of TOF abs, naive inversion does not work. If even we need to subtract 1. 
Now, both -1 and 0 become a mashed TOF bin 0 but they have inverted detectors. 

## Testing performed
New tests introduced to explicitly check the transformation with even and odd TOF bins, with and without mashing. 

  - [X] I have added docstrings/doxygen in line with the guidance in the developer guide
  - [X] I have implemented unit tests that cover any new or modified functionality (if applicable)
  - [X] The code builds and runs on my machine

## Contribution Notes

Please tick the following: 

 - [X] The content of this Pull Request (the Contribution) is intentionally submitted for inclusion in STIR (the Work) under the terms and conditions of the [Apache-2.0](https://www.apache.org/licenses/LICENSE-2.0) License.
 - [X] I (or my institution) have signed the STIR Contribution License Agreement (not required for small changes).

### DIFF
--- a/src/include/stir/ProjDataInfoCylindricalNoArcCorr.inl
+++ b/src/include/stir/ProjDataInfoCylindricalNoArcCorr.inl
@@ -182,7 +182,9 @@ ProjDataInfoCylindricalNoArcCorr::get_det_pos_pair_for_bin(DetectionPositionPair
       dp.pos2().axial_coord() = a1;
     }
 
-  dp.timing_pos() = std::abs(bin.timing_pos_num()) * this->get_tof_mash_factor();
+  dp.timing_pos()
+      = (bin.timing_pos_num() >= 0 ? bin.timing_pos_num() : -bin.timing_pos_num() - (this->get_num_tof_poss() % 2 == 0))
+        * this->get_tof_mash_factor();
 }
 
 END_NAMESPACE_STIR
--- a/src/test/test_time_of_flight.cxx
+++ b/src/test/test_time_of_flight.cxx
@@ -93,11 +93,106 @@ private:
 void
 TOF_Tests::run_tests()
 {
+  info("Test conversion between -N/2 ... +N/2 for odd or -(N/2) ... +(N/2 - 1) for even TOF bins to 0 -> Max Detector "
+       "Positions. ");
+  {
+    // From odd number 351 to odd 9
+    test_scanner_sptr.reset(new Scanner(Scanner::PETMR_Signa));
+    const int test_tof_mashing_factor1 = 1; // to have 9 TOF bins (351/39=9)
+    const int test_tof_mashing_factor = 39; // to have 9 TOF bins (351/39=9)
+    test_proj_data_info_sptr.reset(ProjDataInfo::ProjDataInfoCTI(test_scanner_sptr,
+                                                                 1,
+                                                                 0,
+                                                                 test_scanner_sptr->get_num_detectors_per_ring() / 2,
+                                                                 test_scanner_sptr->get_max_num_non_arccorrected_bins(),
+                                                                 /* arc_correction*/ false));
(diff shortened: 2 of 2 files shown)


src/include/stir/ProjDataInfoCylindricalNoArcCorr.inl
```diff
@@ -182,7 +182,9 @@ ProjDataInfoCylindricalNoArcCorr::get_det_pos_pair_for_bin(DetectionPositionPair
       dp.pos2().axial_coord() = a1;
     }
 
-  dp.timing_pos() = std::abs(bin.timing_pos_num()) * this->get_tof_mash_factor();
+  dp.timing_pos()
+      = (bin.timing_pos_num() >= 0 ? bin.timing_pos_num() : -bin.timing_pos_num() - (this->get_num_tof_poss() % 2 == 0))
```
KrisThielemans, line 186: are we sure a `bool` converts to 1?
probably clearer/safer would be
```c++
(this->get_num_tof_poss() % 2 == 0 ? 1 : 0)
```
Or cryptic
```c++
(1 - (this->get_num_tof_poss() % 2))
```

KrisThielemans (commented):
As this correct?

You put in changes for even number tof bins, but the number of timing_pos could also be even. Do we cope with this? Same convention?

We want the `dp.timing_pos` correspond to the centre of the (mashed) TOF bin.

Examples that seem to give undesirable results

- num_tof_bins is even

| bin.timing_pos_num | centred_range for bin | centred_range for dp.timing_pos | actual (asymmetric) range |
| --- | ---- | ----------- | ----------------- | 
| 0   |  0.5 | 0.5 * mash | 0.5 * mash - 0.5 |

`dp.timing_pos` wouldn't be integer if mash is even

- num_tof_bins is odd, but mash is even

| bin.timing_pos_num | centred_range for bin | centred_range for dp.timing_pos | actual (asymmetric) range |
| --- | --- | --- | ----- | 
| 0   |  0   |  0  |  -0.5 |

src/include/stir/ProjDataInfoCylindricalNoArcCorr.inl
NikEfth: Yeah, fair enought

NikEfth:
I can simplify by droping the convention that dp should be positive. It won't fix everything though.

KrisThielemans:
I don't think that the "positive dps" is the problem. At the moment, I only see the approach followed by @danieldeidda #1334 as correct.

From my table above, I think we should put in some run-time checks and call `error`  if (`scanner.num_tof_bins` is even and  `TOF_mashing_factor`>1), or (`TOF_mashing_factor` is even). That's still better than not allowing even TOF bins at all. (Maybe those restrictions can be relaxed a bit).

I'm not sure what you're testing at the moment. Do you test any of the above cases? Shouldn't the test not fail?

NikEfth:
Now I cleanly test for more cases. Bear with me for a couple of days, please.

